PR #156, opened 2017-04-21
Output more startup config messages in the Manticore CLI

etcd:


japesinator:
This is awesome, thanks! Two small things:

* Would you mind pushing an empty commit to sate Travis? It currently is unaware this branch exists

* When the user specifies `--verbose`, this will show them `Verbosity: 5`, which is mildly counterintuitive. How do you feel about changing that message to something like `Verbose mode on`?

dguido:
Hm, this isn't even in the list of branches on Travis-CI. I'm just going to make a dummy commit to get it going.

etcd:
Hmm, there's 6 different levels of verbosity. Were you thinking of outputting "Verbose mode on" for levels 1-5, and "Verbose mode off" for 0? What about something like "Verbosity (min 0, max 5): 5" or something?

japesinator:
If this is in main.py, the verbosity should only ever get set by the verbosity flag, which is a binary control, so I think it makes sense to have a binary indicator

japesinator:
Also, that commit seems to not work with travis for some reason, so I'd recommend you download the [patch](https://github.com/trailofbits/manticore/pull/156.patch), apply it to a new branch, and issue a PR from that. Sorry about the inconvenience!

mossberg:
Since this will require a new PR anyway, I'm going to close this.

dguido:
Hey @etcd, we're going back and collecting CLAs for past contributors of Manticore but we can't seem to find your email. Can you do us a favor and click through this link to sign the Manticore CLA? Thanks! We're happy to accept any more contributions you want to make :-D

https://cla-assistant.io/trailofbits/manticore

PR #188, opened 2017-04-26
More memory test, autodereferencing mem

feliam:
Extra memory test


tests/test_memory.py
```diff
@@ -1614,6 +1614,28 @@ def test_mmap_file_extra(self):
         m = pickle.loads(pickle.dumps(m))
         self.assertItemsEqual(m[0x10000000:0x10003000], 'X'*0x27f0 + 'Y'*0x20 + '\x00'*0x7f0)
 
+    def testMultiSymbolicDereferences(self):
```
mossberg: snake_case?

PR #194, opened 2017-04-26
Add Google2016 unbreakable challenge to examples repo

ctfhacker:


examples/script/google2016_unbreakable/win.py
```diff
@@ -0,0 +1,62 @@
...
+def hook(state):
+    """ CAUTION: HACK """
+    """ From entry point, jump directly to code performing check """
+
+    # Create a symbolic buffer for our input
+    buffer = state.new_symbolic_buffer(0x43)
+
+    # We are given in the challenge that the flag begins with CTF{
+    # So we can apply constraints to ensure that is true
+    state.constraints.add(buffer[0] == ord('C'))
```
mossberg: does this still work if we use `state.add` instead of `state.constraints.add`? if so, let's use that.

ctfhacker: Absolutely! Changed and pushed.

examples/script/google2016_unbreakable/win.py
```diff
@@ -0,0 +1,62 @@
...
+def hook(state):
+    """ CAUTION: HACK """
+    """ From entry point, jump directly to code performing check """
+
+    # Create a symbolic buffer for our input
+    buffer = state.new_symbolic_buffer(0x43)
+
+    # We are given in the challenge that the flag begins with CTF{
+    # So we can apply constraints to ensure that is true
+    state.add(buffer[0] == ord('C'))
```
mossberg: Could you please update these calls to be `state.constrain` (newly added API) (http://manticore.readthedocs.io/en/latest/api.html#manticore.core.state.State.constrain)

PR #197, opened 2017-04-26
Fix RSB; Add RSC

yan:


manticore/core/cpu/arm.py
```diff
@@ -572,7 +573,16 @@ def ADD(cpu, dest, src, add):
 
     @instruction
     def RSB(cpu, dest, src, add):
-        result, carry, overflow = cpu._ADD(~src.read(), add.read(), 1)
+        inv_src = GetNBits(~src.read(), cpu.address_bit_size)
+        result, carry, overflow = cpu._ADD(inv_src, add.read(), 1)
+        dest.write(result)
+        return result, carry, overflow
```
mossberg: forgot to set flags?

manticore/core/cpu/arm.py
```diff
@@ -572,7 +573,16 @@ def ADD(cpu, dest, src, add):
...
+
+    @instruction
+    def RSC(cpu, dest, src, add):
+        carry = cpu.regfile.read('APSR_C')
+        inv_src = GetNBits(~src.read(), cpu.address_bit_size)
+        result, carry, overflow = cpu._ADD(inv_src, add.read(), carry)
         dest.write(result)
         return result, carry, overflow
```
mossberg: ditto

manticore/core/cpu/arm.py
yan: they get set in `_ADD`

PR #198, opened 2017-04-27
Only make registers appear on verbosity level 5

japesinator:
This PR also sets the default `--verbose` verbosity to 4, which is in my opinion much more sane

### DIFF
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -11,6 +11,7 @@ from functools import wraps
 import types
 import logging
 logger = logging.getLogger("CPU")
+register_logger = logging.getLogger("REGISTERS")
 
 
 
@@ -436,10 +437,10 @@ class Cpu(object):
 
         implementation = getattr(self, name, fallback_to_emulate)
 
-        #log
         if logger.level == logging.DEBUG :
-            for l in str(self).split('\n'):
-                logger.debug(l)
+            logger.debug(self.print_instruction())
+            for l in self.print_registers().split('\n'):
+                register_logger.debug(l)
 
         implementation(*instruction.operands)
         self._icount+=1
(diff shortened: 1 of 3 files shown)


manticore/core/cpu/abstractcpu.py
```diff
@@ -462,22 +463,17 @@ def emulate(self, instruction):
...
-    #Generic string representation
-    def __str__(self):
-        '''
-        Returns a string representation of cpu state
-        
-        :rtype: str
-        :return: name and current value for all the registers. 
-        '''
-        result = ""
+    def print_instruction(self):
```
yan: Since `print_instruction` and `print_registers` don't really print, can we rename them to something more descriptive, like `render_instruction`/`render_registers` or even `registers_str`?

yan: Or even log to `register_logger` directly within those functions, then you won't need to create a multi line str just to `split` it right after. If keeping as is, maybe return a list of strings instead of building a long one and splitting it later.

japesinator:
I decided against logging directly in the rendering functions so `__str__` wouldn't have to touch the logger

manticore/core/cpu/abstractcpu.py
```diff
@@ -470,26 +493,11 @@ def __str__(self):
...
-            else:
-                result += "%3s: %r"%(reg_name, value)
-            result += '\n'
-
-        return result
+        result =  self.render_instruction()
+        result += "\n"
+        for reg in self.render_registers():
+            result += reg
+            result += "\n"
```
yan: maybe:
```
result =  self.render_instruction() + '\n'
result += '\n'.join(self.render_registers())
```
?

Everything else looks good.

manticore/core/cpu/abstractcpu.py
```diff
@@ -470,26 +493,8 @@ def __str__(self):
...
-                result += aux
-            elif isinstance(value, (int, long)):
-                result += "%3s: 0x%016x"%(reg_name, value)
-            else:
-                result += "%3s: %r"%(reg_name, value)
-            result += '\n'
-
-        return result
+        result =  self.render_instruction() + "\n"
+        result += '\n'.join(self.render_registers())
```
yan: `result` isn't returned

PR #203, opened 2017-04-27
Add memory tracing

yan:
When verifying Manticore's emulation integrity, it is beneficial to run a binary in lock-step with a native or qemu-emulated execution. This works well for general instruction semantics, but fails when emulating system calls. Manticore's state might not exactly match that of the reference implementation.

We would like to be able to capture all of Manticore's memory writes between invoking a system call and its return, to mimic them precisely in the emulated instance.

This implementation lets calling code start and stop memory tracing. Stopping a memory trace returns a list of all write actions that occurred since it begun. Start/stop pairs can also be nested, if a future implementation requires it.

### DIFF
--- a/manticore/core/memory.py
+++ b/manticore/core/memory.py
@@ -217,6 +217,7 @@ class AnonMap(Map):
             return map(chr, self._data[index])
         return chr(self._data[index])
 
+
 class FileMap(Map):
     '''
     A file map.
@@ -809,15 +810,16 @@ class Memory(object):
 
     def write(self, addr, buf):
 
-        if self._recording_stack:
-            self._recording_stack[-1].append((addr, buf))
-
         size = len(buf)
         if not self.access_ok(slice(addr, addr + size), 'w'):
             raise MemoryException('No access writing', addr)
         assert size > 0
         stop = addr + size
         start = addr
+
+        if self._recording_stack:
+            self._recording_stack[-1].append((addr, buf))
+
         while addr < stop:
             m = self.map_containing(addr)
             size = min(m.end-addr, stop-addr)
--- a/manticore/core/memory.py
+++ b/manticore/core/memory.py
@@ -773,28 +773,28 @@ class Memory(object):
         return result
 
 
-    def start_write_trace(self):
+    def push_record_writes(self):
         '''
-        Begin recording all writes. Retrieve all writes with `stop_write_trace()`
+        Begin recording all writes. Retrieve all writes with `pop_record_writes()`
 
         :return: None
         '''
-        print '[d] starting trace..'
         self._recording_stack.append([])
 
-    def stop_write_trace(self):
+    def pop_record_writes(self):
         '''
         Stop recording trace and return a `list[(address, value)]` of all the writes
         that occurred, where `value` is of type list[str]. Can be called without
-        intermediate `stop_write_trace()`.
+        intermediate `pop_record_writes()`.
 
-        For example,
-            mem.start_write_trace()
+        For example::
+
+            mem.push_record_writes()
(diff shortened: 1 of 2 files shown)


mossberg:
can we document the use case for this in the pr description?

manticore/core/memory.py
```diff
@@ -771,13 +772,54 @@ def read(self, addr, size):
 
         return result
 
+
+    def start_write_trace(self):
+        '''
+        Begin recording all writes. Retrieve all writes with `stop_write_trace()`
+
+        :return: None
```
mossberg: should we document the return if there is none? I'd say we can leave it out.

manticore/core/memory.py
```diff
@@ -771,13 +772,54 @@ def read(self, addr, size):
...
+        '''
+        print '[d] starting trace..'
+        self._recording_stack.append([])
+
+    def stop_write_trace(self):
+        '''
+        Stop recording trace and return a `list[(address, value)]` of all the writes
+        that occurred, where `value` is of type list[str]. Can be called without
```
mossberg: is it necessarily `list[str]`? couldn't the list also contain Expressions?

manticore/core/memory.py
```diff
@@ -771,13 +772,54 @@ def read(self, addr, size):
...
         return result
 
+
+    def start_write_trace(self):
+        '''
+        Begin recording all writes. Retrieve all writes with `stop_write_trace()`
+
+        :return: None
+        '''
+        print '[d] starting trace..'
```
mossberg: stray print

manticore/core/memory.py
yan: Nope, I added trace support to `Memory.write`; `SMemory.write` records a list of symbols separately and only falls through to `super` if the value is concrete. ( https://github.com/trailofbits/manticore/blob/baa318e308f48b2205e3edbe5aa005025bf4dd18/manticore/core/memory.py#L1019 )

manticore/core/memory.py
```diff
@@ -771,13 +772,54 @@ def read(self, addr, size):
 
         return result
 
+
+    def start_write_trace(self):
```
mossberg: imo naming these in a way that's explicit about a stack being involved would make this, and client code more readable. Maybe `push_record_writes`/`pop_record_writes`

manticore/core/memory.py
```diff
@@ -771,13 +772,54 @@ def read(self, addr, size):
...
+        print '[d] starting trace..'
+        self._recording_stack.append([])
+
+    def stop_write_trace(self):
+        '''
+        Stop recording trace and return a `list[(address, value)]` of all the writes
+        that occurred, where `value` is of type list[str]. Can be called without
+        intermediate `stop_write_trace()`.
+
+        For example,
```
mossberg: if we ever sphinx generate this, i don't think the code will output right, i think a `::`  will specify a code block  http://www.sphinx-doc.org/en/stable/rest.html#source-code

tests/test_memory.py
```diff
@@ -1614,6 +1614,76 @@ def test_mmap_file_extra(self):
...
+        mem.start_write_trace()
+        mem.write(addr+2, 'c')
+        mem.write(addr+3, 'd')
+        inner_writes = mem.stop_write_trace()
+        outer_writes = mem.stop_write_trace()
+
+        # Make sure the first ones aren't in the inner write
+        self.assertNotIn((addr, ['a']), inner_writes)
+        self.assertNotIn((addr+1, ['b']), inner_writes)
+        # Make sure the first two are in the outer write
```
mossberg: this comment is wrong?

PR #205, opened 2017-04-27
Decode instruction early

yan:
This will let hooks access the `cpu.instruction` property

### DIFF
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -187,6 +187,7 @@ class Cpu(object):
         self._regfile = regfile
         self._memory = memory
         self._instruction_cache = {}
+        self._decoded_pc = None
         self._icount = 0
 
         self._md = Cs(self.arch, self.mode)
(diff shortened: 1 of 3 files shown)


manticore/core/cpu/abstractcpu.py
```diff
@@ -416,15 +428,15 @@ def canonicalize_instruction_name(self, instruction):
...
-        if not isinstance(self.PC, (int,long)):
-            raise SymbolicPCException()
+        '''
+        Decode, and execute one instruction pointed by register PC
+        '''
 
-        if not self.memory.access_ok(self.PC,'x'):
-            raise InvalidPCException(self.PC)
+        # Decode the instruction if it wasn't explicitly decoded
+        if self._decoded_pc != self.PC:
```
mossberg: technically this shouldn't happen right, this is just to be safe?

yan: yeah, this shouldn't happen. this can be an assert, i'm just protecting against someone doing `cpu.execute()` without explicitly decoding first.

mossberg: instead of creating a new class member just for this check, can we use `self.instruction.address` for this? `if self.instruction.address != self.PC`

manticore/core/cpu/abstractcpu.py
```diff
@@ -360,12 +360,22 @@ def _wrap_operands(self, operands):
 
     def decode_instruction(self, pc):
         '''
-        This will decode an instruction from memory pointed by @pc
+        This will decode an instruction from memory pointed by `pc` and store
+        it in self.instruction.
 
         :param int pc: address of the instruction
+        :return: None
```
mossberg: rm this? i think it's safe to assume if the docs don't specify a ret there is none

manticore/core/cpu/abstractcpu.py
```diff
@@ -360,12 +360,22 @@ def _wrap_operands(self, operands):
...
         '''
-        #No dynamic code!!! #TODO! 
-        #Check if instruction was already decoded 
+        #No dynamic code!!! #TODO!
+
+        # Make sure PC is concrete and we can access it
+        if not isinstance(pc, (int,long)):
```
mossberg: use issymbolic here?

PR #213, opened 2017-04-28
Update syscall/funcall implementation

yan:
We currently invoke syscalls in an ad-hoc way. This PR wraps system call and function call implementation into arch-specific classes.

Example invocation:
```
def strcmp(a, b):
  # extract strings
  # return cmp

@m.hook(addr_of_strcmp)
def invoke_strcmp(state):
    state.model.invoke_model(strcmp)
```

This PR also implements a common base class for platform models (`Platform`).


manticore/models/linux.py
```diff
@@ -1485,20 +1485,13 @@ def syscall(self):
...
                 }
 
-        index, arguments, writeResult  = self.current.get_syscall_description()
+        index = self.current.ABI.syscall_number()
 
         if index not in syscalls:
             raise SyscallNotImplemented(64, index)
 
-        func = syscalls[index]
+        return self.current.ABI.invoke_syscall(syscalls[index])
```
feliam: Wonder if it's better to decoreate all sys_xxxxx function methods with this so when you browse the class you can detect system calls implementations without relying in the sys_xxx name.

yan: We'd still need to dispatch somewhere though. I'm afraid decorators would fall into "doing too much in a PR". I'd like to decorate these with their numbers at minimum in a different PR.

manticore/core/cpu/arm.py
```diff
@@ -253,6 +253,42 @@ def canonical_registers(self):
...
 
+class Armv7ABI(ABI):
+    '''
+    ARMv7 ABI
+    '''
+    # EABI standards:
+    #  syscall # is in R7
+    #  arguments are passed in R0-R6
+    #  retval is passed in R0
+    # TODO(yan): Stack alignment
```
mossberg: what exactly does stack alignment mean here?

manticore/core/cpu/abstractcpu.py
```diff
@@ -162,7 +165,184 @@ def __contains__(self, register):
 
         :param register: a register name
         '''
-        return self._alias(register) in self.all_registers 
+        return self._alias(register) in self.all_registers
+
+
+class ABI(object):
```
mossberg: pedantic naming thing but for class `Cpu`, we only capitalized the first letter, do the same here to be consistent?

manticore/core/cpu/x86.py
```diff
@@ -5636,99 +5620,46 @@ def VZEROUPPER(cpu):
 ################################################################################
 #Calling conventions
 
-class ABI:
-    '''IA32 Calling conventions
-        https://en.wikipedia.org/wiki/X86_calling_conventions
+class AMD64ABI(ABI):
```
mossberg: I don't think `AMD64ABI` is quite accurate, i feel like this should be `SystemV(ABI)`. the Microsoft x64 ABI is different, for example: https://en.wikipedia.org/wiki/X86_calling_conventions#x86-64_calling_conventions.

manticore/core/cpu/x86.py
```diff
@@ -5744,6 +5675,7 @@ def __init__(self, memory, *args, **kwargs):
         :param machine:  machine code name. Supported machines: C{'i386'} and C{'amd64'}.
         '''
         super(AMD64Cpu, self).__init__(AMD64RegFile(aliases={'PC' : 'RIP', 'STACK': 'RSP', 'FRAME': 'RBP'},  ), memory, *args, **kwargs)
+        self._abi = AMD64ABI(self)
```
mossberg: There are multiple ABIs for x64; in order to support others than SystemV, I feel like this needs to be a ctor parameter, or not be a Cpu attribute at all.

manticore/core/cpu/x86.py
```diff
@@ -5837,6 +5769,42 @@ def XLATB(cpu):
         cpu.AL = cpu.read_int(cpu.RBX + Operators.ZEXTEND(cpu.AL, 64), 8)
 
 
+class I386ABI(ABI):
```
mossberg: this is actually Cdecl?

mossberg:
I'm also wondering if we should not represent both syscall and function ABIs in a single object. For example if we wanted to implement Cdecl and Fastcall ABI classes for x86, both of those share the same syscall ABI since its x86 Linux, so might need to be duplicated code. Although, this can be solved with a class hierarchy like `ABI` -> `LinuxX86(ABI)` (implement syscall stuff) -> `LinuxCdecl(LinuxX86)` (implement func call stuff)

yan:
Re: hierarchy, I've been wanting to create a hierarchy for other reasons also. (i.e. pushing args to stack looks identical in systemv and cdecl save for register sizes)

mossberg:
My current thoughts are that the ABI object should exist as a private attribute on the os model, and should do less than it does now. Rather than accessing cpu registers, it should specify what registers to access, and the model itself should do that.

manticore/models/platform.py
```diff
@@ -0,0 +1,13 @@
...
+import inspect
+
+class Platform(object):
+    '''
+    Base class for all operating system models.
+    '''
+    def __init__(self, path):
+        self._path = path
+
+    def invoke_model(self, implementation, prefix_args=None, varargs=False):
```
yan: Not 100% sure this is the best method name

manticore/core/cpu/x86.py
```diff
@@ -5636,99 +5620,102 @@ def VZEROUPPER(cpu):
...
+            yield reg
+
+    def write_result(self, result):
+        self._cpu.EAX = result
+
+class AMD64LinuxSyscallAbi(SyscallAbi):
+    '''
+    AMD64 Linux system call ABI
+
+    Floating point or wide arguments that deviate from the norm are not yet
```
mossberg: add a TODO comment here?

manticore/core/cpu/x86.py
```diff
@@ -5636,99 +5620,102 @@ def VZEROUPPER(cpu):
...
+    '''
+    def get_arguments(self):
+        bwidth = self._cpu.address_bit_size / 8
+        offset = self._cpu.ESP
+        while True:
+            offset += bwidth
+            yield offset
+
+    def write_result(self, result):
+        if result is not None:
```
mossberg: can we move this check out right before the call site? here i think it's a little unclear why it's necessary at first glance, but given the context of invoke() it becomes much more clear

manticore/core/cpu/x86.py
```diff
@@ -5636,99 +5620,102 @@ def VZEROUPPER(cpu):
...
+            self._cpu.EAX = result
+
+        self._cpu.EIP = self._cpu.pop(self._cpu.address_bit_size)
+        
+class I386StdcallAbi(I386CdeclAbi):
```
mossberg: i see how its useful to inherit like this for get_arguments, but i don't think this inheritance actually makes sense; stdcall and cdecl should be on the same level of the class hierarchy. since arm, x86 and x64 all need some sort of notion of popping args off the stack and there is code duplication for this anyway right now, would we be able to move that up into Abi class itself? remember, cpu.SP is aliased and will work on all cpus.

manticore/core/cpu/x86.py
```diff
@@ -5636,99 +5620,102 @@ def VZEROUPPER(cpu):
...
+        super(I386StdcallAbi, self).write_result(result)
+        bwidth = self._cpu.address_bit_size / 8
+        self._cpu.ESP += self._arguments * bwidth
+        self._arguments = 0
+
+class SystemVAbi(Abi):
+    '''
+    x64 SystemV function call convention
+
+    Floating point or wide arguments that deviate from the norm are not yet
```
mossberg: TODO comment?

manticore/core/cpu/x86.py
```diff
@@ -5636,99 +5620,102 @@ def VZEROUPPER(cpu):
...
+
+        bwidth = self._cpu.address_bit_size / 8
+        offset = self._cpu.RSP
+        while True:
+            offset += bwidth
+            yield offset
+
+    def write_result(self, result):
+        # XXX(yan): Can also return in rdx for wide values.
+        if result is not None:
```
mossberg: same as above, move this check to call site?

manticore/models/linux.py
```diff
@@ -1852,7 +1846,9 @@ def _mk_proc(self, arch):
             mem = SMemory32(self.constraints)
         else:
             mem = SMemory64(self.constraints)
-        return CpuFactory.get_cpu(mem, arch)
```
mossberg: any particular reason this was made more verbose?

manticore/models/windows.py
```diff
@@ -171,6 +174,8 @@ def __init__(self, path, additional_context = None, snapshot_folder=None):
                 self.running.append(self.procs.index(cpu))
         
 
+        self._function_abi = I386StdcallAbi(self.procs[0])
+        #self._syscall_abi =
```
mossberg: clean this comment?

manticore/core/cpu/x86.py
yan: How would this work? `write_result` does its namesake but it also handles `ret` semantics. We need to call it even if the implementation doesn't return anything.

mossberg: oh, i overlooked that. would you be opposed to not having `write_result` also `ret`? i think it makes sense; two are 2 distinct things.

manticore/core/cpu/abstractcpu.py
```diff
@@ -162,7 +165,111 @@ def __contains__(self, register):
...
+        :param tuple prefix_args: Pass these parametrs to implementation before those read from state.
+        :param bool varargs: Whether the function expects a variable number of arguments
+        :return: The result of calling `implementation`
+        '''
+        prefix_args = prefix_args or ()
+
+        spec = inspect.getargspec(implementation)
+
+        if spec.varargs:
+            logger.warning("ABI: Vararg models should implemented as taking a single parameter")
```
mossberg: typo "should implemented"

manticore/core/cpu/abstractcpu.py
```diff
@@ -162,7 +165,111 @@ def __contains__(self, register):
...
+        implementation receives a single argument that is a generator for
+        function arguments. Pass a tuple for `prefix_args` of arguments you'd
+        like to precede the actual arguments.
+
+        :param callable implementation: Python model of the function
+        :param tuple prefix_args: Pass these parametrs to implementation before those read from state.
+        :param bool varargs: Whether the function expects a variable number of arguments
+        :return: The result of calling `implementation`
+        '''
+        prefix_args = prefix_args or ()
```
mossberg: maybe `() if prefix_args is None else prefix_args`?

manticore/models/platform.py
mossberg: TODO rename "implementation" param to "model" or similar

manticore/core/cpu/x86.py
```diff
@@ -5636,99 +5620,106 @@ def VZEROUPPER(cpu):
...
+    # not yet supported.
+
+    def get_arguments(self):
+        # First 6 arguments go in registers, rest are popped from stack
+        reg_args = ('RDI', 'RSI', 'RDX', 'RCX', 'R8', 'R9')
+
+        for reg in reg_args:
+            yield reg
+
+        bwidth = self._cpu.address_bit_size / 8
```
mossberg: use values_from here?

Issue #212, opened 2017-04-28
will_execute_pc signal should only emit itself with 1 state argument


mossberg:
Currently we emit the will_execute_pc signal with both state and PC arguments, however since PC is always easily accessible from the state via state.cpu.PC, this can be refactored so only 1 argument is passed.

Issue #166, opened 2017-04-21
Implement __enter__/__exit__ protocol fully


feliam:
All objects in State need to implement a temporary copy on write mechanism over we could do a fork.

current_state = State()
with current_state as new_state:
             #new_state here most be a COW copy of current state.
             #every action over new_state must not affect current_state
             #current_state must be inmutable during this with-statement 
             #after we do operations over new_state we can save it or forget it.
#As after the with statement new_state disappears then current state can be mutable again 

Currently this is only implemented for the ConstraintSet and State (not all the objects used in State)

PR #214, opened 2017-04-29
refactored will_execute_pc hooks

ManasGeorge:
#212
Fixed assorted typos, an instruction in the the README (running `nosetests` in the toplevel directory without the --exe flag just ignored all of the tests for me, which seems to be default behavior when it encounters executable files, which the tests are). Also refactored the `will_execute_pc` Signal to take just the state as argument, and changed the hooks that call it to accept just the state and derive the program counter from the state.


ManasGeorge:
[This line](https://github.com/ManasGeorge/manticore/blob/54eaa0e0819fb195a8191b927cc2c8ee375e5896/manticore/core/executor.py#L750) in `executor.py` references `current_state.PC`, which throws an exception because `current_state` has no member `PC`. Because of the new exception, the rest of the code in the exception handling block is not executed, so you have fewer states expanded as children. Fixing the presumed typo to `current_state.cpu.PC` causes more states to be expanded as the previously ignored code gets run, and this causes the `simple_parse` test in `test_memdumps` to fail.

README.md
```diff
@@ -97,7 +97,7 @@ You can run the tests with the commands below:
 ```
 cd manticore
 # all tests
-nosetests
+nosetests --exe
```
mossberg: I'm on nosetests 1.3.7 and when i run it in the top level, it actually does run the tests for me! Also, my test files do not have the execute bit on them. i'm not sure what the issue is; could we omit this change for the time being?

manticore/core/executor.py
```diff
@@ -760,7 +760,7 @@ def setstate(state, val):
                     with current_state as new_state:
                         children.append(new_state.co)
                         new_state.add(e.constraint)
-                        self.newerror(new_state.current.PC)
+                        self.newerror(new_state.model.current.PC)
```
mossberg: good catch! instead, can we use `new_state.cpu.PC`?

manticore/manticore.py
```diff
@@ -577,7 +578,8 @@ def terminate(self):
         '''
         self._executor.shutdown()
 
-    def _assertions_callback(self, state, pc):
+    def _assertions_callback(self, state ):
```
mossberg: nit, there is a space after `state`, let's rm it

README.md
ManasGeorge: Sure thing, I'll take it out. I tried cloning the repo again, and I still have the tests being executable; it's probably an OS-specific thing. Do you think it makes sense to add a note saying maybe try the flag if the tests don't seem to run, or make sure that the tests are set to non executable?

manticore/core/executor.py
ManasGeorge: Will do! That is cleaner. Why do we use the the with-statement to make forked states, instead of a fork method or something?

ManasGeorge:
I've made the changes you asked for, thanks for reviewing it so quickly! I was wondering;
- Would it be appropriate to add a remark in the README about the executable bit on the test files? I just cloned the repository fresh and they're all still marked executable, it's probably an OS thing. It's not super important, it'd just save a couple seconds of Googling around to figure out what's going on.
- Why do we use the with-statements as a way to make forked States, as opposed to a `fork()` method?

mossberg:
- I just checked the perms of those files as git sees them, and it seems like there is no execute bit. (also did a fresh clone). Maybe you have something configured to automatically mark python files as executable? I'd say let's omit a readme remark for now.

```
[I] mark forge /p/t/m/tests (master) ❯ git ls-tree -r --full-tree HEAD tests|ag py
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391	tests/__init__.py
100644 blob e545a3a55a9c00879d9ffe16eb10765f365713bf	tests/auto/flags.py
100644 blob 17d7f259f17b5fe40f5008ed86fe62372f76258a	tests/auto/make_dump.py
100644 blob d7ab7665f310e15a80e176bf8d022666b5604777	tests/auto/make_tests.py
100644 blob b8198311ef796608c373b62e120afcf6b8ee1c64	tests/auto/trace.py
100644 blob 7115bd0c186beb5aadafefb0dcbb9d3db839ed14	tests/mockmem.py
100644 blob 33d57f955bf797f062d29a972ec4895f542b7c44	tests/test_allinsts.py
100644 blob 92b3c6c104a952ca45b761bd371feac2d787df19	tests/test_armv7_bitwise.py
100644 blob 9b1d09a6ebdce8a2b0b33258a8caf26666449743	tests/test_armv7cpu.py
...
```

- our forking with the `with` statement is a somewhat experimental interface for some copy on write ideas we had. it's something we've been discussing a lot, and is definitely subject to change! if you have any ideas, we'd love to hear them.

ManasGeorge:
- Oh, I see what the issue is; it looks like PyCharm went ahead and marked them as executable; I cloned it from within PyCharm both times. Doing it on the command line works as expected.
- Ah, that's pretty cool. That should really help cut down on crazy bloat with lots of forking. Where is this discussion happening? #166 is the only relevant issue I see.

mossberg:
Discussion has mostly been in our internal chat, and yup that issue is the only one that really represents this. I definitely agree that the `with` is not a very intuitive way to do this!

PR #217, opened 2017-05-01
Implemented sys_readv system call

saelo:


manticore/models/linux.py
```diff
@@ -1288,6 +1288,52 @@ def sys_getegid(self, cpu):
...
+
+        :param cpu: current CPU.
+        :param fd: the file descriptor of the file to read.
+        :param iov: the buffer where the the bytes to read are stored.
+        :param count: amount of C{iov} buffers to read from the file.
+        :return: the amount of bytes read in total.
+        '''
+        total = 0
+        for i in xrange(0, count):
+            buf = cpu.read_int(iov + i * 16, 32)
```
mossberg: on 32 bit, sizeof(struct iovec) is only 8, so i don't think this math is right here. also, to reduce copy-pasted code, since these two syscall implementations seem identical except for the address math, can we have a single helper function that takes an `is64bit` bool and uses that to decide what math to do?

saelo: Oops! Yes I definitely agree about reducing copy+pasted code. While I'm at it, I can also fix sys_writev. What about using cpu.address_bit_size though? It should also be possible to write a method that takes the bitness as argument and returns an implementation. Would probably look like this in the syscall table: `0xXXXX: self.sys_readv(bits=32)`

mossberg: I like the sound of that!

PR #232, opened 2017-05-04
Formalize API for constraining a State

### LINKED ISSUE #195: Formalize API for adding constraints to State
Currently, we informally have a `State.add` API that is used internally for adding constraints onto a State. It can incidentally be used as a public API, but this was never decided deliberately, so I think we should take some time to actually think if there's anything that should be improved.

imo, it's good except, should be renamed to `State.constrain`.

mossberg:
Fixes #195 

This includes `state.constrain` as an official, documented API. This also adds a CHANGELOG.md to create a single place users can learn about important changes. It essentially copies http://keepachangelog.com/en/0.3.0/

### DIFF
--- a/manticore/core/executor.py
+++ b/manticore/core/executor.py
@@ -560,8 +560,7 @@ class Executor(object):
 
         children = []
         if len(vals) == 1:
-            constraint = symbolic == vals[0]
-            current_state.constrain(constraint)
+            current_state.constrain(symbolic == vals[0])
             setstate(current_state, vals[0])
             #current_state._try_simplify()
         else:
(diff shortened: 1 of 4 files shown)


manticore/core/executor.py
```diff
@@ -561,7 +561,7 @@ def fork(self, current_state, symbolic, vals, setstate=None):
         children = []
         if len(vals) == 1:
             constraint = symbolic == vals[0]
-            current_state.add(constraint, check=True) #We already know it's sat
+            current_state.constrain(constraint)
```
yan: inline the constraint formula? it's inlined in all other invocations and with new naming, reads a tad redundant.

PR #234, opened 2017-05-05
Manticore 0.1.1

mossberg:


mossberg:
oops @ branch name...should've been dev-0.1.1

PR #235, opened 2017-05-06
Added getrlimit syscall

jtang613:
Added getrlimit for x86.
Minor fix for typo in SyscallNotImplemented().

### DIFF
--- a/manticore/models/linux.py
+++ b/manticore/models/linux.py
@@ -9,7 +9,6 @@ from ..core.cpu.abstractcpu import Interruption, Syscall, ConcretizeRegister
 from ..core.cpu.cpufactory import CpuFactory
 from ..core.memory import SMemory32, SMemory64, Memory32, Memory64
 from ..core.smtlib import Operators, ConstraintSet
-from ..models.platform import Platform
 from elftools.elf.elffile import ELFFile
 import logging
 import random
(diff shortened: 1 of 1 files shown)


jtang613:
This provides a close-enough approximation - the 'resource' module will return limits for the python process rather than the process under test. But since these values are typically set at a system or user level, this will still result in sane values for the 90% use-case.

yan:
@jtang613 Looks like `rlim_t` is arch-specific. Can you rewrite your `write_int`s to be dependent on the word width i.e.

```        
word_width = self.current.address_bit_size / 8
self.current.write_int(rlim, rlimit[0], self.current.address_bit_size)
self.current.write_int(rlim+word_width, rlimit[1], self.current.address_bit_size)
```

jtang613:
Thanks for the feedback, changes committed.

yan:
Looks like the last commit touched a lot of lines that aren't related to `getrlimit`. Can you clean up the PR? Other than that, ready to merge.

jtang613:
Yikes! That was unexpected... one moment.

jtang613:
So this seems to have gone sideways on me in a bad way and my git fu is not at a level where I can make it sane again (not really sure where all the extra changes came in).  Mind if I blow this one away a create a new pull from a clean branch?

PR #241, opened 2017-05-09
Decouple functionality via events/signals

feliam:
WIP
* Executor.py refactor
* State search policy decoupled to its own class. Policy() So far only Random() is implemented
* Executor fully equipped with events. Main responsibilities: 
     - backup/restore states from _storage_
     - assign id numbers to new states and testcases
     - report all kid of events via signals
* abstractcpu now signals decode and execute as separated events
* Manticore object now needs to hook on some of the new events to generate testcases and others
* mappings.py moved to utils were it belongs. Yes mappings you belong there.
* events.py now handles forwarding/disable/enable of signals

Things to watch for:
 - Exceptions where are defined, where raised etc
 - Signals, propagation and re-entrance? add/del callbacks in callbacks?
 - Concurrence State.context vs Executor.context
 - If you don't know why is that there it needs doc or deletion


manticore/core/cpu/x86.py
```diff
@@ -5620,6 +5633,102 @@ def VZEROUPPER(cpu):
 ################################################################################
 #Calling conventions
 
+class ABI:
```
mossberg: it looks like the ABI abstraction refactor was undone here, and things are being added back?

manticore/core/cpu/abstractcpu.py
```diff
@@ -293,13 +352,15 @@ def syscall_number(self):
         '''
         raise NotImplementedError
 
+from ...utils.event import Signal
+
 ############################################################################
-# Abstract cpu encapsulating common cpu methods used by platforms and executor.
+# Abstract cpu encapsulating common cpu methods used by models and executor.
```
mossberg: seem "platforms" have been replaced by "models" in some places they shouldn't?

manticore/core/cpu/abstractcpu.py
```diff
@@ -334,7 +409,7 @@ def __getstate__(self):
     def __setstate__(self, state):
         Cpu.__init__(self, state['regfile'], state['memory'])
         self._icount = state['icount']
-        return
+        return
```
mossberg: nit, space

manticore/core/cpu/abstractcpu.py
```diff
@@ -324,6 +385,20 @@ def __init__(self, regfile, memory):
...
+        #####################################################
+        # Signals
+        # signal handlers must have this signature:
+        # handler(cpu, *args, **kwargs)
+        self.will_decode_instruction = Signal()
+        self.will_execute_instruction = Signal()
+        self.did_execute_instruction = Signal()
+        self.will_emulate_instruction = Signal()
+        self.did_emulate_instruction = Signal()
+        self.will_read_register = Signal()
```
mossberg: it seems like some of these are unused? can have documented use cases before adding these in?

manticore/core/cpu/abstractcpu.py
```diff
@@ -479,30 +559,21 @@ def read_bytes(self, where, size):
...
     #######################################
     # Decoder
-    @abstractmethod
     def _wrap_operands(self, operands):
         '''
         Private method to decorate a capstone Operand to our needs. See Operand
         class
         '''
-        pass
+        raise NotImplemented
```
mossberg: should be NotImplementedError

manticore/core/cpu/abstractcpu.py
```diff
@@ -479,30 +559,21 @@ def read_bytes(self, where, size):
...
 
     def decode_instruction(self, pc):
         '''
-        This will decode an instruction from memory pointed by `pc` and store
-        it in self.instruction.
+        This will decode an instruction from memory pointed by @pc
```
mossberg: was this intentional? i think backticks is the reST syntax?

manticore/core/cpu/abstractcpu.py
```diff
@@ -518,7 +589,11 @@ def decode_instruction(self, pc):
                     if isinstance(c, Constant):
                         c = chr(c.value)
                     else:
+                        logger.info('Trying to execute instructions from invalid memory')
```
mossberg: doesn't the existing logger call here already convey basically this information?

manticore/core/executor.py
```diff
@@ -444,401 +382,172 @@ def generate_testcase(self, state, message = 'Testcase generated'):
...
-                    if current_state:
-                        assert solver.check(current_state.constraints)
-
-                    #execute until exception or finish or max consecutive instructions without exception 
-                    max_iters = self.dump_every + count
-
-                    # allow us to terminate manticore processes
-                    while not self.isShutdown():
-                        # Make sure current instruction is decoded so that hooks can access it
-                        current_state.cpu.decode_instruction(current_state.cpu.PC)
```
mossberg: does this undo https://github.com/trailofbits/manticore/pull/205? should it?

manticore/core/cpu/abstractcpu.py
```diff
@@ -570,27 +646,46 @@ def fallback_to_emulate(*operands):
...
+        self.instruction = instruction
         implementation(*instruction.operands)
+        self.instruction = None
         self._icount+=1
 
+        #broadcast event
+        self.did_execute_instruction(instruction)
+
+
+    def get_syscall_description(self):
```
mossberg: this was mistakenly added back?

manticore/core/executor.py
```diff
@@ -69,184 +46,276 @@ def newFunction(self, *args, **kw):
...
-        assert policy in ['random','adhoc','uncovered','dicount','icount','syscount','depth','bucket']
-        self.policy = policy
-
-        self._states = manager.dict()
-        self._test_number = manager.Value('i', 0 )
+        # Signals / Callbacks handlers will be invoked potentially at differeent 
+        # worker processes. State provides a local context to save data.
+
+        #Executor
+        self.will_start_run = Signal()
```
mossberg: are we going a little signal crazy here? a lot of these seem unused. again, i think we should have documented use cases first

manticore/core/executor.py
```diff
@@ -69,184 +46,276 @@ def newFunction(self, *args, **kw):
...
+            It needs a lock. Its used like this:
+
+            with executor.context() as context:
+                vsited = context['visited']
+                visited.append(state.cpu.PC)
+                context['visited'] = visited
+        '''
+        with self._lock:
+            yield self._shared_context
+        #FIXME What if we do a:
```
mossberg: @feliam when you have this kind of fixme thing where you're unsure, can you bring it up explicitly? maybe mention it in the PR description? or in a comment? or slack message?

manticore/manticore.py
```diff
@@ -16,25 +16,47 @@
...
+            elif func_name == 'backup':
+                self.saving_time += func_time
+            elif func_file.endswith('solver.py') and 'setstate' not in func_name and 'getstate' not in func_name and 'ckl' not in func_name:
+                self.solver_time += func_time
+
+
+
 def makeDecree(args):
     constraints = ConstraintSet()
-    platform = decree.SDecree(constraints, ','.join(args.programs))
```
mossberg: mistaken renames?

manticore/core/executor.py
feliam: will_execute_instruction signal now is called only when the instruction was successfully decoded. 
So _I think_ we don't need to force decoding here.

manticore/core/executor.py
feliam: I'm thinking maybe a gui that wants to report what's happening. But yeah! Most Signals around are     a request for comment basically. ( There are several workers that may start or not ). This is definitively a candidate for deletion.

manticore/utils/emulate.py
```diff
@@ -218,11 +218,12 @@ def _step(self, instruction):
             val = self._cpu.read_register(reg)
             if issymbolic(val):
                 from ..core.cpu.abstractcpu import ConcretizeRegister
-                raise ConcretizeRegister(reg, "Concretizing for emulation.",
+                raise ConcretizeRegister(self._cpu, reg, "Concretizing for emulation.",
                                          policy='ONE') 
             self._emu.reg_write(self._to_unicorn_id(reg), val)
 
         # Bring in the instruction itself
+        instruction = self._cpu.decode_instruction(self._cpu.PC)
```
mossberg: also `emulate_next` in test_unicorn.py should be updated

manticore/manticore.py
```diff
@@ -505,7 +537,189 @@ def load_assertions(self, path):
...
                 self._assertions[pc] = ' '.join(line.split(' ')[1:])
 
+    def _backup_state_callback(self, state, state_id):
+        logger.debug("Backup state %r", state_id)
+
+    def _restore_state_callback(self, state, state_id):
+        logger.debug("Restore state %r", state_id)
+
+    def _terminate_state_callback(self, state, state_id, ex):
+        print "Terminate state", state.context
```
mossberg: rm debug print?

manticore/models/system.py
```diff
@@ -0,0 +1,22 @@
+class OSException(Exception):
```
mossberg: what is this file for? models/ is for function models, so should it be somewhere else?

manticore/platforms/decree.py
```diff
@@ -15,13 +15,15 @@
 import logging
 import random
 
-logger = logging.getLogger("PLATFORM")
+logger = logging.getLogger("MODEL")
```
mossberg: can we fix all these model/platform naming issues?

manticore/manticore.py
```diff
@@ -533,32 +747,92 @@ def run(self, procs=1, timeout=0):
...
+        self._executor.will_decode_instruction += self._decode_instruction_callback
+        self._executor.will_backup_state += self._backup_state_callback
+        self._executor.will_restore_state += self._restore_state_callback
+        self._executor.will_fork_state += self._fork_state_callback
+        self._executor.will_terminate_state += self._terminate_state_callback
+        self._executor.will_generate_testcase += self._generate_testcase_callback
+
         if self._hooks:
-            self._executor.will_execute_pc += self._hook_callback
+            self._executor.will_execute_state += self._hook_callback
```
mossberg: executor.will_execute_state is never declared so this breaks hooking

manticore/core/cpu/abstractcpu.py
```diff
@@ -479,89 +554,98 @@ def read_bytes(self, where, size):
...
+        if not self.memory.access_ok(self.PC,'x'):
+            raise InvalidMemoryAccess(self.PC, 'x')
+
+        #broadcast event
+        self.will_decode_instruction()
+
+        instruction = self.decode_instruction(self.PC)
+
+        #broadcast event
+        self.will_execute_instruction(instruction)
```
mossberg: is this the signal that hooks will be fired using? we need to set self.instruction before this so they can access the current instruction?

manticore/manticore.py
```diff
@@ -505,7 +537,189 @@ def load_assertions(self, path):
...
+        state_visited = state.context.get('visited', set())
+        state_instructions_count = state.context.get('instructions_count', 0)
+        with self._executor.locked_context() as context:
+            executor_visited = context.get('visited', set())
+            context['visited'] = executor_visited.union(state_visited)
+
+            executor_instructions_count = context.get('instructions_count', 0)
+            context['instructions_count'] = executor_instructions_count + state_instructions_count 
+
+    def _fork_state_callback(self, state, expression, values, policy):
```
mossberg: imo, we should delete all these callbacks and their signals, if these stub callbacks are the only slots connected to them. that is, until we have a use case for them.

manticore/manticore.py
```diff
@@ -505,7 +537,189 @@ def load_assertions(self, path):
...
+
+
+    def _emulate_instruction_callback(self, state, cpu, instruction):
+        logger.debug("About to emulate instruction")
+
+    def _did_execute_instruction_callback(self, state, cpu, instruction):
+        logger.debug("Did execute an instruction")
+
+    def _execute_instruction_callback(self, state, cpu, instruction):
+        logger.info("exe\n")
```
mossberg: i don't think this is a useful logging statement?

manticore/manticore.py
```diff
@@ -505,7 +537,189 @@ def load_assertions(self, path):
...
+        address = state.cpu.PC
+        if not issymbolic(address):
+            state.context.setdefault('visited', set()).add(address)
+            count = state.context.get('instructions_count', 0)
+            state.context['instructions_count'] = count + 1
+
+
+    def _generate_testcase_callback(self, state, testcase_id, message = 'Testcase generated'):
+        #Fixme split this!
```
mossberg: we've been using a convention of `FIXME(name)`, can we do that here?

manticore/platforms/linux.py
```diff
@@ -1397,7 +1394,7 @@ def sys_exit_group(self, error_code):
         #self.procs[procid] = None
         logger.debug("EXIT_GROUP PROC_%02d %s", procid, error_code)
         if len(self.running) == 0 :
-            raise ProcessExit(error_code)
+            raise Exception('Process exited correctly. Code: {}'.format(error_code))
```
mossberg: why are we using a regular exception here instead of ProcessExit?

manticore/platforms/platform.py
```diff
@@ -2,6 +2,29 @@
 from itertools import islice, imap
 import inspect
 
+class OSException(Exception):
```
mossberg: this is duplicated in models/system.py?

manticore/platforms/windows.py
```diff
@@ -48,9 +47,9 @@ def toStr(state, value):
             value = minmax[0]
     return '{:08x}'.format(value)
 
-class Windows(Platform):
+class Windows(object):
```
mossberg: revert these changes too? windows should be a platform?

manticore/models/system.py
feliam: This system.py is a desperate attempt to abstract the operating system models. Which think its now .. platform. So in the case that exception is actually used it should go somewhere in platform/. tbc..

manticore/platforms/platform.py
feliam: bye bye system.py.

manticore/core/cpu/abstractcpu.py
```diff
@@ -387,7 +458,9 @@ def read_register(self, register):
         :return: register value
         :rtype: int or long or Expression
         '''
-        return self._regfile.read(register)
+        value = self._regfile.read(register)
+        self.will_read_register(register, value)
```
yan: Sort of inaccurate, `will_read_register` should execute before the actual `read`. (i.e. if `read` raises, the signal won't fire)

manticore/core/cpu/abstractcpu.py
```diff
@@ -1,21 +1,76 @@
...
+        self.bytes=bytes
+
+class InstructionNotImplementedError(CpuException):
+    '''
+    Exception raised when you try to execute an instruction that is not yet
+    implemented in the emulator. Add it to the Cpu-specific implementation.
+    '''
+    pass
+
+class DivideError(CpuException):
```
yan: Rename to `DivideByZeroError`?

manticore/core/cpu/abstractcpu.py
```diff
@@ -479,89 +554,98 @@ def read_bytes(self, where, size):
...
+            if issymbolic(c):
+                assert isinstance(c, BitVec) and  c.size == 8
+                if isinstance(c, Constant):
+                    c = chr(c.value)
+                else:
+                    logger.error('Concretize executable memory %r %r', c, text )
+                    raise ConcretizeMemory(self.memory, address = pc,
+                                            size = 8 * self.max_instr_width, 
+                                            policy = 'INSTRUCTION' )
+            assert isinstance(c, str)
```
yan: What else can `c` be here?

manticore/core/cpu/abstractcpu.py
```diff
@@ -479,89 +554,98 @@ def read_bytes(self, where, size):
...
-        instruction = next(self._md.disasm(code, pc))
 
-        #PC points to symbolic memory 
-        if instruction.size > len(text):
-            logger.info("Trying to execute instructions from invalid memory")
-            raise InvalidPCException(pc)
+        #decode the instructtion from code 
+        try:
+            instruction = next(self._md.disasm(code, pc))
+        except Exception as e:
```
yan: Should we catch only Capstone exceptions here?

manticore/core/executor.py
```diff
@@ -69,184 +46,271 @@ def newFunction(self, *args, **kw):
...
+        ''' Executor context is a shared memory object. All workers share this. 
+            It needs a lock. Its used like this:
+
+            with executor.context() as context:
+                vsited = context['visited']
+                visited.append(state.cpu.PC)
+                context['visited'] = visited
+        '''
+        with self._lock:
+            yield self._shared_context
```
yan: What is the `_lock` protecting?

manticore/manticore.py
feliam: Yes! also if not connected the signals wont propagate and stuff would be faster.
Not sure what would be the manticore API to hook on the different signals.

manticore/platforms/linux.py
feliam: Yeah :( .. ideas?
Import and raise state.TerminateState ?
Make a PlatformException for all the platforms and raise that?

manticore/core/cpu/abstractcpu.py
feliam: hmm will/did then ?
Only did_read_register ?
Or on_register_read?
I'm lost with the naming scheme here.

manticore/core/cpu/abstractcpu.py
feliam: Ideally it can't be anything else. 
But it could be an Expression if the concretizing code above failed. 
If that's double checked and we are certain I may just remove this.

manticore/core/executor.py
feliam: So _shared_context is shared between all running workers.
If we grant access to one without a lock and the callback at that worker do some non trivial read writes it may lead to a race. no? (there is already an example hook using shared_context at manticore)

manticore/core/executor.py
```diff
@@ -69,184 +46,271 @@ def newFunction(self, *args, **kw):
...
+                saved_states.append(self._getFilename(filename)) 
+        
+        #We didn't find any saved intermediate states in the workspace
+        if not saved_states:
+            return False
+
+        #search finalized testcases
+        saved_testcases = []
+        for filename in os.listdir(self.workspace):
+            if filename.startsswith('test_') and filename.endswith('.pkl'):
```
yan: Typo? (`startsswith`)

manticore/core/executor.py
```diff
@@ -69,184 +46,271 @@ def newFunction(self, *args, **kw):
...
+                saved_testcases.append(self._getFilename(filename)) 
+
+        if saved_states or saved_testcases:
+            #We are trying to continue a paused analysis
+            if initial is not None:
+                raise Exception('Cant set initial state when continuing from previous workspace: {}'.format(workspace))
```
yan: Raise a more specific exception?

manticore/core/executor.py
```diff
@@ -69,184 +46,271 @@ def newFunction(self, *args, **kw):
...
+            logger.debug("Waiting for available states")
+            self._lock.wait()
+            
+        state_id = random.choice(self._states)
+        del  self._states[self._states.index(state_id)]
+        return state_id
+
+    ###############################################################
+    # File Storage 
+    def backup(self, state):
```
yan: What is state backup?

feliam: Its when you save a State to the secondary storage(in our current implementation is a pickled file on the workspace folder).    backup/restore

manticore/core/cpu/abstractcpu.py
feliam: There is a CsError. But I couldn't make it throw it. 
I got 
StopIteration, ctypes.ArgumentError/TypeError


felipe@localhost /tmp $ python ssss.py 
  File "ssss.py", line 7, in <module>
    i = next( md.disasm(CODE, "SAS"))
  File "/home/felipe/.local/lib64/python2.7/site-packages/capstone/__init__.py", line 849, in disasm
    res = _cs.cs_disasm(self.csh, code, len(code), offset, count, ctypes.byref(all_insn))
ctypes.ArgumentError: argument 4: <type 'exceptions.TypeError'>: wrong type
felipe@localhost /tmp $ python ssss.py 
  File "ssss.py", line 7, in <module>
    i = next( md.disasm(2, 0x1000))
  File "/home/felipe/.local/lib64/python2.7/site-packages/capstone/__init__.py", line 849, in disasm
    res = _cs.cs_disasm(self.csh, code, len(code), offset, count, ctypes.byref(all_insn))
TypeError: object of type 'int' has no len()
felipe@localhost /tmp $ python ssss.py 
  File "ssss.py", line 7, in <module>
    i = next( md.disasm(CODE, 0x1000,-1))
StopIteration

manticore/core/cpu/abstractcpu.py
feliam: This is how the callback for this signal will look like at manticore level. 
   def _execute_instruction_callback(self, state, platform, cpu, instruction):
        logger.info("exe\n")

Instruction will be available as an argument and as cpu.intruction both.

manticore/core/cpu/abstractcpu.py
feliam: All signals here are at least emitted. 
A use case for the read/write-register/memory could be: report the next time you "see" a tainted value. When a tainted value is read from memory or register.

When we emulate an instruction we lost read/write-register//memory so I assume that a user would want to know when those are not available. How? hooking on will_emulate/did_emulate

manticore/core/cpu/abstractcpu.py
```diff
@@ -293,13 +347,15 @@ def syscall_number(self):
         '''
         raise NotImplementedError
 
+from ...utils.event import Signal
```
yan: Why not import during the imports at the beginning of file?

manticore/core/cpu/abstractcpu.py
```diff
@@ -293,13 +347,15 @@ def syscall_number(self):
...
+
 ############################################################################
 # Abstract cpu encapsulating common cpu methods used by platforms and executor.
 class Cpu(object):
     '''
     Base class for all Cpu architectures. Functionality common to all
     architectures (and expected from users of a Cpu) should be here. Commonly
-    used by platforms and py:class:manticore.core.Executor
+    used by plaforms and py:class:manticore.core.Executor
```
yan: typo

manticore/core/cpu/abstractcpu.py
```diff
@@ -446,9 +534,13 @@ def read_int(self, where, size=None):
         if size is None:
             size = self.address_bit_size
         assert size in SANE_SIZES
+        self.will_read_memory(where, size)
+
         data = self.memory[where:where+size/8]
         total_size = 8 * len(data)
```
yan: When will `len(data)*8` not equal `size`?

manticore/core/cpu/abstractcpu.py
```diff
@@ -570,27 +676,40 @@ def fallback_to_emulate(*operands):
...
         if logger.level == logging.DEBUG :
             logger.debug(self.render_instruction())
             for l in self.render_registers():
                 register_logger.debug(l)
-
+        
         implementation(*instruction.operands)
         self._icount+=1
 
+        #broadcast event
```
yan: i'd remove all the 'broadcast event' comments, as the naming convention is pretty clear and they add significant extra lines, imho. (also okay with leaving them in)

manticore/core/cpu/abstractcpu.py
```diff
@@ -605,7 +724,15 @@ def render_instruction(self):
 
     def render_register(self, reg_name):
         result = ""
-        value = self.read_register(reg_name)
+
+        self.will_read_register.disable()
```
yan: `render_register` is only called from `render_registers`. Move the disable/renable stuff there, otherwise we're disabling/reenabling all the signals a ton unnecessarily.

manticore/core/cpu/abstractcpu.py
```diff
@@ -630,114 +761,13 @@ def __str__(self):
...
-
-
 #Instruction decorators
 def instruction(old_method):
     #This should decorate every instruction implementation
     @wraps(old_method)
     def new_method(cpu, *args, **kw_args):
         cpu.PC += cpu.instruction.size
-        return old_method(cpu,*args,**kw_args)
+        value = old_method(cpu,*args,**kw_args)
```
yan: Placeholder from old code? why not return inline?

manticore/core/cpu/abstractcpu.py
```diff
@@ -479,89 +571,103 @@ def read_bytes(self, where, size):
...
-                    else:
-                        logger.error('Concretize executable memory %r %r', c, text )
-                        break
-                assert isinstance(c, str)
-                text += c
-        except MemoryException:
+                text += _read_concrete_byte(pc+i)
+        except MemoryException as mem_except:
+            #lets not raise this exception in case the instruction
+            #was fully read befor the page
```
yan: Perhaps rewrite as:
```
try:
    # Read Instruction from memory
    for i in xrange(0, self.max_instr_width):
        try:
            text += _read_concrete_byte(pc+i)
        except MemoryException:
            # In case of reading into unmapped memory
            break
```
Otherwise, we throw+except for every byte until `max_instr_width`.

manticore/platforms/decree.py
```diff
@@ -88,10 +91,10 @@ class Decree(object):
 
     def __init__(self, programs):
         '''
-        Builds a Decree OS platform
-        :param cpus: CPU for this platform.
-        :param mem: memory for this platform.
-        :todo: generalize for more CPUs.
+        Builds a Decree OS
+        :param cpus: CPU for this platform
+        :param mem: memory for this oplatform
```
yan: typo?

manticore/manticore.py
```diff
@@ -469,20 +497,24 @@ def apply_model_hooks(self, path):
...
         import core.cpu
         import importlib
-        import platforms
+        import models
 
         with open(path, 'r') as fnames:
             for line in fnames.readlines():
                 address, cc_name, name = line.strip().split(' ')
-                fmodel = platforms
+                cc = getattr(core.cpu.x86.ABI, cc_name)
```
yan: A lot of the code here seems reverted to much older code, including naming above (`platform` back to `model`)

manticore/utils/event.py
```diff
@@ -1,13 +1,51 @@
...
+
+def forward_signals(dest, source):
+    ''' 
+        Replicate and forward all the signals from source to dest
+    '''
+    #Import all signals from state
+    for signal_name in dir(source):
+        if not signal_name.startswith('will') and \
+            not signal_name.startswith('did') and \
+            not signal_name.startswith('on'):
```
yan: This might catch a lot of unintended fields. Isn't this unnecessary with the `isinstance` below?

manticore/manticore.py
```diff
@@ -505,7 +537,193 @@ def load_assertions(self, path):
...
+        logger.debug("Decoding stuff instruction not available")
+
+
+    def _emulate_instruction_callback(self, state, platform, cpu, instruction):
+        logger.debug("About to emulate instruction")
+
+    def _did_execute_instruction_callback(self, state, platform, cpu, instruction):
+        logger.debug("Did execute an instruction")
+
+    def _execute_instruction_callback(self, state, platform, cpu, instruction):
```
mossberg: i'm not sure I like how each layer of signal forwarding adds on an extra parameter to the slot/callback. it seems like it could be redundant, for example, given that platform, cpu and instruction and all accessible from the state object?

i also think the way `forward_signal` setattrs makes the code a lot harder to read since it is totally unclear how that signal attribute exists on the class otherwise.

mossberg: would it be possible to have some sort of global event system rather than a hierarchical chain?

manticore/core/cpu/abstractcpu.py
feliam: 100 != len("ABC"[0:100])

manticore/core/cpu/abstractcpu.py
feliam: That's it I'll remove the cli/sti from there. Lets document it and force the user to do it explicitly for now.  I'm not really fixing anything fully with disabling it just here.

manticore/manticore.py
feliam: Forwarding Objects are saved internally at Signal anyway. We could omit them in them in the emit()
We can add an extra keyword argument to Signal.when() to control if the forwarding object has to interpose itself in the callback arguments or not.  
Idea is to have only state + specific arguments at the callback, right?

---- 
For the global event system. You mean that every object holds a reference to .events of type Events()
class Events():
   register_new_signal(str, desc)
   register_callback(str, obj, method)
   list_signals()
   disable()
   enable()

?

manticore/core/cpu/abstractcpu.py
```diff
@@ -144,16 +146,7 @@ def size(self):
...
-    def type(self):
-        ''' This property encapsulate the operand type. 
-            It may be one of the following:
-                register
-                memory
-                immediate
-        '''
-        raise NotImplementedError
-        
+    @abstractmethod
```
feliam: Better not use ABC. So we do not have the overhead of checking in each invocation.

manticore/core/cpu/abstractcpu.py
```diff
@@ -144,16 +146,7 @@ def size(self):
     def reg(self):
         return self._reg_name(self.op.reg)
 
-    @property
-    def type(self):
```
feliam: Better to document in the abstract class so implementations overload this?

PR #244, opened 2017-05-09
Add infrastructure for core models

mossberg:
As opposed to user defined models, core models live in Manticore but can be easily used by user scripts in combination with hooks to model functions.

```
from manticore import Manticore, issymbolic, models

m = Manticore('strcmpplay')

strcmp = 0x42d5f0
@m.hook(strcmp)
def h(state):
    state.invoke_model(models.strcmp)

m.run()
```


manticore/core/state.py
```diff
@@ -234,3 +234,11 @@ def record_branches(self, targets):
...
             except KeyError:
                 self.branches[key] = 1
+
+    def invoke_model(self, model):
+        '''
+        Invoke a model.
+
+        :param callable model: Model to invoke
+        '''
+        self.platform.invoke_model(model, (self,))  # TODO(mark): this can't support varargs core models!
```
yan: Explicitly specify `prefix_args` when passing `(self,)`? Also, move the comment on its own line?

manticore/models.py
```diff
@@ -0,0 +1,75 @@
...
+
+logger = logging.getLogger("MODEL")
+
+
+def _find_zero(cpu, constrs, ptr):
+    '''returns offset from ptr to the first byte that is 0 or an Expression that
+    cannot
+    be
+    nonzero
+    '''
```
yan: Accidental short-line wrapping?

yan: Also, "cannot be nonzero" -> "must be zero"?

mossberg:
I'm planning on adding the actual models in a future PR

PR #247, opened 2017-05-11
Support tracing

yan:
Implement the ability to trace a Manticore execution with qemu/gdb


manticore/platforms/linux.py
```diff
@@ -263,12 +264,21 @@ class Linux(Platform):
...
         :param list argv: The argv array; not including binary.
         :param list envp: The ENV variables.
+        :param \**options: See below
+
+        :Keyword Arguments:
+         * *stack_top* (``int``) --
+           Set the top of the stack to the specified value
+         * *interpreter_base* (``int``) --
+           Set the loading address of the interpreter (if not specified by ELF)
+         * *stack_offset* (``int``) --
```
feliam: stack_offset... If both versions of the program/emulation use getenv/getauxv the trace will diverge. Consider making the stacks match instead of making a gap in one? The gap is also achievable via a fake envp.

scripts/verify.py
```diff
@@ -0,0 +1,182 @@
...
+        if last_instruction != state.cpu.PC:
+            if last_instruction is not None:
+                on_after(m, state, last_instruction)
+
+        last_instruction = state.cpu.instruction
+
+        loc, instr = [x.strip() for x in gdb.getInstruction().split(':')]
+        mnemonic = instr.split('\t')[0]
+            
+        first_param, syscall = gdb.getR('R0'), gdb.getR('R7')
```
mossberg: does `syscall` need to be global? can it be passed into sync_svc?

manticore/platforms/linux.py
```diff
@@ -264,12 +264,21 @@ class Linux(Platform):
     This class emulates the most common Linux system calls
     '''
 
-    def __init__(self, program, argv=None, envp=None):
-        '''
+    def __init__(self, program, argv=None, envp=None, **options):
```
yan: My concern was changing the model API whenever we need to add options. Or do we assume all params after `envp` are options?

mossberg: I feel like using **options at this point isn't changing the API any less, just in a less obvious way. By changing the model (platform, technically) API, do you mean altering this declaration line?

scripts/verify.py
```diff
@@ -0,0 +1,184 @@
...
+
+    @m.hook(None)
+    def on_instruction(state):
+        global initialized, last_instruction, syscall
+        # Initialize our state to QEMU's
+        if not initialized:
+            initialize(state)
+            initialized = True
+
+        if last_instruction is not None and last_instruction.PC != state.cpu.PC:
```
mossberg: last_instruction.address?

scripts/verify.py
```diff
@@ -0,0 +1,184 @@
...
+
+    @m.hook(None)
+    def on_instruction(state):
+        global initialized, last_instruction, syscall
+        # Initialize our state to QEMU's
+        if not initialized:
+            initialize(state)
+            initialized = True
+
+        if last_instruction is not None and last_instruction.address != state.cpu.PC:
```
mossberg: when would the latter part of the check fail? previous instruction is the same as current instruction?

scripts/verify.py
```diff
@@ -0,0 +1,184 @@
...
+            on_after(state, last_instruction)
+
+        last_instruction = state.cpu.instruction
+
+        loc, instr = [x.strip() for x in gdb.getInstruction().split(':')]
+        mnemonic = instr.split('\t')[0]
```
mossberg: could we use `state.cpu.instruction.mnemonic` instead here? it should be the same as what gdb is about to exec?

scripts/verify.py
```diff
@@ -0,0 +1,184 @@
...
+def on_after(state, last_instruction):
+    '''
+    Handle syscalls (import memory) and bail if we diverge
+    '''
+    global icount, in_helper
+
+    icount += 1
+
+    # Synchronize qemu state to manticore's after a system call
+    if last_instruction.mnemonic == 'svc':
```
mossberg: maybe move this syscall stuff into a separate function, and call it right before where `on_after` is called? i think the svc sync would be more clear if both checks for `svc` instruction were in the same function

scripts/verify.py
```diff
@@ -0,0 +1,184 @@
...
+        v = cpu.read_register(name.upper())
+        if should_print:
+            print '{} gdb:{:x} mcore:{:x}'.format(name, vg, v)
+        if vg != v:
+            if should_print:
+                print '^^ unequal'
+            differing = True
+    return differing
+
+def on_after(state, last_instruction):
```
mossberg: maybe rename to something more descriptive like verify_insn?

yan: `on_after` doesn't just verify instruction, it's supposed to be everything we need to handle after an instruction is executed, with the idea to move it to a hook when we get post-exec hooks.

manticore/manticore.py
```diff
@@ -200,7 +200,7 @@ def filter(self, record):
 
         logging.basicConfig(format='%(asctime)s: [%(process)d]%(stateid)s %(name)s:%(levelname)s: %(message)s', stream=sys.stdout)
 
-        for loggername in ['VISITOR', 'EXECUTOR', 'CPU', 'REGISTERS', 'SMT', 'MEMORY', 'MAIN', 'PLATFORM']:
+        for loggername in ['VISITOR', 'EXECUTOR', 'CPU', 'REGISTERS', 'SMT', 'MEMORY', 'MAIN', 'PLATFORM', 'TRACE']:
```
mossberg: Might be nice if we didn't need this, as this makes verify.py not just a normal mcore script

PR #248, opened 2017-05-11
Add input-related features

montyly:
Better handling of symbolic files.

* The user can now define which files are symbolic using `add_symbolic_file`, example:
```python
m = Manticore()
m.add_symbolic_file('my_file.txt')
```

* `generate_testcase()` of `Executor` concretizes the symbolic files.

* A user can generate the input of a `state` at any moment using `generate_inputs()`, example:
```python
@m.hook(0x41414141) 
def generate(state):
  state.generate_inputs(m.workspace, generate_files=True)
```


manticore/platforms/linux.py
```diff
@@ -1735,7 +1735,7 @@ class SLinux(Linux):
     :param list envp: environment variables
     :param tuple[str] symbolic_files: files to consider symbolic
     """
-    def __init__(self, programs, argv=None, envp=None, symbolic_files=()):
+    def __init__(self, programs, argv=None, envp=None, symbolic_files=[]):
```
mossberg: we should not use a list here because it's a mutable type. we should do the `None` thing like with argv/envp http://python-guide-pt-br.readthedocs.io/en/latest/writing/gotchas/#mutable-default-arguments

manticore/manticore.py
```diff
@@ -37,12 +37,13 @@ def makeDecree(args):
     platform.input.transmit(initial_state.symbolicate_buffer('+'*14, label='RECEIVE'))
     return initial_state
 
-def makeLinux(program, argv, env, concrete_start = ''):
+def makeLinux(program, argv, env, concrete_start = '', symbolic_files=[]):
```
mossberg: same with list as default argument (http://python-guide-pt-br.readthedocs.io/en/latest/writing/gotchas/#mutable-default-arguments)

manticore/core/executor.py
```diff
@@ -491,6 +491,15 @@ def generate_testcase(self, state, message = 'Testcase generated'):
         
         file(self._getFilename('test_%08x.syscalls'%test_number),'a').write(repr(state.platform.syscall_trace))
 
+        # save symbolic files
+        if hasattr(state.platform, 'files'):
```
feliam: Consider using files = getattr(state.platform, 'files',None) instead of hasattr. (https://hynek.me/articles/hasattr/)

manticore/core/state.py
```diff
@@ -240,6 +241,40 @@ def record_branches(self, targets):
...
+
+        # save the inputs
+        for symbol in self.input_symbols:
+            buf = solver.get_value(self.constraints, symbol)
+            filename = os.path.join(workspace,'state_%08x.txt'%self.co)
+            file(filename,'a').write("%s: %s\n"%(symbol.name, repr(buf)))
+       
+        # save the symbolic files
+        if generate_files:
+            if hasattr(self.platform, 'files'):
```
feliam: same here

manticore/core/executor.py
```diff
@@ -491,6 +491,15 @@ def generate_testcase(self, state, message = 'Testcase generated'):
...
+            for f in state.platform.files:
+                if hasattr(f,'array'):
+                    buf = solver.get_value(state.constraints, f.array)
+                    filename = os.path.basename(f.array.name)
+                    filename = 'test_file_%s_%d.txt'%(filename, test_number)
```
mossberg: all of the other files we generate have `test_0001.blah` format, and this kind of departs from that in a way that stands out. let's do  `test_0001.filename` instead?

mossberg: also, can we prefer `.format` instead of `%`?

manticore/core/executor.py
```diff
@@ -491,6 +491,15 @@ def generate_testcase(self, state, message = 'Testcase generated'):
...
+                    file(self._getFilename(filename),'a').write("%s"%(buf))
```
mossberg: prefer `open()` to `file()`?

manticore/core/state.py
```diff
@@ -240,6 +241,40 @@ def record_branches(self, targets):
...
+            f.write(str(self.constraints))
+
+        # check that the state is sat
+        assert solver.check(self.constraints)
+
+        # save the inputs
+        for symbol in self.input_symbols:
+            buf = solver.get_value(self.constraints, symbol)
+            filename = os.path.join(workspace,'state_%08x.txt'%self.co)
+            file(filename,'a').write("%s: %s\n"%(symbol.name, repr(buf)))
```
feliam: open() is preferred. Consider using it?
https://slack-files.com/T03A74C1E-F5H2XCNCU-7d40a2b8eb

manticore/core/executor.py
montyly: Should I also directly change all the remaining `%` to `.format` used in `generate_testcase` ?

mossberg: that would be great!

manticore/core/executor.py
```diff
@@ -487,9 +487,20 @@ def generate_testcase(self, state, message = 'Testcase generated'):
...
         for symbol in state.input_symbols:
             buf = solver.get_value(state.constraints, symbol)
-            file(self._getFilename('test_%08x.txt'%test_number),'a').write("%s: %s\n"%(symbol.name, repr(buf)))
+            open(self._getFilename('test_{:08x}.txt'.format(test_number)),'a').write("{:s}: {:s}\n".format(symbol.name, repr(buf)))
         
-        file(self._getFilename('test_%08x.syscalls'%test_number),'a').write(repr(state.platform.syscall_trace))
+        open(self._getFilename('test_{:08x}.syscalls'.format(test_number)),'a').write(repr(state.platform.syscall_trace))
+
+        # save symbolic files
+        files = getattr(state.platform, 'files', None)
```
mossberg: all these getattr checks doesn't feel very pythonic to me. i think we might want to also move this symbolic file saving code into the platform 
1. to avoid copypasting this code to save symbolic files
2. because the executor/state shouldn't be digging too much into the internals of the platform? (.files)
3. because we could remove the getattrs since can be certain that .files and .array exists, etc

however, i think the platform shouldn't necessarily be the one to touch disk (or whatever storage backends we may implement in the future), maybe it should just provide the data and its name (i.e. filename) and return back to the executor or state for writing.

thoughts?

manticore/core/executor.py
```diff
@@ -507,8 +509,8 @@ def generate_testcase(self, state, message = 'Testcase generated'):
                 stdout += ''.join(map(str, data))
             if sysname in ('_transmit', '_write') and fd == 2:
                 stderr += ''.join(map(str, data))
-        file(self._getFilename('test_%08x.stdout'%test_number),'a').write(stdout)
-        file(self._getFilename('test_%08x.stderr'%test_number),'a').write(stderr)
+        open(self._getFilename('test_{:08x}.stdout'.format(test_number)),'a').write(stdout)
```
mossberg: on second thought, this is really low priority, and perhaps not worth anyone's time right now

feliam: Agree. Also may be affected by refactors. Also is easy to read as it is.

manticore/core/executor.py
mossberg: after some discussion in slack, given the impending huge refactor in #241 it should be fine to leave this now and refactor later. my proposal is to implement some method on the Platform for, generating platform specific inputs/data/results. for example `platform.generate_misc_data()` which returns a dict of filename/name of data -> data (str)

manticore/manticore.py
```diff
@@ -37,12 +37,15 @@ def makeDecree(args):
     platform.input.transmit(initial_state.symbolicate_buffer('+'*14, label='RECEIVE'))
     return initial_state
 
-def makeLinux(program, argv, env, concrete_start = ''):
+def makeLinux(program, argv, env, concrete_start = '', symbolic_files=None):
```
mossberg: since makeLinux is only used once, and always called with symbolic files argument, we can keep it simple and not make this a default arg?

PR #249, opened 2017-05-11
Optimizations and bug fixes in smt formulas

montyly:
in `solve.py`: 
* Remove of the incremental mode in `get_all_values()` (two distinct calls to the solver are faster than using incremental mode in the context of `get_all_values()`)

in `visitors.py`:
* Remove duplicated prints in `PrettyPrinter`and `TranslatorSmtLib` (smt formulas are now smaller)
* Fix `visit` in `ArithmeticSimplifier`
* Fix the use of `ConstantFolderSimplifier` (no-trivial operators were not called by `ArithmeticSimplifier`)

### DIFF
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -350,8 +350,12 @@ class Z3Solver(Solver):
             val = None
             while self._check() == 'sat':
                 value = self._getvalue(var)
+                # reset the solver to avoid the incremental mode 
+                # triggered with two consecutive calls to check-sat
+                self._reset(temp_cs.related_to(var) )
                 result.append(value)
-                self._assert( var != value )
+                for value in result:
+                    self._assert( var != value )
                 if len(result) >= maxcnt:
                     if silent:
                         # do not throw an exception if set to silent
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -350,12 +350,19 @@ class Z3Solver(Solver):
             val = None
             while self._check() == 'sat':
                 value = self._getvalue(var)
-                # reset the solver to avoid the incremental mode 
-                # triggered with two consecutive calls to check-sat
-                self._reset(temp_cs.related_to(var) )
                 result.append(value)
-                for value in result:
-                    self._assert( var != value )
+
+                # Reset the solver to avoid the incremental mode 
+                # Triggered with two consecutive calls to check-sat
+                # Yet, if the number of solution is large, sending back
+                # the whole formula is more expensive
+                if len(result) < 50:
+                    self._reset(temp_cs.related_to(var) )
+                    for value in result:
+                        self._assert( var != value )
+                else:
+                    self._assert(var != value)
+                    
                 if len(result) >= maxcnt:
                     if silent:
                         # do not throw an exception if set to silent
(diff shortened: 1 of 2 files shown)


manticore/core/smtlib/visitors.py
```diff
@@ -232,29 +253,35 @@ class ArithmeticSimplifier(Visitor):
...
+            node = stack.pop()
+ 
+            if node in cache:
+                self.push(cache[node])
+            elif isinstance(node, Operation):
+                if node in visited:
+                    operands = [self.pop() for _ in xrange(len(node.operands))]
+                    new_node = self._rebuild(node, operands)
+                    value = self._method(new_node, *operands)
+                    while value is not new_node:
```
feliam: I wonder if we can have a generic  visit() in the Visitor class that can account for this cases. Basically If we can move this thing up in the class hierarchy. Maybe overload _rebuild() and set a class configuration flag to control if i needs to re-runn ._method() until it finds a fixed point?

manticore/core/smtlib/visitors.py
```diff
@@ -147,6 +147,7 @@ def visit_Operation(self, expression, *operands):
         else:
             self._print('...')
         self.indent -= 2
+        return ''
```
montyly: The problem comes from `Visitor._method`, which tries to iterate on all the possible `visit_` methods until one does not return `Ǹone`, that's why I simply added a `return ''` (the return value is never used in fact). If it does make more sense it can return the string representation instead. Another solution could be to write a `_method` for `PettyPrinter` which stops to iterate if a corresponding visitor is found.

For the call to `visit` in `visit_operation`, I guess that the recursive approach was chosen for the PrettyPrinter (instead of using a stack), as it makes easier to keep the consistency of the indentation (but it's just a guess)

manticore/core/smtlib/solver.py
```diff
@@ -351,7 +351,18 @@ def get_all_values(self, constraints, expression, maxcnt=300, silent=False):
             while self._check() == 'sat':
                 value = self._getvalue(var)
                 result.append(value)
-                self._assert( var != value )
+
+                # Reset the solver to avoid the incremental mode
```
montyly: The incremental mode is usually slower (see https://stackoverflow.com/questions/23736269/what-is-the-benifits-of-incremental-solving/23739636#23739636). 
In several cases it is faster to do two formulas:
```
constraints
x != v_0
check-sat
```
```
constraints
x != v_0
x != v_1
check-sat
```
Than

``` :
constraints
x != v_0
check-sat
x != v_1
check-sat
```
Specially this pattern appears every time Manticore wants to compute the PC.

However, if the formula is large and there are a considerable number of solutions, sending the whole set of constraints starts to become more costly than the cost of the incremental mode

feliam: Oh! Unexpected.

manticore/core/smtlib/visitors.py
feliam: Does making _method stop iterating when there is a visit_operation defined change something?  (insted when the visit_operation is defined _and_ returns not None)

montyly: Yes but this has to be applied only for the pretty printer. The optimization needs `_method` to iterate over all the `visit_` methods. But we can define a particular `_method` for the pretty printer with this behavior

PR #250, opened 2017-05-11
Rm unused Slinux ctor params

### LINKED ISSUE #109: SLinux.__init__() ignores the constraints argument
We should either remove the argument, or initialize constraint set correctly

mossberg:
Fixes #109


manticore/platforms/linux.py
```diff
@@ -1730,12 +1730,16 @@ class SLinux(Linux):
     '''
     A symbolic extension of a Decree Operating System Platform.
```
yan: Correct the comment in this PR as well?

PR #251, opened 2017-05-12
Add strcmp model

mossberg:
Add model of strcmp


manticore/models.py
```diff
@@ -0,0 +1,80 @@
...
+
+    cpu = state.cpu
+
+    s1_zero_idx = _find_zero(cpu, state.constraints, s1)
+    s2_zero_idx = _find_zero(cpu, state.constraints, s2)
+    min_zero_idx = min(s1_zero_idx, s2_zero_idx)
+
+    ret = None
+
+    for offset in range(min_zero_idx + 1)[::-1]:
```
yan: `range(min_zero_idx, -1, -1)`?

manticore/models.py
```diff
@@ -0,0 +1,80 @@
...
+    :param int ptr: Address to start searching for a zero from
+    :return: Offset from `ptr` to first byte that is 0 or an `Expression` that must be zero
+    """
+
+    offset = 0
+    while True:
+        byt = cpu.read_int(ptr + offset, 8)
+
+        if issymbolic(byt):
+            if not solver.can_be_true(constrs, byt != 0):
```
yan: Do we want to find the byte that must be zero, or that might be zero?

mossberg: Must be zero. if we did might be zero, the tree strcmp would build would be incomplete if that byte was not zero (and the string kept going) (since strcmp starts at the zero and works its way back)

manticore/models.py
```diff
@@ -0,0 +1,80 @@
+from .utils.helpers import issymbolic
+from .core.smtlib.solver import solver
+from .core.smtlib.operators import ITEBV, ZEXTEND
+
+
+def _find_zero(cpu, constrs, ptr):
```
mossberg: does this belong elsewhere? cpu? platform (we already have _read_string there too)?

yan: I don't think this belongs in `Cpu` (i think `read_string` should though, like `read_int` & co). I think this is a valid place for it; might even remove the leading underscore as it can be an externally-useful function.

mossberg: `models.find_zero` looks a little funny to me. what about helpers.py?

yan: `model_helpers.find_zero`? `models.helpers.find_zero`? I like the latter

mossberg: i'm not sure i want to make models.py a subpackage just to include a helpers file. keep it as is and change later when models.py gets messy or an external user needs _find_zero?

yan:
Should we raise an exception on a symbolic pointer here, or let the exception bubble up? (`TypeError: object of type 'BitVec' has no len()`)

PR #253, opened 2017-05-15
Manticore challenge

ctfhacker:
Simple solution for the manticore challenge from today's blog.


mossberg:
hi @ctfhacker thanks a lot! we made a separate repo for example scripts like this: https://github.com/trailofbits/manticore-examples
could you submit a PR there?

Issue #181, opened 2017-04-25
MemoryException from running `basic`


mossberg:
Probably an emulation bug

[basic.slam.zip](https://github.com/trailofbits/manticore/files/953898/basic.slam.zip)

### OS / Environment
Ubuntu 16.10

skxo:
Exactly the same error
[11605][0] EXECUTOR:ERROR: MemoryException at PC: 0x000000000041c79b. Cause: No access reading
2017-04-25 18:18:05,856: [11605][0] EXECUTOR:INFO: Generating testcase No. 1 for state No.0 - Memory Exception: No access reading <00000008>

OS / Environment

Linux-VirtualBox 4.8.0-49-generic #52-Ubuntu SMP Thu Apr 20 09:38:39 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

please advice will be really nice.

mossberg:
thanks for the bug report @skxo! the binary that got compiled on your system likely is triggering an emulator error we haven't seen before in manticore that is causing the program to incorrectly segfault (it tried to dereference 0x8, which is why it printed Memory Exception). in the meantime if you want to try manticore, look here: https://gist.github.com/ehennenfent/a5ad9746615d1490c618a88b98769c10 . the first link in the writeup is the link to the binary itself, and those manticore scripts in the writeup should work.

PR #259, opened 2017-05-17
Fix x86.PUNPCKL emulation

feliam:
Fixes issue 181.  
https://github.com/trailofbits/manticore/issues/181

### DIFF
--- a/manticore/core/cpu/x86.py
+++ b/manticore/core/cpu/x86.py
@@ -4567,55 +4567,52 @@ class X86Cpu(Cpu):
         '''
         res = dest.write(dest.read() ^ src.read())
 
-
-
-    def _PUNPCKL(cpu, op0, op1, item_size):
+    def _PUNPCKL(cpu, dest, src, item_size):
         '''
         Generic PUNPCKL
         '''
-        assert op0.size == op1.size
-        size = op0.size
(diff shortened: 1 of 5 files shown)


manticore/core/cpu/x86.py
```diff
@@ -4568,8 +4568,73 @@ def PXOR(cpu, dest, src):
...
+            item1 = Operators.ZEXTEND( ( arg1 >> (pos * item_size) )& mask, size)
+            res |= item0 << count
+            count += item_size
+            res |= item1 << count
+            count += item_size
+
+        op0.write(res)
+
+    #                 dest,src
+    def _PUNPCKH(cpu, op0, op1, item_size):
```
yan: rename params instead of the above comment?

tests/auto/make_tests.py
```diff
@@ -14,7 +14,7 @@
 for test in tests:
     try:
         cnt = op_count.get(test['mnemonic'],0)
-        if cnt > 20: #No more than 10000 instructions of each kind
+        if cnt > 1000: #No more than 10000 instructions of each kind
```
yan: Number doesn't match the comment

feliam: It's closer than before! :) improvement!

PR #260, opened 2017-05-17
Add strlen model

mossberg:
Add a model for strlen


manticore/models.py
```diff
@@ -84,3 +84,32 @@ def strcmp(state, s1, s2):
...
+    :param State state: current program state
+    :param int s: Address of string
+    :return: Symbolic strlen result
+    :rtype: Expression or int
+    """
+
+    cpu = state.cpu
+
+    if issymbolic(s):
+        raise ConcretizeArgument(1)
```
feliam: Wonder which policy we want here? maybe even "ALL" ?

mossberg: i'm concerned with using ALL because if the constraints are sufficiently loose, this is a good way to hang mcore. also, it doesn't make very much sense for the string ptr to be symbolic here in the first place; i'd expect this nearly guarantees a segfault in real execution. since we default to minmax, we'll also probably MemoryException, which is the appropriate way to handle this. i'll leave it like this for now.

PR #261, opened 2017-05-17
Update fstat/fstat64, add few syscall wrappers

yan:


manticore/platforms/linux.py
```diff
@@ -1091,16 +1091,28 @@ def sys_newuname(self, old_utsname):
...
-        self.current.write_bytes(old_utsname, uname)
-        logger.debug("sys_newuname(...) -> %r", uname)
+        now = datetime.now().strftime("%a %b %d %H:%M:%S ART %Y")
+
+        if self.current.machine == 'amd64':
+            machine = 'x86_64'
+        elif self.current.machine == 'i386':
+            machine = 'i386'
+        elif self.current.machine == 'armv7':
+            machine = 'armv71'
```
feliam: Consider setting a default to self.current.machine for future machines ?

PR #264, opened 2017-05-18
Linux platform refactoring

yan:
* Move stack-based helpers to `Cpu`
 * Refactor initialization so that we can create a platform model that isn't associated with an ELF file
 * Simplify SLinux sys call handling to rely on normal `ConcretizeArgument` exceptions

### DIFF
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -502,6 +502,50 @@ class Cpu(object):
             where += 1
         return s.getvalue()
 
+    def push_bytes(self, data):
+        '''
+        Write `data` to the stack and decrement the stack pointer accordingly.
+
+        :param str data: Data to write
+        '''
+        self.STACK -= len(data)
+        self.write_bytes(self.STACK, data)
+        return self.STACK
+
+    def pop_bytes(self, nbytes):
+        '''
+        Read `nbytes` from the stack, increment the stack pointer, and return
+        data.
+
+        :param int nbytes: How many bytes to read
+        :return: Data read from the stack
+        '''
+        data = self.read_bytes(self.STACK, nbytes)
+        self.STACK += nbytes
+        return data
+
+    def push_int(self, value):
+        '''
+        Decrement the stack pointer and write `value` to the stack.
+
+        :param int value: The value to write
+        :return: New stack pointer
+        '''
+        self.STACK -= self.address_bit_size / 8
+        self.write_int(self.STACK, value)
+        return self.STACK
+
+    def pop_int(self):
+        '''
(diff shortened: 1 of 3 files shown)


manticore/core/cpu/abstractcpu.py
```diff
@@ -477,6 +478,74 @@ def read_bytes(self, where, size):
...
+            where += 1
+        return s.getvalue()
+
+    def push_bytes(self, data):
+        '''
+        Write `data` to the stack and decrement the stack pointer accordingly.
+
+        :param str data: Data to write
+        '''
+        self.STACK -= len(data)
```
feliam: Should we assert  that ` 'STACK' in self._regfile`  at class constructor?

manticore/core/cpu/abstractcpu.py
```diff
@@ -477,6 +478,74 @@ def read_bytes(self, where, size):
...
+        :param int where: Address to read string from
+        :param int max_length:
+            The size in bytes to cap the string at, or None [default] for no
+            limit.
+        :return: string read
+        :rtype: str
+        '''
+        s = StringIO.StringIO()
+        while True:
+            c = self.read_int(where, 8)
```
feliam: What if c is symbolic?

manticore/core/cpu/abstractcpu.py
yan: do we not assume that `STACK` is a universal alias?

manticore/core/cpu/abstractcpu.py
yan: This function is technically defined to work solely on concrete buffers. What would be the correct behavior? Implement it via Mark's `find_zero`?

manticore/core/cpu/abstractcpu.py
feliam: So aliases are passed to RegisterFile constructor and those may not contain 'STACK' as it is now. If we want STACK to be part of RegisterFile I say we should assert that an specific alias for it is provided. This is at RegisterFile constructor or AbstractCpu I guess.  Same for 'PC'. No?

manticore/core/cpu/abstractcpu.py
feliam: If we are sure it's going to be concrete then just assert not issymbolic(x).

manticore/core/cpu/abstractcpu.py
yan: Added alias check to the `Cpu` ctor

manticore/core/cpu/abstractcpu.py
yan: 👍

yan:
@feliam still approved with the above changes?

PR #267, opened 2017-05-22
Add solve_buffer to api and update google ctf script to use it

### LINKED ISSUE #196: Add API for concretizing a symbolic buffer in memory
A common use case is reading some symbolic data/buffer from memory, and then concretizing it. For example, this frequently appears in CTF challenges and nearly any time you're using the Manticore API for symbolic execution.

Currently, this kind of code is commonly seen in Manticore scripts (credit @ctfhacker)

```
buffer = state.cpu.read_bytes(input_addr, num_bytes)
res = ''.join(chr(state.solve_one(x)) for x in buffer)
```

I think it makes sense to offer a convenience API that does both of these.

Also, this reminds me of "don't make the client do anything the module could do" in this talk on API design (https://youtu.be/aAb7hSCtvGw?t=2730)

japesinator:
Fixes #196


manticore/core/state.py
```diff
@@ -231,6 +231,19 @@ def solve_n(self, expr, nsolves, policy='minmax'):
...
+        Reads `nbytes` of symbolic data from a buffer in memory at `addr` and attempts to
+        concretize it
+
+        :param int address: Address of buffer to conretize
+        :param int nbytes: Size of buffer to conretize
+        :return: Concrete contents of buffer
+        :rtype: list[int]
+        '''
+        buffer = self.cpu.read_bytes(addr, nbytes)
+        return ''.join(chr(self.solve_one(x)) for x in buffer)
```
feliam: This will normally going to work.
Though, this is solving on byte at a time and not stating that fact in the constraint store. So 1st solution _may_ be unrelated to the second one. We should actually handle Array in solver.get_value et. al.
(I could not make a constraint store to show this. May be I'm wrong)

mossberg: Why does it matter that the solutions are related to each other? I figure as long as a solution exists that satisfies the constraints, it is fine?

feliam: array[0] = bitvec(8)
array[1] = bitvec(8)
solver.add_assertion(array[0] >array[1])
solver.get_value(array[0]) - > 0
solver.get_value(array[1]) - > 0

->  := "it could give"

mossberg: I see. Even though this will normally work, I'd say we should still strive to make it as sound as possible. Basically as I understand it, after solving each character, to we need to add a constraint back on the state for the solution, so the solver can take that into account, if there happen to be relationships among the bytes in the buffer.

We could implement this right here in `solve_buffer`, but perhaps also modify a deeper interface like solver.get_value as felipe suggested.

feliam: If we want to get away with doing it here for now. We do not  want to add the assertions  to the actual state but to a temporary ConstraintSet.Something like...
```
result = []
with state.constraints as temp_cs:
   for c in array:
     result.append(solve_one(c))
     temp.cs.add(c==result[-1])
```

mossberg: I think I actually prefer having it in the State, and not the solver; it feels more correct to me to have the solver provide the 'low level' functionality to solve a single value, and then have the state apply that in higher level ways, such as this.

manticore/core/state.py
```diff
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from copy import deepcopy
```
mossberg: stray import here

examples/script/google2016_unbreakable/win.py
```diff
@@ -51,8 +51,7 @@ def hook(state):
 def hook(state):
     print("Hit the final state.. solving")
 
-    buffer = state.cpu.read_bytes(input_addr, num_bytes)
-    res = ''.join(chr(state.solve_one(x)) for x in buffer)
+    res = ''.join(map(chr, state.solve_buffer(input_addr, num_bytes)))
```
mossberg: now that we have manticore-examples, we can remove this script and directory entirely? we can update the script there, and maybe edit the docstring to reflect that the original submitted script has been modified

manticore/core/state.py
```diff
@@ -33,6 +34,7 @@ def __init__(self, constraints, platform):
         self.forks = 0
         self.co = self.get_new_id()
         self.constraints = constraints
+        self._constraint_stack = []
```
mossberg: stray variable also

PR #273, opened 2017-05-25
Check returncode of manticore integration tests

mossberg:
Without this, the manticore subprocess can fail, but the test still passes


japesinator:
👍

PR #274, opened 2017-05-25
Add real test for m.hook API

mossberg:
Other tests verify that the hook infrastructure is set up, but not that the hook actually runs.


japesinator:
👍

PR #276, opened 2017-05-30
Implement user variadic model interface

### LINKED ISSUE #275: Implement user defined variadic model interface
As decided by the poll in slack (https://empirehacking.slack.com/archives/C3PTWK7UM/p1494966837390179, https://gist.github.com/yan/265a14ad90c2715fac31238c5cae95cb), we will use an `@variadic` decorator for marking user defined function models as taking a variable number of arguments. This will cause the function to take a single argument, which is a generator object for the arguments to the function.

```
# 1 ------------------------------------------------
# Option 1, "@variadic" attribute on the model itself, and it takes a single
# parameter which acts as an iterator for all functions. Calling next() on the
# args object gets you arguments. Invoking vararg and constant arity functions
# is identical.

from manticore import variadic

@variadic
def printf(state, args):
    fmt = next(args)
    pass
@m.hook(printfaddr)
def hk(state):
    state.invoke_model(printf)
def strcmp(state, arg1, arg2):
    pass
@m.hook(strcmpaddr)
def hk(state):
    state.invoke_model(strcmp)
```

mossberg:
Fixes #275

### DIFF
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -222,14 +222,14 @@ class Abi(object):
 
     def invoke(self, model, prefix_args=None):
         '''
-        Invoke a callable `model` as if it was a native function. If `variadic`
-        is true, model receives a single argument that is a generator for
-        function arguments. Pass a tuple of arguments for `prefix_args` you'd
-        like to precede the actual arguments.
+        Invoke a callable `model` as if it was a native function. If
+        :func:`~manticore.models.isvariadic` returns true for `model`, `model` receives a single
+        argument that is a generator for function arguments. Pass a tuple of
+        arguments for `prefix_args` you'd like to precede the actual
+        arguments.
 
         :param callable model: Python model of the function
         :param tuple prefix_args: Parameters to pass to model before actual ones
-        :param bool variadic: Whether the function expects a variable number of arguments
         :return: The result of calling `model`
         '''
         prefix_args = prefix_args or ()
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -220,7 +220,7 @@ class Abi(object):
             yield base
             base += word_bytes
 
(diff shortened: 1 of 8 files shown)


manticore/core/state.py
```diff
@@ -248,5 +249,4 @@ def invoke_model(self, model):
 
         :param callable model: Model to invoke
         '''
-        # TODO(mark): this can't support varargs core models!
-        self.platform.invoke_model(model, prefix_args=(self,))
+        self.platform.invoke_model(model, prefix_args=(self,), variadic=isvariadic(model))
```
yan: If we're standardizing on `@variadic` attribute, do we want to remove the `variadic` parameter to `invoke_model`? It sounds like we'll always pass `isvariadic()` to it anyway, might as well move that check inside and simplify the interface.

manticore/models.py
```diff
@@ -3,6 +3,19 @@
...
+
+def isvariadic(func):
+    return getattr(func, VARIADIC_FUNC_ATTR, False)
+
+def variadic(func):
+    """
+    A decorator used to mark a function model as variadic. This function should
+    take two parameters: a :class:`~manticore.core.state.State` object, and
+    a generator object for the arguments.
+    """
```
yan: Add `:param` descriptor to docstring?

manticore/core/state.py
mossberg: good idea, was going to bring that up

manticore/core/cpu/abstractcpu.py
```diff
@@ -255,8 +255,10 @@ def resolve_argument(arg):
         descriptors = self.get_arguments()
         argument_iter = imap(resolve_argument, descriptors)
 
+        from ...models import isvariadic
```
mossberg: Moving this import to the top with the others introduces a circular dependency error, which would require some restructuring of things and moving around to fix. Perhaps leave a comment, and gh issue, but do later?

PR #279, opened 2017-06-05
Add better forking to memory

japesinator:


japesinator:
We should probably do COW via some more general method (i.e. a Forkable class) instead of the class-specific stuff partially implemented here. Hence, I'm closing this PR and looking into that

PR #289, opened 2017-06-08
Make Linux write(2) work with Files

mossberg:
Currently, Linux.sys_write assumes `is_full` and `transmit` methods on file descriptor objects, however File has neither of these. Add `is_full` to File, and refactor so both Files and Sockets use `write`

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -241,9 +241,6 @@ class Socket(object):
         return ret
 
     def write(self, buf):
-        return self.transmit(buf)
-
-    def transmit(self, buf):
         assert self.is_connected()
         return self.peer._transmit(buf)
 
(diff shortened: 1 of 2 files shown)


manticore/platforms/linux.py
```diff
@@ -51,15 +51,19 @@ def tell(self, *args):
         return self.file.tell(*args)
     def seek(self, *args):
         return self.file.seek(*args)
-    def write(self, *args):
-        return self.file.write(*args)
+    def write(self, buf):
+        for c in buf:
+            self.file.write(c)
```
yan: Why write it a byte at a time?

mossberg: I was mostly mirroring `Socket._transmit`. This method is called with a list in sys_write, the loop allows this to work with lists or strs.

PR #290, opened 2017-06-08
fsync, rename, and a few cleanups

yan:


manticore/platforms/linux.py
```diff
@@ -1196,6 +1202,45 @@ def sys_open(self, buf, flags, mode):
...
+        :param int newnamep: pointer to newname
+        '''
+        oldname = self.current.read_string(oldnamep)
+        newname = self.current.read_string(newnamep)
+
+        ret = 0
+        try:
+            os.rename(oldname, newname)
+        except OSError as e:
+             ret = -e.errno
```
mossberg: -1 is returned also

manticore/platforms/linux.py
```diff
@@ -1196,6 +1202,45 @@ def sys_open(self, buf, flags, mode):
...
+
+        ret = 0
+        try:
+            f = self.files[fd]
+            if isinstance(f, Socket):
+                ret = -errno.EINVAL
+            else:
+                f.sync()
+        except IndexError:
+            ret = -errno.EBADF
```
mossberg: same -1 thing

manticore/platforms/linux.py
```diff
@@ -1047,12 +1053,12 @@ def sys_write(self, fd, buf, count):
 
             if not self._is_open(fd):
                 logger.error("WRITE: Not valid file descriptor. Returning EBADFD %d", fd)
-                return errno.EBADF
+                return -errno.EBADF
```
mossberg: i just checked the man page and write(2) returns -1 on error, setting `errno` to the error code; it doesn't return the err code.

PR #292, opened 2017-06-08
fixes #291 for incorrect syscall returns and some indent issues

### LINKED ISSUE #291: Linux incorrect syscall error returns
In some system calls, we incorrectly return positive errno numbers instead of negative ones. Examples: sys_lseek, sys_read

nettrino:


manticore/platforms/linux.py
```diff
@@ -1,18 +1,20 @@
...
-import cgcrandom
-import weakref
+import logging
+import random
 import errno
-import os, struct
+import os
+import struct
+
+import cgcrandom
```
yan: Since we're reformatting, should probably use relative project import for `cgcrandom`

manticore/platforms/linux.py
```diff
@@ -1,21 +1,22 @@
...
 import fcntl
+import logging
+import os
+import random
+import struct
+
+from elftools.elf.elffile import ELFFile
 
-import cgcrandom
-import weakref
```
mossberg: looks like `weakref` was used in 1 place that is commented out. i don't see a point in keeping the commented code, so lets clean that too

PR #298, opened 2017-06-09
make SymbolicFile inherit from File

nettrino:
Dropped SymbolicFile methods that don't do anything.
There should be test to exercise this functionality

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -34,36 +34,9 @@ def perms_from_protflags(prot_flags):
 
 class File(object):
     def __init__(self, *args, **kwargs):
-        #Todo: assert file is seekable otherwise we should save what was
-        #read/write to the state
+        # TODO: assert file is seekable otherwise we should save what was
+        # read/write to the state
         self.file = file(*args,**kwargs)
-    def stat(self):
-        return os.fstat(self.fileno())
-    def ioctl(self, request, argp):
-        #argp ignored..
-        return fcntl.fcntl(self, request)
-    @property
-    def name(self):
-        return self.file.name
-    @property
-    def mode(self):
-        return self.file.mode
-    def tell(self, *args):
-        return self.file.tell(*args)
-    def seek(self, *args):
-        return self.file.seek(*args)
-    def write(self, buf):
-        for c in buf:
-            self.file.write(c)
-    def read(self, *args):
-        return self.file.read(*args)
-    def close(self, *args):
-        return self.file.close(*args)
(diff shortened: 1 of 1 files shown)


manticore/platforms/linux.py
```diff
@@ -96,31 +106,37 @@ def __init__(self, constraints, path="sfile", mode='rw', max_size=100, wildcard=
...
         '''
-        # XXX(yan): wildcard isn't used; check callers and remove.
         assert 'r' in mode
-        if isinstance(path, str):
-            path = File(path, mode)
-        assert isinstance(path, File)
+        assert isinstance(path, str)
+        super(SymbolicFile, self).__init__(path, mode)
+        # read the concrete data using the parent the read() form the File class
+        data = super(SymbolicFile, self).read()
```
mossberg: what do you think of using `self.file.read()` instead?

manticore/platforms/linux.py
```diff
@@ -320,9 +308,9 @@ def _init_fds(self):
         self.input.peer = stdin
         #A receive on stdout or stderr will return no data (rx_bytes: 0)
 
-        assert self._open(stdin) == 0
```
nettrino: I don't mind. It had taken me a while to get https://github.com/trailofbits/manticore/blob/a1a2dfd3ec68294c4a2a84b1b026bc80d3b84de6/manticore/platforms/linux.py#L1615, so I figured if we are to use STDIN, STDOUT etc better to name them somehow, thus I used sys.__stdin__.fileno() so as to not define a new var. I overall avoid any sort of number which is not 0, 1, -1, so as to give more context, but no hard feelings on this

manticore/platforms/linux.py
```diff
@@ -96,31 +110,37 @@ def __init__(self, constraints, path="sfile", mode='rw', max_size=100, wildcard=
         :param str mode: the access permissions of the symbolic file
         :param max_size: Maximun amount of bytes of the symbolic file
         '''
-        # XXX(yan): wildcard isn't used; check callers and remove.
         assert 'r' in mode
-        if isinstance(path, str):
-            path = File(path, mode)
-        assert isinstance(path, File)
+        assert isinstance(path, str)
```
mossberg: let's also update the docstring for `path`. it can now be something like 

`:param str path: ...`

nettrino: Should we have the assert as well or just let the parent handle these (for isinstance and mode)?

mossberg: good point, i'm inclined to let the parent handle them

PR #299, opened 2017-06-09
Move pretty printing into helpers

japesinator:
This should make benchmarking scripts easier, as we can just call this method from inside the script


scripts/benchmark.py
```diff
@@ -1,13 +1,11 @@
 from manticore import Manticore
+from manticore.utils.helpers import pretty_print_results
```
mossberg: what if instead of a helper, we make a `.dump()` or `.to_str()` method on `ProfilingResults` itself? (i'm not sure of the most idiomatic name)

PR #301, opened 2017-06-12
nits for common Socket/File FD operations

nettrino:


manticore/platforms/linux.py
```diff
@@ -250,6 +255,11 @@ def _transmit(self, buf):
             self.buffer.insert(0,c)
         return len(buf)
 
+    def sync(self):
+        raise BadFd("Invalid sync() operation on Socket")
```
mossberg: we actually do not want to raise here, we don't want analysis to crash/stop if the program has an fsync, which is why it was just a nop earlier

manticore/platforms/linux.py
```diff
@@ -986,20 +1000,18 @@ def sys_lseek(self, fd, offset, whence):
...
-
-        if isinstance(self.files[fd], Socket):
-            logger.info("LSEEK: Not valid file descriptor on lseek. Fd not seekable. Returning EBADF")
+        '''
+        try:
+            # Read the data and put it in memory
+            if not self._is_open(fd):
+                raise BadFd()
+            self.files[fd].seek(offset)
+        except BadFd:
```
mossberg: couldn't we also just catch IndexError, rather than raising and immediately catching?

manticore/platforms/linux.py
```diff
@@ -93,7 +97,7 @@ def sync(self):
         '''
         Flush buffered data. Currently not implemented.
         '''
-        return
+        raise NotImplementedError
```
mossberg: i think we actually want the parent implementation to be a nop

manticore/platforms/linux.py
```diff
@@ -232,6 +236,7 @@ def connect(self, peer):
             peer.peer = self
 
     def read(self, size):
+        assert not self.is_empty(), "Attempting read() from empty Socket buf"
```
mossberg: i don't think an assert is the right way to handle this error case? i think we should handle it by returning an empty list, not AssertionError

manticore/platforms/linux.py
nettrino: For this we were checking if we have a Socket and returning 
ret = -errno.EINVAL particularly for that. Will we ever do a sync on a socket?

manticore/platforms/linux.py
nettrino: Yep I don't like this handling logic either.. this was there just since both a seek()
on Socket and a non-open fd both return a EBADF. Essentially we would like to ignore what type of fd we have and let the classes handle this, this is why I wanted to remove all the isinstance() cases in handling logic. Unfortunately, for _is_open, self.files can have an idx with a None value for a closed fd, so IndexError will require us tweaking the self.files logic. What do you think?

manticore/platforms/linux.py
mossberg: ah i see, sorry about that, yes this looks fine

manticore/platforms/linux.py
mossberg: maybe catch both IndexError and AttributeError (None.seek(offset))? or introduce a helper function for accessing an open fd that takes care of indexing into `self.files` but raises a exception if there was None there, or it was out of range

manticore/platforms/linux.py
```diff
@@ -1225,15 +1247,13 @@ def sys_fsync(self, fd):
...
         try:
-            f = self.files[fd]
-            if isinstance(f, Socket):
-                ret = -errno.EINVAL
-            else:
-                f.sync()
+            self.files[fd].sync()
         except IndexError:
             ret = -errno.EBADF
+        except BadFd:
```
mossberg: it's confusing that BadFd returns EINVAL but right above IndexError returns EBADF? i would expect BadFd to return EBADF

manticore/platforms/linux.py
```diff
@@ -1663,14 +1693,13 @@ def execute(self):
...
         try:
             self.current.execute()
             self.clocks += 1
             if self.clocks % 10000 == 0:
                 self.check_timers()
                 self.sched()
-        except (Interruption, Syscall) as e:
+        except (Interruption, Syscall):
             try:
                 syscallret = self.syscall()
```
mossberg: since `syscallret` is unused, we can rm this assignment

manticore/platforms/linux.py
```diff
@@ -1790,7 +1819,8 @@ def _stat(self, path, buf, is64bit):
         if is64bit:
             ret = self.sys_fstat64(fd, buf)
         else:
-            ret = self.sys_fstat(fd, buf)
+            # FIXME (theo) is that self.sys_fstat for SLinux?
```
mossberg: what do you mean by this? i think we should keep sys_fstat here

manticore/platforms/linux.py
```diff
@@ -939,21 +953,22 @@ def _dup(self, fd):
...
         :rtype: int
-        :param fd: the file descriptor to close.
+        :param fd: the file descriptor to duplicate.
         :return: C{0} on success.
         '''
         return self._open(self.files[fd])
 
-    def _is_open(self, fd):
-        return fd >= 0 and fd < len(self.files) and self.files[fd] is not None
+    def _assert_open(self, fd):
```
nettrino: Agreed

manticore/platforms/linux.py
nettrino: Yep, I was confused a bit by the IndexError comment at https://github.com/trailofbits/manticore/pull/301/files/89afa2325a62db323730cee42c1481201c55c593

Ok so we raise Badfd a and that should do it

manticore/platforms/linux.py
nettrino: sys_fstat only exists in SLinux() so wanted to double check this is not a typo

mossberg:
Ah! good point, I think it is not a typo, but now that I look more closely, I think sys_fstat actually does belong in Linux and not SLinux
--
Mark Mossberg
mark@trailofbits.com

manticore/platforms/linux.py
mossberg: Ah! good point, I think it is not a typo, but now that I look more closely, I think sys_fstat actually does belong in Linux and not SLinux'

manticore/platforms/linux.py
```diff
@@ -1076,7 +1076,7 @@ def sys_write(self, fd, buf, count):
                 raise RestartSyscall()
 
             data = cpu.read_bytes(buf, count)
-            self.files[fd].write(data)
+            write_fd.write(data)
```
mossberg: we can replace `self.files[fd]` above with `write_fd` also

manticore/platforms/linux.py
```diff
@@ -263,11 +272,14 @@ def __init__(self, program, argv=None, envp=None):
...
         :param list argv: The argv array; not including binary.
         :param list envp: The ENV variables.
+        :ivar files: List of active file descriptors
+        :type files: list[Socket] or list[File]
         '''
         super(Linux, self).__init__(program)
 
         self.program = program
         self.clocks = 0
+        # List of active file descriptors.
```
mossberg: can rm this comment now

manticore/platforms/linux.py
```diff
@@ -986,43 +1004,43 @@ def sys_lseek(self, fd, offset, whence):
...
-         '''
-        if not self._is_open(fd):
-            logger.info("LSEEK: Not valid file descriptor on lseek. Returning EBADF")
-            return -errno.EBADF
-
-        if isinstance(self.files[fd], Socket):
-            logger.info("LSEEK: Not valid file descriptor on lseek. Fd not seekable. Returning EBADF")
+        '''
+        try:
+            # Read the data and put it in memory
```
mossberg: this comment is an incorrect copy paste i think, we can rm

PR #304, opened 2017-06-12
(wip) removing capstone dependencies

nettrino:


manticore/core/cpu/abstractcpu.py
```diff
@@ -1,64 +1,71 @@
...
-        operand.
+        operands and immediates based on the dissasembler operand descriptor in
+        use. This class knows how to browse an operand and get its details.
 
         It also knows how to access the specific Cpu to get the actual values
         from memory and registers.
 
         :param Cpu cpu: A Cpu instance
-        :param op: A Capstone operand
+        :param op: An operand instance
```
yan: Use Sphinx annotations here, i.e.: `:param Operand op: A wrapped Instruction operand`

manticore/core/cpu/abstractcpu.py
```diff
@@ -1,64 +1,71 @@
...
         :type op: X86Op or ArmOp
         '''
         assert isinstance(cpu, Cpu)
-        assert isinstance(op, (X86Op, ArmOp))
+        # Check against the list of supported operands
+        assert isinstance(op, (capstone.x86.X86Op, capstone.arm.ArmOp))
```
yan: Does this belong here? Would we have to add a new class to the list as we add support for non-Capstone operands? Sounds like this should just be `isinstance(op, Operand)` (or even assumed, the assert might be extraneous)

manticore/core/cpu/abstractcpu.py
```diff
@@ -102,15 +109,15 @@ def write(self, value):
         ''' It writes the value of specific type to the registers or memory '''
         raise NotImplementedError
 
-# Basic register file structure not actually need to abstract as it's used only from the cpu implementation
 class RegisterFile(object):
-
+    """
```
yan: This should probably be a comment, not a docstring.

manticore/core/cpu/abstractcpu.py
```diff
@@ -552,8 +555,8 @@ def pop_int(self):
     @abstractmethod
     def _wrap_operands(self, operands):
         '''
-        Private method to decorate a capstone Operand to our needs. See Operand
-        class
+        Private method to decorate an Operand to our needs based on the
+        underlying architecture. See Operand class
```
yan:
```
See :class:`~manticore.core.cpu.abstractcpu.Operand` class.
```

manticore/core/cpu/disasm.py
```diff
@@ -0,0 +1,126 @@
+from abc import abstractmethod
+
+from capstone import Cs
+
+class Insn(object):
```
yan: Rename to `Instruction`?

manticore/core/cpu/disasm.py
```diff
@@ -0,0 +1,126 @@
...
+        pass
+
+class Disasm(object):
+    """Abstact class for different disassembler interfaces"""
+
+    def __init__(self, disasm):
+        self.disasm = disasm
+
+    @abstractmethod
+    def get_insn(self, code, pc):
```
yan: Rename to `get_instruction` or `disassemble_instruction`?

manticore/core/cpu/abstractcpu.py
```diff
@@ -582,35 +585,37 @@ def decode_instruction(self, pc):
...
-                c = self.memory[pc+i]
+                c = self.memory[pc + i]
                 if issymbolic(c):
-                    assert isinstance(c, BitVec) and  c.size == 8
+                    assert isinstance(c, BitVec) and c.size == 8
                     if isinstance(c, Constant):
                         c = chr(c.value)
                     else:
-                        logger.error('Concretize executable memory %r %r', c, text )
+                        logger.error('Concretize executable memory %r %r',
```
yan: Collapse to single line? Spacing feels gratuitous :)

manticore/core/cpu/abstractcpu.py
nettrino: That was basically left in place, but once our own Instruction/Operand class is complete, for each disassembler we will retrofit the instructions to fit that, so this will be removed once capstone operands are also ported to the Manticore operands

PR #308, opened 2017-06-15
Changed Socket parameters to accept seek paramaters

Sheidbrink:
If you run `manticore crackme` you'll get the error `seek() takes exactly 1 argument (2 given)`, this is because the `seek` function in the `Socket` class of `linux.py` is implemented without accepting any parameters. I changed this to match the other seek declarations.

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -257,7 +257,7 @@ class Socket(object):
     def sync(self):
         raise BadFd("Invalid sync() operation on Socket")
 
-    def seek(self):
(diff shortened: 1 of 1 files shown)


dguido:
Hey @Sheidbrink, we're going back and collecting CLAs for past contributors of Manticore but we can't seem to find your email. Can you do us a favor and click through this link to sign the Manticore CLA? Thanks! We're happy to accept any more contributions you want to make :-D

https://cla-assistant.io/trailofbits/manticore

Issue #311, opened 2017-06-19
Nonuniform state numbers


mossberg:
When manticore forks, state id's are not assigned in a uniform, incremental way. In the below run of `examples/linux/basic`, we can see that the state ids are 0, 3, and 5, when they should intuitively be 0, 1, and 2. This makes the state id and incorrect source for the actual number of states that were created.

This happens because

1. In the State ctor, `get_new_id` is always called. When state are loaded (unpickled), this ctor executes, incrementing the global state count unnecessarily.
2. In State `__enter__` (copying mechanism), `get_new_id` is explicitly called, in addition to the ctor, which unnecessarily bumps the state count.

There are fixes for these in the dev-events branch, as a result of refactoring this and moving state counting to the executor.

```
2017-06-19 19:10:32,984: [24704] EXECUTOR:DEBUG: Workspace set: /mnt/hgfs/code/manticore/examples/linux/mcore_fjK04M
2017-06-19 19:10:33,006: [24704] EXECUTOR:DEBUG: Saving state 0 to file state_000000.pkl
2017-06-19 19:10:33,013: [24704] EXECUTOR:DEBUG: Adding state state_000000.pkl to processing list. State list size: 1
2017-06-19 19:10:33,013: [24704] EXECUTOR:DEBUG: STAT:      name  branches     dicount    filesize       forks      icount   nsyscalls          pc        proc    received        smem     symbols    syscount  transmited
2017-06-19 19:10:33,016: [24704] EXECUTOR:DEBUG: STAT:         0OrderedDict()           0      203837           0           0           0        None        None           0           0           0           0           0
2017-06-19 19:10:33,031: [24761] EXECUTOR:DEBUG: Starting Manticore Symbolic Emulator mainloop (pid 24761).
2017-06-19 19:10:33,033: [24761] EXECUTOR:DEBUG: Prioritizing metric max [1.0]
2017-06-19 19:10:33,034: [24761] EXECUTOR:DEBUG: Selecting a new state to analyze /mnt/hgfs/code/manticore/examples/linux/mcore_fjK04M/state_000000.pkl. Using policy: random. Processing list size is 1
2017-06-19 19:10:33,043: [24761] EXECUTOR:DEBUG: Removing state 0 from storage
2017-06-19 19:10:33,044: [24761] EXECUTOR:DEBUG: Selected state: 0 (1.0)
[...]

PR #312, opened 2017-06-19
Fix state count increment bug

mossberg:
This addresses the second issue #311.


mossberg:
fixed in 241 / dev-events which we are merging soon

Issue #333, opened 2017-06-19
Add a multifunctiona context dict to State


feliam:
A state context is used in the wiki about exploit generation. Lets handle it in manticore nicely/
Also move all statistic and other variablesfrom State to a state context.  Need to handle that via callbacks or plugins.

       # Stats I'm not sure we need in general
        self.last_pc = (None, None)
        self.visited = set()
        self.branches = OrderedDict()

PR #336, opened 2017-06-20
Add a context dict to state

japesinator:
As described in #333, but doesn't yet refactor `last_pc` etc. into `state.context`


japesinator:
Should be unnecessary once #341 is merged

PR #337, opened 2017-06-20
Feature changes to support concrete execution.

douggard:
- Add ctypes to lseek for signed seeks. Pass whence to seek. Fixes edge case bug where 64 bit int is too big to fit in Python's long.
- Move mmap and related functions from SLinux to Linux.
- Introduce function _sys_open_get_file to isolate symbolic code in sys_open to SLinux.

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -4,6 +4,7 @@ import logging
 import os
 import random
 import struct
+import ctypes
 
 from elftools.elf.elffile import ELFFile
 
@@ -276,7 +277,7 @@ class Linux(Platform):
         :type files: list[Socket] or list[File]
         '''
         super(Linux, self).__init__(program)
-
+        
         self.program = program
         self.clocks = 0
         self.files = []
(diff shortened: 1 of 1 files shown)


manticore/platforms/linux.py
```diff
@@ -1004,14 +1005,20 @@ def sys_lseek(self, fd, offset, whence):
         :return: 0 (Success), or EBADF (fd is not a valid file descriptor or is not open)
 
         '''
+        addressbitsize = {'i386':32, 'x86_64':64, 'armv71': 32}[self._uname_machine]
+        if addressbitsize == 32:
```
yan: The canonical way to access the width of the machine from the platform is via `self.current.address_bit_size`

manticore/platforms/linux.py
```diff
@@ -1945,6 +2035,16 @@ def __setstate__(self, state):
...
         super(SLinux, self).__setstate__(state)
 
+    def _sys_open_get_file(self, filename, flags, mode):
+        if filename in self.symbolic_files:
+            logger.debug("%s file is considered symbolic", filename)
+            assert flags & 7 == os.O_RDWR or flags & 7 == os.O_RDONLY, (
+                "Symbolic files should be readable?")
+            f = SymbolicFile(self.constraints, filename, mode)
+        else:
+            f = File(filename, mode) # TODO (theo) modes, flags
```
yan: call `super._sys_open_get_file()` here instead of reimplementing?

manticore/platforms/linux.py
```diff
@@ -276,7 +277,7 @@ def __init__(self, program, argv=None, envp=None):
         :type files: list[Socket] or list[File]
         '''
         super(Linux, self).__init__(program)
-
+
```
yan: Extra whitespace

douggard:
Changes made in last commit.

PR #341, opened 2017-06-20
Dev - events

feliam:
Deja vu!


manticore/core/cpu/abstractcpu.py
```diff
@@ -1,22 +1,79 @@
...
+            
+        self.cpu = cpu
+        self.reg_name = reg_name
+        self.policy = policy
+
+class ConcretizeArgument(CpuException):
+    '''
+    Raised when a symbolic argument needs to be concretized.
+    '''
+    def __init__(self, cpu, argnum, policy='MINMAX'):
```
mossberg: in decree, linux, models, and windows, we forgot to pass the cpu argument

manticore/core/executor.py
```diff
@@ -69,23 +46,32 @@ def newFunction(self, *args, **kw):
...
-        self.solver_time = 0
-        for (func_file, _, func_name), (_, _, _, func_time, _) in raw_stats.stats.iteritems():
-            if func_name == '_getState':
-                self.loading_time += func_time
-            elif func_name == '_putState':
-                self.saving_time += func_time
-            elif func_file.endswith('solver.py') and 'setstate' not in func_name and 'getstate' not in func_name and 'ckl' not in func_name:
-                self.solver_time += func_time
 
     def __str__(self):
```
mossberg: forgot to move this function?

manticore/core/executor.py
```diff
@@ -98,164 +84,243 @@ def __str__(self):
...
+        self.put(state_id)
+        return state_id
+
+    def load_workspace(self):
+        #Browse and load states in a workspace in case we are trying to 
+        # continue from paused run
+        saved_states = []
+        for filename in os.listdir(self.workspace):
+            if filename.startswith('state_') and filename.endswith('.pkl'):
+                saved_states.append(self._getFilename(filename))
```
mossberg: there is no more `_getFilename`

manticore/core/executor.py
```diff
@@ -98,164 +84,243 @@ def __str__(self):
...
+
+        #search finalized testcases
+        saved_testcases = []
+        for filename in os.listdir(self.workspace):
+            if filename.startswith('test_') and filename.endswith('.pkl'):
+                saved_testcases.append(self._getFilename(filename)) 
+
+        if saved_states or saved_testcases:
+            #We are trying to continue a paused analysis
+            if initial is not None:
```
mossberg: initial never assigned

manticore/core/executor.py
```diff
@@ -98,164 +84,243 @@ def __str__(self):
...
+                raise ArgumentError('Cant set initial state when continuing from previous workspace: {}'.format(workspace))
```
mossberg: ArgumentError never declared

mossberg: workspace also not declared

manticore/core/executor.py
```diff
@@ -98,164 +84,243 @@ def __str__(self):
...
+            #with open(filename, "r") as f:
+            #    temp_state = cPickle.loads(f.read())
+            #    self.policy.measure(temp_state)
+            state_id = int(filename[6:-4])
+            self._states.append(state_id)
+
+        #reset test and states counter 
+        for filename in saved_states:
+            state_id = int(filename[6:-4])
+            self._state_counter.value = max(self._state_counter.value, state_id)
```
mossberg: self._state_counter doesn't exist

manticore/core/executor.py
```diff
@@ -98,164 +84,243 @@ def __str__(self):
...
+
+        for filename in saved_testcases:
+            state_id = int(filename[6:-4])
+            self._test_counter.value = max(self._test_counter.value, state_id)
```
mossberg: self._test_counter doesn't exist either

manticore/core/executor.py
```diff
@@ -98,164 +84,243 @@ def __str__(self):
...
+    # Synchronization helpers
+    @sync
+    def _start_run(self):
+        #notify siblings we are about to start a run()
+        self._running.value+=1
 
     @sync
-    def delState(self, state):
-        return self._delState(state)
+    def _stop_run(self, count=0):
```
mossberg: count unused?

manticore/core/executor.py
```diff
@@ -268,162 +333,30 @@ def _putState(self, state):
             filesize = f.tell()
```
mossberg: filesize unused now

manticore/core/executor.py
```diff
@@ -455,412 +387,167 @@ def generate_testcase(self, state, message = 'Testcase generated'):
...
         '''
-        if self.profiling:
-            self._profile.enable()
 
-        policy_order=self.policy_order
-        policy=self.policy
+        #policy_order=self.policy_order
+        #policy=self.policy
 
         count = 0
```
mossberg: count unused

manticore/core/executor.py
```diff
@@ -455,412 +387,167 @@ def generate_testcase(self, state, message = 'Testcase generated'):
...
-            except Exception as e:
-                import traceback
-                trace = traceback.format_exc()
-                logger.error("Exception: %s\n%s", str(e), trace)
-                for log in trace.splitlines():
-                    logger.error(log) 
+            if current_state != None:
+                #Notify this worker is done
+                self.will_terminate_state(current_state, current_state_id, 'Shutdown')
                 current_state = None
```
mossberg: current_state now unused

manticore/core/state.py
```diff
@@ -64,37 +100,48 @@ def cpu(self):
...
-
     def __enter__(self):
         assert self._child is None
         new_state = State(self.constraints.__enter__(), self.platform)
-        new_state.visited = set(self.visited)
-        new_state.forks = self.forks + 1
-        new_state.co = State.get_new_id()
         new_state.input_symbols = self.input_symbols
-        new_state.branches = self.branches
+        import copy
```
mossberg: import above?

manticore/core/cpu/abstractcpu.py
```diff
@@ -549,89 +641,99 @@ def pop_int(self):
...
         '''
         Decode, and execute one instruction pointed by register PC
         '''
+        if issymbolic(self.PC):
+            raise ConcretizeRegister(self, 'PC', policy='ALL')
+
+        if not self.memory.access_ok(self.PC,'x'):
+            raise InvalidMemoryAccess(self.PC, 'x')
+
+        self.will_decode_instruction()
```
feliam: Problem is that cpu.instruction is a property that decodes from where the PC points to at that moment. And it may be used in odd places. All signal emits are in question. Lets do a pass after merge and maybe after used a bit.

manticore/core/executor.py
feliam: The semantics of putstate are divided into put() and store() so I'm not sure that is still relevant. I think jp was profiling via common cProfiler, no? I vote to remove.

manticore/core/cpu/abstractcpu.py
mossberg: I see, `decode_instruction` might be used in other places, including when the emulator is not "officially" decoding the instruction. 👍

manticore/core/cpu/abstractcpu.py
mossberg: (addressed)

manticore/platforms/windows.py
```diff
@@ -174,14 +168,17 @@ def __init__(self, path, additional_context = None, snapshot_folder=None):
                 self.running.append(self.procs.index(cpu))
         
 
-        self._function_abi = I386StdcallAbi(self.procs[0])
```
mossberg: (addressed)

manticore/manticore.py
```diff
@@ -6,25 +6,46 @@
...
+        self.raw_stats = raw_stats
+        self.instructions_executed = instructions_executed
+
+        self.time_elapsed = raw_stats.total_tt
+
+        self.loading_time = 0
+        self.saving_time = 0
+        self.solver_time = 0
+        for (func_file, _, func_name), (_, _, _, func_time, _) in raw_stats.stats.iteritems():
+            if func_name == 'load':
```
mossberg: is this going to track every function call named `load` or `store`, even if they're not related to states? for example `Linux.load`?

manticore/core/executor.py
mossberg: why remove? don't we still want to have profiling data on how long state loading and storing takes?

feliam: Full standard profiling information is still gathered in a single file. Then you can use pstat.py util to analyze it.

manticore/manticore.py
feliam: yes. that's bad. I think we should not do any of that self profiling here.

manticore/core/executor.py
mossberg: can we add a regression test for these small bugs? even the most basic test of this function would have failed because the bugs were very simple (nonexistent functions/variables)

manticore/manticore.py
```diff
@@ -534,43 +787,102 @@ def run(self, procs=1, timeout=0):
...
+        if self.should_profile:
+
+            class PstatsFormatted:
+                def __init__(self, d):
+                    self.stats = dict(d)
+                def create_stats(self):
+                    pass
+
+            ps = None
+            for item in self._stats:
```
mossberg: `_stats` is undeclared. this breaks when you use `manticore --profile`

feliam: This will need a test too.

manticore/core/executor.py
mossberg: Let's discuss this in a new PR? Since master is no longer stable, imo it's fine if we merge some slightly buggy profiling code, and make a new PR to refactor and fix .

mossberg:
In general, this is looking fine to me! I would like to see this nitpick fixed (https://github.com/trailofbits/manticore/pull/341#discussion_r123361383), and I think it would be good to add some regression tests for a lot of these very basic bugs since the tests clearly don't test the code those were in.

Also let's not forget to make an issue for profiling refactoring/discussion.

manticore/core/executor.py
mossberg: either way, this `__str__` method definitely doesn't belong under `Random`, so if `ProfilingResults` was removed, it should be too, right?

feliam: AAHHH

tests/test_binaries.py
```diff
@@ -57,19 +57,13 @@ def testTimeout(self):
...
-        2017-06-21 17:14:30,369: [111762][3] EXECUTOR:INFO: Generating testcase No. 1 for state No.3 - Program finished correctly
-        2017-06-21 17:14:31,294: [111762][5] EXECUTOR:INFO: Generating testcase No. 2 for state No.5 - Program finished correctly
+        Tests that default verbosity produces the expected volume of output
         """
 
         dirname = os.path.dirname(__file__)
         filename = os.path.join(dirname, 'binaries/basic_linux_amd64')
         output = subprocess.check_output(['python', '-m', 'manticore', filename])
-        self.assertEqual(len(output.splitlines()), 4)
+        self.assertEqual(len(output.splitlines()), 25)
```
mossberg: This an excerpt of the default output I get from `manticore basic`

```
2017-06-23 19:24:00,012: [48164] MAIN:INFO: Loading program: ['basic']
2017-06-23 19:24:00,015: [48164] MAIN:INFO: Workspace: /mnt/hgfs/code/manticore/examples/linux/mcore_P1jq2y
2017-06-23 19:24:00,015: [48164] MANTICORE:INFO: Loading program basic
2017-06-23 19:24:00,808: [48164] MANTICORE:INFO: Starting 1 processes.
2017-06-23 19:24:00,819: [48200] EXECUTOR:INFO: Restoring state: 1 from /mnt/hgfs/code/manticore/examples/linux/mcore_P1jq2y/state_000001.pkl
2017-06-23 19:24:00,829: [48200] EXECUTOR:INFO: Removing state 1 from storage
2017-06-23 19:24:02,706: [48200] EXECUTOR:INFO: Generic state fork on condition
2017-06-23 19:24:02,762: [48200] EXECUTOR:INFO: Restoring state: 3 from /mnt/hgfs/code/manticore/examples/linux/mcore_P1jq2y/state_000003.pkl
2017-06-23 19:24:02,772: [48200] EXECUTOR:INFO: Removing state 3 from storage
2017-06-23 19:24:03,427: [48200] MANTICORE:INFO: Terminate_state_callback for 3
2017-06-23 19:24:03,429: [48200] EXECUTOR:INFO: Generic terminate state
2017-06-23 19:24:03,429: [48200] EXECUTOR:INFO: Generating testcase No. 1  - ProcessExit 0
2017-06-23 19:24:03,431: [48200] MANTICORE:INFO: Generating testcase No. 1 - ProcessExit 0
```

- we have two "loading program" log messages, let's rm one
- imo messages like "Generic state fork on condition", "Terminate_state_callback for 3", "Generic terminate state" and possible "Restoring/removing state" are a little too low level and obtuse for no verbosity mode. Especially if we are considering a new user unfamiliar with the idea of states or forking. i'd vote to keep the default output as primarily just the "Generating testcase" messages

PR #347, opened 2017-06-22
Add basic binja viz plugin

mossberg:


scripts/binaryninja/README.md
```diff
@@ -0,0 +1,18 @@
+To use a plugin:
+
+1. Symlink the desired plugin into the directory specified here: https://github.com/Vector35/binaryninja-api/tree/dev/python/examples#loading-plugins
```
nettrino: Perhaps it is cleaner to have an inline link with 

[into the BinaryNinja plugin directory](https://github.com/Vector35/binaryninja-api/tree/dev/python/examples#loading-plugins)

scripts/binaryninja/README.md
```diff
@@ -0,0 +1,18 @@
...
+```
+$ ln -s $PWD/viz.py ~/Library/Application\ Support/Binary\ Ninja/plugins/
+
+```
+
+From script console:
+
+```
+>>> import viz
+>>> viz.viz(bv, '/mnt/hgfs/code/manticore/examples/linux/mcore_1vCAKM')
```
nettrino: mcore_ is a file that is generated by manticore but is not in the examples/ by default right?
It would be super helpful if we include an actual example_dir for this script particularly,
and also a screenshot of the plugin?

scripts/binaryninja/viz.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+MBASE = 0x0007fffffdb7000
+
+
+def viz(view, fname, base=0):
+    """
+    Given a Manticore workspace, or trace file, highlight the basic blocks.
+    """
+    fname = fname.replace('/mnt/hgfs', '~')
```
nettrino: Is this specific to one machine?

scripts/binaryninja/README.md
nettrino: Also we should mention at some point that you need to have the same binary loaded in Binja
as the one that the manticore analysis was run on :P

scripts/binaryninja/viz.py
mossberg: Yeah, this is just a convenience because I run Manticore in a ubuntu vm and binja in my mac. I can use `realpath` to get the global path in my vm and can use it directly in binja.

scripts/binaryninja/README.md
mossberg: I improved the instructions, i do feel like it's a bit overkill with the example dir/screenshot though?

PR #348, opened 2017-06-22
Add ldrd and strd instructions and tests

GarretReece:
Does what it says on the tin.


mossberg:
@GarretReece feel free to merge at will

PR #350, opened 2017-06-22
lsl.w in thumb mode fixed, plus tests

GarretReece:
Fixes lsl.w in thumb mode and includes a test that exercised the problematic behavior.

The BX instruction now sets the processor mode appropriately for disassembly.

### DIFF
--- a/manticore/core/cpu/arm.py
+++ b/manticore/core/cpu/arm.py
@@ -331,9 +331,9 @@ class Armv7Cpu(Cpu):
         self._force_next = state['_force_next']
 
     def _set_mode(self, new_mode):
-		assert new_mode in (CS_MODE_ARM, CS_MODE_THUMB)
-		self.mode = new_mode
-		self._md.mode = new_mode
+        assert new_mode in (CS_MODE_ARM, CS_MODE_THUMB)
+        self.mode = new_mode
+        self._md.mode = new_mode
 
     def _swap_mode(self):
         #swap from arm to thumb or back
--- a/manticore/core/cpu/arm.py
+++ b/manticore/core/cpu/arm.py
@@ -335,6 +335,15 @@ class Armv7Cpu(Cpu):
 		self.mode = new_mode
 		self._md.mode = new_mode
 
+    def _swap_mode(self):
+        #swap from arm to thumb or back
+        assert self.mode in (CS_MODE_ARM, CS_MODE_THUMB)
+        if self.mode == CS_MODE_ARM:
+            self._set_mode(CS_MODE_THUMB)
+        else:
+            self._set_mode(CS_MODE_ARM)
+
+
     # Flags that are the result of arithmetic instructions. Unconditionally
     # set, but conditionally committed.
     #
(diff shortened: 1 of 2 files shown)


manticore/core/cpu/arm.py
```diff
@@ -731,6 +740,10 @@ def B(cpu, dest):
     # XXX How should we deal with switching Thumb modes?
```
mossberg: we can rm this comment now, right?

manticore/core/cpu/arm.py
```diff
@@ -760,13 +773,7 @@ def BLX(cpu, dest):
         ## The `blx <label>` form of this instruction forces a state swap
```
mossberg: is the XXX/TODO comment above still accurate? since we do support thumb can it be rm'd?

manticore/core/cpu/arm.py
```diff
@@ -907,14 +914,20 @@ def _SR(cpu, insn_id, dest, op, *rest):
         if insn_id == ARM_INS_ASR:
             srtype = ARM_SFT_ASR_REG
         elif insn_id == ARM_INS_LSL:
-            srtype = ARM_SFT_LSL_REG
+            if rest and rest[0].type == 'immediate':
```
mossberg: since this issue occurs for asr and lsr also, can we fix those insns here also?

manticore/core/cpu/arm.py
```diff
@@ -907,14 +914,20 @@ def _SR(cpu, insn_id, dest, op, *rest):
         if insn_id == ARM_INS_ASR:
```
mossberg: can we update the docstring to say that what's there only applies to arm, not thumb?

manticore/core/cpu/arm.py
```diff
@@ -335,6 +335,15 @@ def _set_mode(self, new_mode):
 		self.mode = new_mode
 		self._md.mode = new_mode
 
+    def _swap_mode(self):
+        #swap from arm to thumb or back
+        assert self.mode in (CS_MODE_ARM, CS_MODE_THUMB)
+        if self.mode == CS_MODE_ARM:
+            self._set_mode(CS_MODE_THUMB)
```
mossberg: in the future (not this pr), perhaps we could use a property instead of a setter for mode

manticore/core/cpu/arm.py
mossberg: (addressed)

tests/test_armv7cpu.py
```diff
@@ -1244,21 +1264,10 @@ def test_lsl_imm_carry(self):
...
-        self.stack = self.mem.mmap(0xf000, 0x1000, 'rw')
-        start = self.code + 4
-        self.mem.write(start, asm)
-        self.rf.write('PC', start)
-        self.rf.write('SP', self.stack + 0x1000)
-        self.rf.write('R5', 0x1)
-        self.rf.write('R6', 0x2)
-        self.cpu._set_mode(CS_MODE_THUMB)
+    @itest_setregs("R5=1", "R6=2")
+    @itest_custom_thumb("lsl.w r5, r6, #3")
```
mossberg: 👏

mossberg: if you wanted, you could even make an `itest_thumb` as a counterpart to regular `itest`. the only difference is that the decorator calls cpu.execute so the test can simply assert.

PR #368, opened 2017-06-29
New Policy class

feliam:
Example PoC of a Policy implementation:: Uncovered


manticore/core/executor.py
```diff
@@ -48,29 +48,69 @@ def newFunction(self, *args, **kw):
...
+        ''' Save prepare(state) on policy shared context before 
+            the state is stored
+        '''
+        with self.locked_context() as ctx:
+            metric = self.prepare(state)
+            if metric is not None:
+                ctx[state_id] = metric
+
+    def prepare(self, state):
+        ''' Process a state and keep enough data to later decide it's
```
japesinator: "Extract the relevant information from a state for later prioritization"?

manticore/core/executor.py
```diff
@@ -48,29 +48,69 @@ def newFunction(self, *args, **kw):
...
+            priority #fixme rephrase
+        '''
+        return None
 
-    def priority(self, state_id):
-        ''' A numeric value representing likelihood to reach the interesting program spot '''
-        return 1.0
+    def choice(self, state_ids):
+        ''' Select a state id from states_id.
```
japesinator: Nit, should be `state_ids`

manticore/core/executor.py
```diff
@@ -48,29 +48,69 @@ def newFunction(self, *args, **kw):
...
+        #hook on the necesary executor signals 
+        #on callbacks save data in executor.context['policy']
+
+    def prepare(self, state):
+        ''' this is what we need to save for choosing later '''
+        return state.cpu.PC
+
+    def choice(self, state_ids):
+        # Use executor.context['uncovered'] = state_id -> stats
+        # am
```
japesinator: What's this line?

manticore/core/executor.py
```diff
@@ -48,29 +48,69 @@ def newFunction(self, *args, **kw):
...
+    def __init__(self, executor, *args, **kwargs):
+        super(Policy, self).__init__(*args, **kwargs)
+        self._executor = executor
+        self._executor.did_add_state += self._add_state_callback
+        
+    @contextmanager
+    def locked_context(self):
+        ''' Policy shared context dictionary '''
+        with self._executor.locked_context() as ctx:
+            policy_context = ctx.get('policy', None)
```
feliam: I think that will create a dict even if we do not need it.

japesinator: I just tested and it will do that. As such, I think we should just use `get('policy')` as it's equivalent to `get('policy', None)`

manticore/core/executor.py
```diff
@@ -48,29 +48,84 @@ def newFunction(self, *args, **kw):
...
+    def locked_context(self, key=None, default=dict):
+        ''' Policy shared context dictionary '''
+        keys = ['policy']
+        if key is not None:
+            keys.append(key)
+        with self._executor.locked_context('.'.join(keys), default) as policy_context:
+            yield policy_context
+
+    def _add_state_callback(self, state_id, state):
+        ''' Save sumarize(state) on policy shared context before
```
japesinator: Nit, should be summarize

manticore/core/executor.py
```diff
@@ -48,29 +48,84 @@ def newFunction(self, *args, **kw):
...
+        ''' Save the last pc before storing the state '''
+        return state.cpu.PC
+
+    def choice(self, state_ids):
+        interesting = set()
+        with self.locked_context() as policy_ctx: 
+            visited = policy_ctx.get('visited', set())
+            lastpc = policy_ctx.get('summaries', dict())
+            for _id in state_ids:
+                if lastpc.get(_id, None) not in visited:
```
japesinator: `get(x, None)` is equivalent to just `get(x)` (this happens in a number of places)

manticore/core/executor.py
```diff
@@ -145,8 +207,24 @@ def locked_context(self):
...
                 context['visited'] = visited
         '''
+        assert default in (list, dict, set)
         with self._lock:
-            yield self._shared_context
+            if key is None:
+                yield self._shared_context
+            elif '.' in key:
+                keys = key.split('.')
+                with self.locked_context('.'.join(keys[:-1])) as sub_context:
```
japesinator: I'm not sure string concatenation is the best way to handle the multiple key situation, can we try a list of keys?

yan: (Commenting for review) The concatenation is part of a shortcut syntax to use nested dicts. I.e. `locked_context('foo.bar')` will write to `_shared_context['foo']['bar']` and initialize `_shared_context['foo']` to `{}` if it doesn't exist.

PR #376, opened 2017-07-07
New state serialization

yan:


manticore/core/executor.py
```diff
@@ -299,76 +250,23 @@ def get(self):
...
 
     def list(self):
         ''' Returns the list of states ids currently queued '''
         return list(self._states)
 
     def generate_testcase(self, state, message='Testcase generated'):
         '''
-        Create a serialized description of a given state.
+        Simply announce that we're going to generate a testcase. Actual generation
+        should be handled by the driver class (such as :class:`~manticore.Manticore`)
```
feliam: Maybe better to just remove this one and just emit the event?

manticore/core/workspace.py
```diff
@@ -0,0 +1,458 @@
...
+        return state_ids
+
+    def _get_id(self):
+        '''
+        Get a unique state id.
+
+        :rtype: int
+        '''
+        try:
+            self._lock.acquire()
```
feliam: I wonder if we need to use the  global executor lock for this or if there is a locking mechanism more close to the workspace implementation..

feliam: Also I think "with lock as.." instead of try-finally is used elsewhere. no?

manticore/core/workspace.py
```diff
@@ -0,0 +1,458 @@
...
+        Save the environment from `state` to storage. Return a state id
+        describing it, which should be an int or a string.
+
+        :param State state: The state to serialize
+        :param int testcase_id: Identifier for the state
+        :param str state: Optional message to include
+        :return: A state id representing the saved state
+        '''
+
+        self.save_summary(state, message)
```
feliam: All this should be plug-able from manticore to enable platforms that do not have FDs or configurations that didn't want to spend time maintaining a trace.

manticore/core/workspace.py
```diff
@@ -0,0 +1,458 @@
...
+        self.save_trace(state)
+        self.save_constraints(state)
+        self.save_input_symbols(state)
+        self.save_syscall_trace(state)
+        self.save_fds(state)
+        key = 'test_{:08x}.pkl'.format(self._last_id)
+        self._store.save_state(state, key)
+
+        self._last_id += 1
```
feliam: Lock?

manticore/core/workspace.py
```diff
@@ -0,0 +1,458 @@
...
+                else:
+                    summary.write("  Instruction: {symbolic}\n")
+
+    def save_trace(self, state):
+        with self._named_stream('trace') as f:
+            for pc in state.context['visited']:
+                f.write('0x{:08x}\n'.format(pc))
+
+    def save_constraints(self, state):
+        # XXX(yan): We want to conditionally enable this check
```
feliam: We may also just remove it(assert) as it's going to fail later when trying to solve the paths constraint anyway

manticore/core/workspace.py
yan: `Manticore` locks the context before invoking `generate_testcase`

feliam: Uh is that right? Do we want to lock the executor every time we solve a testcase? Sounds like a test case generation should work only over a state. (except for last id?)

manticore/core/memory.py
```diff
@@ -139,6 +139,9 @@ def __repr__(self):
         '''
         return '<%s 0x%016x-0x%016x %s>'%(self.__class__.__name__, self.start, self.end, self.perms)
 
+    def __cmp__(self, other):
+        return cmp(self.start, other.start)
```
feliam: Shouldn't we check end and perms and name. 
m1 == m2 ?

PR #377, opened 2017-07-07
Fix memdump tests broken by dev-events

japesinator:
`testWin32API` is still skipped because it takes about 300s on my machine


manticore/platforms/windows.py
```diff
@@ -175,14 +175,17 @@ def __init__(self, path, additional_context = None, snapshot_folder=None):
...
         self._current = self.running[0]
-        self._function_abi = I386StdcallAbi(self.procs[0])
 
         #Install event forwarders
         for proc in self.procs:
             forward_signals(self, proc)
         
 
     @property
+    def _function_abi(self):
```
nettrino: perhaps rename removing "_" so that it is not private?

japesinator: I think the non-private method for this is [`invoke_model`](https://github.com/trailofbits/manticore/blob/520a9be47d4df0cdb4990c3b4f81318574bde8ed/manticore/platforms/platform.py#L32)

PR #380, opened 2017-07-11
Fix small bugs in evm 

montyly:
Bugs in opcodes (too many arguments + wrong LOG name)
Fix wrong call parameters + typo


dguido:
I'm just merging this considering it's a development branch and not master.

Issue #379, opened 2017-07-11
32 bit ELF not parsed correctly


sinaelgl:
### OS / Environment

Distributor ID:	Debian
Description:	Debian GNU/Linux 9.0 (n/a)
Release:	9.0
Codename:	n/a

### Python version

Python 2.7.12+

### Summary of the problem
Some 32 bit binaries crash manticore. I compiled the same source for 32 and 64 and the 64 version generates correct input under manticore but the 32 version just errors out.

[binary.zip](https://github.com/trailofbits/manticore/files/1138196/binary.zip)

### Step to reproduce the behavior
unzip binary.zip
manticore ./hash1_gcc49_32

### Expected behavior
$ manticore ./hash1_gcc6_64 
2017-07-11 11:26:51,840: [14851] MANTICORE:INFO: Loading program ./hash1_gcc6_64
2017-07-11 11:26:52,951: [14851] MANTICORE:INFO: Starting 1 processes.
2017-07-11 11:27:51,550: [14882] EXECUTOR:INFO: Generating testcase No. 1 - Program finished with exit status: 0L
2017-07-11 11:27:52,913: [14882] EXECUTOR:INFO: Generating testcase No. 2 - Program finished with exit status: 0L
2017-07-11 11:28:31,458: [14882] EXECUTOR:INFO: Generating testcase No. 3 - Program finished with exit status: 0L
2017-07-11 11:28:32,067: [14851] MANTICORE:WARNING: Using shared context without a lock
2017-07-11 11:28:32,081: [14851] MANTICORE:INFO: Results in example/mcore_EBOvrw
2017-07-11 11:28:32,081: [14851] MANTICORE:INFO: Instructions executed: 308358
2017-07-11 11:28:32,081: [14851] MANTICORE:INFO: Coverage: 8549 different instructions executed
2017-07-11 11:28:32,081: [14851] MANTICORE:INFO: Total time: 99.1297779083
2017-07-11 11:28:32,082: [14851] MANTICORE:INFO: IPS: 3110

### Actual behavior
$ manticore ./hash1_gcc49_32
2017-07-11 11:30:21,265: [15023] MANTICORE:INFO: Loading program ./hash1_gcc49_32
2017-07-11 11:30:22,344: [15023] MANTICORE:INFO: Starting 1 processes.
2017-07-11 11:30:26,051: [15054] EXECUTOR:ERROR: Exception: 'NoneType' object has no attribute 'stat'
Traceback (most recent call last):
[...]

PR #384, opened 2017-07-11
Check for negative fds in fstat

japesinator:
Should help #379 stop crashing


manticore/platforms/linux.py
```diff
@@ -1822,8 +1822,13 @@ def sys_newfstat(self, fd, buf):
         :rtype: int
         :param fd: the file descriptor of the file that is being inquired.
         :param buf: a buffer where data about the file will be stored.
-        :return: C{0} on success.
+        :return: C{0} on success, EBADF when called with bad fd
         '''
+
+        if fd < 0:
```
nettrino: Perhaps better to use `_get_fd` similar to `sys_write`, `sys_read`?

Issue #378, opened 2017-07-10
`state.generate_inputs()` broken


mossberg:
`state.generate_inputs` crashes because there is no `state.co` for the state id anymore.

```
2017-07-10 16:42:02,835: [85802] EXECUTOR:ERROR: Exception: 'State' object has no attribute 'co'
Traceback (most recent call last):
  File "/mnt/hgfs/code/manticore/manticore/core/executor.py", line 463, in run
    if not current_state.execute():
  File "/mnt/hgfs/code/manticore/manticore/core/state.py", line 119, in execute
    result = self.platform.execute()
  File "/mnt/hgfs/code/manticore/manticore/platforms/linux.py", line 1803, in execute
    self.current.execute()
  File "/mnt/hgfs/code/manticore/manticore/core/cpu/abstractcpu.py", line 735, in execute
    self.will_execute_instruction(instruction)
  File "/mnt/hgfs/code/manticore/manticore/utils/event.py", line 77, in __call__
    return self.emit(*args, **kwargs)
  File "/mnt/hgfs/code/manticore/manticore/utils/event.py", line 99, in emit
    results.append(f(obj, *args, **kwargs))
  File "/mnt/hgfs/code/manticore/manticore/utils/event.py", line 99, in emit
    results.append(f(obj, *args, **kwargs))
  File "/mnt/hgfs/code/manticore/manticore/utils/event.py", line 99, in emit
    results.append(f(obj, *args, **kwargs))
  File "/mnt/hgfs/code/manticore/manticore/utils/event.py", line 138, in <lambda>
    method = MethodType(lambda *args, **kwargs: self.emit(*args, **kwargs), obj)
  File "/mnt/hgfs/code/manticore/manticore/utils/event.py", line 99, in emit
    results.append(f(obj, *args, **kwargs))
  File "/mnt/hgfs/code/manticore/manticore/manticore.py", line 898, in _hook_callback
    cb(state)
  File "lads-baby-re.py", line 27, in myhook
    state.generate_inputs(m.workspace)
  File "/mnt/hgfs/code/manticore/manticore/core/state.py", line 323, in generate_inputs
    smtfile = 'state_{:08x}.smt'.format(self.co)
AttributeError: 'State' object has no attribute 'co'

```

mossberg:
More discussion about this in #373

japesinator:
`get_inputs` is deprecated in favor of the new `ManticoreOutput` features as of #385

mossberg:
[...]

Issue #373, opened 2017-07-06
State counter / identifier


montyly:
The commit 520a9be47d4df0cdb4990c3b4f81318574bde8ed removed the [state](https://github.com/trailofbits/manticore/blob/master/manticore/core/state.py) counter variable (`state.co`, assigned by the function `state.get_new_id()`). This variable is used as an identifier by `state.generate_inputs`, to generate unique filenames per state (see [smtfile](https://github.com/trailofbits/manticore/blob/master/manticore/core/state.py#L323), [input](https://github.com/trailofbits/manticore/blob/master/manticore/core/state.py#L333) and [input file](https://github.com/trailofbits/manticore/blob/master/manticore/core/state.py#L345)).

Is there any other identifier that can be used to differentiate the states?

feliam:
A State may contain a platform that does not have files. So maybe better to do it at manticore.py via events ?
Something like this: https://github.com/trailofbits/manticore/blob/master/manticore/core/executor.py#L344-L371

montyly:
Yes, it makes sense. 
Or maybe we can add a generic `generate_symbolic_inputs()` in [plateform](https://github.com/trailofbits/manticore/blob/master/manticore/platforms/platform.py), and then let each platform handling the possible symbolic inputs (which can be files, network, ..) for the targeted system?

mossberg:
I agree with @montyly, I think we could move the platform specific code here (https://github.com/trailofbits/manticore/blob/master/manticore/core/state.py#L340-L352) to the Platform, and replace that code with something like `self.platform.generate_symbolic_inputs()` (or some other name). That platform function would either actually write to the workspace itself, or return some data which is written to the workspace by `state.generate_inputs`.

mossberg:
@feliam what do you think of the above idea?

mossberg:
#438 is some work at reimplementing `state.generate_inputs` using the new workspace, output and events infrastructure. It does not separate platform specific code from testcase generation code yet. @feliam can you let me know what you think?

mossberg:
@montyly As of current Manticore master, `state.generate_testcase` exists (effectively replaces the old `state.generate_inputs`), so you can use that now!

PR #385, opened 2017-07-12
Remove `state.generate_inputs`

japesinator:
As of right now, this functionality is broken (see: #378, #373), and should arguably be more platform-specific (not all platforms necessarily have files)

### DIFF
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -311,43 +311,6 @@ class State(object):
             except KeyError:
                 self.branches[key] = 1
 
-    def generate_inputs(self, workspace, generate_files=False):
-        '''
-        Save the inputs of the state
-
-        :param str workspace: the working directory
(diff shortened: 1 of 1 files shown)


japesinator:
Fixes #378

manticore/core/state.py
```diff
@@ -311,43 +311,6 @@ def record_branches(self, targets):
             except KeyError:
                 self.branches[key] = 1
 
-    def generate_inputs(self, workspace, generate_files=False):
```
feliam: Instead of removing it consider moving this ->  https://github.com/trailofbits/manticore/blob/master/manticore/core/workspace.py#L389
Or is it already there?

(Then in another PR remove that class)

feliam: Nvm it seems that the same functionality is in ManticoreOutput

PR #386, opened 2017-07-12
Add basic documentation on function modeling

### LINKED ISSUE #362: Document function modeling and APIs
We should write documentation in the wiki or `docs/` about how to use these features.

mossberg:
Fixes #362

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -2,10 +2,10 @@ Function Models
 ===============
 
 The Manticore function modeling API can be used to override a certain
-function in the target program with an custom implementation in Python.
+function in the target program with a custom implementation in Python.
 This can greatly increase performance.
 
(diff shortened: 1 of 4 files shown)


docs/models.rst
```diff
@@ -0,0 +1,29 @@
+Function Models
+===============
+
+The Manticore function modeling API can be used to override a certain
+function in the target program with an custom implementation in Python.
```
yan: "a custom"

docs/models.rst
```diff
@@ -0,0 +1,29 @@
...
+This can greatly increase performance.
+
+Manticore comes with implementations of function models for some popular library routines (core models),
```
yan: is "popular" the right term? "common" maybe?

PR #394, opened 2017-07-17
Expose symbolic files to command line interface, implement whence for symbolic seek

251:
As requested in Slack:

* a new parameter `--file` to add symbolic files
* a very simple symbolic seek implementation (current `seek` lacks a `whence` parameter and fails with TypeError)

Edit: changed `symfile` to `file` for consistent naming

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -172,23 +172,28 @@ class SymbolicFile(File):
 
     def seek(self, offset, whence = os.SEEK_SET):
         '''
-        Returns the read/write file offset
+        Repositions the file C{offset} according to C{whence}.
+        Returns the resulting offset or -1 in case of error.
         :rtype: int
-        :return: the read/write file offset.
+        :return: the file offset.
         '''
         assert isinstance(offset, (int, long))
-        assert isinstance(whence, (int, long))
+        assert whence in (os.SEEK_SET, os.SEEK_CUR, os.SEEK_END)
 
-        # horribly simplified (omits mode checks, error handling)
+        new_position = 0
         if whence == os.SEEK_SET:
-            self.pos = offset
+            new_position = offset
         elif whence == os.SEEK_CUR:
-            self.pos += offset
+            new_position = self.pos + offset
         elif whence == os.SEEK_END:
-            self.max_size += offset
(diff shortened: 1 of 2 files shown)


manticore/platforms/linux.py
```diff
@@ -170,14 +170,25 @@ def tell(self):
...
+        assert isinstance(offset, (int, long))
+        assert isinstance(whence, (int, long))
+
+        # horribly simplified (omits mode checks, error handling)
+        if whence == os.SEEK_SET:
+            self.pos = offset
+        elif whence == os.SEEK_CUR:
+            self.pos += offset
+        elif whence == os.SEEK_END:
+            self.max_size += offset
```
japesinator: I'm not sure this is correct behavior, I think this should just set `pos` to `offset` bytes from the end of the file and not affect `max_size`

manticore/platforms/linux.py
```diff
@@ -170,14 +170,25 @@ def tell(self):
         '''
         return self.pos
 
-    def seek(self, pos):
+    def seek(self, offset, whence = os.SEEK_SET):
         '''
         Returns the read/write file offset
```
japesinator: This docstring isn't correct anymore, we should probably update it

manticore/platforms/linux.py
251: What about a more complex version:

```Python
        new_position = 0
        if whence == os.SEEK_SET:
            new_position = offset
        elif whence == os.SEEK_CUR:
            new_position = self.pos + offset
        elif whence == os.SEEK_END:
            new_position = self.max_size + offset

        if new_position < 0:
            return -1

        if new_position < self.max_size:
            self.pos = new_position
        else:
            logger.warning("Symbolic seek: maximum file length reached")
            self.pos = self.max_size

        return self.pos
```

I'm not sure on how to handle the last case. I guess one can drop the last if/else. Technically there is no problem setting the position behind max_size. A read/write access should warn about it.

japesinator: At least in the python repl, `seek` will let you move the cursor as far past the end of the file as you'd like, and I'm fine being consistent with that.

japesinator:
This looks great, thanks!

PR #396, opened 2017-07-17
Make save_stream pass through kwargs

japesinator:
Should fix use cases like https://github.com/trailofbits/manticore/blob/master/manticore/manticore.py#L623

### DIFF
--- a/manticore/core/workspace.py
+++ b/manticore/core/workspace.py
@@ -437,8 +437,8 @@ class ManticoreOutput(object):
         self.save_fds(state)
         self._store.save_state(state, self._named_key('pkl'))
 
-    def save_stream(self, *rest, **kwargs):
-        return self._store.save_stream(*rest, **kwargs)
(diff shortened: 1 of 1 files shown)


manticore/core/workspace.py
```diff
@@ -94,7 +94,7 @@ def load_value(self, key):
             return s.read()
 
     @contextmanager
-    def save_stream(self, key):
+    def save_stream(self, key, *rest, **kwargs):
```
nettrino: Do we need both `key` and `rest` here?

japesinator: Nothing currently uses `*rest`, but it's used in e.g. https://github.com/trailofbits/manticore/blob/master/manticore/core/workspace.py#L440-L441

nettrino: Can we only leave `*rest, **kwargs` and have `key = rest[0]` or it's not clean enough?

PR #399, opened 2017-07-18
Add optional redis dependency to setup.py

### LINKED ISSUE #398: Add redis dependency to setup.py
We should be able to use a setuptools "extra" to have redis automatically installed by pip, if a user specifies they want a redis install. This command line would looke something like

```
pip install manticore[redis]
```

http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-dependencies

japesinator:
Fixes #398

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,8 @@ nosetests --processes=8 --process-timeout=120 tests/test_binaries.py
 ### Redis
 
 If you'd like to use redis for state serialization (instead of disk), install
-redis using your host package manager, then run
+redis using your host package manager, then install manticore as above, but
+with `[redis]` appended to the name of the package, e.g.
 
 ```
 pip install manticore[redis]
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -114,6 +114,18 @@ period via the --process-timeout flag. E.g.,
 nosetests --processes=8 --process-timeout=120 tests/test_binaries.py
 ```
 
+### Redis
+
+If you'd like to use redis for state serialization (instead of disk), install
+redis using your host package manager, then run
+
+```
+pip install manticore[redis]
```
mossberg: The reader interested in this section really should be appending `[redis]` to whichever `pip install` command of the above three they chose. Could we change the language to reflect this to avoid people blindly doing `pip install` even if they did `pip install --user` above, for example? I still like keeping the code block to show an example of how this looks, since it's somewhat uncommon i think.

setup.py
```diff
@@ -19,6 +19,9 @@
             'coverage',
             'nose',
             'Sphinx',
+        ],
+        'redis': [
```
mossberg: we also probably want to add redis to the dev deps

PR #404, opened 2017-07-20
Refactor linux platform to use manticore.binary

### LINKED ISSUE #335: Use manticore.binary
Use manticore.binary instead of ElfFile at linux.py

https://github.com/trailofbits/manticore/blob/master/manticore/binary/__init__.py#L89

japesinator:
Fixes #335


manticore/binary/__init__.py
```diff
@@ -105,31 +107,9 @@ def __init__(self, filename):
...
-            assert elf_segment.header.p_filesz != 0 or elf_segment.header.p_memsz != 0 
-            yield((elf_segment.header.p_vaddr,
-                  elf_segment.header.p_memsz,
-                  perms, 
-                  elf_segment.stream.name, elf_segment.header.p_offset, elf_segment.header.p_filesz))
-
     def getInterpreter(self):
         return self.interpreter
 
-    def threads(self):
```
feliam: Why remove this? When binary is a process snapshot you have threads.

japesinator: I only removed it because it was unused, if you feel like it's a useful API I can definitely add it back

PR #407, opened 2017-07-20
Fix forking state from memory

### LINKED ISSUE #406: Incorrect exception in memory.py
https://github.com/trailofbits/manticore/blob/45d5b9f6d24ea69ac2238821a99c068452ab4295/manticore/core/memory.py#L957

`ForkState` is defined in state.py, which depends on memory.py

japesinator:
Fixes #406, but uses a deferred circular import to do so

### DIFF
--- a/manticore/core/memory.py
+++ b/manticore/core/memory.py
@@ -955,7 +955,7 @@ class SMemory(Memory):
                 for base in e.solutions:
                     condition = Operators.OR(address == base, condition )
                 from .state import ForkState
-                raise ForkState(condition)
+                raise ForkState("Forking state on incomplete result", condition)
 
             #So here we have all potential solutions to address
             assert len(solutions) > 0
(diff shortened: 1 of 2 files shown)


manticore/core/state.py
```diff
@@ -3,7 +3,7 @@
 import logging
 from collections import OrderedDict
 
-from .smtlib import solver
+from .smtlib import solver, Bool
```
feliam: Bool is now used ad was not before?

manticore/core/memory.py
```diff
@@ -954,7 +954,8 @@ def read(self, address, size):
                 condition = False
                 for base in e.solutions:
                     condition = Operators.OR(address == base, condition )
-                raise ForkState(condition)
+                from .state import ForkState
```
feliam: move up ?

manticore/core/state.py
japesinator: Bool was always used, it just stopped being imported for a bit (https://github.com/trailofbits/manticore/blob/42b462fc6c2c9ee66393991d692215475b09fcd4/manticore/core/state.py#L58)

manticore/core/memory.py
japesinator: If we import initially, we fail because of circular imports.

PR #414, opened 2017-07-26
Add Taint Parameters

ehennenfent:
Adds parameters allowing the user to specify an ID for taint tracking to `state.new_symbolic_buffer` and `state.symbolicate_buffer`. Also adds a property to `smtlib.expression.ArrayProxy` that allows the taint of the proxied ArrayVariable to be returned.

### DIFF
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -176,7 +176,8 @@ class State(object):
         :return: :class:`~manticore.core.smtlib.expression.Expression` representing the buffer.
         '''
         name = options.get('name', 'buffer')
-        expr = self.constraints.new_array(name=name, index_max=nbytes)
+        taint = options.get('taint', frozenset())
+        expr = self.constraints.new_array(name=name, index_max=nbytes, taint=taint)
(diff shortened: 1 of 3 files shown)


mossberg:
hi, thanks for the PR! I'll get to looking at this soon

manticore/core/state.py
```diff
@@ -176,7 +176,8 @@ def new_symbolic_buffer(self, nbytes, **options):
         :return: :class:`~manticore.core.smtlib.expression.Expression` representing the buffer.
         '''
         name = options.get('name', 'buffer')
-        expr = self.constraints.new_array(name=name, index_max=nbytes)
+        taint = options.get('taint', frozenset())
```
mossberg: Let's also update the docstrings for new_symbolic_buffer and symbolicate_buffer. adding something like this

```
        :param taint: Taint identifier of this buffer
        :type taint: tuple or frozenset
```

mossberg: addressed

ehennenfent:
Thanks, happy to help!

Issue #415, opened 2017-07-26
More flexible first state loading mechanism.


feliam:
Let Manticore be initialized without a program.
Add an API to let the user synthetized states after Manticore() constructor and add them afterwards.

mossberg:
related: #70

PR #421, opened 2017-07-27
Refactor platform details out from Manticore

feliam:
Manticore() refactor
https://github.com/trailofbits/manticore/issues/415
Main goal is to have Executor instantiated in Manticore __init__ and being able to start without a state. It's a Manticore() refactor.


manticore/__main__.py
```diff
@@ -110,11 +112,12 @@ def main():
     if args.assertions:
         m.load_assertions(args.assertions)
 
-    m.verbosity = args.v
+    print args.programs
```
yan: Left in debug print?

manticore/manticore.py
```diff
@@ -113,23 +177,23 @@ def makeWindows(args):
...
+def make_initial_state(binary_path, *args, **kwargs):
+    magic = file(binary_path).read(4)
     if magic == '\x7fELF':
-        return 'ELF'
+        # Linux
+        state = make_linux(binary_path, *args, **kwargs)
     elif magic == 'MDMP':
-        return 'PE'
+        # Windows
+        raise NotImplementedError("Binary {} not supported.".format(binary_path))
```
yan: Intentional `raise` here?

manticore/manticore.py
```diff
@@ -27,8 +27,69 @@
...
+manticore_verbosity=''
+
+def set_verbosity(setting):
+    global manticore_verbosity
+    global loggers
+
+    if isinstance(setting, (int, long)):
+        #10             20            30               40             50
+        #logging.DEBUG, logging.INFO, logging.WARNING, logging.ERROR, logging.CRITICAL
+        levels = ['',
```
yan: `MANTICORE` and `MEMORY` can't be set individually this way

feliam: :(

feliam: register_logger(name, letter_id=None) and let modules register themself here somehow?
Change "MANTICORE"  to "MAIN"

mossberg:
Cool ideas, some comments below. this is mostly high level feedback and more brainstorming about the example evm script felipe posted in chat. Would love more discussion and feedback.

### initial state registration API

I'm not a fan of this pattern:

```
initial_state = ...
m = Manticore()
m.add(initial_state)
```

I think it's an example of making the client do something the module could do. instead, what about

```
initial_state = 
m = Manticore(initial_state)

m2 = Manticore('/bin/ls') # also works
```

the first argument to `Manticore` would take advantage of dynamic typing by either accepting a string or a `State`, but i don't think this is a bad thing. I don't think this would be too confusing for a user, as long as it was documented and there were examples. we could potentially require the state to be kwarg only `Manticore(state=initial_state)`

another potential idea:

```
initial_state = ...
m = Manticore(initial_state)

m2 = Manticore.frompath('/bin/ls')

m3 = Manticore('/bin/ls') # doesn't work
```

here, `.frompath` would be a classmethod. this pattern exists in the stdlib, for example with the datetime class. however, I'm not a fan of it, as it's pretty extreme API breakage, and the `Manticore('/binary/path')` pattern is widely engrained, primarily because of angr. i do think that is as intuitive and convenient as it gets for the most common use case of analyzing a binary.


### APIs for custom initial state creation

I think we should be pretty careful about the APIs we expose for doing the manual state creation. I feel like exposing the `State`, and `ConstraintSet` classes directly is a little bit too low level. What about something like this:

- we expose the platform module in the API, and function in the module for creating a state given whatever inputs the platform needs
- we have an init callback that the user registers. they are passed an initial state, and can perform whatever custom state initialization they like in this callback. custom state initialization basically boils to 2 things: calling platform APIs, and calling cpu APIs. both the platform and cpu are available from the state.
- with this, manticore itself will assume very little about the platform. the `<platform>.create_state()` function will take the platform specific initial inputs, and the init hook will allow the user to do platform specific customizations using the APIs of the platform and cpu objects (see imaginary embedded system below)
- note: instead of using the init callback, it would be possible for the user to do customization in between calling the platform function for creating a state, and passing that state into the Manticore ctor. However, this would not be possible for the convenience/legacy interface `Manticore('/bin/ls')` because there is no intermediate state variable - they would need to use the init hook. Also even though this would be possible, I don't think it should be encouraged, to keep congruence with our `@m.hook` pattern

Examples:

detailed interface: evm

```
from manticore.platforms import evm


evm_bytecode = '...'

m = Manticore(evm.create_state())  # TODO this is somewhat an example of making the client to what the module shouldl do, but I don't have a good solution right now. I think it's acceptable since this is not a primary use case

@m.init
def inithook(state):
    world = state.platform
    # custom evm state setup
    world.create_account()
    world.create_contract(evm_bytecode)
    ...
```

possible, but discouraged and undocumented:

```
from manticore.platforms import evm


evm_bytecode = '...'

state = evm.create_state()

world = state.platform
# custom evm state setup
world.create_account()
world.create_contract(evm_bytecode)
evmcpu = state.cpu
evmcpu.PC = 0xsomething

m = Manticore(state)

```

detailed interface: linux

```
from manticore.platforms import linux


m = Manticore(linux.create_state('/my/binary', argv=[], envp=[]))

@m.init
def inithook(state):
    linux = state.platform
    # custom linux state setup
    linux.open_fd
    linux.stdin.transmit(...)
    cpu = state.cpu
    cpu.PC = interesting_func_addr
    cpu.R0 = arg1
    ...
```

detailed interface: imaginary weird embedded system

```
from manticore.platforms import weird_embedded


m = Manticore(weird_embedded.create_state('firmware.blob', hw_config=dict(dict with some configs))

@m.init
def inithook(state):
    embedded_sys = state.platform
    # custom state setup
    embedded_sys.interrupts_stuff()
    embedded_sys.mmio_registers_stuff()

    cpu = state.cpu
    cpu.PC = interesting_func_addr
    cpu.R0 = arg1
    ...
```

convenience/legacy interface

```


m = Manticore('/bin/ls')

@m.init
def inithook(state):
    linux = state.platform
    # custom linux state setup
    linux.open_fd
    linux.stdin.transmit(...)
    cpu = state.cpu
    cpu.PC = interesting_func_addr
    cpu.R0 = arg1
    ...
```


```


m = Manticore(unpickle('mystatefile.pkl'))

```

yan:
I like this proposal, including overloading Manticore's constructor. I don't think it's a problem to dispatch based on ctor's argument type -- overloading the ctor is a very common pattern. However, I'm not sure which is better -- adding a new hook type or just exposing the entrypoint from Manticore.

i.e.:

```
@m.hook(m.entry)
def init(state):
   ...
```

vs

```
@m.init
def init(state):
  ...
```

I feel like exporting the entrypoint is a lighter-weight change to the API that doesn't create a new hook type.

mossberg:
Good point! A random thought is if we define a `m.init`, it doesn't
*necessarily* have to have the same semantics as `m.hook(m.entry)`. That
said, a user initialization hook should probably execute after Manticore
has done its initialization (in the case of linux, loading the binary; in
the case of evm, setting up the EVMWorld(?) (@feliam does this sound
right?)), and before Manticore begins executing the program, which is
precisely the semantics of hooking on the first instruction.

In short, I'm in support of only adding an `m.entry` API for now, instead
of a full on `m.init` hook.

However, I can see in the future that we will probably want to add `m.init`
simply as a convenience interface, given that `m.hook(m.entry)` will likely
be a very common pattern for mcore scripts.

mossberg:
Idea:

Rather than importing a platform module with a `create_state()` function, what if we expose classmethods on `Manticore` for this? This will reduce required imports.

```
m = Manticore.evm()
m = Manticore.linux('/binary', argv=[], envp=[])
m = Manticore.weird_embedded('firmware.blob', hw_config=dict(dict with some configs))
```

mossberg:
Note: an example of a similar dynamic typed API from the stdlib is here: https://docs.python.org/2/library/ast.html#ast.literal_eval

feliam:
What about:
We accept that a user may add be states from multiple sources. 


if flavor=1:
    m = Manticore(config=global_config)
    assert len(m) == 0, "There should be no states in queue"

    m.linux_state('/bin/ls')
    m.add(state)
else:
    m = Manticore(initial_state='/bin/ls', config=global_config)

assert len(m) == 1, "There should be a single state"

#or @hook for "will_execute_instruction" 
#This will evolve into a more contained plugin framework
m.subscribe('some_event', my_hook)  

m.run()

len(m) -> ?  normally zero unless ManticorePause is implemented

mossberg:
I'm a little unsure about `m.add()`. Initializing analysis from multiple states is not currently supported, so it would be a new feature. What exactly are some use cases for this?

> B) If we do not care for m.add(state) I'll need a way to do the reinsertion of a finished state into the queue

can you elaborate more on the use case? from what context, a mcore script, or in the core?

feliam:
A use case for m.add()

There is the need to re-insert a state after it finishes. 
In EVM you need to emulate 2 'transactions'. The first would run over the evm world (platform) and reach a final stage/instruction. Then without resetting the memory and storage or any other platform internal data you make the second transaction. (Which is like a function call).  One way to do this is to subscribe to 'will_terminate_state' and reinsert the state into the queue via a m.add() (and prevent the testcase generation via e.testcase=False) ... Like this.

    #Callbacks
    def terminate_transaction_callback(self, state, state_id, e):
            do_something_else(state)  #Apply the next function call/transaction
            self.add(state)  #Re add state to queue
            e.testcase = False   #prevent the testcase generation

mossberg:
Ok, I'm fine with `m.add()`, but I think we should rename it to better describe what it does, and also keep it as an undocumented/experimental API.

I still think Manticore should only init with 1 state, unless there's another clear use case for multiple states init?

> we check somehow at each add(state) that the state we are adding is a relative of the initial state.

we probably only need to do this if we implement the binary tree state data structure, right? if it's a simple list, we don't need to worry about this?

feliam:
>we probably only need to do this if we implement the binary tree state data structure, right? if it's a simple list, we don't need to worry about this?"

Normally all the states in the working queue are a fork from a parent and that makes a tree up until the first state. 
If we allow m.add() the previous statement is not necessarily true (unless we check somehow the new addition is in the family )

japesinator:
If we want to both support a "tree" of states, and arbitrary state addition (which I think is a good idea), I think my preference would be to have a "0" state, which starts out fully analyzed and is only used for tree-traversal purposes, from which our initial state, as well as anything added with just `m.add(state)`, descends from, and then have `add` take an optional kwarg for parent when that isn't desired.

For instance, if we load some binary, our initial state tree is:

```
[0 state] -> [makeLinux(binary)]
```

and running `m.add(s)` gets us:

```
[0 state] -> [makeLinux(binary)
          -> [s]
```

but we can also call `m.add(s', parent=s)` to get:

```
[0 state] -> [makeLinux(binary)
          -> [s] -> [s']
```

mossberg:
We decided to accept the new (experimental/undocumented) API for users adding states to Manticore list of states to explore: `m.enqueue()`

mossberg:
> I like this proposal, including overloading Manticore's constructor. I don't think it's a problem to dispatch based on ctor's argument type -- overloading the ctor is a very common pattern. However, I'm not sure which is better -- adding a new hook type or just exposing the entrypoint from Manticore.

We've recently realized that `m.hook(m.entry)` does not make sense for all platforms (EVM for example, has no single entry point. with evm, we consider multiple contracts/programs, and we really want some sort of hook after init but before any kind of execution, which in the case of evm is semantically different from hooking on first insn of a particular contract)

This brings us back to using `@m.init`

mossberg:
# Manticore Initial State Customization Proposal v2

## initial state registration API

First argument to `Manticore` is either:
- `str`:  path to binary (convenience/legacy interface)
- `State: initial State object of analysis. this is the initial state registration API. API users would not really be using this, since the `State` is too low level of an object for them to be manually constructing and dealing with. They will primarily be using the platform classmethods

## manticore initialization API changes

To support initializing Manticore for different platforms cleanly, classmethods corresponding to platforms are proposed.

```
m = Manticore.evm()
m = Manticore.linux('/binary', argv=[], envp=[])
m = Manticore.weird_embedded('firmware.blob', hw_config=dict(dict with some configs))
```

For API compatibility, and as a convenience interface for Linux (which is expected to be a primary use case), we will maintain the conventional Manticore constructor

```
m = Manticore('/binary', argv)
```

## initial state customization

A hook will be provided for after Manticore's internal state init, but before analysis has begun (before instructions are executed). In this hook, platform and CPU apis can be called as needed in order to perform custom state initialization

```
@m.init
def inithook(state):
  platform = state.platform
  cpu = state.cpu
```

## API for dynamically enqueueing arbitrary states

We will export `m.enqueue(state)` as an undocumented api to dynamically enqueue states. Users should typically not need to do this, as they should use the classmethods, however it is required for evm changes we wish to introduce. this api is relevant to this rfc as it is relevant to initializing manticore with States.

manticore/__main__.py
```diff
@@ -79,51 +78,38 @@ def parse_arguments():
 def main():
     args = parse_arguments()
 
-    m = Manticore(args.argv[0], args.argv[1:], disasm=args.disasm)
-    m.policy = args.policy
-    m.args = args
+    log.set_verbosity(args.v)
```
mossberg: I think we should still opt to use m.verbosity here. `log` isn't meant to be a public API :/

manticore/__main__.py
```diff
@@ -79,51 +78,38 @@ def parse_arguments():
...
+
+
+    #TODO 
+    initial_state = make_initial_state(args.argv[0], argv=args.argv[1:],  disasm=args.disasm)
+    m = Manticore(initial_state, workspace_url=args.workspace,  policy=args.policy)
```
mossberg: should policy be an argument to m.run()?

manticore/__main__.py
```diff
@@ -79,51 +78,38 @@ def parse_arguments():
...
     args = parse_arguments()
 
-    m = Manticore(args.argv[0], args.argv[1:], disasm=args.disasm)
-    m.policy = args.policy
-    m.args = args
+    log.set_verbosity(args.v)
+
+
+    #TODO 
+    initial_state = make_initial_state(args.argv[0], argv=args.argv[1:],  disasm=args.disasm)
```
mossberg: I don't think we should use `make_initial_state` here, i think we should stick to public APIs

manticore/__main__.py
```diff
@@ -79,51 +78,38 @@ def parse_arguments():
...
-    if args.maxstates != 0:
-        m.maxstates = args.maxstates
-
-    if args.coverage:
-        m.coverage_file = args.coverage
-
     if args.names is not None:
         m.apply_model_hooks(args.names)
 
     if args.env:
```
mossberg: we can move this stuff to the init hook instead of requiring `m.initial_state`?

manticore/__main__.py
```diff
@@ -79,51 +78,38 @@ def parse_arguments():
...
     if args.data:
         m.concrete_data = args.data
 
-    if args.workspace:
-        m.workspace = args.workspace
-
     if args.profile:
         m.should_profile = args.profile
 
-    if args.dumpafter != 0:
```
mossberg: for all of these that we remove, we should remove them from the --help documentation also

feliam: We should also add an issue to reimplement using events/plugin

manticore/__main__.py
```diff
@@ -79,51 +78,38 @@ def parse_arguments():
...
-    if args.coverage:
-        m.coverage_file = args.coverage
-
     if args.names is not None:
         m.apply_model_hooks(args.names)
 
     if args.env:
         for entry in args.env:
             name, val = entry[0].split('=')
-            m.env_add(name, val)
```
mossberg: i think both of these will crash, state doesn't have either env_add or add_symbolic_file. these would fit perfectly on the platform though, which we can call from the init hook

mossberg: actually env_add will be challenging, because Linux needs that when loading the binary

feliam: m.initial_state.platform.env_add(name, val) ?

manticore/__main__.py
feliam: We currently need m.initial_state to publish the event for the init hook.

manticore/__main__.py
feliam: Agree

manticore/__main__.py
feliam: Loggers are global singleton, manticore is an object of which you can have a bunch. m.set_verbosity is confusing. It'll let you think you control the logging of m while you are controlling a global thing that affects loggers for all the manticores alive. Unless we use a static method? Manticore.set_verbosity ?

manticore/__main__.py
feliam: It sounds like it but it's _currently_? needed at executor constructor.

mossberg: ok, we can address later

manticore/manticore.py
```diff
@@ -157,53 +172,24 @@ class Manticore(object):
     :type args: list[str]
     :ivar dict context: Global context for arbitrary data storage
     '''
-    def __init__(self, path_or_state, args=None, disasm='capstone'):
```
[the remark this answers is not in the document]
feliam: Let's remove all positional arguments except for path_or_state?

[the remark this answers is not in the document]
feliam: There are none. Easy!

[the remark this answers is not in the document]
mossberg: i don't understand, how does that fix the api breakage concern?

manticore/manticore.py
```diff
@@ -157,53 +169,27 @@ class Manticore(object):
...
-            self._initial_state_type = 'MCORE'
-
-        self._check_disassembler_present(disasm)
-        self._disasm = disasm
-        # yan's comment
-        self._argv = args
-        self._env = {}
-        # Will be set to a temporary directory if not set before running start()
-        self._policy = 'random'
+            ws_path = None
```
mossberg: I think we should error here, rather than silently ignore and keep going

feliam: I think ManticoreOutput will pick a random fresh workspace when None.

mossberg: everything will be fine, but a workspace_url that is not a string _is_ an error. i just got confused by this for a bit trying to test that api compat stuff because i was expecting some kind of error when i gave a list for workspace_url

PR #425, opened 2017-07-31
Fix context serialization

feliam:
replace __reduce__ by  collaborative __set/getstate__

### DIFF
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -163,7 +163,7 @@ class State(Eventful):
         return self._constraints
 
     @constraints.setter
-    def constrains(self, constraints):
+    def constraints(self, constraints):
         self._constraints = constraints
         self.platform.constraints = constraints
 
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -70,50 +70,49 @@ class State(Eventful):
 
     def __init__(self, constraints, platform, **kwargs):
         super(State, self).__init__(**kwargs)
-        self.platform = platform
-        self.forks = 0
-        self.constraints = constraints
-        self.platform.constraints = constraints
-
-        self.input_symbols = list()
+        self._platform = platform
+        self._constraints = constraints
+        self._platform.constraints = constraints
+        self._input_symbols = list()
         self._child = None
-
-        self.context = dict()
+        self._context = dict()
         ##################################################################33
-        # Signals are lost in serialization and fork !!
-        #self.will_add_constraint = Signal()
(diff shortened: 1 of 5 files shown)


manticore/core/state.py
```diff
@@ -70,50 +70,49 @@ class State(Eventful):
...
+        self._child = state['child']
+        self._context = state['context']
+        ##################################################################33
+        # Events are lost in serialization and fork !!
+        self.forward_events_from(self._platform)
 
+    #Fixme(felipe) change for with state.cow_copy() as st_temp:.
     def __enter__(self):
         assert self._child is None
         new_state = State(self.constraints.__enter__(), self.platform)
```
yan: `_constraints` and `_platform`?

manticore/core/state.py
```diff
@@ -147,6 +146,27 @@ def setstate(state, value):
...
+    @property
+    def platform(self):
+        return self._platform
+
+    @property
+    def constraints(self):
+        return self._constraints
+
+    @constraints.setter
+    def constrains(self, constraints):
```
yan: typo?

manticore/core/state.py
```diff
@@ -143,16 +142,37 @@ def setstate(state, value):
...
-        assert self.mem.constraints is self.constraints
+        assert self._platform._constraints is self._constraints
+        assert self.mem._constraints is self._constraints
         return result
 
+    @property
+    def input_symbols(self):
+        return self._input_symbols
+
+    @property
```
mossberg: are these properties to prevent users from reassigning `state.context`?

feliam: for reasigning _constraints, _input_symbols _platform _context yes

manticore/core/state.py
```diff
@@ -143,16 +142,37 @@ def setstate(state, value):
...
+    def context(self):
+        return self._context
+
+    @property
+    def platform(self):
+        return self._platform
+
+    @property
+    def constraints(self):
```
yan: Do we want to return `_platform.constraints` here and remove the cached `_constraints_ reference altogether?

feliam: I'm not sure. There may be platforms that won't need to have a constraintSet in it?
Current feeling is that State should be the ContraintSet owner. 
And that platform should only be notified of what the ConstraintSet is just in case.

yan:
👍 . Only comment is `test_record_branches` should probably be removed as well, not just skipped

PR #426, opened 2017-08-01
EVM support for manticore

### LINKED ISSUE #416: Initial EVM support

feliam:
Fixes #416

* Supports most EVM instructions concrete and symbolic
* Supports inter contract calls and recursive calls
* Supports arbitrary symbolic multi-contract attacks/analysis 
* Fair symbolic SHA3 concretization mechanism

Lacks:
* GAS support
* Good reporting


manticore/core/workspace.py
```diff
@@ -454,13 +454,6 @@ def save_testcase(self, state, prefix, message=''):
 
         self._named_key_prefix = prefix
         self._increment_id()
-
-        self.save_summary(state, message)
```
mossberg: this should not be removed?

manticore/manticore.py
```diff
@@ -628,6 +708,15 @@ def terminate(self):
     #############################################################################
     # Move all the following elsewhere Not all manticores have this
 
+    def add_symbolic_file(self, symbolic_file):
```
mossberg: this may have gotten added back to manticore.py in the merge; it should be in linux.py

setup.py
```diff
@@ -12,6 +12,7 @@
         'pyelftools',
         'unicorn',
         'ply',
+        'sha3',
```
mossberg: i think we needed pysha3

manticore/manticore.py
feliam: What about _get_symbol_address(self, symbol) ?

manticore/core/workspace.py
feliam: What to save have nothing to do with the workspace. Not sure where I put it.

manticore/manticore.py
feliam: Let's leave it under the "Move all the following elsewhere Not all manticores have this FIXME"

manticore/core/workspace.py
mossberg: workspace directories look pretty empty, not sure it was put somewhere else

```
[I] mark ubuntu /m/h/c/m/e/linux (dev-evm) ❯ manticore basic
2017-09-13 18:46:38,504: [99283] MANTICORE:INFO: Loading program basic
2017-09-13 18:46:39,373: [99283] MANTICORE:INFO: Starting 1 processes.
2017-09-13 18:46:43,445: [99283] MANTICORE:INFO: Generated testcase No. 0 - Program finished with exit status: 0L
2017-09-13 18:46:44,429: [99283] MANTICORE:INFO: Generated testcase No. 1 - Program finished with exit status: 0L
2017-09-13 18:46:44,443: [99283] MANTICORE:INFO: Results in /mnt/hgfs/code/manticore/examples/linux/mcore_7zPYZL
2017-09-13 18:46:44,444: [99283] MANTICORE:INFO: Instructions executed: 12464
2017-09-13 18:46:44,444: [99283] MANTICORE:INFO: Coverage: 3143 different instructions executed
2017-09-13 18:46:44,444: [99283] MANTICORE:INFO: Total time: 5.07086396217
2017-09-13 18:46:44,444: [99283] MANTICORE:INFO: IPS: 2457
[I] mark ubuntu /m/h/c/m/e/linux (dev-evm) ❯ l /mnt/hgfs/code/manticore/examples/linux/mcore_7zPYZL
total 864
-rw-r--r-- 1 501 dialout     37 Sep 13 18:46 command.sh
-rw-r--r-- 1 501 dialout 424234 Sep 13 18:46 test_00000000.pkl
-rw-r--r-- 1 501 dialout 424469 Sep 13 18:46 test_00000001.pkl
-rw-r--r-- 1 501 dialout  34573 Sep 13 18:46 visited.txt
```

feliam: Hmm ideas on where to put them? We do not want (for example)  test_000001.stdin for EVM

mossberg: In general, the platform somewhere... platform has a method returning a dict of X -> file content, where X is what is now passed to self._named_stream()?

mossberg: paging @yan on this also

yan: I'm not sure Platform needs to return a dict that a Workspace can serialize, it should probably implement a `dump_summary(workspace, state)`  method and serialize what needs to be serialized.

mossberg: I think if that function returns a dict, it will make for easier testing in per platform tests.

Since this change will take place in `ManticoreOutput.save_testcase`, the invocation would be `state.platform.dump_summary(self)`, which seems slightly code smelly

manticore/manticore.py
```diff
@@ -229,10 +270,22 @@ def decree(cls, path, concrete_data='', **kwargs):
         except KeyError:  # FIXME(mark) magic parsing for DECREE should raise better error
             raise Exception('Invalid binary: {}'.format(path))
 
+    @classmethod
+    def evm(cls, path, concrete_data='', **kwargs):
```
mossberg: can clean up some of these unused arguments

manticore/core/executor.py
```diff
@@ -496,8 +494,9 @@ def run(self):
                     trace = traceback.format_exc()
                     print str(e), trace
                     logger.error("Exception: %s\n%s", str(e), trace)
+                    print str(e), trace
```
mossberg: let's find all the stray prints and rm

manticore/core/smtlib/visitors.py
```diff
@@ -448,27 +450,25 @@ def visit_ArraySelect(self, expression, *operands):
...
+            else:
+                return arr.array.select(index)
 
     def visit_Expression(self, expression, *operands):
         assert len(operands) == 0
         assert not isinstance(expression, Operation)
         return expression
 
+#FIXME this should forget old expressions lru?
+arithmetic_simplifier_cache = {}
```
mossberg: this appears unused, maybe it should be removed?

manticore/manticore.py
```diff
@@ -52,6 +52,45 @@ def _check_disassembler_present(disasm):
     initial_state = State(constraints, platform)
     return initial_state
 
+def make_evm(args):
```
mossberg: what is the fate of this code? I like for now, we shouldn't try to devise some kind of disk evm binary format, and have evm features be API only. and from what this code looks like, the evm binary format would consist of pseudo-python anyway.

manticore/manticore.py
```diff
@@ -335,11 +387,47 @@ def wrapper(*args, **kwargs):
                 self._workers.append(p)
                 p.start()
 
+
+    ###########################################################################
+    # Workers                                                                 #
+    ###########################################################################
+    def _start_workers(self, num_processes, profiling=False):
```
mossberg: merge issue? this function is duplicated.

manticore/utils/helpers.py
```diff
@@ -1,4 +1,4 @@
-
+import collections
```
mossberg: stray import here?

mossberg: nevermind, it is used

manticore/core/state.py
```diff
@@ -76,10 +77,12 @@ def __init__(self, constraints, platform, **kwargs):
         self._input_symbols = list()
         self._child = None
         self._context = dict()
-        self._init_context()
         ##################################################################33
         # Events are lost in serialization and fork !!
         self.forward_events_from(platform)
+        
+        #FIXME(felipe) This should go into some event callback in a plugin (start_run?)
+        self._init_context()
```
yan: Should we update this to use new plugin events?

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+
+TT256 = 2 ** 256
+TT256M1 = 2 ** 256 - 1
+TT255 = 2 ** 255
+TOOHIGHMEM = 0x100
+
+def to_signed(i):
+    return i if i < TT255 else i - TT256
+
+def pack(i, size=32):
```
yan: This looks like it this takes an int and packs it into a 32 byte string, msb first. We should either update the name or document this in the docstring; the implementation isn't obvious.

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+    assert size >=1
+    o = [0] * size
+    for x in range(size):
+        o[(size-1) - x] = i & 0xff
+        i >>= 8
+    return ''.join(map(chr,o))
+
+class EVMMemory(object):
```
yan: Probably not in this PR, but should we eventually make our regular `Memory` be more generic so we can use it as a base class for arbitrary VMs?

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+    '''
+    The EVM symbolic memory manager.
+    '''
+    def __init__(self, constraints, address_size=256, value_size=8, *args, **kwargs):
+        '''
+        Builds a memory.
+
+        :param constraints:  a set of constraints
+        :param symbols: Symbolic chunks
```
yan: Docstring isn't up to date with the prototype.

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+        new_max = address % 32
+        self._allocated = Operators.ITEBV(256, self._allocated < new_max, new_max, self._allocated)
+
+    def _concrete_read(self, address):
+        return self._memory.get(address, 0)
+
+    def _concrete_write(self, address, value):
+        assert not issymbolic(address) 
+        assert not issymbolic(value)
+        assert value & ~(pow(2,self._value_size)-1) == 0 , "Not the correct size sor a value"
```
yan: Typo.

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+            return 'SWAP%d'%(self.pops-1)
+        elif self._name == 'LOG':
+            return 'LOG%d'%(self.pops-2)
+        return self._name
+
+    def __str__(self):
+        bytes = self.bytes.encode('hex')
+
+        output = '<%s> '%bytes + self.name + (' 0x%x'%self.operand if self.has_operand else '')
+        if True:
```
yan: Left in debug code?

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+
+    @property
+    def description(self):
+        return self._description
+
+    @property
+    def bytes(self):
+        bytes = []
+        bytes.append(chr(self._opcode))
+        for offset in reversed(xrange(self.operand_size)):
```
yan: maybe `xrange(self.operand_size-1, -1, -1)` instead of multiple iterators?

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+                0xfe: ('INVALID', 0, 0, 0, 0, 'Designated invalid instruction.'),
+                0xff: ('SELFDESTRUCT', 0, 1, 0, 0, 'Halt execution and register account for later deletion.')
+            }
+
+
+    @staticmethod
+    def decode_one(bytecode):
+        '''
+        '''
+        bytecode = iter(bytecode)
```
yan: This looks redundant, it's already made into an iterator in `decode_all`

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+        memory, the initial memory
+        address, the address of the account which owns the code that is executing.
+        origin, the sender address of the transaction that originated this execution. A 160-bit code used for identifying Accounts.
+        price, the price of gas in the transaction that originated this execution.
+        data, the byte array that is the input data to this execution
+        caller, the address of the account which caused the code to be executing. A 160-bit code used for identifying Accounts
+        value, the value, in Wei, passed to this account as part of the same procedure as execution. One Ether is defined as being 10**18 Wei.
+        bytecode, the byte array that is the machine code to be executed.
+        header, the block header of the present block.
+        depth, the depth of the present message-call or contract-creation (i.e. the number of CALLs or CREATEs being executed at present).
```
yan: Update format to sphinx docstrings?

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+
+    #Memory related
+    def _allocate(self, address):
+        #print pretty_print (address)
+        #if address > 100000:
+        #    raise NotEnoughGas()
+        self.memory.allocate(address)
+
+    def _store(self, address, value):
+        #CHECK ADDRESS IS A 256 BIT INT OR BITVEC
```
yan: Should these be asserts?

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+            expression = self.pc
+            def setstate(state, value):
+                state.platform.current.pc = value
+            raise Concretize("Concretice PC",
+                                expression=expression, 
+                                setstate=setstate,
+                                policy='ALL')
+
+        self.publish('will_decode_instruction', self.pc)
+        current = self.instruction
```
yan: Should we publish `did_decode_instruction` here and `did_execute_instruction` below?

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+
+        self.publish('did_execute_instruction', self, current)
+
+        
+        #Check result (push)
+        if current.pushes > 1:
+            assert len(result) == current.pushes
+            for value in reversed(result):
+                self._push(value)
+        elif current.pushes == 1:
```
yan: Why does this need a special-cased conditional?

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+    def REVERT(self, offset, size):
+        data = self.read_buffer(offset, size)
+        raise Revert(data)
+
+    def SELFDESTRUCT(self, to):
+        '''Halt execution and register account for later deletion'''
+        raise SelfDestruct(to)
+
+    def __str__(self):
+        def hexdump(src, length=16):
```
yan: Maybe move this to `manticore.utils`?

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1760 @@
...
+                    return False
+            return cond
+
+        if any(map(issymbolic, data)):
+            logger.info("SHA3 Searching over %d known hashes", len(self._sha3))
+            logger.info("SHA3 TODO save this state for future explorations with more known hashes")
+            results = []
+            known_hashes = False
+            for key, value in self._sha3.items():
+                cond = compare_buffers(key
```
yan: Left in print here and below

manticore/platforms/evm.py
feliam: Agree. We should explore the notion. Though there are differences.  LinuxMemory has mmap and EVMMemory is compacted fro 0 to a variable max. 
We should document our memory backends everywhere and even provide alternatives like:
* Eager+ITE  <- we are using this and its undocumented 
* Lazy+Array

manticore/manticore.py
feliam: Maybe this should load the json generated by something like:

```solc --combined-json abi,asm,ast,bin,bin-runtime,clone-bin,compact-format,devdoc,hashes,interface,metadata,opcodes,srcmap,srcmap-runtime,userdoc simple```

and perform a basic symbolic exploration on it? We can remove for now.

manticore/core/state.py
feliam: Not sure we'll have the time to merge plugins. (I vote yes anyway!)

manticore/manticore.py
mossberg: i like that idea, but i think we should remove for now.

manticore/platforms/evm.py
yan: This should probably live in a different file.

manticore/platforms/evm.py
feliam: Unless it's called from other func. I can make it private if this is not acceptable.

manticore/platforms/evm.py
feliam: We don't have that. Maybe we should. Though, in that case we need to upgrade abstract cpu et.at. I think that's for a different PR.

manticore/platforms/evm.py
feliam: With single PUSH instruction handlers wont return a list with 1 element but the element itself.

PR #428, opened 2017-08-01
Add basic socket support

yan:
This PR implements basic socket support. We assume a socket has a number of bytes of symbolic data in it, and some operations like `sendfile(2)` are nops. Incomplete, but sufficient to emulate basic network operations.


manticore/platforms/linux.py
```diff
@@ -2143,3 +2248,64 @@ def sys_write(self, fd, buf, count):
...
+    def sys_accept(self, sockfd, addr, addrlen, flags):
+        #TODO(yan): Transmit some symbolic bytes as soon as we start. 
+        # Remove this hack once no longer needed.
+
+        fd = super(SLinux, self).sys_accept(sockfd, addr, addrlen, flags)
+        if fd < 0:
+            return fd
+        sock = self._get_fd(fd)
+        nbytes = 32
+        symb = self.constraints.new_array(name='socket', index_max=nbytes)
```
feliam: Maybe maintain an family, address  -> symbolic array 
built in the constructor like symbolic_files ?

PR #438, opened 2017-08-05
Implement state.generate_testcase

### LINKED ISSUE #373: State counter / identifier

mossberg:
Proposed implementation related to #373 

Fixes #373 

This implements `state.generate_testcase(name, description)` which can be used by hooks to arbitrary generate testcases to the workspace.

discussion
- api name? `state.generate_testcase`, `state.generate_inputs`, `state.dump_inputs`, `state.get_inputs`, `concretize_inputs`

### DIFF
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -366,5 +366,5 @@ class State(Eventful):
         self.context['branches'] = dict()
 
     # XXX(mark) naming? generate_testcase? to keep with testcase parlance
-    def generate_inputs(self, name):
-        self.publish('state_generate_inputs', self, name=name)
+    def generate_inputs(self, name, message='State generated inputs'):
+        self.publish('state_generate_inputs', message=message, name=name)
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -364,3 +364,7 @@ class State(Eventful):
 
     def _init_context(self):
         self.context['branches'] = dict()
+
+    # XXX(mark) naming? generate_testcase? to keep with testcase parlance
+    def generate_inputs(self):
+        self.publish('state_generate_inputs', self)
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -366,4 +366,4 @@ class State(Eventful):
         self.context['branches'] = dict()
 
     def generate_testcase(self, name, message='State generated testcase'):
-        self.publish('state_generate_inputs', message=message, name=name)
+        self.publish('state_generate_inputs', name, message)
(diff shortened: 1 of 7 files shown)


manticore/core/state.py
```diff
@@ -364,3 +364,6 @@ def mem(self):
 
     def _init_context(self):
         self.context['branches'] = dict()
+
+    def generate_testcase(self, name, message='State generated testcase'):
+        self.publish('state_generate_inputs', message=message, name=name)
```
yan: I think the `state_generate_inputs` event should order its args as `(name, message)`, in the same order as the `generate_testcase` api.

manticore/utils/event.py
```diff
@@ -46,8 +46,9 @@ def _get_signal_bucket(self, name):
         #A bucket is a dictionary obj -> set(method1, method2...)
         return self._signals.setdefault(name,  dict())
 
-    def publish(self, name, *args, **kwargs):   
-        bucket = self._get_signal_bucket(name)
+    # The underscore _name is to avoid naming collisions with callback params
```
yan: How would this collide with callback params?

mossberg: Just if I wanted to have a `name=` kwarg in my callback handler. `self.publish('some_event', arg1, name='somename')` would fail because `name` would be passed twice.

manticore/core/state.py
```diff
@@ -364,3 +364,12 @@ def mem(self):
...
         self.context['branches'] = dict()
+
+    def generate_testcase(self, name, message='State generated testcase'):
+        """
+        Generate a testcase for this state and place in the analysis workspace.
+
+        :param str name: Short string identifying this testcase used to prefix workspace entries.
+        :param str message: Longer description
+        """
+        self.publish('state_generate_inputs', name, message)
```
yan: I'm not sure this fits with our existing event naming scheme of will/did. Maybe `should_generate_inputs` is more appropriate here?

mossberg: I just realized it really doesn't make any sense to create an entirely new `state_generate_inputs` event, and then connect it to an existing handler `_generate_testcase_callback`. I should be able to use the existing `will_generate_inputs` here

yan: That's even cleaner; still a'd

PR #443, opened 2017-08-09
Use the same workspace URI in Executor.__init__

### LINKED ISSUE #442: Incorrect workspace creation
Currently, when a workspace isn't specified, Manticore creates two workspaces on the filesystem. This is due to [manticore/manticore.py:708](https://github.com/trailofbits/manticore/blob/master/manticore/manticore.py#L708), as `ws_path` is `None`.

The fix is two fold: 

1. Have `Workspace.uri` return a full, type-qualified path (i.e. `fs:/path`), and
2. Pass the uri to the `Executor`

yan:
Fixes #442


yan:
(references issue #442)

manticore/core/workspace.py
```diff
@@ -58,6 +58,28 @@ class Store(object):
 
     Implement either save_value/load_value in subclasses, or save_stream/load_stream, or both.
     """
+    @classmethod
+    def store_type(cls):
```
mossberg: I think you meant to use staticmethod here? Alternatively, we could just class variables?

manticore/core/workspace.py
```diff
@@ -58,6 +58,28 @@ class Store(object):
...
+        raise NotImplementedError("Store subclass must define store_type() class method")
+
+    @classmethod
+    def create_store(cls, desc):
+	"""
+	Create a :class:`~manticore.core.workspace.Store` instance depending on the descriptor.
+
+	Valid descriptors:
```
mossberg: this formatting might mess up when rendered with Sphinx, does it?

manticore/core/workspace.py
```diff
@@ -58,6 +58,28 @@ class Store(object):
 
     Implement either save_value/load_value in subclasses, or save_stream/load_stream, or both.
     """
+    @classmethod
+    def store_type(cls):
+        raise NotImplementedError("Store subclass must define store_type() class method")
+
+    @classmethod
+    def create_store(cls, desc):
```
mossberg: naming here? Maybe `Store.fromdescriptor`? The `.from...` idiom is from the datetime module which is the primary stdlib module I know of that uses classmethods, so maybe we should copy that

manticore/core/workspace.py
```diff
@@ -401,7 +410,8 @@ def __init__(self, desc=None):
         :param desc: A descriptor ('type:uri') of where to write output.
         """
         self._named_key_prefix = 'test'
-        self._store = _create_store(desc)
+        self._descriptor = desc
```
mossberg: Can we simplify this `self._descriptor` vs `self.descriptor` stuff? Can't we just do `self.descriptor = desc` and be done? I don't see a reason to have a property that computes that on every access, or even recompute it once from the `_store`, since the end result should be the same as `desc`?

manticore/core/workspace.py
```diff
@@ -410,6 +420,18 @@ def __init__(self, desc=None):
...
+    def descriptor(self):
+        """
+        Return a descriptor that created this workspace. Descriptors are of the
+        format <type>:<uri>, where type signifies the medium. For example,
+          fs:/tmp/workspace
+          redis:127.0.0.1:6379
+
+        :rtype: str
+        """
+        return ':'.join([self._store.store_type(), self._store.uri])
```
mossberg: this would break with `MemoryStore`, since it doesn't actually save `.uri`

manticore/core/workspace.py
yan: I had class variables initially, but wanted to be able to raise when unimplemented. I've reverted to variables and just check in `fromdescriptor` now.

manticore/core/workspace.py
yan: We can't because `desc` is `None` when we should create a new tmp workspace. I updated it to cache the descriptor so we don't recalc every time.

manticore/core/workspace.py
yan: `uri` is defined in `Store`, not in `MemoryStore`

manticore/core/workspace.py
```diff
@@ -59,6 +59,27 @@ class Store(object):
...
+	  * fs:<path>
+	  * redis:<hostname>:<port>
+	  * mem:
+
+	:param str desc: Store descriptor
+	:return: Store instance
+	"""
+        type_, uri = ('fs', None) if desc is None else desc.split(':', 1)
+        for subclass in cls.__subclasses__():
+            if not hasattr(subclass, 'store_type'):
```
mossberg: i'm wondering if we need an explicit check here. accessing `subclass.store_type` in the next line will error and implicitly communicate that same message to a Store subclass implementer that forgot it?

docs/index.rst
```diff
@@ -9,6 +9,7 @@ Manticore is a prototyping tool for dynamic binary analysis, with support for sy
 
    api
    models
+   internal
```
mossberg: i'm not sure we should have internal classes documented on readthedocs, I sort of feel like it will only encourage people to use these classes/methods that they shouldn't. the main people that should be reading those internal docs anyway are developers, who should be able to read in the source in their editor.

manticore/core/workspace.py
```diff
@@ -410,6 +413,21 @@ def __init__(self, desc=None):
...
+        """
+        Return a descriptor that created this workspace. Descriptors are of the
+        format <type>:<uri>, where type signifies the medium. For example,
+          fs:/tmp/workspace
+          redis:127.0.0.1:6379
+
+        :rtype: str
+        """
+        if self._descriptor is None:
+            self._descriptor = ':'.join([self._store.store_type, self._store.uri])
```
mossberg: style comment, `'{}:{}'.format()`?

docs/index.rst
yan: I mostly agree with you. I ended up adding them here when you asked about a docstring rendering correctly in Sphinx. Vote to remove?

manticore/core/workspace.py
yan: We don't need it, added it thinking explicit is better than implicit, but I don't think it's unclear without it. I don't feel strongly about this, can rm if you think it's a wart.

mossberg: I think if we want to be super idiomatic, i think excluding it is more along the lines of "Easier to ask for forgiveness than permission". i vote to remove, but also document the necessity of `store_type` in the `Store` docstring.

PR #444, opened 2017-08-09
Updated visualization plugin with coverage info & live update option

nettrino:


mossberg:
When I tried using the highlight plugin, I get this error in the log (but highlighting did still show up)
image

Also, when I tried highlighting, then clearing, then highlighting again, it didn't ask me for a trace file the second time, and seemed to just use the first one? Do you get this behavior? Also using the clear item seems to beachball binja for a little bit on my machine.

scripts/binaryninja/manticore_viz/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+interval = 1.5
+
+class Singleton(type):
+    _instances = {}
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            cls._instances[cls] = super(Singleton, cls).__call__(*args, **kwargs)
+        return cls._instances[cls]
+
+class TraceVisualizer(object):
```
mossberg: why should this be a singleton?

scripts/binaryninja/manticore_viz/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+            t = threading.Thread(target=self.highlight_from_file,
+                                 args=(self.workspace,))
+        elif os.path.isdir(self.workspace):
+            t = threading.Thread(target=self.highlight_from_dir,
+                                 args=(self.workspace,))
+        t.start()
+
+    def highlight_from_file(self, tracefile):
+        while True:
+            time.sleep(interval)
```
mossberg: move this to the bottom of the loop, so we don't unnecessarily sleep in non live mode? also below

scripts/binaryninja/manticore_viz/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+            for f in os.listdir(workspace_dir):
+                if f.endswith('trace'):
+                    self.highlight_trace(os.path.join(workspace_dir, f))
+            self.compute_coverage()
+            if not self.live_update:
+                break
+
+    def highlight_trace(self, tracefile):
```
mossberg: `highlight_trace` and `highlight_from_file` are a little confusing, readability-wise since they take the same argument and would seem to do the same thing

scripts/binaryninja/manticore_viz/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+        f = open(tracefile, "r")
+        fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
```
mossberg: is this locking stuff necessary? can't we read it all at once, and then do the highlighting/deduplication after?

scripts/binaryninja/manticore_viz/__init__.py
nettrino: The idea was that you only want a single class to be called whenever you instantiate things. E.g. you want to recall what you highlighted (also perhaps user's selection of a workspace?) so that you can run `clear` on it

scripts/binaryninja/manticore_viz/__init__.py
nettrino: yes because if it is a live  large trace you can write to it while it is being read no?

scripts/binaryninja/manticore_viz/plugin.json
```diff
@@ -0,0 +1,15 @@
+{
+	"plugin": {
+		"name": "ManticoreTrace",
+		"type": ["ui"],
+		"api": "python2",
+		"description": "Manticore Trace Visualization.",
+		"longdescription": "This plugin visualizes Manticore execution traces.\n",
+		"license": {
```
mossberg: i'm not sure what license we should have on this exactly, can we omit this for now?

scripts/binaryninja/manticore_viz/__init__.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def highlight_from_file(self, tracefile):
+        while True:
+            self.process_trace(tracefile)
+            if not self.live_update:
+                break
+            time.sleep(interval)
+
+    def highlight_from_dir(self, workspace_dir):
+        while True:
```
mossberg: since it looks like this live update thread will infinite loop thread, can we have a way to let the user end live visualization mode? maybe there could be "start live update" and "end live update" buttons

nettrino: `Clear` should suffice for that?

scripts/binaryninja/manticore_viz/plugin.json
nettrino: Perhaps remove the license overall and keep the json as it is recommended by the binja plugin format?

mossberg: sure, remove license and keep json sounds fine

mossberg:
awesome, this plugin is so much better now :)

PR #445, opened 2017-08-09
Implement logging presets

### LINKED ISSUE #431: Logging presets
We should decide on the logging presets that Manticore will come configured with as a convenience logging interface. This will allow users to have coarse control over logging without having to deal directly with the python logging infrastructure.

We can possibly implement a second, fine grain logging interface later for users that desire more control than the presets allow.

nettrino:
Fixes #431

### DIFF
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -21,7 +21,7 @@ def parse_arguments():
     parser = argparse.ArgumentParser(description='Symbolically analyze a program')
     parser.add_argument('--workspace', type=str, default=None,
                         help='A folder name for temporaries and results. (default mcore_?????)')
-    parser.add_argument('-v', action='count', default=1, help='Specify verbosity level from -v to -vvvv')
+    parser.add_argument('-v', action='count', default=0, help='Specify verbosity level from -v to -vvvv')
     parser.add_argument('--profile', action='store_true', help='Enable profiling mode.')
 
     parser.add_argument('--buffer', type=str, help='Specify buffer to make symbolic')
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -71,8 +71,6 @@ def main():
     args = parse_arguments()
 
     m = Manticore(args.programs[0], args.programs[1:])
-
-    set_verbosity(args.v)
     m.policy = args.policy
     m.args = args
 
@@ -112,6 +110,7 @@ def main():
     if args.assertions:
         m.load_assertions(args.assertions)
 
+    m.verbosity = args.v
 
     m.run(args.procs, args.timeout)
 
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -9,7 +9,7 @@ from multiprocessing import Manager, Pool, Process
 from threading import Timer
 from core.smtlib import Expression
 from manticore import Manticore
-from utils.mlogging import set_verbosity
+from .utils.logging import set_verbosity
 try:
     import psutil
 except ImportError:
@@ -22,7 +22,7 @@ def parse_arguments():
     parser = argparse.ArgumentParser(description='Symbolically analyze a program')
     parser.add_argument('--workspace', type=str, default=None,
                         help='A folder name for temporaries and results. (default mcore_?????)')
-    parser.add_argument('-v', action='count', default=0, help='Specify verbosity level from -v to -vvvv')
+    parser.add_argument('-v', action='count', default=1, help='Specify verbosity level from -v to -vvvv')
     parser.add_argument('--profile', action='store_true', help='Enable profiling mode.')
 
     parser.add_argument('--buffer', type=str, help='Specify buffer to make symbolic')
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -9,7 +9,7 @@ from multiprocessing import Manager, Pool, Process
 from threading import Timer
 from core.smtlib import Expression
 from manticore import Manticore
-from .utils.logging import set_verbosity
+from .utils.log import set_verbosity
 try:
     import psutil
 except ImportError:
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -9,6 +9,7 @@ from multiprocessing import Manager, Pool, Process
 from threading import Timer
 from core.smtlib import Expression
 from manticore import Manticore
+from utils.mlogging import set_verbosity
 try:
     import psutil
 except ImportError:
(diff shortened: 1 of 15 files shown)


manticore/utils/mlogging.py
```diff
@@ -0,0 +1,81 @@
...
+
+def loggerSetState(logger, stateid):
+    logger.filters[0].stateid = stateid
+
+for l in loggers:
+    logging.getLogger(l).addFilter(ctxfilter)
+    logging.getLogger(l).setState = types.MethodType(loggerSetState,
+                                                     logging.getLogger(l))
+
+def set_verbosity(setting):
```
mossberg: this should also be exposed to the api via `from manticore import set_verbosity`, which requires modification of manticore's `__init__.py`

manticore/__main__.py
```diff
@@ -9,6 +9,7 @@
 from threading import Timer
 from core.smtlib import Expression
 from manticore import Manticore
+from utils.mlogging import set_verbosity
```
mossberg: naming here, maybe just `logging` instead of `mlogging`?

nettrino: There is the logging built-in module.. should we change it to manticore_logging?

manticore/__main__.py
```diff
@@ -21,7 +22,7 @@ def parse_arguments():
     parser = argparse.ArgumentParser(description='Symbolically analyze a program')
     parser.add_argument('--workspace', type=str, default=None,
                         help='A folder name for temporaries and results. (default mcore_?????)')
-    parser.add_argument('-v', action='count', default=1, help='Specify verbosity level from -v to -vvvv')
+    parser.add_argument('-v', action='count', default=0, help='Specify verbosity level from -v to -vvvv')
```
mossberg: this probably shouldn't change, we decided that the cli default should be verbosity level 1 (https://github.com/trailofbits/manticore/issues/431#issuecomment-320022337)

manticore/manticore.py
```diff
@@ -294,37 +268,6 @@ def maxstorage(self):
     def maxstorage(self, max_storage):
         self._maxstorage = max_storage
 
-    @property
-    def verbosity(self):
```
japesinator: Can we keep this API? `m.verbosity = x` feels more idiomatic to me than `m.set_verbosity(x)`

manticore/utils/log.py
```diff
@@ -0,0 +1,81 @@
...
+    clamped = clamp(setting, 0, len(levels) - 1)
+    if clamped != setting:
+        logger.debug("%s not between 0 and %d, forcing to %d",
+                     setting,
+                     len(levels) - 1,
+                     clamped)
+    for level in range(clamped + 1):
+        for log_type, log_level in levels[level]:
+            logging.getLogger(log_type).setLevel(log_level)
+    # manticore_verbosity = clamped
```
japesinator: Was this left in by accident?

manticore/manticore.py
mossberg: Hm, we also never did reach a clear decision on whether to make this interface global, as opposed to associated with the Manticore class (@feliam mentioned it here https://github.com/trailofbits/manticore/issues/431#issuecomment-320025228), although to be honest, I'm starting to reconsider whether we should. Between the options of

```
m = Manticore('/bin/ls')
m.verbosity = 3
m.run()
```
and
```
from manticore import Manticore, set_verbosity
m = Manticore('/bin/ls')
set_verbosity(3)
m.run()
```
I do prefer the former, mostly because I don't like the namespace clutter that `set_verbosity` adds. Thoughts @trailofbits/manticore?

japesinator: As an example, I feel like https://github.com/trailofbits/manticore/pull/445/files#diff-1906be2ef8b0ab48f2d00bca1d684961R75 feels a little inconsistent with the two lines immediately after

nettrino: We can still leave it out of the class and just expose the setters and getters?

mossberg:
sidenote, for this PR name, may I suggest "Implement logging presets" which I think is a little more accurate than "logging template nits" :)

manticore/manticore.py
```diff
@@ -574,10 +528,11 @@ def _fork_state_callback(self, state, expression, values, policy):
             manticore_context['visited'] = manticore_visited.union(state_visited)
         state.context['visited_since_last_fork'] = set()
 
-        logger.info("Forking, about to store. (policy: %s, values: %s)", policy,
-                ', '.join('0x{:x}'.format(pc) for pc in values))
+        logger.debug("Forking, about to store. (policy: %s, values: %s)",
```
mossberg: can we have this message appear at verbosity=2 also? i realize this is somewhat troublesome to only log this message and not all of the other debug messages. one solution could be to introduce a custom log level in between info and debug, but i think an easier and pragmatic solution can simply be to move this logger call into the executor right before we do `self.publish('will_fork_state', ...)` with log level INFO. i also think it is more correct to emit this event from the executor since it really is more of an executor level event, rather than a higher level manticore event

manticore/manticore.py
```diff
@@ -290,40 +266,23 @@ def dumpafter(self, dump_after):
...
-        self._maxstorage = max_storage
-
     @property
     def verbosity(self):
-        '''
-        Convenience interface for setting logging verbosity to one of several predefined
-        logging presets. Valid values: 0-5
-        '''
-        return self._verbosity
+        """Interface for getting Manticore logging verbosity.
```
mossberg: super nit here: for properties, our auto generated docs only show the docstring for the getter, not the setter. so we need the getter docstring to also express that you can set it here. 
 (http://manticore.readthedocs.io/en/latest/api.html#manticore.Manticore.verbosity) Can we copy verbatim what we used to have, since I don't see a reason to change the language?

manticore/manticore.py
```diff
@@ -290,40 +266,23 @@ def dumpafter(self, dump_after):
...
-
-        clamped = clamp(setting, 0, len(levels) - 1)
-        if clamped != setting:
-            logger.debug("%s not between 0 and %d, forcing to %d", setting, len(levels) - 1, clamped)
-        for level in range(clamped + 1):
-            for log_type, log_level in levels[level]:
-                logging.getLogger(log_type).setLevel(log_level)
-        self._verbosity = clamped
+        """A call used to modify the level of output verbosity
+        :param int level: the level of verbosity to be used
```
mossberg: nit, the docstring uses `level` but the parameter is `setting`

PR #451, opened 2017-08-12
Fix state.generate_testcase

### LINKED ISSUE #450: state.generate_testcase() broken
`state.generate_testcase` has a typo which causes it to break. it publishes the wrong event, `will_generate_inputs`, instead of the existing `will_generate_testcase`.

https://github.com/trailofbits/manticore/blob/36a58810f905e299ff63a6208dc4e08eeb891869/manticore/core/state.py#L375

The fix for this is trivial, we just need to correct the string, but it's concerning that this bug was able to slip by without triggering test failures, or anything. To correct against these errors, we should probably declare these event key strings as constants somewhere and use those instead of literal strings.

mossberg:
Fixes #450 

Fixes the typo that caused the bug, and adds a unit test to check that `state.generate_testcase` publishes the proper event. It's not easy to write a unit test that actually checks that a testcase is generated due to how the State is coupled with the Executor for this functionality.

### DIFF
--- a/tests/test_state.py
+++ b/tests/test_state.py
@@ -199,10 +199,12 @@ class StateTest(unittest.TestCase):
 
 
     def _test_state_gen_helper(self, name, msg):
-        self.assertEqual(name, self._tmp_name)
-        self.assertEqual(msg, self._tmp_msg)
+        self.assertEqual(name, 'statename')
+        self.assertEqual(msg, 'statemsg')
+        raise Exception('callback')
+
     def test_state_gen(self):
         self.state.subscribe('will_generate_testcase', self._test_state_gen_helper)
-        self._tmp_name = 'statename'
-        self._tmp_msg = 'statemsg'
-        self.state.generate_testcase(self._tmp_name, self._tmp_msg)
+        with self.assertRaises(Exception) as cm:
+            self.state.generate_testcase('statename', 'statemsg')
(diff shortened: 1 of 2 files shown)


tests/test_state.py
```diff
@@ -196,4 +200,14 @@ def testContextSerialization(self):
         self.assertEqual( new_state.context['step'], 20)
         new_new_state = pickle.loads(new_new_file)
         self.assertEqual( new_new_state.context['step'], 30)
-        
+
+
+    def _test_state_gen_helper(self, name, msg):
+        self.assertEqual(name, 'statename')
+        self.assertEqual(msg, 'statemsg')
+        raise _CallbackExecuted
```
feliam: The exception stops the processing in the event handling. For a single event this is not important. Though..

Consider setting a flag in the callback or assertMethodIsCalled (from https://stackoverflow.com/questions/3829742/assert-that-a-method-was-called-in-a-python-unit-test)

mossberg: Since the assertMethodIsCalled in the link isn't stdlib, I'd prefer not to add it to our codebase when simpler solutions work. To me an exception seemed cleaner than adding some kind of flag variable to the class just for the purpose of this particular test. the fact that the exception stops event handling processing isn't a big issue here since all we care about is if the helper executed, and don't care about anything else, so i'm going to merge as is.

PR #459, opened 2017-08-17
Add gotchas docs

### LINKED ISSUE #419: Document gotchas
Manticore has a number of "gotchas": quirks or little things you need to do in a certain way otherwise you'll have crashes and other unexpected results. Known gotchas should be documented so users can be aware of them, and so we can be aware of them and strive to eliminate them.

mossberg:
Fixes #419

### DIFF
--- a/docs/gotchas.rst
+++ b/docs/gotchas.rst
@@ -1,0 +1,33 @@
+Gotchas
+=======
+
+Manticore has a number of "gotchas": quirks or little things you need to do in a certain way otherwise you'll have crashes and other unexpected results.
+
+Mutable context entries
+-----------------------
+
+Something like ``m.context['flag'].append('a')`` inside a hook will not work. You need to (unfortunately, for now) do ``m.context['flag'] += ['a']``. This is related to
+Manticore's built in support for parallel analysis and use of the `multiprocessing` library. This gotcha is specifically related to this note from the Python
+`documentation <https://docs.python.org/2.7/library/multiprocessing.html#multiprocessing.managers.SyncManager.list>`_ :
+
+    "Note: Modifications to mutable values or items in dict and list proxies will not be propagated through the manager, because the proxy has no way of knowing when its values or items are modified. To modify such an item, you can re-assign the modified object to the container proxy"
+
+
+Context locking
+---------------
+
+Manticore natively supports parallel analysis; if this is activated, client code should be careful to properly lock the global context when accessing it.
+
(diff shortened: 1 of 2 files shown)


docs/gotchas.rst
```diff
@@ -0,0 +1,33 @@
...
+Manticore's built in support for parallel analysis and use of the `multiprocessing` library. This gotcha is specifically related to this note from the Python
+`documentation <https://docs.python.org/2.7/library/multiprocessing.html#multiprocessing.managers.SyncManager.list>`_ :
+
+    "Note: Modifications to mutable values or items in dict and list proxies will not be propagated through the manager, because the proxy has no way of knowing when its values or items are modified. To modify such an item, you can re-assign the modified object to the container proxy"
+
+
+Context locking
+---------------
+
+Manticore natively supports parallel analysis; if this is activated, client code should be careful to properly lock the global context when accessing it.
```
feliam: I'd say client code should be always careful to properly lock.

docs/gotchas.rst
```diff
@@ -0,0 +1,33 @@
...
+
+An example of a global context race condition, when modifying two context entries.::
+
+    m.context['flag1'] += ['a']
+    --- interrupted by other worker
+    m.context['flag2] += ['b']
+
+Client code should use the :meth:`~manticore.Manticore.locked_context` API::
+
+    with m.locked_context() as global_context:
```
feliam: You have also 
with m.locked_context('feature_list', list) as feature_list:
    feature_list.append(1)

But it helps when you update a single the in the global_context

feliam: https://github.com/trailofbits/manticore/blob/5d87fc83b36c1405cffff7fb7193eb1359f077d3/manticore/manticore.py#L221

docs/gotchas.rst
```diff
@@ -0,0 +1,33 @@
...
+        global_context['flag1'] += ['a']
+        global_context['flag2'] += ['b']
```
feliam: maybe flags_1 instead of flag1 to hint for a list? (I can;t think of a better feature name)

branches? trace? feature_list?

docs/gotchas.rst
mossberg: I was thinking this should be documented in the `locked_context` documentation, not in our gotchas documentation?

feliam: sure!

mossberg: #460

PR #461, opened 2017-08-17
Update readme

### LINKED ISSUE #448: Update readme
Before the next release, we should update our readme.

- remove "partial" description of armv7 support, because we can do thumb now!
- remove pip 7.1.0 requirement, since we no longer require the --no-binary flag for regular installs because of the capstone version we specify. we do still require that flag for dev installs, but in general most people are not doing that, and mentioning a pip 7.1.0 requirement is just noise and unnecessary
- consider updating our example usage? we wrote that initial one very ad hoc, and we've added a bunch of apis

mossberg:
Various readme cleanups. Dev install instructions moved into the wiki.

Fixes #448

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,37 +82,7 @@ sudo pip install manticore
 
 Once installed, the `manticore` CLI tool and its Python API will be available.
 
-### For developers
-
-For a dev install that includes dependencies for tests, run:
-
-```
-git clone https://github.com/trailofbits/manticore.git && cd manticore
-pip install --no-binary keystone-engine -e .[dev]
-```
-
-You can run the tests with the commands below:
-
-```
-cd manticore
-# all tests
-nosetests
-# just one file
-nosetests tests/test_armv7cpu.py
-# just one test class
-nosetests tests/test_armv7cpu.py:Armv7CpuInstructions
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,12 +19,12 @@ Manticore supports binaries of the following formats, operating systems, and
...
 Examples of practical manticore usage are also [on github](https://github.com/trailofbits/manticore-examples).
 
-- OS/Formats: Linux ELF, Windows Minidump
-- Architectures: x86, x86_64, ARMv7 (partial)
+- OS/Formats: Linux ELF
+- Architectures: x86, x86_64, ARMv7
 
 ## Requirements
 
-Manticore is supported on Linux and requires Python 2.7, pip 7.1.0 or higher, and the [Z3 Theorem Prover](https://github.com/Z3Prover/z3/releases). Ubuntu 16.04 is strongly recommended.
```
yan: Leave in `pip`, but without a specified version? Not sure if we can assume `pip` is installed

mossberg: _technically_ manticore doesn't require pip to actually run and do binary analysis. I think it's fair to assume `pip` is available.

PR #466, opened 2017-08-21
Seed random on policy init to make results deterministic

### LINKED ISSUE #465: Pseudo-random state selection
Currently, the `Random` state selection policy is completely random. It would be useful to add support for providing an initial seed for analysis replayability.

japesinator:
Fixes #465

### DIFF
--- a/docs/gotchas.rst
+++ b/docs/gotchas.rst
@@ -30,4 +30,7 @@ Client code should use the :meth:`~manticore.Manticore.locked_context` API::
         global_context['flag1'] += ['a']
         global_context['flag2'] += ['b']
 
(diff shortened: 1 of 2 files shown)


japesinator:
There was some discussion in the internal slack about whether `random`'s default behavior should be truly random (as it was before), or take a user-defined seed (and also log the seed on each run). We came to the conclusion that replayability by default was our preferred UX, and having different seeds between runs solves almost no problems we were aware of (exceptions: users expecting truly random behavior based on the name, analyses sometimes "getting lucky" and bypassing a problematic state, pathological anti-manticore programs that deliberately abuse our prng algorithm)

PR #467, opened 2017-08-21
Basic Caching for Binja IL

nettrino:
Currently not caching IL instructions due to the `il_queue`, however this commit gives a speedup of 35%. It will be useful to minimize the calls to binja overall

### DIFF
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -723,36 +723,28 @@ class Cpu(Eventful):
         if insn.address != self.PC:
             return
 
+        name = self.canonicalize_instruction_name(insn)
+
+        def fallback_to_emulate(*operands):
+            text_bytes = ' '.join('%02x'%x for x in insn.bytes)
+            logger.info("Unimplemented instruction: 0x%016x:\t%s\t%s\t%s",
+                        insn.address, text_bytes, insn.mnemonic, insn.op_str)
+
+            self.publish('will_emulate_instruction', insn)
+            self.emulate(insn)
+            self.publish('did_emulate_instruction', insn)
+
+        implementation = getattr(self, name, fallback_to_emulate)
+
         if logger.level == logging.DEBUG :
             logger.debug(self.render_instruction(insn))
             for l in self.render_registers():
                 register_logger.debug(l)
 
-        self._insn_implementation(insn)
-        self._icount += 1
-        self.publish('did_execute_instruction', instruction)
-
-    def fallback_to_emulate(self, *operands):
-        insn = self.instruction
-        text_bytes = ' '.join('%02x'%x for x in insn.bytes)
-        logger.info("Unimplemented instruction: 0x%016x:\t%s\t%s\t%s",
-                insn.address, text_bytes, insn.mnemonic, insn.op_str)
-
-        self.publish('will_emulate_instruction', insn)
-        self.emulate(insn)
-
-        self.publish('did_emulate_instruction', insn)
-
-    def _insn_implementation(self, insn):
-        name = self.canonicalize_instruction_name(insn)
-
-        implementation = getattr(self, name, self.fallback_to_emulate)
         implementation(*insn.operands)
-        self.update_pc()
+        self._icount += 1
 
-    # to be overriden if needed
-    def update_pc(self):
-        pass
+        self.publish('did_execute_instruction', insn)
 
     def emulate(self, insn):
         '''
--- a/manticore/core/cpu/binja.py
+++ b/manticore/core/cpu/binja.py
@@ -427,8 +427,9 @@ class BinjaCpu(Cpu):
             for l in self.render_registers():
                 register_logger.debug(l)
 
-        assert ((hasattr(insn, "sets_pc") and insn.sets_pc) or
-                self.PC == self._last_pc)
+        assert (self.PC == self._last_pc or
+                (isinstance(insn, BinjaILDisasm.BinjaILInstruction) and
+                 insn.sets_pc))
 
         implementation(*insn.operands)
 
@@ -448,7 +449,7 @@ class BinjaCpu(Cpu):
         #
         # we might be executing a Capstone instruction at this point
         # if we context-switched, so check the sets_pc attr
-        if not (hasattr(insn, "sets_pc") and
+        if not (isinstance(insn, BinjaILDisasm.BinjaILInstruction) and
                 (insn.sets_pc or self.disasm.il_queue)):
             self.PC = self._last_pc + insn.size
 
--- a/manticore/core/cpu/binja.py
+++ b/manticore/core/cpu/binja.py
@@ -17,6 +17,7 @@ from ..smtlib import Operators, BitVecConstant, operator
 from ...utils.helpers import issymbolic
 
 logger = logging.getLogger("CPU")
+register_logger = logging.getLogger("REGISTERS")
 
 
 class BinjaRegisterFile(RegisterFile):
@@ -343,7 +344,7 @@ class BinjaCpu(Cpu):
             c = self.memory[address]
 
             if issymbolic(c):
-                assert isinstance(c, BitVec) and  c.size == 8
+                assert isinstance(c, BitVec) and c.size == 8
                 if isinstance(c, Constant):
                     c = chr(c.value)
                 else:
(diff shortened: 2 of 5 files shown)


manticore/core/cpu/disasm.py
```diff
@@ -141,6 +138,18 @@ def unimplemented(self, il):
         return (il.operation == enums.LowLevelILOperation.LLIL_UNIMPL or
                 il.operation == enums.LowLevelILOperation.LLIL_UNIMPL_MEM)
 
+    def _get_current_llil_func(self, code, pc):
+        if pc in self.llil_func_cache:
```
joshwatson: One thing you might want to consider is checking the target of an `LLIL_LOAD` operation, and removing any address from the cache that would be modified by that `LLIL_LOAD`. Otherwise, you could end up with cached instructions that do not match the underlying binary data.

nettrino: You mean for cases like self-modifying code? I think one of the biggest issues right now is the fact that we maintain the il_queue for instructions like `div`, `if` etc. Also there are multiple calls to binja which should be minimized, because just the mere task of fetching one 
instruction resolves into multiple calls. Perhaps it is worth it adding a feature where we load the IL directly from a serialized object, with all the functions and indexes stored there, so that we don't do any lookups.

manticore/core/cpu/binja.py
```diff
@@ -380,6 +381,79 @@ def decode_instruction(self, pc):
...
+                self.emulate(insn)
+
+        implementation = getattr(self, name, fallback_to_emulate)
+
+        if logger.level == logging.DEBUG :
+            logger.debug(self.render_instruction(insn))
+            for l in self.render_registers():
+                register_logger.debug(l)
+
+        assert ((hasattr(insn, "sets_pc") and insn.sets_pc) or
```
japesinator: `getattr(insn, 'sets_pc', False)`?

manticore/core/cpu/binja.py
```diff
@@ -1225,6 +1268,11 @@ def make_symbolic(flag_expr):
 def insn_flags(cpu):
     # FIXME temporary hack due to binja issue with flags
     f = cpu.disasm.current_func
+    #  print "------"
```
japesinator: These should probably either be removed or converted to debug logs

manticore/core/cpu/disasm.py
```diff
@@ -141,6 +138,18 @@ def unimplemented(self, il):
...
+            return self.llil_func_cache[pc]
+
+        from binaryninja import Architecture, LowLevelILFunction
```
japesinator: Move to top?

manticore/core/cpu/disasm.py
```diff
@@ -141,6 +138,18 @@ def unimplemented(self, il):
...
+        # FIXME
+        func = LowLevelILFunction(Architecture['x86_64'])
+        func.current_address = pc
+        size = self.view.arch.get_instruction_low_level_il(code, pc, func)
+        self.llil_func_cache[pc] = (func, size)
+        return func, size
```
japesinator: It might be counterintuitive to return both `func` and `size` from a function called `_get_current_llil_func`, we should consider either splitting the function in two or renaming it

manticore/core/cpu/binja.py
```diff
@@ -380,6 +381,79 @@ def decode_instruction(self, pc):
...
+
+        # don't bump the PC if we are in an LLIL that has set it,
+        # or if there are pending IL insn in the queue. This is because
+        # for cases where we have other il instructions in the queue,
+        # such as when we get a divu insn, the PC + size will point
+        # to the next assembly instruction and not the next LLIL
+        #
+        # we might be executing a Capstone instruction at this point
+        # if we context-switched, so check the sets_pc attr
+        if not (hasattr(insn, "sets_pc") and
```
japesinator: maybe just `not (getattr(insn, 'sets_pc', False) or self.disasm.il_queue)`? I think as-is, if there's a queue but the current insn doesn't have a `sets_pc` attr, this conditional won't pass

manticore/core/cpu/binja.py
```diff
@@ -380,6 +381,79 @@ def decode_instruction(self, pc):
...
+                self.FALLBACK(name, *operands)
+                # XXX after this point self.PC != self._last_pc but that is
+                # OK because we will update the PC  properly
+            else:
+                text_bytes = ' '.join('%02x'%x for x in insn.bytes)
+                logger.info("Unimplemented instruction: 0x%016x:\t%s\t%s\t%s",
+                            insn.address, text_bytes, insn.mnemonic, insn.op_str)
+
+                self.publish('will_emulate_instruction', insn)
+                self.emulate(insn)
```
japesinator: We should publish `'did_emulate_instruction'` here, right?

manticore/core/cpu/binja.py
```diff
@@ -380,6 +381,79 @@ def decode_instruction(self, pc):
...
+
+        self.publish('will_decode_instruction', self.PC)
+
+        insn = self.decode_instruction(self.PC)
+        self._last_pc = self.PC
+
+        self.publish('will_execute_instruction', insn)
+
+        # FIXME (theo) why just return here?
+        if insn.address != self.PC:
```
japesinator: I'm kinda confused about when this happens, but I think we should at least log when it does

manticore/core/cpu/abstractcpu.py
```diff
@@ -723,36 +723,28 @@ def execute(self):
...
-        self._icount += 1
-        self.publish('did_execute_instruction', instruction)
+        name = self.canonicalize_instruction_name(insn)
 
-    def fallback_to_emulate(self, *operands):
-        insn = self.instruction
-        text_bytes = ' '.join('%02x'%x for x in insn.bytes)
-        logger.info("Unimplemented instruction: 0x%016x:\t%s\t%s\t%s",
-                insn.address, text_bytes, insn.mnemonic, insn.op_str)
+        def fallback_to_emulate(*operands):
```
japesinator: This should still take `self` as the first argument, right?

manticore/core/cpu/binja.py
nettrino: There will never be a case where you will have an `il_queue` but not a `sets_pc`. Essentially
the only reason you check for `sets_pc` is because you could be executing a captstone instruction there, perhaps it makes more sense to check for that.

manticore/core/cpu/disasm.py
nettrino: We want to keep this internal to the function since a user might not have binaryninja installed, thus we don't want to do get import errors. You mean in the top of the function? If it is cached won't we be doing an import that is not used?

manticore/core/cpu/abstractcpu.py
nettrino: It is internal to the function, so no need for a `self`?

manticore/core/cpu/disasm.py
joshwatson: There's a couple of other cases, as well. First, if you're dealing with JIT code, your cache would be invalid if a page containing JIT code was freed and replaced with a new page of JIT code. Second, if a dynamic library is unloaded and another is loaded, it could potentially be placed in the same location as the old one. In both cases, the cache would then be invalid.

nettrino: Hm I see. So right now, we only cache the Functions and LLILFunctions, and not the instructions. This is again a problem right?

manticore/core/cpu/abstractcpu.py
japesinator: Whoops, my bad

manticore/core/cpu/binja.py
japesinator: +1 for refactoring to check explicitly for that

manticore/core/cpu/binja.py
mossberg: I remember being kind of confused at this recently also, but i suspect this check is not necessary anymore -- insn.address should always be == self.PC

i think this check was first introduced in this pr when we restructured when the instruction was decoded (moved it to executor) so that `state.cpu.instruction` in a hook would refer to the instruction about to be executed. https://github.com/trailofbits/manticore/pull/205/files#diff-6ad9911cdf40511ac94cf12c0ea2b36dR432

however, i'm pretty that check was there just in case; the body shouldn't ever have actually been executed in any normal operation of manticore

a little later, in the dev-events merge we reverted some of this, and no longer decode instructions in the executor, and move it back to the cpu. https://github.com/trailofbits/manticore/pull/341/files#diff-cb6130b98f84db152ec7d0078d15e4e4L689

in that pr, this code got refactored (https://github.com/trailofbits/manticore/pull/341/files#diff-6ad9911cdf40511ac94cf12c0ea2b36dR738), but i think since we reverted the original reason that check was introduced, i think it should be ok to remove this

@yan does this sound right to you?

yan: That does sound right; it's likely not needed anymore. If I recall correctly, it was a temporary fix to get around instruction caching issues when we separated decoding from execution.

Issue #342, opened 2017-06-21
Initialize Executor in Manticore ctor


mossberg:
We currently init `Executor` in `Manticore.run`, however this is not ideal. For example, it complicates `Manticore.locked_context`. A problem with initializing Executor in Manticore ctor is that Manticore has APIs that allow modification of parameters that Executor is initialized with, therefore Manticore and Executor could potentially become out of sync.

Will removing the APIs for modifying that state that's passed into Executor solve this? If so, let's do this.

PR #471, opened 2017-08-23
Initialize executor at manticore initialization

japesinator:
This should make setting policy or adding callbacks much easier

### DIFF
--- a/manticore/manticore.py
+++ b/manticore/manticore.py
@@ -20,7 +20,7 @@ from .core.executor import Executor
 from .core.parser import parse
 from .core.state import State, TerminateState
 from .core.smtlib import solver, ConstraintSet
-from .core.workspace import ManticoreOutput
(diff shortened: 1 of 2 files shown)


japesinator:
Fixes #342

manticore/manticore.py
```diff
@@ -315,6 +337,18 @@ def verbosity(self, setting):
     def maxstorage(self, max_storage):
         self._maxstorage = max_storage
 
+    @property
+    def workspace(self):
+        return self._workspace
+
+    @workspace.setter
+    def workspace(self, ws):
```
feliam: Do we want to change the workspace after creation? Is there a case we do not know the workspace at constructor time ?

japesinator: I think that changing the workspace after analysis has started could potentially do some pretty weird things (what if states are already written to that workspace?) and doesn't seem like something a user would need to do.

PR #475, opened 2017-08-31
Allow Manticore initialization with State object

japesinator:
This allows initialization of `Manticore` objects with a `State` object instead of just a path to a binary

### DIFF
--- a/manticore/manticore.py
+++ b/manticore/manticore.py
@@ -156,15 +156,21 @@ class Manticore(object):
     :type args: list[str]
     :ivar dict context: Global context for arbitrary data storage
     '''
-    def __init__(self, binary_path, args=None, disasm='capstone'):
-        assert os.path.isfile(binary_path)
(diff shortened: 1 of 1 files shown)


manticore/manticore.py
```diff
@@ -156,15 +156,21 @@ class Manticore(object):
...
     '''
-    def __init__(self, binary_path, args=None, disasm='capstone'):
-        assert os.path.isfile(binary_path)
+    def __init__(self, path_or_state, args=None, disasm='capstone'):
 
         args = [] if args is None else args
 
+        if isinstance(path_or_state, str):
+            assert os.path.isfile(path_or_state)
+            self._binary = path_or_state
```
feliam: Do we need this ? Is'nt this in the State too ?

mossberg:
just a quick note, if we decide that being able to pass a state into `Manticore` is part of the public API, we should update the docstring

PR #476, opened 2017-08-31
Update Manticore docstring

japesinator:


manticore/manticore.py
```diff
@@ -151,7 +151,7 @@ class Manticore(object):
     '''
     The central analysis object.
 
-    :param str binary_path: Path to binary to analyze
+    :param str path_or_state: Path to a binary to analyze or `State` object
```
mossberg: nice - we should also update the `str` type annotation next to `param`. we should use a `:type ...` block to do the multiple types thing

mossberg:
my comment got hidden, but i also commented: 

also, it's verbose but can we do

```
:class:`~manticore.core.state.State`
```

so we get a nice link in the docs? (and can you please build the docs to confirm?)

PR #477, opened 2017-09-01
Update examples

### LINKED ISSUE #424: Update example scripts to not use `dump_stats`
We don't currently have `dump_stats` in the API, but it's used in a [bunch](https://github.com/trailofbits/manticore/search?utf8=%E2%9C%93&q=dump_stats&type=) of example scripts

yan:
The few examples in README.md are either not entirely correct or produce warnings. This fixes warnings and output examples.

Fixes #424

### DIFF
--- a/examples/script/multi_arch_sym.py
+++ b/examples/script/multi_arch_sym.py
@@ -35,6 +35,3 @@ if __name__ == '__main__':
     m.verbosity = 2
     m.run()
 
-    # Print high level statistics
-    m.dump_stats()
-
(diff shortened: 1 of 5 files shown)


examples/script/multi_arch_sym.py
```diff
@@ -35,6 +35,3 @@ def entered_func(state):
     m.verbosity = 2
```
mossberg: while we're here, can we clean up the `start()` typo on the above line (also in the other script)? maybe we can even remove "start() returns when Manticore finishes" and put something equivalent in the API docs.

yan: Was adding something similar to the API docs, then took a moment -- isn't this implied?

Issue #464, opened 2017-08-19
command-line symbolic argv?


Owlz:
I've been having trouble getting command-line symbolic argv to work. For instance:

```bash
$ manticore ./test AAAAAAAAAAAAAAAAA
```

The "A" values do not appear to become symbolic. Nor does the argv[0] ("./test") string.

feliam:
I don't think you can make argv[0] symbolic from the command line. For that you'll need to dig into the API. 
As for argv[N+1] you can do something like this ...

manticore ./test +++++++++++++++

'+' will be replaced by free symbolic bytes.

Owlz:
Ah, you're right the + worked for that. Could also be useful to have argv[0] be symbolic as well. My guess is to accomplish that you would actually need a flag on the command line, like --symbolic-argv [N [N] [N]].

mossberg:
symbolic arg[0] is an interesting thought, but i can't say it feels like a common use case to me. it would be possible currently from the API by hooking and overwriting argv[0] with a `State.new_symbolic_buffer()`

since the original question seemed to be answered, going to close

PR #482, opened 2017-09-06
Document cli better

### LINKED ISSUE #472: Better common usage documentation
We should have more documentation on use of the command line tool, which essentially does not exist. Doing so would help the likely many users that run into questions like #464 .

At a minimum, it should describe invoke the manticore cli, and how to pass concrete or symbolic arguments.

mossberg:
Add more cli documentation so people don't run into issues like #464 

Fixes #472

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,8 +101,8 @@ have to manually set the workspace to the redis URI.
 
 ```
 $ manticore ./path/to/binary        # runs, and creates a mcore_* directory with analysis results
-$ manticore ./path/to/binary ab cd  # use concrete strings "ab", "cd" as argv
-$ manticore ./path/to/binary ++ ++  # use two symbolic strings of length two as argv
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -100,7 +100,9 @@ have to manually set the workspace to the redis URI.
 ## Usage
 
 ```
-$ manticore ./path/to/binary  # runs, and creates a mcore_* directory with analysis results
+$ manticore ./path/to/binary        # runs, and creates a mcore_* directory with analysis results
+$ manticore ./path/to/binary ab cd  # use concrete strings "ab", "cd" as argv
+$ manticore ./path/to/binary ++ ++  # use two symbolic strings of length two as argv
```
yan: "as argv" implies as the entire argv array. How about "Use symbolic strings of length two for `argv[1]` and `argv[2]`?" or "Use two symbolic strings of length two as program arguments". (Ditto for above)

PR #483, opened 2017-09-07
Implemented getcwd sys call, along with a new helper function that wr…

johnfxgalea:
…ites a NULL terminated string to memory.

The reason for this commit is that the sys call was not implemented.

### DIFF
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -554,22 +554,19 @@ class Cpu(Eventful):
 
     def write_string(self, where, string, max_length=None):
         '''
-        Writes a str to memory, appending a NULL-terminator at the end.
+        Writes a string to memory, appending a NULL-terminator at the end.
         :param int where: Address to write the string to
+        :param str string: The string to write to memory
         :param int max_length:
             The size in bytes to cap the string at, or None [default] for no
             limit. This includes the NULL terminator.
-        :type string: str
         '''
         
         if max_length is not None:
             string = string[:max_length-1]
         
-        # Conver the string to a list, and append 0 to it.
-        data = list(string)
-        data.append(0)
+        self.write_bytes(where, string + '\x00')
         
-        self.write_bytes(where, data)
 
     def read_string(self, where, max_length=None):
         '''
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -552,6 +552,25 @@ class Cpu(Eventful):
             result.append(Operators.CHR(self.read_int(where + i, 8)))
         return result
 
+    def write_string(self, where, string, max_length=None):
+        '''
+        Writes a str to memory, appending a NULL-terminator at the end.
+        :param int where: Address to write the string to
+        :param int max_length:
+            The size in bytes to cap the string at, or None [default] for no
+            limit. This includes the NULL terminator.
+        :type string: str
+        '''
+        
+        if max_length is not None:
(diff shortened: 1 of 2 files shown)


yan:
Thank you for this PR 👍 ! I have a few comments inline:

manticore/core/cpu/abstractcpu.py
```diff
@@ -552,6 +552,25 @@ def read_bytes(self, where, size):
...
+        :param int max_length:
+            The size in bytes to cap the string at, or None [default] for no
+            limit. This includes the NULL terminator.
+        :type string: str
+        '''
+        
+        if max_length is not None:
+            string = string[:max_length-1]
+        
+        # Conver the string to a list, and append 0 to it.
```
yan: Two comments, one stylistic and one structural: we generally prefer not to comment _what_ the code does, but _how_ it does it, so commenting that you're converting and appending is probably not needed. Second, this feels verbose and since `write_string` is a function that might see regular use in the future, should not perform unneeded conversions. I'd be inclined to replace following three lines with just:

`self.write_bytes(string + chr(0))` (or `'\x00'` instead of the cast)

manticore/core/cpu/abstractcpu.py
```diff
@@ -552,6 +552,25 @@ def read_bytes(self, where, size):
             result.append(Operators.CHR(self.read_int(where + i, 8)))
         return result
 
+    def write_string(self, where, string, max_length=None):
+        '''
+        Writes a str to memory, appending a NULL-terminator at the end.
+        :param int where: Address to write the string to
+        :param int max_length:
```
yan: Add `string` in the param list as the other params? Mostly for consistency :)

manticore/platforms/linux.py
```diff
@@ -1029,6 +1029,38 @@ def sys_chdir(self, path):
...
+        
+            if size > 0 and size < (len(current_dir) + 1):
+                logger.info("GETCWD: size is greater than 0, but is smaller than the length"  
+                            "of the +path + 1. Returning ERANGE")
+                return -errno.ERANGE
+        
+            # Write path to string
+            self.current.write_string(buf, current_dir)
+            
+            assert self.current.read_string(buf) == current_dir
```
yan: Under what circumstances would this assert be false?

manticore/platforms/linux.py
```diff
@@ -1029,6 +1029,38 @@ def sys_chdir(self, path):
...
+            
+            logger.debug("getcwd(0x%08x, %u) -> <%s> (Size %u)", buf, size, current_dir, size)
+
+            return 0
+        except OSError as e:
+            return e.errno
```
yan: `e.errno` is a positive value, `sys_getcwd` should return a negative error code. Probably a forgotten negation.

manticore/platforms/linux.py
```diff
@@ -1029,6 +1029,38 @@ def sys_chdir(self, path):
...
+            if not os.path.isdir(current_dir):
+                logger.info("GETCWD: Working directory does not exist. Returning ENOENT")
+                return -errno.ENOENT
+                      
+            if size > 0 and size < length:
+                logger.info("GETCWD: size is greater than 0, but is smaller than the length"  
+                            "of the path + 1. Returning ERANGE")
+                return -errno.ERANGE
+        
+            if not (buf in self.current.memory and buf+length in self.current.memory):
```
yan: We actually provide an api for this check, you can rewrite the above check as:
`if not self.current.memory.access_ok(slice(buf, buf+length), 'w'):`

PR #490, opened 2017-09-08
Added implementation for sys_dup and sys_dup2

johnfxgalea:


johnfxgalea:
Let me know of any suggestions. Happy to look into them.

manticore/platforms/linux.py
```diff
@@ -1331,7 +1331,43 @@ def sys_rt_sigprocmask(self, cpu, how, newset, oldset):
...
+    
+    def sys_dup(self, fd):
+        '''
+        Duplicates an open file descriptor
+        :rtype: int
+        :param fd: the open file descriptor to duplicate.
+        :return: the new file descriptor.
+        '''
+        
+        if self.files[fd] is None:
```
yan: Thank you again for these, really great to see these syscalls finally get added. A few notes on this PR:

1) `fd` might not be in the `files` list and can raise; we made `_get_fd` for those extra checks; check how `lseek` is implemented.

manticore/platforms/linux.py
```diff
@@ -1331,7 +1331,43 @@ def sys_rt_sigprocmask(self, cpu, how, newset, oldset):
...
+        :param fd: the open file descriptor to duplicate.
+        :param newfd: the file descriptor to alias the file described by fd.
+        :return: newfd.
+        '''
+        
+        if self.files[fd] is None:
+            logger.info("DUP2: Passed fd is not open. Returning EBADF")
+            return -errno.EBADF
+            
+        if self.files[newfd] is not None:
```
yan: 2) Both of the checks in dup2 should account for `fd` not existing in `files`.
3) Make sure that `newfd` is within bounds of the `files` array. (might have to create it; consult kernel's dup2 implementation)

manticore/platforms/linux.py
```diff
@@ -1331,7 +1337,57 @@ def sys_rt_sigprocmask(self, cpu, how, newset, oldset):
...
+        Duplicates an open fd to newfd. If newfd is open, it is first closed
+        :rtype: int
+        :param fd: the open file descriptor to duplicate.
+        :param newfd: the file descriptor to alias the file described by fd.
+        :return: newfd.
+        '''
+        #TODO(yan): Once getrlimit(2) is implemented for getdtablesize, move
+        # this to a platform instance var. 256 is a conservative size that gets
+        # returned if rlimit() errors.
+        MAX_DTABLE_SIZE = 256
```
mossberg: can we move this to an instance variable now? and move this comment to a stub getrlimit(2) that just raises NotImplementError? also out of curiosity where is the dtable terminology used? i don't see it in the getrlimit man page or kernel source around getrlimit.

yan: Check man pages for getdtablesize(3) and sure.

mossberg:
@yan your commit looks good to me 👍

manticore/platforms/linux.py
```diff
@@ -312,6 +315,11 @@ def __init__(self, program, argv=None, envp=None, disasm='capstone', **kwargs):
         self.programs = program
         self.disasm = disasm
 
+        # dict of [int -> (int, int)] where tuple is (soft, hard) limits
+        self.rlimits = {
```
mossberg: nit, we could probably prefix with `_`

manticore/platforms/linux.py
```diff
@@ -1716,8 +1726,16 @@ def sys_futex(self, uaddr, op, val, timeout, uaddr2, val3):
         logger.debug("sys_futex(...) -> -1")
         return -1
     def sys_getrlimit(self, resource, rlim):
-        logger.debug("sys_getrlimit(%x, %x) -> -1", resource, rlim)
-        return -1
+        ret = -1
+        if resource in self.rlimits:
+            rlimit_tup = self.rlimits[resource]
+            # 32 bit values on both 32 and 64 bit platforms. For more info,
+            # see the BUGS section in rlimit(2) man page.
```
mossberg: `man 2 rlimit` returns nothing for me on ubuntu 16.04, is this a typo?

manticore/platforms/linux.py
```diff
@@ -292,6 +292,9 @@ class Linux(Platform):
     This class emulates the most common Linux system calls
     '''
 
+    # from /usr/include/asm-generic/resource.h
+    RLIMIT_NOFILE = 7 #/* max number of open files */
```
mossberg: just a question: 7 is the max number of file a process can have open? doesn't this seem really low?

manticore/platforms/linux.py
```diff
@@ -1331,7 +1347,57 @@ def sys_rt_sigprocmask(self, cpu, how, newset, oldset):
...
+        Duplicates an open fd to newfd. If newfd is open, it is first closed
+        :rtype: int
+        :param fd: the open file descriptor to duplicate.
+        :param newfd: the file descriptor to alias the file described by fd.
+        :return: newfd.
+        '''
+        #TODO(yan): Once getrlimit(2) is implemented for getdtablesize, move
+        # this to a platform instance var. 256 is a conservative size that gets
+        # returned if rlimit() errors.
+        MAX_DTABLE_SIZE = 256
```
mossberg: we can remove this variable now, and just reference self.rlimit?

manticore/platforms/linux.py
yan: No, 7 is the magic value that refers to `RLIMIT_NOFILE ` constant; i didn't include others from the header since they're not used yet.

manticore/platforms/linux.py
yan: Sorry, meant `getrlimit(2)`

manticore/platforms/linux.py
```diff
@@ -312,6 +315,11 @@ def __init__(self, program, argv=None, envp=None, disasm='capstone', **kwargs):
         self.programs = program
         self.disasm = disasm
 
+        # dict of [int -> (int, int)] where tuple is (soft, hard) limits
```
mossberg: if we wanted to get extra fancy, we could use a NamedTuple here and omit the comment describing the tuple. now or later, either is fine though.

manticore/platforms/linux.py
```diff
@@ -1331,7 +1347,53 @@ def sys_rt_sigprocmask(self, cpu, how, newset, oldset):
...
+        :param newfd: the file descriptor to alias the file described by fd.
+        :return: newfd.
+        '''
+        try:
+            file = self._get_fd(fd)
+        except BadFd:
+            logger.info("DUP2: Passed fd is not open. Returning EBADF")
+            return -errno.EBADF
 
+        soft_max, hard_max = self._rlimits.get(self.RLIMIT_NOFILE, (256, 1024))
```
mossberg: do we need to use `.get()` here? can't we be confident `self.RLIMIT_NO_FILE` will exist in `_rlimits`?

mossberg:
@yan i've looked through your commits, and they look good to me

PR #491, opened 2017-09-08
Fix logger verbosity and test

mossberg:
Fixes a minor logging verbosity regression (load state messages were being logged on verbosity=1) (see #431)

Also improve test to check for this.

### DIFF
--- a/manticore/manticore.py
+++ b/manticore/manticore.py
@@ -586,10 +586,10 @@ class Manticore(object):
                 self._assertions[pc] = ' '.join(line.split(' ')[1:])
 
     def _store_state_callback(self, state, state_id):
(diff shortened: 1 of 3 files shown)


manticore/manticore.py
```diff
@@ -585,11 +584,8 @@ def load_assertions(self, path):
                     logger.debug("Repeated PC in assertions file %s", path)
                 self._assertions[pc] = ' '.join(line.split(' ')[1:])
 
-    def _store_state_callback(self, state, state_id):
-        logger.info("store state %r", state_id)
-
     def _load_state_callback(self, state, state_id):
```
yan: Remove callback altogether if it's not doing anything?

PR #494, opened 2017-09-12
Use z3-solver package

### LINKED ISSUE #219: Use z3-solver pypi package for z3 installation
Why not use python z3? The wrapper does basically the same things that z3 command line does, and anything that doesn't directly have a function in the library is exposed through other means.

This would be helpful because you could then pip install z3-solver as well, which would remove another package being installed directly to the system.

mossberg:
Updates setup.py to install z3 via pip, and removes docs references to installing z3 via apt.

Fixes #219

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Examples of practical manticore usage are also [on github](https://github.com/tr
 
 ## Requirements
 
-Manticore is supported on Linux, and requires Python 2.7 and the [Z3 Theorem Prover](https://github.com/Z3Prover/z3/releases). Ubuntu 16.04 is strongly recommended.
+Manticore is supported on Linux, and requires Python 2.7. Ubuntu 16.04 is strongly recommended.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -119,7 +117,7 @@ m = Manticore('./path/to/binary')
 def hook(state):
   cpu = state.cpu
   print 'eax', cpu.EAX
-  print cpu.read_int(cpu.SP)
+  print cpu.read_int(cpu.ESP)
```
yan: Any reason for updating to `ESP`? As opposed to leaving it or updating to `STACK`.

mossberg: The intention here is stack pointer for 32 bit mode, but x86 actually has a register called `SP` which is the low 16 bits of the stack ptr, so that's what this was actually referencing which is incorrect. I think using native reg name here is better than `STACK` as it's more intuitive.

PR #495, opened 2017-09-13
Fix stderr file creation and test for it going forwards

### LINKED ISSUE #484: stderr output file not generated
There is a typo on this line that causes Manticore to not generate files for stderr.

https://github.com/trailofbits/manticore/blob/d83fa737a79937fc8ee639aab601743c38707b54/manticore/core/workspace.py#L537

The string `stdout` should be `stderr`.

japesinator:
Fixes #484

### DIFF
--- a/tests/test_workspace.py
+++ b/tests/test_workspace.py
@@ -1,3 +1,5 @@
+import os
+import shutil
 import signal
 import unittest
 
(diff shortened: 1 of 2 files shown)


tests/test_workspace.py
```diff
@@ -108,3 +110,10 @@ def test_output(self):
         self.assertIn('messages', keys)
```
mossberg: could it be possible to simply add another assert In line here for stderr instead of a separate test function?

PR #498, opened 2017-09-13
Refactor CLI, and Manticore high level interfaces

mossberg:
- Manticore ctor
- Manticore classmethods
- Controversial Manticore.verbosity() refactor as felipe's been suggesting: what do yall think? I added it because `m.verbosity = 2`, while a nice interface, doesn't let you control logging verbosity of what happens during Manticore ctor. and now, in Linux, the binary is loaded during the Manticore ctor (vs when you call m.run())

### DIFF
--- a/manticore/manticore.py
+++ b/manticore/manticore.py
@@ -211,9 +211,7 @@ class Manticore(object):
 
         if isinstance(path_or_state, str):
             assert os.path.isfile(path_or_state)
-            # self._initial_state = make_initial_state(path_or_state, **kwargs)
             self._initial_state = make_initial_state(path_or_state, argv=linux_argv, env=linux_envp, **kwargs)
-            # self._initial_state = make_linux(path_or_state, linux_argv, linux_envp)
         elif isinstance(path_or_state, State):
             self._initial_state = path_or_state
         else:
@@ -233,7 +231,7 @@ class Manticore(object):
     def decree(cls, path, concrete_data='', **kwargs):
         try:
             return cls(make_decree(path, concrete_data), **kwargs)
-        except KeyError:
+        except KeyError:  # FIXME(mark) magic parsing for DECREE should raise better error
             raise Exception('Invalid binary: {}'.format(path))
 
     @property
--- a/manticore/manticore.py
+++ b/manticore/manticore.py
@@ -170,7 +170,7 @@ class Manticore(object):
     :ivar dict context: Global context for arbitrary data storage
     '''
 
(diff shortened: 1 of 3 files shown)


manticore/manticore.py
```diff
@@ -169,8 +170,11 @@ class Manticore(object):
     :ivar dict context: Global context for arbitrary data storage
     '''
 
-    def __init__(self, path_or_state, workspace_url=None, policy='random', **kwargs):
-        
+    def __init__(self, path_or_state, linux_argv=None, linux_envp=None, workspace_url=None, policy='random', **kwargs):
+
+        linux_argv = [] if linux_argv is None else linux_argv
```
feliam: Why not move this to kwargs  and let Linux() do the check

manticore/manticore.py
```diff
@@ -271,19 +296,12 @@ def _real_context():
                 yield ctx
                 context[key] = ctx
 
-    @property
-    def verbosity(self):
+    @staticmethod
+    def verbosity(level):
```
yan: Should this be `set_verbosity`?

manticore/manticore.py
mossberg: I like keeping the actual arguments, because then it is easy to tell what arguments this function takes. but yes, omit the checks here, and save them for later

feliam: But we don't really know what arguments this may take as other platforms are be added. I say.. kwargs to the rescue. Like the cooperative classes thing. no?

mossberg: the thought is when platforms are added, they should use their respective classmethod, not the Manticore ctor -- this is a legacy interface only for linux?

manticore/manticore.py
mossberg: Perhaps, I personally find `Manticore.verbosity(2)` more aesthetically pleasing than `Manticore.set_verbosity(2)`, but it's a mild preference

manticore/manticore.py
mossberg: btw, kwargs is still here, and forwarded to make_initial_state

mossberg: another thought is that we should just break the API - it's mildly inconvenient for some people, but perhaps we are too young to care about legacy cruft?

yan: break the API in what way? just in the prototype/how it's documented, but keep old code working?

mossberg: > break the API in what way?

no, i was thinking of _actually_ breaking it. remove how the `Manticore.linux()` interface is currently superimposed over the `Manticore` ctor. break `Manticore('binary', args, envp)` and force use of platform classmethods (unless you already have a State) (`Manticore.linux(binary, args, envp)`)

PR #505, opened 2017-09-14
Refactor platform specific code out of ManticoreOutput

mossberg:
Refactor of ManticoreOutput.

These changes can be tested by running manticore on a regular binary, and then on the examples/evm/coverage.py script and looking at the produces workspaces.

Includes some irrelevant changes 
- in tests/ to get nosetests to not import error
- in examples/evm: fixes coverage.py as a simple way to test that these changes work


examples/evm/coverage.py
```diff
@@ -11,14 +11,14 @@ def transaction(self, transaction):
...
     def __init__(self):
-        super(ManticoreEVM, self).__init__()
         #Make the constraint store
         constraints = ConstraintSet()
         #make the ethereum world state
         world = evm.EVMWorld(constraints)
 
         #Let's start with the symbols
         self.temp_initial_state = State(constraints, world)
+        super(ManticoreEVM, self).__init__(self.temp_initial_state)
```
yan: Move `super()` to last line of the ctor? ctor can either establish state for subclass, or depend on it, keeping it in the middle can potentially be confusing.

PR #506, opened 2017-09-14
Manticore plugins

feliam:
Add a simple plugin registration mechanism
Move platform or feature specific functionality to a plugin.


manticore/core/plugin.py
```diff
@@ -0,0 +1,158 @@
+import logging
+from ..utils.helpers import issymbolic
+logger = logging.getLogger('MANTICORE')
+
+class Plugin(object):
+    event_names = [
+        'did_enqueue_state','did_execute_instruction',
```
yan: We should eventually mark all the events a class can generate as a class constant and aggregate them here, otherwise we're duplicating these.

manticore/manticore.py
```diff
@@ -212,8 +201,52 @@ def __init__(self, path_or_state, argv=None, workspace_url=None, policy='random'
...
+            callback_name = event_name+'_callback'
+            if hasattr(plugin, callback_name):
+                self.subscribe(event_name, getattr(plugin, callback_name))
+            
+        for callback_name in dir(plugin):
+            if callback_name.endswith('_callback'):
+                event_name = callback_name[:-9]
+                if event_name not in Plugin.event_names:
+                    logger.info("There is no event name %s for callback on plugin type %s", event_name, type(plugin) ) 
+                    print "There is no event name %s for callback on plugin type %s" %( event_name, type(plugin)  )
```
yan: Left-in `print`

manticore/manticore.py
```diff
@@ -212,8 +201,52 @@ def __init__(self, path_or_state, argv=None, workspace_url=None, policy='random'
...
+            
+
+
+    def unregister_plugin(self, plugin):
```
yan: I think the adjective is `unregistered` but the verb is `deregister`.

manticore/core/plugin.py
```diff
@@ -0,0 +1,158 @@
...
+        'did_write_register','did_fork_state','will_decode_instruction',
+        'will_execute_instruction','will_finish_run','will_fork_state','will_generate_testcase',
+        'will_load_state','will_read_memory','will_read_register','will_start_run',
+        'will_store_state','will_terminate_state','will_write_memory','will_write_register',]
+
+    def __init__(self): 
+        self.manticore = None
+
+class Tracer(Plugin):
+    def will_start_run_callback(self, state):
```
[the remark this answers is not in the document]
feliam: on_will_start_run()
 will_start_run()
 will_start_run_callback()
I'd go with any.
"foo" is a little less safe to detect as a callback in the plugin registering thing
And I'm not sure how it reads "on_will_blah". But I know nothing.

manticore/manticore.py
```diff
@@ -226,22 +227,28 @@ def register_plugin(self, plugin):
...
-        
-        for event_name in Plugin.event_names:
-            callback_name = event_name+'_callback'
-            if hasattr(plugin, callback_name):
-                self.subscribe(event_name, getattr(plugin, callback_name))
-            
+
+        #XXX(yan): keeping track of all_events is potentially too heavyweight just for warnings.
+        all_events = set()
+        for src in (Executor, State, Cpu, Manticore):
```
feliam: This means we cant make plugin that hooks on events emitted from the other subclasses of Eventful ?

manticore/core/plugin.py
feliam: We could explore the Eventful.__subclassess__() idea

PR #507, opened 2017-09-17
Add PCMPXSTRX Instructions

HighW4y2H3ll:


manticore/core/cpu/x86.py
```diff
@@ -4881,6 +4881,219 @@ def PCMPEQB(cpu, op0, op1):
...
+            oecx = 0
+            tres = res
+            while ((tres & 1) != 0):
+                oecx += 1
+                tres >>= 1
+            return oecx
+        else:
+            oecx = 0
+            tres = res
+            msbmask = (1 << (self.regfile._table['XMM0'].size/stepsize - 1))
```
feliam: This is already great but I'll leave this comment here anyway.
Is self.regfile._table['XMM0'].size going to vary? If not, why not replace it with the actual constant ?
Other idea. Add regfile.sizeof?(self, reg): self.regfile._table[reg].size

HighW4y2H3ll: I agree changing self.regfile._table['XMM0'].size to actual constant. It's not going to change anyway.

HighW4y2H3ll:
Ahh, thanks for pointing out that unittest directory :) I was struggling how to test these instructions. Let me check this out first.

HighW4y2H3ll:
[test.c.TXT](https://github.com/trailofbits/manticore/files/1327253/test.c.TXT)

ok, here's the testing file i'm using. Do you mind checking this out?
I'm not sure what's going on but I got these errors when running the unittest. I did not have ADD instructions there.

error

HighW4y2H3ll:
OK, Here's the code I used to do unittest.
[test_pcmpxstrx.c.txt](https://github.com/trailofbits/manticore/files/1327389/test_pcmpxstrx.c.txt)

tests/auto/make_tests.py
```diff
@@ -16,6 +16,8 @@
         cnt = op_count.get(test['mnemonic'],0)
         if cnt > 1000: #No more than 1000 instructions of each kind
```
feliam: Can you also momentarily tweak this counter to generate ~10? tests for each instruction and send them with the PR.
Add a file like tests/test_x86pcmpx.py  or another name you prefer.

HighW4y2H3ll: Sure, just pushed the full test cases, there's 64 test cases for each instruction. And I also had one more test case for memory type operand ("pcmpistri $0b00011011, %q1, %%xmm1\n").

PR #510, opened 2017-09-21
Require Z3 in instructions

mike-bailey:
Followed instructions including installed via pip only to find out it requires z3, which was clarified in the Dockerfile. Specifying in the README that z3 is necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Install and try Manticore in a few shell commands (see an [asciinema](https://as
 
 ```
 # Install system dependencies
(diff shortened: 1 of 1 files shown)


yan:
We are changing how z3 is installed in the upcoming 0.1.5 release. Check PR#501. Try installing Manticore from `master` by following [the development installation guide](https://github.com/trailofbits/manticore/wiki/Hacking-on-Manticore#developer-installation), it should install z3 automatically.

PR #511, opened 2017-09-22
Pythonic logger naming

### LINKED ISSUE #432: Use python logging idioms
Currently, Manticore's logging uses custom logging strings, e.g. `logger = logging.getLogger('MANTICORE')`

Per the python documentation [here](https://docs.python.org/2.7/library/logging.html#logger-objects), the recommended pattern is `logging.getLogger(__name__)`. This lets us organize our logger hierarchy just like our module hierarchy.

yan:
Fixes #432

### DIFF
--- a/manticore/utils/log.py
+++ b/manticore/utils/log.py
@@ -11,63 +11,86 @@ class ContextFilter(logging.Filter):
             record.stateid = '[%d]' % self.stateid
         else:
             record.stateid = ''
+
+        # Summarize logger name
+        components = record.name.split('.')
+        name = '.'.join(c[0] for c in components[:-1]) + '.' + components[-1]
+        record.name = name
+
         return True
 
-loggers = ['MANTICORE',
-           'VISITOR',
-           'EXECUTOR',
-           'CPU',
-           'REGISTERS',
-           'SMT',
-           'MEMORY',
-           'PLATFORM']
 manticore_verbosity = 0
+all_loggers = []
 
 def init_logging():
+    global all_loggers
+    all_loggers = logging.getLogger().manager.loggerDict.keys()
+        
     ctxfilter = ContextFilter()
     logfmt = ("%(asctime)s: [%(process)d]%(stateid)s %(name)s:%(levelname)s:"
               " %(message)s")
     logging.basicConfig(format=logfmt, stream=sys.stdout, level=logging.ERROR)
-    for l in loggers:
-        logging.getLogger(l).setLevel(logging.WARNING)
-        logging.getLogger(l).addFilter(ctxfilter)
-        logging.getLogger(l).setState = types.MethodType(loggerSetState,
-                                                         logging.getLogger(l))
+    for name in logging.getLogger().manager.loggerDict.keys():
+        logger = logging.getLogger(name)
+        if not name.startswith('manticore'):
+            next
+        logger.setLevel(logging.WARNING)
+        logger.addFilter(ctxfilter)
+        logger.setState = types.MethodType(loggerSetState, logger)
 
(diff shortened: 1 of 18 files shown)


manticore/utils/log.py
```diff
@@ -11,70 +11,95 @@ def filter(self, record):
...
         [
-            ('MANTICORE', logging.DEBUG),
-            ('SMT', logging.DEBUG)
+            ('manticore.manticore', logging.DEBUG),
+            ('manticore.core.smtlib', logging.DEBUG),
+            ('manticore.core.smtlib.*', logging.DEBUG)
          ]
     ]
 
+    def match(name, pattern):
```
japesinator: Do we need this custom match function? I think we can do everything above with just vanilla regex

japesinator: Alternatively, why not use https://docs.python.org/2.7/library/glob.html ?

manticore/utils/log.py
```diff
@@ -11,70 +11,95 @@ def filter(self, record):
...
+        logger.addFilter(ctxfilter)
+        logger.setState = types.MethodType(loggerSetState, logger)
 
 def loggerSetState(logger, stateid):
     logger.filters[0].stateid = stateid
 
 def set_verbosity(setting):
-    global manticore_verbosity
-    zero = map(lambda x: (x, logging.WARNING), loggers)
+    global manticore_verbosity, all_loggers
```
japesinator: How would you feel about `all_loggers` being a property of `Manticore` instead of a global? That seems handy if we'd ever like to e.g. change which loggers there are at times other than initialization (like a plugin might?).

manticore/utils/log.py
```diff
@@ -11,70 +11,95 @@ def filter(self, record):
             record.stateid = '[%d]' % self.stateid
         else:
             record.stateid = ''
+
+        # Summarize logger name
+        components = record.name.split('.')
+        name = '.'.join(c[0] for c in components[:-1]) + '.' + components[-1]
```
japesinator: I think it might be worth breaking this out into a slightly more verbose function like we did with `clamp`

manticore/utils/log.py
yan: using `logging.getLogger` is package-scoped, not Manticore scoped. Moving logging to Manticore will require changing how we log in general. Not saying it's a bad idea, just far outside the scope of this PR.

manticore/utils/log.py
yan: `glob`is solely for files; I looked into using [fnmatch](https://docs.python.org/2/library/fnmatch.html) to do the actual matching, but that's still file-centric. I wanted a simple matching function that matches on a single wildcard. I felt something small and specific was cleaner than a regex, but I can change it to a regex if you disagree.

manticore/utils/log.py
yan: 👍

manticore/utils/log.py
```diff
@@ -6,75 +6,105 @@ class ContextFilter(logging.Filter):
...
     logging.basicConfig(format=logfmt, stream=sys.stdout, level=logging.ERROR)
-    for l in loggers:
-        logging.getLogger(l).setLevel(logging.WARNING)
-        logging.getLogger(l).addFilter(ctxfilter)
-        logging.getLogger(l).setState = types.MethodType(loggerSetState,
-                                                         logging.getLogger(l))
+    for name in logging.getLogger().manager.loggerDict.keys():
+        logger = logging.getLogger(name)
+        if not name.startswith('manticore'):
+            next
```
mossberg: is this a bug? should this be `continue`?

PR #513, opened 2017-09-28
Trace following mode

yan:


manticore/core/plugin.py
```diff
@@ -15,6 +36,110 @@ def will_start_run_callback(self, state):
...
+    def did_execute_instruction_callback(self, state, last_pc, pc, insn):
+        if not self.active:
+            return
+        event = self.get_next('regs')
+        self.last_instruction = event['values']
+        if issymbolic(pc):
+            state.constrain(state.cpu.RIP == self.last_instruction['RIP'])
+        else:
+            for start, stop in self.symbolic_ranges:
+                if start <= pc <= stop:
```
mossberg: I don't think this is worth fixing now, but wanted to point out that this could be buggy. what if the stop symex pc is lower than the start symex pc? this could happen if the function you want to stop symexing in is simple very low in memry.

yan: Good point, noted.

manticore/manticore.py
```diff
@@ -615,9 +616,21 @@ def create_stats(self):
 
     def _start_run(self):
         assert not self.running
+
+        path_or_state, argv, kwargs = self._run_args
```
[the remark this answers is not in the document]
yan: I needed to register plugins first so that they can add hooks, before there was any state or elf loaded.

PR #521, opened 2017-10-13
EVM support

### LINKED ISSUE #416: Initial EVM support

feliam:
Fixes #416

* Supports most EVM instructions concrete and symbolic
* Supports inter contract calls and recursive calls
* Supports arbitrary symbolic multi-contract attacks/analysis
* Fair symbolic SHA3 concretization mechanism

Lacks:

* Gas support
* Good reporting


examples/evm/seth.py
```diff
@@ -0,0 +1,449 @@
...
+from manticore.core.smtlib import ConstraintSet, Operators, solver, issymbolic, Array, Expression, Constant
+from manticore.core.smtlib.visitors import arithmetic_simplifier
+from manticore.platforms import evm
+from manticore.platforms.evm import pack_msb
+from manticore.core.state import State
+import tempfile
+from subprocess import Popen, PIPE
+import sha3
+
+solc = "/usr/bin/solc"
```
yan: Should this be looked up in $PATH?

examples/evm/seth.py
```diff
@@ -0,0 +1,449 @@
...
+    def serialize(value):
+        if isinstance(value, str):
+            return ManticoreEVM.serialize_string(value)
+        if isinstance(value, list):
+            return ManticoreEVM.serialize_array(value)
+        if isinstance(value, (int, long)):
+            return ManticoreEVM.serialize_uint(value)
+        if isinstance(value, ManticoreEVM.SByte):
+            return ManticoreEVM.serialize_uint(value.size) + (None,)*value.size + (('\x00',)*(32-(value.size%32)))
+        if isinstance(value, type(None)):
```
yan: should this just be `if value is None`?

examples/evm/seth.py
```diff
@@ -0,0 +1,449 @@
...
+
+    @staticmethod
+    def serialize_uint(value, size=32):
+        '''takes an int and packs it into a 32 byte string, msb first''' 
+        assert size >=1
+        bytes = []
+        for position in range(size):
+            bytes.append( Operators.EXTRACT(value, position*8, 8) )
+        chars = map(Operators.CHR, bytes)
+        return tuple(reversed(chars))
```
yan: maybe?:

```
chars = []
for position in range(size-1,-1,-1):
    chars.append(Operators.CHR(Operators.EXTRACT(value, position*8,8)))
return tuple(chars)
```

examples/evm/seth.py
```diff
@@ -0,0 +1,449 @@
...
+        with self.locked_context('seth') as context:
+            context['_pending_transaction'] = ('CALL', caller, address, value, data)
+        return self.run()
+
+    def run(self, **kwargs):
+        #Check if there is a pending transaction
+        with self.locked_context('seth') as context:
+            assert context['_pending_transaction'] is not None
+            #there is at least one states in seth saved states
+            assert len(context['_saved_states']) + int(self.initial_state is not None) > 0
```
yan: `assert context['_saved_states'] or self.initial_state` ?

examples/evm/seth.py
```diff
@@ -0,0 +1,449 @@
...
+            assert len(self._executor.list()) == 0
+
+            for state_id in context['_saved_states']:
+                self._executor.put(state_id)
+    
+            context['_saved_states'] = []
+
+        #A callback will use _pending_transaction and 
+        #issue the transaction in each state
+        result = super(ManticoreEVM, self).run(procs = 6, **kwargs)
```
yan: Pass along `procs` from `kwargs`? or does 6 have significance?

examples/evm/seth.py
```diff
@@ -0,0 +1,449 @@
...
+            world.transaction(address=address, caller=caller, data=data, value=value)
+        else:
+            assert ty == 'CREATE_CONTRACT'
+            world.create_contract(caller=caller, address=address, balance=value, init=data)
+
+    def will_execute_instruction_callback(self, state, instruction):
+        assert state.constraints == state.platform.constraints
+        assert state.platform.constraints == state.platform.current.constraints
+
+        #print state.platform.current
```
yan: Remove comments?

manticore/core/smtlib/expression.py
```diff
@@ -603,11 +607,12 @@ def index(self):
     def byte(self):
         return self.operands[2]
 
-class ArrayProxy(ArrayVariable):
+class ArrayProxy(Array):
```
yan: Why the change and not just keep `ArrayVariable` and the name it provides?

manticore/core/smtlib/expression.py
```diff
@@ -622,16 +627,15 @@ def select(self, index):
...
-        self._array.select = None
         self._array = auxiliar
         return auxiliar
 
     def _get_size(self, index):
         size = index.stop - index.start
         if isinstance(size, BitVec):
             import visitors
-            size = visitors.ArithmeticSimplifier(expression=size)
+            from manticore.core.smtlib.visitors import pretty_print, arithmetic_simplifier
```
yan: `pretty_print` is unused

manticore/core/state.py
```diff
@@ -35,7 +36,7 @@ class Concretize(StateException):
         #Fixme Doc.
 
     '''
-    _ValidPolicies = ['MINMAX', 'ALL', 'SAMPLED', 'ONE']
+    _ValidPolicies = ['ALL32', 'MINMAX', 'ALL', 'SAMPLED', 'ONE']
```
yan: We should document what these policies are, or at least the new one.

manticore/manticore.py
```diff
@@ -203,9 +214,9 @@ def __init__(self, path_or_state, argv=None, workspace_url=None, policy='random'
             self._initial_state = make_initial_state(path_or_state, argv=argv, **kwargs)
         elif isinstance(path_or_state, State):
             self._initial_state = path_or_state
-        else:
-            raise TypeError("Manticore must be intialized with either a State or a path to a binary")
 
+        if not isinstance(self._initial_state, State):
```
yan: Move to top of func and update to: `if not isinstance(path_or_state, (State, str))`? self._initial_state isn't technically an argument

manticore/manticore.py
```diff
@@ -242,7 +253,7 @@ def register_plugin(self, plugin):
             if callback is not None:
                 self.subscribe(event_name, callback)
 
-        if logger.isEnabledFor(logging.DEBUG):
+        if logger.isEnabledFor(logging.DEBUG) or True:
```
yan: Left in `or True`

manticore/core/smtlib/expression.py
feliam: With current implementation after you assign some key-value to this ArrayProxy it stops being an ArrayVariable. And it has no name assigned to it.


X = ArrayVariable()
Y = Store(X, an_address, a_value)  <- this is an Array without a name assigned to it. 

As an alternative implementation we can still return a *.name* for an ArrayProxy by creating a new variable when needed. Though it needs access to the constrain store from the ArrayProxy.
X = ArrayVariable()
Y = Store(X, an_address, a_value)
Z = ArrayVariable()
assert( Z == Y )

manticore/manticore.py
feliam: But I wanted to be sure the object created by make_initial_state is a State also.

yan: under what situations can `make_initial_state` return a non-`State` object?

feliam: Hopefully none. That would be a bug. Are we sure there is no bug in current or future implementations of make_initial_state and its constituents parts ?

This one check guards for both a parameter or a make_initial_state bug.

manticore/platforms/evm.py
```diff
@@ -0,0 +1,1877 @@
...
+from ..utils.helpers import issymbolic
+from ..utils.event import Eventful
+from ..core.smtlib.visitors import pretty_print, arithmetic_simplifier, translate_to_smtlib
+from ..core.state import Concretize,TerminateState
+import logging
+import sys, hashlib
+if sys.version_info < (3, 6):
+    import sha3
+
+logger = logging.getLogger('PLATFORM')
```
yan: s/`'PLATFORM'`/`__name__`/?

PR #525, opened 2017-10-18
Simple event signature fixes

feliam:


manticore/platforms/evm.py
```diff
@@ -1387,7 +1388,8 @@ def p1 (x):
             r =  ' '*75 + hd[i]
             result.append(r)
 
-        result.append( '-'*147)
+        result =  ['-'*147] + result
```
yan: just initialize `result` with `['-'*147]` on line 1361?

Issue #527, opened 2017-10-20
evm: swarm hash in bytecode misinterpreted as INVALID opcode


alxiong:
Essence of the issue emerges from **post 0.4.7** solidity compiler appends swarm hashes at the end of the deployed compiled bytecode. 

[See official 0.4.7 major release summary:]( https://github.com/ethereum/solidity/releases/tag/v0.4.7)

Namely, normally actual bytecode ends with `5600` or equivalently `Jump`, `Stop` or equivalently "Alter the program counter, then halts execution"
After this, a following pattern should be observed _(post 0.4.7 of course)_:
`a165627a7a72305820` +<32bytes swarm hash> + '0029' 

[See solc bytecode metadata documentation](http://solidity.readthedocs.io/en/develop/metadata.html?highlight=swarm#encoding-of-the-metadata-hash-in-the-bytecode) 

Currently, in evm examples, all `coverage.log` recognize the metadata as INVALID opcode, or misinterpret them if case of collisions (i.e. "a1" used here doesn't translate to "Log 1 item"). To discuss in a more specific case, running
`python simple_mapping.py`
terminal output ends with:
```shell
  0x01b3 <56> JUMP                                                                       Alter the program counter.
   0x01b4 <00> STOP                                                                       Halts execution.
   0x01b5 <a1> LOG1                                                                       Append log record with one topic.
   0x01b6 <65627a7a723058> PUSH6 0x627a7a723058                                           Place 6-byte item on stack.
   0x01bd <20> SHA3                                                                       Compute Keccak-256 hash.
   0x01be <cd> INVALID                                                                    Unknown opcode
   0x01bf <7423ee0b5f90785ff9114882a877fcb42f9b90a142c9> PUSH21 0x23ee0b5f90785ff9114882a877fcb42f9b90a142c9Place 21-byte item on stack.
   0x01d5 <28> INVALID                                                                    Unknown opcode
   0x01d6 <21> INVALID                                                                    Unknown opcode
   0x01d7 <d7> INVALID                                                                    Unknown opcode
   0x01d8 <0b> SIGNEXTEND                                                                 Extend length of two's complement signed integer.
   0x01d9 <dc> INVALID                                                                    Unknown opcode
   0x01da <3f> INVALID                                                                    Unknown opcode
   0x01db <a9> INVALID                                                                    Unknown opcode
   0x01dc <10> LT                                                                         Less-than comparision.
   0x01dd <a4> LOG4                                                                       Append log record with four topics.
[...]

PR #529, opened 2017-10-20
Fix coverage calculation vs. metadata

feliam:
Fixes first part of https://github.com/trailofbits/manticore/issues/527
Swarm hash in bytecode was misinterpreted as INVALID opcode.


examples/evm/seth.py
```diff
@@ -44,7 +44,7 @@ def compile_code(source_code):
         temp.flush()
         p = Popen([solc, '--bin', temp.name], stdout=PIPE)
         outp = p.stdout.readlines()
-        return parse_bin(outp)
+        return parse_bin(outp)[:]
```
yan: This syntax feels weird. Maybe wrap the return value of `parse_bin()` in `list()` or at the call site?

PR #530, opened 2017-10-20
Various bug fixes.

### LINKED ISSUE #147: Use consistent naming in API
`State.symbolicate_buffer` uses the `label` parameter to tag symbolic values, where as `State.new_symbolic_buffer` uses `name`.

### LINKED ISSUE #272: Unhandled sys_write exception
### OS / Environment
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.2 LTS
Release:        16.04
Codename:       xenial

### Python version
Python 2.7.12

### Summary of the problem
The Linux platform incorrectly tries to use the `is_full()` method in the `sys_write` implementation, which is only defined on sockets.

```
2017-05-24 13:42:52,455: [83565][0] EXECUTOR:ERROR: Exception: 'File' object has no attribute 'is_full'
  File "/mnt/hgfs/projects/manticore/manticore/core/executor.py", line 672, in run
  File "/mnt/hgfs/projects/manticore/manticore/core/state.py", line 88, in execute
    result = self.platform.execute()
  File "/mnt/hgfs/projects/manticore/manticore/platforms/linux.py", line 1628, in execute
    syscallret = self.syscall()
  File "/mnt/hgfs/projects/manticore/manticore/platforms/linux.py", line 1486, in syscall
  File "/mnt/hgfs/projects/manticore/manticore/core/cpu/abstractcpu.py", line 263, in invoke
    result = model(*argument_tuple)
  File "/mnt/hgfs/projects/manticore/manticore/platforms/linux.py", line 1973, in sys_write
    return super(SLinux, self).sys_write(fd, buf, count)
  File "/mnt/hgfs/projects/manticore/manticore/platforms/linux.py", line 1055, in sys_write
    if fd > 2 and self.files[fd].is_full():
AttributeError: 'File' object has no attribute 'is_full'
```

### LINKED ISSUE #503: on syscall not implemented, print out actual name of syscall
This would be convenient for users experiencing syscall not implemented errors.

The modification would be in linux.py, `Linux.syscall()`. The syscall idx -> string name data is located in linux_syscalls.py

### LINKED ISSUE #509: State new_symbolic_buffer vs new_symbolic_value
state.py: new_symbolic_buffer vs new_symbolic_value.
One uses the keyword argument `name` and the other `label` for the same thing

### LINKED ISSUE #515: Syscalls do not emit did_execute_instruction events.
Since syscalls are implemented with exceptions, they do not emit a `did_execute_instruction` event.

yan:
This should fix #503, fix #515, fix #272, fix #147, and fix #509.

### DIFF
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -756,10 +756,11 @@ class Cpu(Eventful):
             for l in self.render_registers():
                 register_logger.debug(l)
 
-        implementation(*insn.operands)
-        self._icount += 1
-
-        self._publish('did_execute_instruction', self._last_pc, self.PC, insn)
+        try:
+            implementation(*insn.operands)
+            self._icount += 1
+        finally:
+            self._publish('did_execute_instruction', self._last_pc, self.PC, insn)
 
     def emulate(self, insn):
         '''
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -756,11 +756,21 @@ class Cpu(Eventful):
             for l in self.render_registers():
                 register_logger.debug(l)
 
+        #FIXME(yan): In the case the instruction implementation invokes a system call, we would not be able to
+        # publish the did_execute_instruction event from here, so we capture and attach it to the syscall
+        # exception for the platform to emit it for us once the syscall has successfully been executed.
+        def did_exec():
+            self._icount += 1
+            self._publish('did_execute_instruction', self._last_pc, self.PC, insn)
+
         try:
             implementation(*insn.operands)
-            self._icount += 1
-        finally:
-            self._publish('did_execute_instruction', self._last_pc, self.PC, insn)
+        except (Interruption, Syscall) as e:
+            e.on_handled = did_exec
+            raise e
+        else:
+            did_exec()
+
 
     def emulate(self, insn):
         '''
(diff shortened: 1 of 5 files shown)


manticore/core/cpu/abstractcpu.py
```diff
@@ -756,10 +756,11 @@ def fallback_to_emulate(*operands):
...
 
-        implementation(*insn.operands)
-        self._icount += 1
-
-        self._publish('did_execute_instruction', self._last_pc, self.PC, insn)
+        try:
+            implementation(*insn.operands)
+            self._icount += 1
+        finally:
+            self._publish('did_execute_instruction', self._last_pc, self.PC, insn)
```
feliam: But did it execute the instruction ?
This is a problem. If the instruction raises a Concretize(..) it will re-enter in all the child states. Then both will_execute_instruction and did_execute_instructions are going to be executed 3 times for a single instruction ?

manticore/platforms/linux.py
```diff
@@ -1296,14 +1311,12 @@ def sys_open(self, buf, flags, mode):
...
-                         filename, mode, f.fileno())
-        # FIXME(theo) generic exception
-        except Exception as e:
-            logger.info("Could not open file %s. Reason %s" % (filename, str(e)))
+
+            f = self._sys_open_get_file(filename, flags)
+            logger.debug("Opening file %s for real fd %d",
+                         filename, f.fileno())
+        except IOError as e:
+            logger.info("Could not open file %s. Reason: %s" % (filename, str(e)))
```
feliam: Just saw this..

Is better not to use the % explicitly there.
logger.info("Could not open file %s. Reason: %s", filename, str(e)) 
This will not format the string if not  log is not active.

PR #535, opened 2017-10-25
Remove win32 support

hugin:
Remove old and incomplete support for Windows minidumps and associated tests.


tests/test_memdumps.py
```diff
@@ -84,31 +84,10 @@ def _runManticore(self, dumpname):
 
         self.assertGreaterEqual(actual, expected)
 
-    @unittest.skip('Windows temporarily unmaintained')
```
yan: It looks like the rest of this file isn't used anymore with these tests deleted.

PR #538, opened 2017-10-27
Move inline function definitions out of the AbstractCPU's execute method

pgoodman:
This change moves from inline function definitions out of the critical path of AbstractCpu's execute method.

### DIFF
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -42,6 +42,12 @@ class InstructionNotImplementedError(CpuException):
     '''
     pass
 
+class InstructionEmulationError(CpuException):
+    '''
+    Exception raised when failing to emulate an instruction outside of Manticore.
+    '''
+    pass
+
 class DivideByZeroError(CpuException):
     ''' A division by zero '''
     pass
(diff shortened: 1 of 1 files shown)


manticore/core/cpu/abstractcpu.py
```diff
@@ -742,36 +742,36 @@ def execute(self):
...
+            try:
+                getattr(self, name)(*insn.operands)
+            except AttributeError:
+                text_bytes = ' '.join('%02x'%x for x in insn.bytes)
+                logger.info("Unimplemented instruction: 0x%016x:\t%s\t%s\t%s",
+                            insn.address, text_bytes, insn.mnemonic, insn.op_str)
+                self.emulate(insn)
         except (Interruption, Syscall) as e:
-            e.on_handled = did_exec
+            e.on_handled = lambda: self._publish_instruction_as_executed(insn)
```
feliam: We should consider handling syscalls with an event callback instead of using an exception. So it continues here after the handler execution?

pgoodman: That decision is beyond the scope of my change. My change is mostly superficial, but has a minor performance impact.

manticore/core/cpu/abstractcpu.py
```diff
@@ -781,13 +781,15 @@ def emulate(self, insn):
...
-
-
-        # We have been seeing occasional Unicorn issues with it not clearing
-        # the backing unicorn instance. Saw fewer issues with the following
-        # line present.
-        del emu
+        try:
+            emu.emulate(insn)
+        
+        #FIXME(pag): Catch `UcError`, e.g. executing a UD2 instruction.
```
yan: Catch the error and log to a `logger` before throwing away?

pgoodman: So I wasn't too sure how to get at the right class name, i.e. what imports to make, and also what the right way to propagate the info is. In my mind, `UcError` should be caught and converted into some kind of more manticore-specific exception.

PR #551, opened 2017-11-08
New API for EVM analysis

feliam:
A much more easy to use solidity biased API. 

* Contracts can be created directly from solidity sourcecode
`contract_account = seth.solidity_create_contract(source_code, owner=user_account)
`
* Better ABI support. Call specific contract function by name, python objects are serialized automatically
`#Potential overflow
contract_account.add(1, 100)
`
* Better reporting of function names and parameter
`TRANSACTION  3 - CALL
	 From: 0x1bfa530d5d685155e98cd7d9dd23f7b6a801cfef
	 To: 0x1bfa530d5d685155e98cd7d9dd23f7b6a801cfef
	 Value: 0 wei
	 Data: 1003e2d2000000000ffffffffffff88060900000e3b0bfffb8b004080000000000000000
	 Function:  add ( 0x000000000ffffffffffff88060900000e3b0bfffb8b004080000000000000000 )
`
* Link exceptions with solidity source_code
`REPORT: THROW  at:
     13    assert(sellerBalance >= value)
`
* Use symbolic values in transactions

.. and much much more..


examples/evm/seth.py
```diff
@@ -334,13 +402,74 @@ def compare_buffers(a, b):
...
+        e = state.context['last_exception']
         if ty is not None:
             if str(e) != ty:
                 return
         print "="*20
-        print "REPORT:", e, "\n"
+        print "REPORT:", e, 
+
+        try:
+            asm = list(evm.EVMDecoder.decode_all(runtime_bytecode[:-9-33-2]))
```
mossberg: what is this magic number?

examples/evm/simple_functions.py
```diff
@@ -0,0 +1,48 @@
...
+}
+'''
+#Initialize accounts
+user_account = seth.create_account(balance=1000)
+contract_account = seth.solidity_create_contract(source_code, owner=user_account)
+
+
+symbolic_data = seth.SByte(4) 
+symbolic_value = None 
+seth.transaction(  caller=user_account,
```
mossberg: can we replace this with the new python method call api? also what does this script do? can we add a comment at the top?

manticore/core/state.py
```diff
@@ -111,7 +111,7 @@ def __enter__(self):
         assert self._child is None
         new_state = State(self._constraints.__enter__(), self._platform)
         new_state._input_symbols = list(self._input_symbols)
-        new_state._context = copy.deepcopy(self._context)
+        new_state._context = copy.copy(self._context)
```
mossberg: what effect does this have?

examples/evm/reentrancy_concrete.py
```diff
@@ -87,85 +84,52 @@
...
-attacker_account = seth.create_account(balance=10)
-user_account = seth.create_account(balance=1000)
-contract_account = seth.create_contract(owner=user_account, 
-                                        init=contract_bytecode)
+#Initialize user and contracts
+user_account = seth.create_account(balance=100000000000000000)
+attacker_account = seth.create_account(balance=100000000000000000)
+
+contract_account = seth.solidity_create_contract(contract_source_code, owner=user_account) #Not payable
+seth.world[int(contract_account)]['balance']=1000000000000000000  #give it some ether
```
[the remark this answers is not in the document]
feliam: As it is now we should make the constructor `payable` for that. That `balance` is the value on the create transaction.

examples/evm/reentrancy_concrete.py
```diff
@@ -87,85 +84,52 @@
...
 
-print "\t Setting attack string"
 
+print "\t Setting attack string"
 #'\x9d\x15\xfd\x17'+pack_msb(32)+pack_msb(4)+'\x5f\xd8\xc7\x10',
-seth.transaction(  caller=attacker_account,
-                        address=exploit_account,
-                        data=seth.make_function_call("set_reentry_attack_string(bytes)", seth.make_function_id('withdrawBalance()')),
-                        value=0)
+reentry_string = seth.make_function_id('withdrawBalance()')
```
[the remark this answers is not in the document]
feliam: For the concrete version that's a yes, But then the auto exploit generation in the symbolic version will be even a little harder to understand.

examples/evm/reentrancy_concrete.py
```diff
@@ -87,85 +84,52 @@
...
+exploit_account.set_reentry_attack_string(reentry_string)
+
 
 print "[+] Initial world state"
-print " attacker_account %x balance: %d"% (attacker_account, seth.world.storage[attacker_account]['balance'])
-print " exploit_account %x balance: %d"%  (exploit_account, seth.world.storage[exploit_account]['balance'])
-print " user_account %x balance: %d"%  (user_account, seth.world.storage[user_account]['balance'])
-print " contract_account %x balance: %d"%  (contract_account, seth.world.storage[contract_account]['balance'])
+print " attacker_account %x balance: %d"% (attacker_account, seth.get_balance(attacker_account))
```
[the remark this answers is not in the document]
feliam: Nice idea.  What if the attacker_account has a function named like that?

examples/evm/reentrancy_symbolic.py
```diff
@@ -85,79 +82,57 @@
...
-                        address=exploit_account,
-                        data=seth.make_function_call("set_reentry_attack_string(bytes)", seth.SByte(32)),
-                        value=0)
+exploit_account.set_vulnerable_contract(contract_account)
+
+print "\t Setting 30 reply reps"
+exploit_account.set_reentry_reps(30)
+
+print "\t Setting reply string"
+exploit_account.set_reentry_attack_string(seth.SByte(38))
```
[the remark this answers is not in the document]
feliam: No idea. It should be big enough to contain all possible inputs. (I think 4 will do here :) )

examples/evm/seth.py
feliam: http://solidity.readthedocs.io/en/develop/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode

feliam: I wanted to show that we can handle generic transactions. That's an evm transaction.
With 4 bytes of symbolic data (input)
And symbolic value

examples/evm/simple_functions.py
feliam: I wanted to show that we can handle generic transactions. That's an evm transaction.
With 4 bytes of symbolic data (input)
And symbolic value

manticore/core/state.py
feliam: This is probably wrong. We want to deepcopy everything (list/dict/etc) except the Expressions saved in the context. Expressions should be copied just by reference.
So there are no 2 Variables with the same name.

examples/evm/seth.py
```diff
@@ -7,44 +7,55 @@
...
+        self._hashes = {}
+        self._caller = None
+        self._value = 0
+
+        name, source_code, init_bytecode, metadata, metadata_runtime, hashes = self._seth.context['seth']['metadata'][address]
+        for signature in hashes.keys():
+            func_name = str(signature.split('(')[0])
+            self._hashes[func_name] = signature, hashes[signature]
+
+    def __int__(self):
```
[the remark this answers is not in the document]
feliam: Interesting! `address` can not be a solidity function name.

The problem right now is that other accounts not created via solidity will be just an integer.
We may need to replace all the accounts and addresses with an EVMAccount.
...

Issue #550, opened 2017-11-08
Manticore EVM generates extra workspace folder


feliam:
cd examples/evm
python minimal.py
ls mcore* |wc

PR #557, opened 2017-11-10
[fix] Issue #550: generates extra workspace folder

cole-lightfighter:
This change fixes Issue #550, where the Manticore workspace dumps state onto the filesystem. The fix works by using the `workspace_url` knob in the `ManticoreEVM` constructor. 

Signed-off-by: Cole Lightfighter <cole@onicsla.bz>


yan:
@cole-lightfighter thank you for your pr! i modified it to change how we handle stores and workspaces to make the fix more permanent and work with other store types.

Issue #546, opened 2017-11-06
AttributeError: 'NoneType' object has no attribute 'runtime_bytecode'


ggrieco-tob:
### OS / Environment
ArchLinux

### Manticore version

baf2d769a624306f91fcca44bcff0b087f1bba69

### Python version

2.7

### Summary of the problem

TypeError: _hook_callback() when you try to hook every instruction executed in a smart contract.

### Step to reproduce the behavior

Execute this script from the examples/evm folder:

```
from seth import *

seth = ManticoreEVM()
seth.verbosity(3)
user_account = seth.create_account(balance=1000)

bytecode = '`'
#Initialize contract
contract_account = seth.create_contract(owner=user_account, 
                                          balance=0, 
                                          init=bytecode)

def explore(state):
    pass

seth.add_hook(None, explore)
seth.transaction(  caller=user_account,
                    address=contract_account,
                    value=None,
                    data='',
                 )
```

### Expected behavior

It shouldn't crash.

### Actual behavior

```2017-11-06 17:01:06,960: [25750] m.manticore:INFO: Starting 1 processes.
2017-11-06 17:01:06,963: [25750] m.c.executor:INFO: load state 0
2017-11-06 17:01:06,969: [25750] m.manticore:INFO: Results in /home/g/Work/Code/manticore/examples/evm/mcore_YLgrYD
2017-11-06 17:01:06,970: [25750] m.manticore:INFO: Total time: 0.00987696647644
2017-11-06 17:01:06,972: [25750] m.manticore:INFO: Starting 1 processes.
2017-11-06 17:01:06,975: [25750] m.c.executor:INFO: load state 2
2017-11-06 17:01:07,016: [25750] m.c.executor:ERROR: Exception: _hook_callback() takes exactly 4 arguments (3 given)
Traceback (most recent call last):
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.5-py2.7.egg/manticore/core/executor.py", line 456, in run
    if not current_state.execute():
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.5-py2.7.egg/manticore/core/state.py", line 126, in execute
    result = self._platform.execute()
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.5-py2.7.egg/manticore/platforms/evm.py", line 1544, in execute
    self.current.execute()
[...]

PR #558, opened 2017-11-11
Assert valid operand length

cole-lightfighter:
~~EVM bytecode comes in 32-byte chunks, and contracts require at least one
segment to be created~~.

An example, originally written by @ggrieco-tob in issue #546, is included.
Unsure what expected behavior should be, but initial state has a problem when a contract is
created with a single byte. 

Also added some exception handling in `parse_operand()` to catch invalid
instructions, e.g. a PUSH1 instruction followed by no bytes.


examples/evm/seth.py
```diff
@@ -190,6 +190,9 @@ def create_contract(self, owner, balance=0, init=None, address=None):
         with self.locked_context('seth') as context:
             assert context['_pending_transaction'] is None
         assert init is not None
+        # bytecode is only valid in 32-byte segments
```
feliam: Do you have any reference for this? I'm failing to find that.

cole-lightfighter: > Definition: len(a) is the number of bytes in a binary string a. The type of len(a) is assumed to be uint256.
from the [ABI](https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI) specification.

I took this to mean that properly encoded instructions would each take at least one uint256 in memory (32-bytes). After re-reading, I'm sure I misread. All the ABI encoded stuff I have seen has instructions padded out to 32 bytes, though.

After thinking about it, this still doesn't solve the underlying problem, since encoding invalid data may be a feature for Manticore (thinking fuzzing like @ggrieco-tob is doing). Thoughts?

Issue #543, opened 2017-10-27
KeyError accessing the global storage in SLOAD


ggrieco-tob:
### OS / Environment
ArchLinux

### Manticore version

baf2d769a624306f91fcca44bcff0b087f1bba69

### Python version

2.7

### Summary of the problem

KeyError accessing the global storage in SLOAD

### Step to reproduce the behavior

I'm using [this script](https://gist.github.com/ggrieco-tob/3f3491956197e8ba19f88694af803bab) to test manticore.

$ ./evm.py [QuickFuzz.80617981580.352.5119381167503483594.150.evm.txt](https://github.com/trailofbits/manticore/files/1423626/QuickFuzz.80617981580.352.5119381167503483594.150.evm.txt)

### Expected behavior

It shouldn't crash.

### Actual behavior

Traceback (most recent call last):
  File "./evm.py", line 27, in <module>
    evm.execute() 
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.5-py2.7.egg/manticore/platforms/evm.py", line 841, in execute
    result = implementation(*arguments)
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.5-py2.7.egg/manticore/platforms/evm.py", line 1210, in SLOAD
    return self.global_storage[self.address]['storage'][offset]
KeyError: 71748523475265

mossberg:
going to close for now, that script shows kind of an old way of using mcore evm

PR #562, opened 2017-11-14
Fix SLOAD invalid memory access

cole-lightfighter:
Check global_storage includes an entry at address before trying to load storage.

This fixes the SLOAD memory issue from #543, but it leads to another deeper bug:
```
assert isinstance(value,(int,long)) or isinstance(value, BitVec) and value.size==256
```
It seems to occur when the error condition in SLOAD uses a short-circuit return, i.e. doesn't return a value.

If SLOAD returns a WORD-length null byte-string ("0x00" * 32), then a stack underflow occurs:
```
evm.execute()
File manticore/platforms/evm.py, line 830, in execute
    arguments.append(self._pop())
File manticore/platforms/evm.py, line 792, in _pop
    raise StackUnderflow()
manticore.platforms.evm.StackUnderflow
```
Should the error in SLOAD raise an exception instead of returning a null byte-string?

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -249,8 +249,11 @@ class EVMInstruction(object):
     def parse_operand(self, buf):
         operand = 0
         for _ in range(self.operand_size):
-            operand <<= 8
-            operand |= ord(next(buf))
+            try:
+                operand <<= 8
+                operand |= ord(next(buf))
+            except:
+                raise TerminateState("Operand has insufficient bytes")
         self._operand = operand
 
     @property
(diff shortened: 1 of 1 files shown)


manticore/platforms/evm.py
```diff
@@ -249,8 +249,11 @@ def __init__(self, opcode, name, operand_size, pops, pushes, fee, description, o
     def parse_operand(self, buf):
         operand = 0
         for _ in range(self.operand_size):
-            operand <<= 8
-            operand |= ord(next(buf))
+            try:
+                operand <<= 8
+                operand |= ord(next(buf))
+            except:
+                raise TerminateState("Operand has insufficient bytes")
```
feliam: We are currently using the disassembler in other projects that wont't have TerminateState. So we should handle this iterator exception upstream and leave StopIteration here.

manticore/platforms/evm.py
```diff
@@ -1207,6 +1210,8 @@ def MSTORE8(self, address, value):
 
     def SLOAD(self, offset):
         '''Load word from storage'''
+        if not self.address in self.global_storage:
+            return
```
feliam: I think this should return zero, and it may consume gas.

PR #563, opened 2017-11-14
EVM assembler/disassembler doc and cleanup

feliam:


manticore/platforms/evm.py
```diff
@@ -261,101 +265,286 @@ def write(self, address, value):
...
-
         
-class EVMDecoder(object):
+class EVMAssembler(object):
     ''' 
         EVM Instruction factory
+        
+        Example use::
+
+            >>> from manticore.platforms.evm import *
```
mossberg: similar comment to above, can we avoid `import *`?

manticore/platforms/evm.py
```diff
@@ -1,14 +1,9 @@
-'''
-Solidity / Smart contract VM
-Implements the yellow paper: http://gavwood.com/paper.pdf
-Get example contracts from here:
-https://ethereum.github.io/browser-solidity/#version=soljson-latest.js
-'''
+''' Symbolic EVM implementation based the yellow paper: http://gavwood.com/paper.pdf '''
```
mossberg: "based on the"

manticore/platforms/evm.py
```diff
@@ -261,101 +265,286 @@ def write(self, address, value):
...
+            return self._name
+
+        @property
+        def description(self):
+            ''' Coloquial description of the instruction '''
+            return self._description
+
+        @property
+        def bytes(self):
+            ''' Encoded insttruction '''
```
mossberg: "instruction"

manticore/platforms/evm.py
```diff
@@ -261,101 +265,286 @@ def write(self, address, value):
...
+            return self._pushes
+
+        @property
+        def size(self):
+            ''' Size of the encoded instruction '''
+            return self._operand_size + 1
+
+        @property
+        def fee(self):
+            ''' Tha basic gass fee of the instruction '''
```
mossberg: "gass"

manticore/platforms/evm.py
```diff
@@ -261,101 +265,286 @@ def write(self, address, value):
...
+
+            Example use::
+
+                instruction = EVMAssembler.assemble_one('PUSH1 0x10')
+                print 'Instruction: %s'% instruction
+                print '\tdescription:', instruction.description
+                print '\tgroup:', instruction.group
+                print '\taddress:', instruction.offset
+                print '\tsize:', instruction.size
+                print '\thas_operand:', instruction.has_operand
```
mossberg: "thas"

manticore/platforms/evm.py
mossberg: also "tha"

manticore/platforms/evm.py
```diff
@@ -261,101 +265,286 @@ def write(self, address, value):
...
+                raise Exception("Not enough data for decoding")
+
+        @property
+        def operand_size(self):
+            ''' The immediate operand size '''
+            return self._operand_size
+
+        @property
+        def has_operand(self):
+            ''' Flag that indicates if the instruction uses an immediate operand'''
```
mossberg: use similar phrasing to all the other flags? "True if ..."

manticore/platforms/evm.py
```diff
@@ -261,101 +265,286 @@ def write(self, address, value):
...
+            return self.operand_size > 0
+
+        @property
+        def operand(self):
```
mossberg: we have a lot of properties here that simple return an underscore attribute. is this necessary? can we simply use public class attributes? we can document them with Sphinx `:ivar:`

manticore/platforms/evm.py
```diff
@@ -500,35 +689,244 @@ class EVMDecoder(object):
...
+        ''' Assemble one EVM instruction from its textual representation. 
+            
+            :param assembler: assembler code for one instruction
+            :param offset: offset of the instruction in the bytecode (optional)
+
+            :return: An Instruction object
+
+            Example use::
+
+                >>> print evm.EVMAssembler.encode_one('LT')
```
mossberg: wrong function name here

manticore/platforms/evm.py
```diff
@@ -500,35 +689,244 @@ class EVMDecoder(object):
...
+                                     )
+                ...
+                "0x6060604052600261010"
+        '''
+        return ''.join(map(lambda x:x.bytes, EVMAssembler.assemble_all(asmcode, offset=offset)))
+
+    @staticmethod
+    def disassemble_hex(bytecode, offset=0):
+        ''' Disassemble an EVM bytecode 
+            :param bytecode: canonical representation of an evm bytecode (hexadecimal)
```
mossberg: needs newline before this, or it's rendered weird

manticore/platforms/evm.py
```diff
@@ -261,101 +265,286 @@ def write(self, address, value):
...
+                return 'DUP%d'%self.pops
+            elif self._name == 'SWAP':
+                return 'SWAP%d'%(self.pops-1)
+            elif self._name == 'LOG':
+                return 'LOG%d'%(self.pops-2)
+            return self._name
+            
+        def parse_operand(self, buf):
+            ''' Parses an operand from buf 
+                :param buf: a buffer
```
mossberg: need newline here between text and param

mossberg:
Some general comments:

- watch out for the newline between the docstring text and metadata

image

- be careful that in example code blocks inside docstrings, the function names are right (right now there are some inconsistencies between encode/decode vs assemble/disassemble)

manticore/platforms/evm.py
```diff
@@ -500,35 +689,244 @@ class EVMDecoder(object):
...
+            offset += instr.size
+
+    @staticmethod
+    def disassemble_one(bytecode, offset=0):
+        ''' Decode a single instruction from a bytecode
+
+            :param bytecode: the bytecode stream 
+            :param offset: offset of the instruction in the bytecode(optional)
+            :type bytecode: iterator/sequence/str
+
```
mossberg: forgot the `:return:`!

manticore/platforms/evm.py
```diff
@@ -500,35 +689,244 @@ class EVMDecoder(object):
...
+                ...
+                "0x6060604052600261010"
+        '''
+        return ''.join(map(lambda x:x.bytes, EVMAssembler.assemble_all(asmcode, offset=offset)))
+
+    @staticmethod
+    def disassemble_hex(bytecode, offset=0):
+        ''' Disassemble an EVM bytecode 
+            :param bytecode: canonical representation of an evm bytecode (hexadecimal)
+            :param offset: offset of the first instruction in the bytecode(optional)
```
mossberg: also it's helpful to put the types here, you can do `:param int offset:`

manticore/platforms/evm.py
feliam: \t has

manticore/platforms/evm.py
```diff
@@ -500,35 +689,244 @@ class EVMDecoder(object):
...
 
     @staticmethod
-    def decode_one(bytecode):
+    def assemble_all(assembler, offset=0):
+        ''' Assemble a sequence of textual representation of EVM instructions 
+
+            :param assembler: assembler code for any number of instructions
+            :param offset: offset of the first instruction in the bytecode(optional)
+
+            :return: An generator of Instruction objects
```
mossberg: We can do some really fancy stuff and get links in the docs if we want

see: https://github.com/trailofbits/manticore/blob/master/manticore/manticore.py#L450

manticore/platforms/evm.py
feliam: The idea of a property returning _some_attribute is to make the some_attribute kind of read_only.
Can we dosument @properties with :ivar: ?

mossberg:
I forgot about the read-only thing, never mind then!

manticore/platforms/evm.py
mossberg: oops, sorry

docs/api.rst
```diff
@@ -36,3 +36,12 @@ Models
    .. function:: strlen
 
    .. function:: strcmp
+
+EVM
+---
+.. automodule:: manticore.platforms.evm
+.. autoclass:: manticore.platforms.evm::EVMAssembler.Instruction
```
mossberg: this is broken now because of the `EVMAsm` rename

PR #575, opened 2017-11-18
Updating README with solc binary requirements

### LINKED ISSUE #565: ethereum: document that solc is required
We need to edit the README to communicate that the `solc` binary is required as a separate dependency for certain ethereum API functionality (compiling sol source)

sidhant-gupta-004:
Fixes #565


mossberg:
Thanks for the contribution! I slightly modified the phrasing.

PR #588, opened 2017-11-22
Fixes for issue 587

GarretReece:
Fixes to the SUB, BLX, and MOV instruction implementations
Gettid syscall implementation


manticore/core/cpu/arm.py
```diff
@@ -815,8 +823,12 @@ def RSC(cpu, dest, src, add):
         return result, carry, overflow
 
     @instruction
-    def SUB(cpu, dest, src, add):
-        result, carry, overflow = cpu._ADD(src.read(), ~add.read(), 1)
+    def SUB(cpu, dest, src, add=None):
+        if add:
```
yan: Be explicit here, `if add is not None`, as `0` will cause this to fail.

PR #589, opened 2017-11-22
EVM api refactor

feliam:
This is a first cleanup and documentation pass towards a more usable manticore-EVM API
* Sphinx Documented
* moved seth to manticore
* moved static abi helpers to its own class
* added helpers to seth
* encapsulated storage/balance access in evm.py
* renamed {pop,push} to (pop,push}_vm and others


docs/api.rst
```diff
@@ -38,8 +38,14 @@ Models
    .. function:: strcmp
 
 EVM
----
+====
```
mossberg: EVM should still be under the API reference level, not on its own level

image

docs/api.rst
```diff
@@ -38,8 +38,14 @@ Models
...
 .. automodule:: manticore.platforms.evm
+API
```
mossberg: this should be named "Seth", not "API" for now, and updated when we decide on a non placeholder name

manticore/seth.py
```diff
@@ -1,97 +1,52 @@
...
+from .core.state import State
 import tempfile
 from subprocess import Popen, PIPE
 import sha3
 import json
+import StringIO
 
 
-class EVMContract(object):
+class ABI(object):
```
mossberg: this should be a module, not a class -let's fix in a later PR though

manticore/seth.py
```diff
@@ -147,20 +103,114 @@ def make_function_arguments(*args):
...
+            return object.__getattribute__(self, name)            
+
+
+
+class ManticoreEVM(Manticore):
+    ''' Manticore EVM manager
+    
+        Usage Ex::
+
+            from seth import ManticoreEVM, ABI
```
mossberg: this should from `from manticore.seth`

we need to be _very_ careful with these inline code blocks since they are not tested! it really sucks to encounter broken code snippets in documentation.

manticore/seth.py
```diff
@@ -302,6 +390,13 @@ def run(self, **kwargs):
         return result
           
     def save(self, state, final=False):
+        ''' Save a state in secundary storage and add it to running or final lists
```
mossberg: "secundary"

examples/evm/simple_transaction.py
```diff
@@ -1,4 +1,4 @@
-from seth import *
+from manticore.seth import *
```
mossberg: we should not encourage `import *`, can you please remove all instances of this?

manticore/platforms/evm.py
```diff
@@ -1626,6 +1626,7 @@ def SLOAD(self, offset):
 
     def SSTORE(self, offset, value):
         '''Save word to storage'''
+        #self.world.set_storage_data(self.address, offset, value)
```
mossberg: can rm this comment

manticore/platforms/evm.py
```diff
@@ -1902,14 +1903,72 @@ def storage(self):
...
+
+    def get_code(self, address):
+        return self.storage[address]['code']
+
+    def set_code(self, address, data):
+        self.storage[address]['code'] = data
+
+    def log(self, address, topic, data):
+        pass
+    '''
```
mossberg: we shouldn't keep this comment here like this. what should we do?

manticore/platforms/evm.py
```diff
@@ -1902,14 +1903,72 @@ def storage(self):
...
+        self.storage[dst]['balance'] += value
+        self.storage[src]['balance'] -= value
+
+    def get_code(self, address):
+        return self.storage[address]['code']
+
+    def set_code(self, address, data):
+        self.storage[address]['code'] = data
+
+    def log(self, address, topic, data):
```
mossberg: rm this func if it does nothing?

manticore/platforms/evm.py
```diff
@@ -2073,14 +2136,14 @@ def transaction(self, address, origin=None, price=0, data='', caller=None, value
             for i in range(len(data)):
                 data_symb[i] = Operators.ORD(data[i])
             data = data_symb
-        bytecode = self.storage[address]['code']
+        bytecode = self.get_code(address)
```
mossberg: not for this PR but change this to `self.account(address).code`?? same with the other `get_` functions?

manticore/seth.py
```diff
@@ -220,29 +269,60 @@ def running_state_ids(self):
...
+        ''' last returned buffer for state `state_id` '''
+        state = self.load(state_id)
+        return state.world.last_return
 
     def solidity_create_contract(self, source_code, owner, balance=0, address=None, args=()):
-        name, source_code, init_bytecode, metadata, metadata_runtime, hashes = self.compile(source_code)
-        address = self.create_contract(owner=owner, address=address, balance=balance, init=tuple(init_bytecode)+tuple(ManticoreEVM.make_function_arguments(*args)))
+        ''' Creates a solidity contract 
+
+            :param source_code: solidity source code
```
mossberg: add types here?

`:param type source_code:`

manticore/seth.py
```diff
@@ -314,8 +409,32 @@ def save(self, state, final=False):
                 context['_saved_states'].append(state_id)
         return state_id
 
+    def load(self, state_id=None):
+        ''' Load one of the running or final states.
+            
+            :param state_id: a state id or None. If None it assumes there is a single running state
```
mossberg: for super sphinx fu:

```
:param state_id: blah
:type state_id: int or None
```

manticore/platforms/evm.py
feliam: What would self.account(address) return? 
An EVMAccount()?   All the attributes of an EVMAccount may be used as a method name.
So, is .code is the code of the contract or the contract method named 'code' ?

manticore/seth.py
feliam: If we go for a multi file solution then yes.

PR #599, opened 2017-11-28
Log unimplemented instructions (issue 163)

### LINKED ISSUE #163: Better log invalid instructions from unicorn
Make it print the problematic instruction when it's not handled even at unicorn.

$ manticore some_app
2017-04-21 10:30:55,198: [28132] MAIN:INFO: Loading program: ['some_app']
2017-04-21 10:30:55,199: [28132] MAIN:INFO: Workspace: ./mcore_ViYAIf
2017-04-21 10:33:28,328: [28182][0] EXECUTOR:ERROR: Exception: Invalid instruction (UC_ERR_INSN_INVALID)
  File "/home/user/Projects/manticore/manticore/core/executor.py", line 669, in run
  File "/home/user/Projects/manticore/manticore/core/state.py", line 88, in execute
    result = self.model.execute()
  File "/home/user/Projects/manticore/manticore/models/linux.py", line 1702, in execute
    self.current.execute()
  File "/home/user/Projects/manticore/manticore/core/cpu/abstractcpu.py", line 444, in execute
    implementation(*instruction.operands)
  File "/home/user/Projects/manticore/manticore/core/cpu/abstractcpu.py", line 435, in fallback_to_emulate
    self.emulate(instruction)
  File "/home/user/Projects/manticore/manticore/core/cpu/abstractcpu.py", line 459, in emulate
    emu.emulate(instruction)
  File "/home/user/Projects/manticore/manticore/utils/emulate.py", line 184, in emulate
    self._step(instruction)
  File "/home/user/Projects/manticore/manticore/utils/emulate.py", line 230, in _step
    self._emu.emu_start(self._cpu.PC, self._cpu.PC+instruction.size, count=1)
  File "build/bdist.linux-x86_64/egg/unicorn/unicorn.py", line 288, in emu_start
    raise UcError(status)
UcError: Invalid instruction (UC_ERR_INSN_INVALID)

roachspray:
Possible fix for #163 -- create error log message and re-raise an exception. I would use .format(), but I wanted to stay consistent with what is being used elsewhere.

Fixes #163


manticore/core/cpu/abstractcpu.py
```diff
@@ -789,6 +790,11 @@ def emulate(self, insn):
         emu = UnicornEmulator(self)
         try:
             emu.emulate(insn)
+        except unicorn.UcError as e:
+            text_bytes = ' '.join('%02x'%x for x in insn.bytes)
```
mossberg: we should check here to make sure that it actually is an invalid insn error, instead of another unicorn error. we should be able to do this by checking that the `e.errno` value is `unicorn.UC_ERR_INSN_INVALID`!

PR #601, opened 2017-11-28
Refactor concolic implementation

yan:


examples/script/concolic.py
```diff
@@ -95,174 +97,149 @@ def eq(a, b):
...
-    
-
-def eqls(a, b):
-	# a b and 2 iterables of branch constraints
-    for aa, bb in zip(a, b):
-        if not eq(aa, bb):
-            return False
-    return True
 
+def perm(lst, func):
```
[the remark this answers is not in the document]
yan: yep. if you think of the index of the loop below as the elements of `lst` to flip, it skips the first bitmask (all zeroes, i.e. no flips).

examples/script/concolic.py
```diff
@@ -95,174 +97,149 @@ def eq(a, b):
...
         with m2.locked_context() as ctx:
-            ctx['sss'] = s
-            ctx['readdata'] = []
+            readdata = []
             for name, fd, data in s.platform.syscall_trace:
                 if name in ('_receive', '_read') and fd == 0:
-                    ctx['readdata'] += [data]
+                    readdata.append(data)
+            ctx['readdata'] = readdata
+            ctx['constraints'] = list(s.constraints.constraints)
```
mossberg: nice!

examples/script/concolic.py
```diff
@@ -272,31 +249,15 @@ def main():
...
-        # from the symex. sounds pretty hard
-        #
-        # but maybe a dumb way where we blindly permute the constraints
-        # and just check if they're sat before queueing will work
-
-        to_queue = get_new_constrs_for_queue(cons, newcons)
-
-        log('permuting constraints and adding {} constraints sets to queue'.format(len(to_queue)))
+        to_queue, new_datas = concrete_input_to_constraints(inp, cons)
+        if len(new_datas) > 0:
```
mossberg: oh cool, is this change meant to handle the case when a discovered path gets more symbolic input? if so, how was it tested?

yan: This was more of a refactoring from previous code to wrap `symbolic_run_get_cons`/`get_new_constrs_for_queue` in a function that was made sure to capture side effects. Testing, just made sure to run it on the binaries I was using with the original script with and made sure it found the same states and produced identical output.

examples/script/concolic.py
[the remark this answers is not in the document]
mossberg: super clever! can we add just a few words in the docstring to document the cleverness, since i get the feeling i will wut pretty hard at this code in 6 months

PR #602, opened 2017-11-28
Centralize system call logging

### LINKED ISSUE #440: Move individual syscall logging to dispatcher
We currently have syscalls log themselves, which is repetitive and can leave out some invocations. We should move the `logger.debug` line to the syscall dispatcher to standardize on format and be consistent.

yan:
We currently log syscall invocation from each syscall individually. This PR fixes that and adds a helper to expand strings.

Fixes #440.

### DIFF
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -1,6 +1,7 @@
 import inspect
 import logging
 import StringIO
+import string
 import sys
 import types
 
@@ -376,26 +377,36 @@ class SyscallAbi(Abi):
         return self._last_arguments
 
     def invoke(self, model, prefix_args=None):
-        max_arg_expansion = 32
         # invoke() will call get_argument_values()
         self._last_arguments = ()
 
         ret = super(SyscallAbi, self).invoke(model, prefix_args)
 
         if platform_logger.isEnabledFor(logging.DEBUG):
+            # Try to expand strings up to max_arg_expansion
+            max_arg_expansion = 32
+            # Add a hex representation to return if greater than min_hex_expansion
+            min_hex_expansion = 0x80
+
             args = []
             for arg in self._last_arguments:
                 arg_s = "0x{:x}".format(arg)
                 if self._cpu.memory.access_ok(arg, 'r'):
-                    s = self._cpu.read_string(arg, max_arg_expansion).translate(None, '\n')
-                    if len(s) == max_arg_expansion:
-                        s = s + '..'
-                    if len(s) > 4: # don't expand very short strings
-                        arg_s = arg_s + ' ({})'.format(s)
+                    s = self._cpu.read_string(arg, max_arg_expansion)
+                    if all(c in string.printable for c in s):
+                        if len(s) == max_arg_expansion:
+                            s = s + '..'
+                        if len(s) > 2:
+                            arg_s = arg_s + ' ({})'.format(s.translate(None, '\n'))
                 args.append(arg_s)
 
             args_s = ', '.join(args)
-            platform_logger.debug('%s(%s) -> %d', model.im_func.func_name, args_s, ret)
+
+            ret_s = '{}'.format(ret)
+            if ret > min_hex_expansion:
+                ret_s = ret_s + '({:x})'.format(ret)
+
+            platform_logger.debug('%s(%s) -> %s', model.im_func.func_name, args_s, ret_s)
 
 ############################################################################
 # Abstract cpu encapsulating common cpu methods used by platforms and executor.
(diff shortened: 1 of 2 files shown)


manticore/core/cpu/abstractcpu.py
```diff
@@ -591,9 +637,7 @@ def read_string(self, where, max_length=None):
         while True:
             c = self.read_int(where, 8)
 
-            assert not issymbolic(c)
-
-            if c == 0:
+            if c == 0 or issymbolic(c):
```
mossberg: we should check `issymbolic` first; otherwise symbolic needs to implement `__nonzero__`

```
2017-11-28 17:23:25,775: [37608] m.c.executor:ERROR: Exception: __nonzero__ for Bool
  File "/mnt/hgfs/code/manticore/manticore/core/executor.py", line 460, in run
  File "/mnt/hgfs/code/manticore/manticore/core/state.py", line 126, in execute
  File "/mnt/hgfs/code/manticore/manticore/platforms/linux.py", line 2010, in execute
    self.syscall()
  File "/mnt/hgfs/code/manticore/manticore/platforms/linux.py", line 1855, in syscall
  File "/mnt/hgfs/code/manticore/manticore/core/cpu/abstractcpu.py", line 390, in invoke
    s = self._cpu.read_string(arg, max_arg_expansion).translate(None, '\n')
  File "/mnt/hgfs/code/manticore/manticore/core/cpu/abstractcpu.py", line 640, in read_string
    if c == 0 or issymbolic(c):
  File "/mnt/hgfs/code/manticore/manticore/core/smtlib/expression.py", line 119, in __nonzero__
    raise NotImplementedError("__nonzero__ for Bool")
NotImplementedError: __nonzero__ for Bool

```

manticore/core/cpu/abstractcpu.py
```diff
@@ -351,6 +371,32 @@ def syscall_number(self):
...
+        return self._last_arguments
+
+    def invoke(self, model, prefix_args=None):
+        max_arg_expansion = 32
+        # invoke() will call get_argument_values()
+        self._last_arguments = ()
+
+        ret = super(SyscallAbi, self).invoke(model, prefix_args)
+
+        if platform_logger.isEnabledFor(logging.DEBUG):
```
mossberg: this code will probably produce slightly incorrect output sometimes
- if a pointer doesn't point to a string. maybe we can check whether the bytes lie in ASCII
- if an argument is actually a negative signed int, it will print it out as unsigned

i think the benefits are worth those, likely infrequent, misprints, so we don't necessarily have to do anything.

maybe we can print the return code as both decimal and hex? for cases like:

```
sys_brk(0x6f01c0) -> 7274944
sys_brk(0x6f1000) -> 7278592

```

manticore/platforms/linux.py
```diff
@@ -1753,30 +1716,23 @@ def sys_exit_group(self, error_code):
         self.sched()
         self.running.remove(procid)
         #self.procs[procid] = None
-        logger.debug("EXIT_GROUP PROC_%02d %s", procid, ctypes.c_int32(error_code).value)
         if len(self.running) == 0 :
```
mossberg: with this, sys_exit no longer prints out a log because that syscall implementation bails with an exception and doesn't return. this might be fine though, since we get a testcase generated msg immediately after 

```
2017-11-28 17:25:38,376: [37716] m.c.executor:INFO: Forking, about to store. (policy: ALL, values: 0x400a9a, 0x400ab6)
2017-11-28 17:25:38,425: [37716] m.c.executor:INFO: load state 3
2017-11-28 17:25:38,838: [37716] m.manticore:INFO: Generated testcase No. 0 - Program finished with exit status: 0
```

manticore/core/cpu/abstractcpu.py
feliam: To implement __nonzero__ from Expression you need to reference current _constraints_ whatever that is from the expression.
Also we probably do not want that as __nonzero__ normally should fork. 
We can though improve the __nonzero__ not implemented exception.   

((( Or the craziness catch the __nonzero__ exception and replace the value in the python call stack with one of the possible concretizations of the exp )))

manticore/core/cpu/abstractcpu.py
yan: So at one point, I considered producing `hexdump`-like output, but realized it might be adding complexity where it's not needed; I ended up adding a minimum length on strings in its place, with the assumption that it might catch some misprints.

Re: both decimal and hex output, one thing I considered is keeping it decimal if it's < 128, and hex otherwise. Or print it in decimal always, but add a parenthetical if it's big enough (`65280 (0xff00)`).

manticore/core/cpu/abstractcpu.py
yan: +1, think checking for `issymbolic` first is the right way to go here.

manticore/platforms/linux.py
yan: I think this is okay, `exit(2)` is defined to not return.

Issue #581, opened 2017-11-20
CI should run scripts in examples/


mossberg:
We should make sure the manticore scripts in examples/ work, or at least don't crash.

todo
- modify scripts/travis_test.sh to run all the manticore scripts in examples/ before proceeding with regular the regular tests

PR #604, opened 2017-11-28
Travis ci includes tests for examples\

chowdaryd:
issue #581


mossberg:
Hi @chowdaryd, thanks a lot for the PR but it doesn't quite fix #581; we should run all the manticore _scripts_ in examples/ , not manticore on all the binaries in examples/

chowdaryd:
Hi @mossberg, do you mean the sh script should run every script in example\ independently?

mossberg:
Yes, that’s the idea. We should not have broken manticore scripts in the examples/ directory, that sucks for anyone trying to learn. The examples should all be small scripts, so it shouldn’t take too long to run them all

PR #609, opened 2017-11-30
Fixed missing exception definition of Deadlock in decree platform

ggrieco-tob:
The actual deadlock is still there, but the exception is defined.

### DIFF
--- a/manticore/platforms/decree.py
+++ b/manticore/platforms/decree.py
@@ -10,6 +10,7 @@ from ..core.executor import TerminateState
 from ..utils.helpers import issymbolic
 from ..binary import CGCElf
 from ..platforms.platform import Platform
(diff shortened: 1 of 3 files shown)


manticore/platforms/decree.py
```diff
@@ -10,6 +10,7 @@
 from ..utils.helpers import issymbolic
 from ..binary import CGCElf
 from ..platforms.platform import Platform
+from ..platforms.platform import RestartSyscall, Deadlock, BadFd, SyscallNotImplemented
```
feliam: So the idea is that all(or some) the platform exceptions end up at platorm.py and then that's managed by carefully importing what we need in the different places? 
I know that some of them will not make sense for other platforms and non should be used upstream or in State or Executor.

PR #610, opened 2017-11-30
Updates for thumb mode

### LINKED ISSUE #605: Incorrect invalid memory access for nc on arm
### Manticore version
commit e50cc6b on master

### Summary of the problem
Emulation of an armv7 compilation of the nc utility results in an incorrect invalid memory access error

`manticore -- ./nc-android-ndk-armv7-debug -l -p 1234`

### Expected behavior
Emulation succeeds

### Actual behavior
Emulation fails with invalid memory access error

### Any relevant logs
including just the tail end here for brevity's sake
```
2017-11-29 07:56:48,024: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x0000000000012634:      adds    r2, r7, #0
2017-11-29 07:56:48,025: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x0000000000012636:      adds    r0, r4, #0
2017-11-29 07:56:48,026: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x0000000000012638:      blx     #0x1d6d0
2017-11-29 07:56:48,027: [21608] m.c.c.arm:DEBUG: swapping mode due to BLX at inst 0x1263c
2017-11-29 07:56:48,027: [21608] m.c.c.arm:DEBUG: swapping into ARM mode
2017-11-29 07:56:48,028: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000001d6d0:      push    {r4, r5, r6, lr}
2017-11-29 07:56:48,029: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000001d6d4:      subs    lr, r2, #0
2017-11-29 07:56:48,030: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000001d6d8:      sub     sp, sp, #8
2017-11-29 07:56:48,032: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000001d6dc:      mov     r4, r0
2017-11-29 07:56:48,033: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000001d6e0:      mov     r5, r1
2017-11-29 07:56:48,034: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000001d6e4:      mov     r6, r3
2017-11-29 07:56:48,035: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000001d6e8:      addne   ip, r0, #8
2017-11-29 07:56:48,036: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000001d6ec:      beq     #0x1d78c
2017-11-29 07:56:48,037: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000001d6f0:      ldm     lr!, {r0, r1, r2, r3}
2017-11-29 07:56:48,038: [21608] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000001d6f4:      stm     ip!, {r0, r1, r2, r3}
2017-11-29 07:56:48,135: [21608] m.manticore:INFO: Generated testcase No. 0 - Invalid memory access (mode:w) <a98c2642>
```
[nc-android-ndk-armv7-debug.tar.gz](https://github.com/trailofbits/manticore/files/1514432/nc-android-ndk-armv7-debug.tar.gz)

GarretReece:
Fixes #605 

Adds correct handling of PC relative addressing for thumb mode.
Adds correct handling of mode swaps in more cases, as well as more thumb specific instruction implementations. 

Adds test cases for instruction implementations.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/610)


manticore/core/cpu/arm.py
```diff
@@ -900,6 +926,12 @@ def CMP(cpu, reg, compare):
     def POP(cpu, *regs):
         for reg in regs:
             val = cpu.stack_pop(cpu.address_bit_size / 8)
+            if reg.reg in ('PC', 'R15'):
+                if val & 0x1:
```
yan: How will this behave if `val` is symbolic? It might make sense to switch mode on executing, instead of switching prior to jumping. Might make forked states that end up in different exec modes easier to implement.

mossberg:
status of this pr: still some work to do, but no one working on it presently

mossberg:
Reviewed 2 of 3 files at r4.
Review status: 2 of 3 files reviewed at latest revision, 8 unresolved discussions.

---

*[manticore/core/cpu/arm.py, line 14 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/610#-L5LFLx50Usv87BEvZcS:-L5LFLx50Usv87BEvZcT:b-sidx3f) ([raw file](https://github.com/trailofbits/manticore/blob/300e735a43f5f977b76d91f527dbbf3c14a1afb4/manticore/core/cpu/arm.py#L14)):*
> ```Python
> from .abstractcpu import (
>     Abi, SyscallAbi, Cpu, RegisterFile, Operand,
>     CpuException, Interruption
> ```

unused import?

---

*[manticore/core/cpu/arm.py, line 90 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/610#-L5LOv_8XiITJ2kvZf0j:-L5LOv_8XiITJ2kvZf0k:b-712poz) ([raw file](https://github.com/trailofbits/manticore/blob/300e735a43f5f977b76d91f527dbbf3c14a1afb4/manticore/core/cpu/arm.py#L90)):*
> ```Python
>         if self.type == 'register':
>             value = self.cpu.regfile.read(self.reg)
>             # We set PC to point to the next instruction after the current
> ```

I believe the code change is correct here, but the comment is not; for both ARM and Thumb, reading PC returns the address of the instruction _after_ the next instruction. The reason we only add 1 instruction width here is at this point, PC has already been incremented by an instruction width as part of the `@instruction` decorator, which executes before this point.

I think we should rephrase the comment to something like "We need PC to read here as current instruction + 8 (ARM) or current instruction + 4 (Thumb). At this point, it will already read as current instruction + 4 (ARM) or current instruction + 2 (Thumb), so we just need to add one extra instruction width"

relevant docs:



---

*[manticore/core/cpu/arm.py, line 170 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/610#-L5LTsRr56FuhoqmR1vD:-L5LTsRr56FuhoqmR1vE:b-d9bune) ([raw file](https://github.com/trailofbits/manticore/blob/300e735a43f5f977b76d91f527dbbf3c14a1afb4/manticore/core/cpu/arm.py#L170)):*
> ```Python
>         #
>         # Regardless of mode, our implementation of read(PC) will return the address
>         # of the instruction following the current instruction.
> ```

ditto above, not sure the comment is correct? should it be "following the next instruction"?

---

*[manticore/core/cpu/arm.py, line 179 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/610#-L5LTe8PMlLDpNsWpmG1:-L5LTe8PMlLDpNsWpmG2:bvmvlkt) ([raw file](https://github.com/trailofbits/manticore/blob/300e735a43f5f977b76d91f527dbbf3c14a1afb4/manticore/core/cpu/arm.py#L179)):*
> ```Python
>                 #we need (PC & 0xFFFFFFFC) + 4
>                 #thus:
>                 new_base = (base - len(self.cpu.instruction.bytes)) & 0xFFFFFFFC
> ```

could we use `cpu.instruction.size` instead of `len(self.cpu.instruction.bytes)`?

---

*[manticore/core/cpu/arm.py, line 430 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/610#-L5Lh6DFQh1O2-rJeCgv:-L5Lh6DFQh1O2-rJeCgw:bnl2htb) ([raw file](https://github.com/trailofbits/manticore/blob/300e735a43f5f977b76d91f527dbbf3c14a1afb4/manticore/core/cpu/arm.py#L430)):*
> ```Python
> 
>         elif _type in range(cs.arm.ARM_SFT_ASR_REG, cs.arm.ARM_SFT_RRX_REG + 1):
>             amount = Operators.EXTRACT(amount.read(), 0, 8)
> ```

it looks like this function is an exactly copy of `_Shift`, except for around this line. can we avoid copypasting it, and introduce a boolean flag? what exactly is the underlying problem?

---

*[manticore/core/cpu/arm.py, line 1154 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/610#-L5LeLA-_4KskCMNQc-j:-L5LeLA-_4KskCMNQc-k:b-sz3ezg) ([raw file](https://github.com/trailofbits/manticore/blob/300e735a43f5f977b76d91f527dbbf3c14a1afb4/manticore/core/cpu/arm.py#L1154)):*
> ```Python
>             amount = rest[0].read()
>             result, carry = cpu._Shift(op.read(), srtype, amount, carry)
>         elif cpu.mode == cs.CS_MODE_THUMB:
> ```

would it be possible to not modify this here, but in `op.read()`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/610)*

yan:
Reviewed 3 of 3 files at r4.

---

*[manticore/core/cpu/arm.py, line 1154 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/610#-L5LeLA-_4KskCMNQc-j:-L5PSzXZ2DzwN2RMoWMa:b-gpc351) ([raw file](https://github.com/trailofbits/manticore/blob/300e735a43f5f977b76d91f527dbbf3c14a1afb4/manticore/core/cpu/arm.py#L1154)):*

would it be possible to not modify this here, but in `op.read()`?
</blockquote></details>

I'm not sure, because the two variants read from different parameters (`op` vs `dest`)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/610)*

mossberg:
:lgtm:

---

Reviewed 2 of 2 files at r5, 1 of 1 files at r6.

---

*[manticore/core/cpu/arm.py, line 1154 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/610#-L5LeLA-_4KskCMNQc-j:-L5PwwnHWY-3xYWLvZBn:b32wtd) ([raw file](https://github.com/trailofbits/manticore/blob/300e735a43f5f977b76d91f527dbbf3c14a1afb4/manticore/core/cpu/arm.py#L1154)):*

I'm not sure, because the two variants read from different parameters (`op` vs `dest`)
</blockquote></details>

ok

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/610#-:-L5Q0d4ooKLOaYW3lqku:bnfp4nl)*

yan:
:lgtm:

---

Reviewed 2 of 2 files at r5, 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/610#-:-L5Q13DIpLB_lTBvqC3t:bnfp4nl)*

mossberg:
great work @GarretReece! and those tests ❤️

Issue #621, opened 2017-12-03
'DivideError' is used but not defined in manticore/core/cpu/x86.py


ggrieco-tob:
### OS / Environment

Distributor ID:	ArchLinux
Description:	ArchLinux
Release:	rolling-release

### Manticore version

Last revision (be0141e501d3f40b77b111f9451cb2be8a2d33d4)

### Python version

Python 2.7.13

### Summary of the problem

'DivideError' is used but not defined in [manticore/core/cpu/x86.py](https://github.com/trailofbits/manticore/blob/8002a1d682a14c857f687754b754dabdee1d12c8/manticore/core/cpu/x86.py). Maybe we should use 'DivideByZeroError' defined in [manticore/core/cpu/abstractcpu.py](https://github.com/trailofbits/manticore/blob/8002a1d682a14c857f687754b754dabdee1d12c8/manticore/core/cpu/abstractcpu.py) instead?

ggrieco-tob:
Fixed in 976f31446d5719895735db599db2e36d83cab61c.

PR #624, opened 2017-12-04
Fixed issue #621: DivideError is used but not defined in manticore/core/cpu/x86.py

ggrieco-tob:
I decided to use 'DivideError' instead of 'DivideByZeroError' because it is a more general error.

### DIFF
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -50,8 +50,8 @@ class InstructionEmulationError(CpuException):
     '''
     pass
 
-class DivideByZeroError(CpuException):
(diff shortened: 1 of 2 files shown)


yan:
`DivideError` is more general, but what other divide errors can happen by an instruction implementation? Maybe `ArithmeticError` is what we're after if we want a generic exception? Do we want a more generic div zero exception?

yan:
@ggrieco-tob Sorry that we sat on this for a while; I think `DivideByZeroError` is more correct than just `DivideError` since there should be no other division errors. Can you revert that change, but leave the other fixes in place? We'll merge right after.

PR #632, opened 2017-12-07
Introduce a 'force' parameter to memory access functions

### LINKED ISSUE #549: Provide an API for modifying memory that side-steps permissions
Currently, using `Cpu` and `Memory` methods for writing memory respect permissions set on memory mappings. This is correct for emulation, but can be a hinderance for API usage and custom analysis. Rewriting read-only memory from a hook is currently not possible without using internal methods, but can be useful to do.

This issue proposes adding either another set of methods to the public API to write memory, or an optional keyword parameter (such as `force` or `ignore_permissions`) to existing methods.

yan:
Introduces a `force` parameter to impacted Memory and Cpu methods. 

Fixes #549


manticore/core/memory.py
```diff
@@ -999,13 +994,14 @@ def read(self, address, size):
...
         '''
         Write a value at address.
         :param address: The address at which to write
         :type address: int or long or Expression
         :param value: Bytes to write
         :type value: str or list
+        :param force: Whether to ignore permissions
         '''
         size = len(value)
         if issymbolic(address):
```
feliam: assert force is False in the symbolic address path?

yan: The reason i hesitated to add force to that path is because this is used to find cases of writing to unmapped memory, if we honor force, we can miss those. One way to side-step that is to only consider unmapped memory as crashing conditions, not just write failure.

feliam: You are probably not using force=True on symbolic addresses. 
So I propose you just add the assert there.
assert force is False 
?

yan: Addressed in most recent commit

mossberg:
An alternative to the optional kwarg parameter `force=` interface: add boolean state to cpu specifying whether to enforce memory permissions. as suggested by @feliam

1. that state is controlled with some cpu methods.

```
cpu.disable/enable_memory_checks()
```

2. use a context manager. this could alleviate the issue of potentially forgetting to reenable memory checks with the above interface.

```
with cpu.disable_memory_permissions() as cpu:
    cpu.write()
```

mossberg:
correction: the boolean state would be on Memory, not on Cpu though (i.e. memory will have traditional enable/disable methods, cpu’s context manager would enable/disable them)

PR #636, opened 2017-12-11
evm: Make 'not enough funds' states finalized states

mossberg:
Currently, when we terminate a state because there were not enough funds in the sender's balance for the transaction,  we keep that state alive, causing it to produce new states when another transaction is executed. This results in redundant states. We can safely terminate a 'not enough funds' state because it has no effect on blockchain state.

### DIFF
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -97,8 +97,8 @@ def ethereum_cli(args):
         print "[+] There are %d reverted states now"% len(m.terminated_state_ids)
         print "[+] There are %d alive states now"% len(m.running_state_ids)
 
-    for state in m.all_states:
-        print str(state.context['last_exception'])
+        for state in m.all_states:
+            print str(state.context['last_exception'])
 
     # for state in seth.all_states:
     #     blockchain = state.platform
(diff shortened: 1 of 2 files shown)


manticore/seth.py
```diff
@@ -656,7 +656,9 @@ def _terminate_state_callback(self, state, state_id, e):
             our private list 
         '''
         state.context['last_exception'] = e
-        if e.message != 'REVERT':
+        # TODO(mark): This will break if we ever change the message text. Use a less
+        # brittle check.
+        if e.message not in {'REVERT', 'Not Enough Funds for transaction'}:
```
feliam: TXERROR ?

manticore/__main__.py
```diff
@@ -97,8 +97,8 @@ def ethereum_cli(args):
         print "[+] There are %d reverted states now"% len(m.terminated_state_ids)
         print "[+] There are %d alive states now"% len(m.running_state_ids)
 
-    for state in m.all_states:
-        print str(state.context['last_exception'])
+        for state in m.all_states:
```
feliam: Will not this re-print the states from previous iterations ?

mossberg: yes it will, i'll rm this from the pr, it is not related and only for debugging

PR #637, opened 2017-12-11
Detectors ()

feliam:
Special kind of Plugin that early detect/hint dangerous evm traces.
3 types implemented:

IntegerOverflow: Detects any it overflow on instructions ADD and SUB.
UnitializedMemory: detects the use of not initialized memory
UnitializedStorage: detects the use of not initialized storage


manticore/platforms/evm.py
```diff
@@ -1153,6 +1166,7 @@ def _store(self, address, value):
         #CHECK VALUE IS A 256 BIT INT OR BITVEC
         self._allocate(address)
         self.memory.write(address, [value])
+        self._publish('did_evm_write_memory', address, value)
```
mossberg: this is not consistent with `_published_events` above?

manticore/seth.py
```diff
@@ -7,7 +7,100 @@
...
+class Detector(Plugin):
+    @property
+    def name(self):
+        return self.__class__.__name__.split('.')[-1]
+
+    def get_findings(self, state):
+        return state.context.setdefault('seth.findings.%s'%self.name,set())
+
+    def add_finding(self, state, finding):
+        name = self.__class__.__name__.split('.')[-1]
```
mossberg: can use self.name

manticore/platforms/evm.py
feliam: Yes  'evm_write_memory' is declared. It does not complain about that

PR #638, opened 2017-12-11
Implement initial evm workspace, fix small bugs

mossberg:
- Fixes a bug in tuple size in sol compile helper
- generate .tx.# files for tx in each state
- save the compiled bytecode in workspace as a convenience, if solidity was compiled.

### DIFF
--- a/manticore/core/plugin.py
+++ b/manticore/core/plugin.py
@@ -32,7 +32,7 @@ class Tracer(Plugin):
     def will_start_run_callback(self, state):
         state.context['trace'] = []
 
-    def will_execute_instruction_callback(self, state, pc, instruction):
+    def did_execute_instruction_callback(self, state, pc, target_pc, instruction):
         state.context['trace'].append(pc)
 
 class ExtendedTracer(Plugin):
--- a/manticore/core/plugin.py
+++ b/manticore/core/plugin.py
@@ -32,7 +32,7 @@ class Tracer(Plugin):
     def will_start_run_callback(self, state):
         state.context['trace'] = []
 
-    def did_execute_instruction_callback(self, state, pc, target_pc, instruction):
+    def will_execute_instruction_callback(self, state, pc, instruction):
         state.context['trace'].append(pc)
 
 class ExtendedTracer(Plugin):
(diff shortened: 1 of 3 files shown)


manticore/core/plugin.py
```diff
@@ -32,7 +32,7 @@ class Tracer(Plugin):
     def will_start_run_callback(self, state):
         state.context['trace'] = []
 
-    def did_execute_instruction_callback(self, state, pc, target_pc, instruction):
+    def will_execute_instruction_callback(self, state, pc, instruction):
```
feliam: Does this works on forking instructions ? Does this get executed twice on a fork ?

manticore/platforms/evm.py
```diff
@@ -2381,3 +2382,27 @@ def compare_buffers(a, b):
             
         self.current._push(value)
         self.current.pc += self.current.instruction.size
+
+    def generate_workspace_files(self):
+        def solve_data(data):
+            ret = ''
+            try:
```
feliam: Why not state.solve_one   ?

mossberg: there is no state accessible from the platform

feliam: Ah.
So why not 
`solver.get_value(self._constraints, data)`
?

mossberg: ah, there is no reason, i wasn't paying attention :( just quickly copying `SLinux.generate_workspace_file`

manticore/core/plugin.py
mossberg: yes it probably will. i don't have time to really think too much about a proper solution for this right now, and it's not that high priority, so i'll omit this diff

mossberg: good catch, thanks!

manticore/platforms/evm.py
```diff
@@ -2381,3 +2382,18 @@ def compare_buffers(a, b):
...
+
+    def generate_workspace_files(self):
+        ret = {}
+        for i, tx in enumerate(self.transactions):
+            name = 'tx.{}'.format(i)
+            data = {
+                'to': tx.address,
+                'from': tx.caller,
+                'data': solver.get_value(self.constraints, tx.data).encode('hex'),
+                'value': tx.value if not issymbolic(tx.value) else '{symbolic!}'
```
feliam: For `data` it provides a concrete solution but for `value` `{symbolic!}`.
Maybe provide a concrete solution for value too ?

PR #639, opened 2017-12-11
Added PUSHA,PUSHAD,POPA,POPAD x86 instructions

ramiscoding:
Test cases for these instructions are being worked on. The auto testcase generation tool doesn't handle these instructions.

### DIFF
--- a/manticore/core/cpu/x86.py
+++ b/manticore/core/cpu/x86.py
@@ -3068,6 +3068,78 @@ class X86Cpu(Cpu):
         cpu.push(cpu.RFLAGS, 64)
 
     @instruction
+    def PUSHA(cpu):
+        '''
+        Pushes contents of all the registers onto the stack.
+
+        The order in which the register values are pushed into the stack
+        is the same as mentioned in the list below.The value of SP pushed
+        is the value prior to the first push.
+
+        :param cpu: current CPU.
+        '''
+        temp_SP = cpu.read_register('SP')
+        registers = ['AX', 'CX', 'DX', 'BX', temp_SP, 'BP', 'SI', 'DI']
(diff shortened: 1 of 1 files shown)


manticore/core/cpu/x86.py
```diff
@@ -3067,6 +3067,78 @@ def PUSHFQ(cpu):
...
+
+        The order in which the register values are pushed into the stack
+        is the same as mentioned in the list below.The value of SP pushed
+        is the value prior to the first push.
+
+        :param cpu: current CPU.
+        '''
+        temp_SP = cpu.read_register('SP')
+        registers = ['AX', 'CX', 'DX', 'BX', temp_SP, 'BP', 'SI', 'DI']
+        for reg in registers:
```
yan: This code reads the value of `SP` and uses it as register name, probably not what you intended. Ditto for `PUSHAD`.

pgoodman:
I think that the code would be simpler if you manually wrote out the actual operations on the various registers/stack, as opposed to using a loop.

manticore/core/cpu/x86.py
ramiscoding: That was intended behavior. The value of SP pushed into the stack should be the original value of SP before the instruction is executed. That is why I saved the SP value into another variable and pushed it to the stack.

yan: The original value of SP isn't what's being pushed however. Try to write a simple test case to confirm these semantics.

mossberg:
hi @ramiscoding! since it looks like progress on this has slowed, i am going to close this pr for the time being. feel free to reopen if you look into it again, and of course feel free to ask any questions for clarification :)

PR #640, opened 2017-12-12
System calls addition - sys_getrandom and sys_openat

### LINKED ISSUE #707: SyscallNotImplemented: Syscall index "257" (sys_openat) not implemented.
### OS / Environment
kali-current

### Manticore version
Version: 0.1.6

### Python version
Python 2.7.14+

I have the most simplest program:

```
#include <stdio.h>
void main(int argc, char *argv[]) {
  char buf[8];
  strcpy(buf, argv[1]);
  if (strlen(buf) > 12)
    printf("done\n");
  return 0;
}
````

but when run with manticore I get:
````
# manticore ./strcpy +++
2018-01-14 15:20:54,228: [6828] m.manticore:INFO: Loading program ./strcpy
2018-01-14 15:20:58,744: [6828] m.c.executor:ERROR: Exception: Syscall index "257" (sys_openat) not implemented.
  File "/usr/local/lib/python2.7/dist-packages/manticore/core/executor.py", line 463, in run
  File "/usr/local/lib/python2.7/dist-packages/manticore/core/state.py", line 126, in execute
  File "/usr/local/lib/python2.7/dist-packages/manticore/platforms/linux.py", line 2015, in execute
    self.syscall()
  File "/usr/local/lib/python2.7/dist-packages/manticore/platforms/linux.py", line 1856, in syscall
    raise SyscallNotImplemented(index, name)
SyscallNotImplemented: Syscall index "257" (sys_openat) not implemented.

2018-01-14 15:20:58,749: [6828] m.manticore:INFO: Results in /prg/misc/vulndev/mcore_9ivqcB
2018-01-14 15:20:58,750: [6828] m.manticore:INFO: Total time: 3.67245984077
````

on the examples its working though

### PR DESCRIPTION
**Pull request by Srinivas11789**
**Summary:**
System call addition at linux.py 
* sys_getrandom
* sys_openat

**Progress:** Tested with some binaries making the respective system calls

**Pending tasks -** Unit test code

Fixes #707

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/640)

### CODE REVIEW

File: manticore/platforms/linux.py [hunk 1]
```diff
@@ -1314,6 +1314,39 @@ def sys_open(self, buf, flags, mode):
...
+            return self.sys_open(buf, flags, mode)
+
+        try:
+            directory_file_descriptor = self._get_fd(dirfd)
+        except BadFd:
+            logger.info("OPENAT: Not valid file descriptor. Returning EBADF")
+            return -errno.EBADF
+
+        if os.path.isdir(directory_file_descriptor.name):
+                buf = os.path.relpath(buf,directory_file_descriptor.name)
```
yan: Fix indent?

manticore/platforms/linux.py
```diff
@@ -2408,6 +2441,80 @@ def sys_open(self, buf, flags, mode):
...
+        if issymbolic(buf):
+            logger.debug("Ask to generate random to a symbolic buffer")
+            raise ConcretizeArgument(self, 1)
+
+        if issymbolic(size):
+            logger.debug("Ask to generate random of symbolic number of bytes ")
+            raise ConcretizeArgument(self, 2)
+
+        try:
+          flag = ['', 'GRND_NONBLOCK', 'GRND_RANDOM'][flags & 7]
```
yan: Assign names to `flag` values and use those in the body, as opposed to turning them into a string and turning flag checking into a string comparison? Also, the way it's written right now, the code is incorrect in case multiple flags are set. (line 2499 is never reached)

i.e.
```
flags = flags & 7
GRND_NONBLOCK = 1
GRND_NONBLOCK = 2
...
if flags & GRND_NONBLOCK:
    ..
```

manticore/platforms/linux.py
Srinivas11789: fixed the indent.

manticore/platforms/linux.py
Srinivas11789: Will rework the flags logic.

yan: Any updates on the above rework? We can also help with implementing it.

yan:
@Srinivas11789 Heya, did you have a chance to address the comments re: flags? We can work on the fixes ourselves if you're short on time. Thanks!

manticore/platforms/linux.py
```diff
@@ -2408,6 +2481,47 @@ def sys_open(self, buf, flags, mode):
...
+        A version of openat that includes a symbolic path and symnbolic directory file descriptor
+
+        :param dirfd: directory file descriptor
+        :param buf: address of zero-terminated pathname
+        :param flags: file access bits
+        :param mode: file permission mode
+        '''
+
+        if issymbolic(dirfd):
+            logger.debug("Ask to read from a symbolic directory file descriptor!!")
```
feliam: Should not we constraint the dirfd so it can only _range_ into the already opened fd set (+ 1 unopened)?

mossberg:
Review status: 0 of 1 files reviewed at latest revision, 8 unresolved discussions.

---

*[manticore/platforms/linux.py, line 34 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5B1D6BhQ3S12dMz6TQ:-L5B1D6BhQ3S12dMz6TR:b-vwa7mt) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L34)):*
> ```Python
> 
> class FdError(Exception):
>     def __init__(self, message='', err=None):
> ```

we can directly make `err=errno.EBADF` here

---

*[manticore/platforms/linux.py, line 1092 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5B02Xnog4SV0dvrVzt:-L5B02Xnog4SV0dvrVzu:bw33kr3) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1092)):*
> ```Python
>     def _get_fd(self, fd):
>         if not self._is_fd_open(fd):
>             raise FdError()
> ```

nit, can omit the ()

---

*[manticore/platforms/linux.py, line 1369 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5BBtugyHbQOw6xTpdw:-L5BBtugyHbQOw6xTpdx:bxgw5x) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1369)):*
> ```Python
>         except IOError as e:
>             logger.info("Could not open file %s. Reason: %s", filename, str(e))
>             return -e.errno if e.errno else -errno.EINVAL
> ```

we lost the `is not None` in the original, we should keep it

---

*[manticore/platforms/linux.py, line 1963 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5B5z5G3Sa-vW-wh2bZ:-L5B5z5G3Sa-vW-wh2b_:b482cun) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1963)):*
> ```Python
> 
>         try:
>             if flags & GRND_RANDOM:
> ```

we should do a check for invalid flags and ret einval
http://elixir.free-electrons.com/linux/latest/source/drivers/char/random.c#L1912

---

*[manticore/platforms/linux.py, line 1967 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5B8igf-WhJCbCN9EQN:-L5B8igf-WhJCbCN9EQO:bru5zkb) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1967)):*
> ```Python
>                     data = rnd.read(size)
>             else:
>                 data = os.urandom(size)
> ```

it occurs to me that we don't need to actually provide random data here, we could just do something like `data = 'A'*size`, if we wanted to. this would make this code simpler, and also reduce the ability of the analyzed program to influence the analysis machine (affecting its entropy pool). it would also make using manticore more deterministic; different runs of manticore on the same binary that uses getrandom would behave the same way, which can be nice.

a con would be that manticore would be somewhat less fit for just running programs, but i don't find this particularly compelling.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640)*

mossberg:
Review status: 0 of 1 files reviewed at latest revision, 12 unresolved discussions.

---

*[manticore/platforms/linux.py, line 1044 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5BEKyR-Ifb6iCZWCIS:-L5BEKyR-Ifb6iCZWCIT:b-xfbf8o) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1044)):*
> ```Python
>         if self.current.address_bit_size == 32:
>             sdword = ctypes.c_int32(dword).value
>         else:
> ```

alternatively explicitly check for 32 and 64 bits, and error out with a clear message otherwise? 

---

*[manticore/platforms/linux.py, line 1347 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5BDYAcbkLfK49FMd7B:-L5BDYAcbkLfK49FMd7C:brzcl7y) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1347)):*
> ```Python
>         if os.path.abspath(filename).startswith('/proc/self'):
>             if filename == '/proc/self/exe':
>                 filename = os.path.abspath(self.program)
> ```

can we add an else that errors out if it was a /proc/self file but not /proc/self/exe? currently we would just try to open it, which would have wrong results, since it would open the /proc/self for the manticore process?

---

*[manticore/platforms/linux.py, line 1392 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5BEbYjiNbGhoPmEkCs:-L5BEbYjiNbGhoPmEkCt:b4jchz7) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1392)):*
> ```Python
>         dirfd = self._to_signed_dword(dirfd)
>         if dirfd == self.FCNTL_FDCWD:
>             dir_path = os.path.curdir
> ```

right here, could we also just pass it to `self.sys_open`, since it's effectively the same behavior now?

---

*[manticore/platforms/linux.py, line 1412 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5BEstknzuBkpOBYo8U:-L5BEstknzuBkpOBYo8V:bydihqh) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1412)):*
> ```Python
>         except IOError as e:
>             logger.info("Could not open file %s. Reason: %s", filename, str(e))
>             return -e.errno if e.errno else -errno.EINVAL
> ```

same `is not None` thing from above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640)*

yan:
Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions.

---

*[manticore/platforms/linux.py, line 34 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5B1D6BhQ3S12dMz6TQ:-L5BDyRpEjtlggQPorjh:bbt6o99) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L34)):*

we can directly make `err=errno.EBADF` here
</blockquote></details>

🤦‍♂️

---

*[manticore/platforms/linux.py, line 1044 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5BEKyR-Ifb6iCZWCIS:-L5BFjP-ya_sJNwfrxEq:b-gikbgv) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1044)):*

alternatively explicitly check for 32 and 64 bits, and error out with a clear message otherwise? 
</blockquote></details>

Agree with sentiment, and will update code, but I wonder what would have to happen to the cpu state where `address_bit_size` is incorrect.

---

*[manticore/platforms/linux.py, line 1392 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5BEbYjiNbGhoPmEkCs:-L5BGZh_72bHnV3cAojK:bkrxxo9) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1392)):*

right here, could we also just pass it to `self.sys_open`, since it's effectively the same behavior now?
</blockquote></details>

It's not exactly the same behavior, as we have to pass sys_open a pointer to the path. To make use of sys_open directly, we'd have to mmap a buffer, write the new path into it, pass an address, and clean up. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640)*

mossberg:
Review status: 0 of 1 files reviewed at latest revision, 13 unresolved discussions.

---

*[manticore/platforms/linux.py, line 1072 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5BPhfAuB52nolpDKvd:-L5BPhfAuB52nolpDKve:b-u69s84) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1072)):*
> ```Python
>         :return: C{0} on success.
>         '''
>         self.files[fd] = None
> ```

i noticed it seems like File.close or Directory.close never actually gets called? while we're here, maybe we can clean this up

---

*[manticore/platforms/linux.py, line 1053 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5BQ5XhflmPrHT465_G:-L5BQ5XhflmPrHT465_H:bq97eku) ([raw file](https://github.com/trailofbits/manticore/blob/9c7c133284aaa95ad9c937b525bb71f5ba5f4345/manticore/platforms/linux.py#L1053)):*
> ```Python
>             sdword = ctypes.c_int64(dword).value
>         else:
>             raise EnvironmentError("Corrupted internal CPU state (arch width is {})".format(arch_width))
> ```

this could also potentially just trigger down the line when someone is implementing a 16 bit cpu or something

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640)*

yan:
Reviewed 1 of 1 files at r9, 1 of 1 files at r10.
Review status: 0 of 1 files reviewed at latest revision, 13 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640)*

mossberg:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions.

---

*[manticore/platforms/linux.py, line 2495 at r8](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L4I4HUyIwSuoCVrc8j7-r8-2495:-L5Fkp-cpvfgi9XFlxKo:b-cuhfbx) ([raw file](https://github.com/trailofbits/manticore/blob/21cf46021cc865ce008b2d14bbcdc4f676753cee/manticore/platforms/linux.py#L2495)):*

Should not we constraint the dirfd so it can only _range_ into the already opened fd set (+ 1 unopened)?
</blockquote></details>

yes, i think we should definitely do this. otherwise, if it's a free symbolic, it'll use the default minmax policy, and the max won't be a meaningful fd. we should constrain `dirfd` like felipe said, and use the ALL concretization policy

---

*[manticore/platforms/linux.py, line 1974 at r11](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5FlT97RvgJsvWXdrlR:-L5FlT97RvgJsvWXdrlS:b-8fo681) ([raw file](https://github.com/trailofbits/manticore/blob/c186efb1c69494a55c5e3b0b987008b1e42479b6/manticore/platforms/linux.py#L1974)):*
> ```Python
>             return -errno.EFAULT
> 
>         if flags & ~(GRND_NONBLOCK|GRND_RANDOM) != 0:
> ```

nit, rm the `!= 0` ?

---

*[manticore/platforms/linux.py, line 1978 at r11](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5FmHKbqD-8uXAiF3BG:-L5FmHKbqD-8uXAiF3BH:b-c0df4q) ([raw file](https://github.com/trailofbits/manticore/blob/c186efb1c69494a55c5e3b0b987008b1e42479b6/manticore/platforms/linux.py#L1978)):*
> ```Python
> 
>         try:
>             if self._true_random:
> ```

on second thought, do you feel strongly about retaining the true random functionality? if we don't, we can seriously reduce the complexity of this code (this whole block will turn into 1 line!). imo, given how rarely we expect true random to be needed/used, that is a worthwhile tradeoff

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640)*

yan:
:sweat_smile: 

---

Reviewed 1 of 1 files at r11.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[manticore/platforms/linux.py, line 2495 at r8](https://beta.reviewable.io/reviews/trailofbits/manticore/640#-L4I4HUyIwSuoCVrc8j7-r8-2495:-L5Fthns26CSWsnLnB3-:b-u5erce) ([raw file](https://github.com/trailofbits/manticore/blob/21cf46021cc865ce008b2d14bbcdc4f676753cee/manticore/platforms/linux.py#L2495)):*

yes, i think we should definitely do this. otherwise, if it's a free symbolic, it'll use the default minmax policy, and the max won't be a meaningful fd. we should constrain `dirfd` like felipe said, and use the ALL concretization policy
</blockquote></details>

but then we won't explore any cases where `dirfd` is invalid. I guess to compromise we can constraint to [0, maxfd+1].

---

*[manticore/platforms/linux.py, line 1392 at r9](https://beta.reviewable.io/reviews/trailofbits/manticore/640#-L5BEbYjiNbGhoPmEkCs:-L5Fv6GK5QvKdKp4NOeB:b-g7e7n1) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1392)):*

isn't `buf ` a pointer to the path? based on the man page, my understanding is that if the path is relative, and the FDCWD flag is passed, behavior is identical to open().
</blockquote></details>

you're right; even if the path is absolute, behavior should be identical to open() if FDCWD is passed.

---

*[manticore/platforms/linux.py, line 1978 at r11](https://beta.reviewable.io/reviews/trailofbits/manticore/640#-L5FmHKbqD-8uXAiF3BG:-L5FtvqQATP3XLzNoqUm:bhls0m1) ([raw file](https://github.com/trailofbits/manticore/blob/c186efb1c69494a55c5e3b0b987008b1e42479b6/manticore/platforms/linux.py#L1978)):*

on second thought, do you feel strongly about retaining the true random functionality? if we don't, we can seriously reduce the complexity of this code (this whole block will turn into 1 line!). imo, given how rarely we expect true random to be needed/used, that is a worthwhile tradeoff
</blockquote></details>

I vote to leave it; the code is already written and 4 simple lines isn't a ton of cognitive load imho. Also don't feel super strongly about it, so can remove it.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/640#-:-L5FtSqhEdOIX-GToqxE:be2l6su)*

mossberg:
Reviewed 1 of 1 files at r11.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[manticore/platforms/linux.py, line 2495 at r8](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L4I4HUyIwSuoCVrc8j7-r8-2495:-L5Fw7WWYMA55-iDnDVD:b-9rbi39) ([raw file](https://github.com/trailofbits/manticore/blob/21cf46021cc865ce008b2d14bbcdc4f676753cee/manticore/platforms/linux.py#L2495)):*

but then we won't explore any cases where `dirfd` is invalid. I guess to compromise we can constraint to [0, maxfd+1].
</blockquote></details>

mhm, i believe that's what felipe suggested (opened fd set + 1 unopened)

---

*[manticore/platforms/linux.py, line 1392 at r9](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5BEbYjiNbGhoPmEkCs:-L5FyN5e8hcHeiimDM_M:b-sv3tty) ([raw file](https://github.com/trailofbits/manticore/blob/487d5e8adfdc654a1f4ef8a95fca3cbedfc50dbf/manticore/platforms/linux.py#L1392)):*

you're right; even if the path is absolute, behavior should be identical to open() if FDCWD is passed.
</blockquote></details>

ok great. in this case, i think we should do the replacement then. it will allow us to eliminate the `else:` block, and unindent all that code, which should help make the code easier to read.

---

*[manticore/platforms/linux.py, line 1978 at r11](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5FmHKbqD-8uXAiF3BG:-L5Fx6wpBCAXOgD1SRhW:be9m6xg) ([raw file](https://github.com/trailofbits/manticore/blob/c186efb1c69494a55c5e3b0b987008b1e42479b6/manticore/platforms/linux.py#L1978)):*

I vote to leave it; the code is already written and 4 simple lines isn't a ton of cognitive load imho. Also don't feel super strongly about it, so can remove it.
</blockquote></details>

this is what i'm talking about, we can replace 15 lines of complex control flow and exception handling, with 1 super easy to grok line! we can always get this code back if we really need it, but i don't think we'll miss it..i would like to remove it :)



---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640)*

yan:
Reviewed 2 of 2 files at r12.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640)*

mossberg:

---

*[manticore/platforms/linux.py, line 2564 at r12](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5L3FDmldgWolBRWLex:-L5L3FDmldgWolBRWLey:b-l0yf7t) ([raw file](https://github.com/trailofbits/manticore/blob/3ddbb7deaaf10e82d13c42064e528ab0ecc80809/manticore/platforms/linux.py#L2564)):*
> ```Python
>         if issymbolic(dirfd):
>             logger.debug("Ask to read from a symbolic directory file descriptor!!")
>             # Constraint to a valid fd and one past the end of fds
> ```

sp, "Constrain"

---

*[manticore/platforms/linux.py, line 2567 at r12](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5L35sWDwSCCUUxBG68:-L5L35sWDwSCCUUxBG69:b3w7ktg) ([raw file](https://github.com/trailofbits/manticore/blob/3ddbb7deaaf10e82d13c42064e528ab0ecc80809/manticore/platforms/linux.py#L2567)):*
> ```Python
>             self.constraints.add(dirfd >= 0)
>             self.constraints.add(dirfd <= len(self.files))
>             raise ConcretizeArgument(self, 0)
> ```

is MINMAX the appropriate policy to use here? perhaps we should use ALL or SAMPLED?

---

*[tests/test_linux.py, line 141 at r12](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5L0q7GjKjNYT-hlvGN:-L5L0q7GjKjNYT-hlvGO:b-n4c38w) ([raw file](https://github.com/trailofbits/manticore/blob/3ddbb7deaaf10e82d13c42064e528ab0ecc80809/tests/test_linux.py#L141)):*
> ```Python
> 
>         # Create a minimal state
>         model = self.symbolic_linux
> ```

s/model/platform throughout?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640)*

yan:
Review status: 1 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[manticore/platforms/linux.py, line 2567 at r12](https://reviewable.io:443/reviews/trailofbits/manticore/640#-L5L35sWDwSCCUUxBG68:-L5L69uZJxHrwLV2ioXQ:bonx9iy) ([raw file](https://github.com/trailofbits/manticore/blob/3ddbb7deaaf10e82d13c42064e528ab0ecc80809/manticore/platforms/linux.py#L2567)):*

is MINMAX the appropriate policy to use here? perhaps we should use ALL or SAMPLED?
</blockquote></details>

I went with default because `ALL` would be everything including 0-2 and `SAMPLED` might miss the out of bounds fd.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640)*

mossberg:
:lgtm:

---

Reviewed 2 of 2 files at r13.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640#-:-L5L8zPvbN84Dx7MOrrc:bnfp4nl)*

yan:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/640#-:-L5LIIqT91zmR745DCVd:bnfp4nl)*

PR #641, opened 2017-12-13
Better EVM workspace output

feliam:
Better EVM specific workspace output.
Examples API updated to use that ^
seth.finalize() makes testcases of still running states at the end of analisys

https://asciinema.org/a/nkSvLzlIF5Xi2AkEB3GKdJtL4


examples/evm/solidse.py
```diff
@@ -65,7 +65,7 @@
 
         if tx.result in ('THROW', 'REVERT', 'SELFDESTRUCT'):
             if metadata is not None:
-                address, offset = state.context['seth.trace'][-1]
+                address, offset = state.context['m.trace'][-1]
```
mossberg: is this intended?

manticore/core/state.py
```diff
@@ -310,6 +310,8 @@ def solve_one(self, expr):
         :return: Concrete value
         :rtype: int
         '''
+        if not issymbolic(expr):
```
mossberg: `solver.get_value` should already handle this case. is the intention to give `state.solve_one(not_symbolic)` and `solver.get_value(cs, not_symbolic)` different behaviors?

manticore/core/workspace.py
```diff
@@ -393,6 +393,16 @@ def save_state(self, state):
         self._store.save_state(state, '{}{:08x}{}'.format(self._prefix, id_, self._suffix))
         return id_
 
+    def rm_state(self, state_id):
+        """
+        Remove a state from storage identified by `state_id`.
+
+        :param state_id: The state reference of what to load
+        :return: The deserialized state
+        :rtype: State
```
mossberg: docstring bug, this doesn't return anything

mossberg:
I _really_ like the `(*)` thing to hint concretized values!

manticore/seth.py
```diff
@@ -885,188 +977,253 @@ def global_findings(self):
...
-            for address, memlog, topics in world.logs:
-                try:
-                    res = memlog
-                    if isinstance(memlog, Expression):
-                        res = state.solve_one(memlog)
-                        if isinstance(memlog, Array):
-                            state.constrain(compare_buffers(memlog, res))
-                        else:
-                            state.constrain(memlog== res)
+    def _generate_testcase_callback(self, state, name, message):
```
mossberg: is there any particular reason we override `_generate_testcase_callback` vs modifying `EVMWorld.generate_workspace_files()`?

manticore/seth.py
```diff
@@ -885,188 +977,253 @@ def global_findings(self):
...
+                        offset = state.solve_one(offset)
+                        value = state.solve_one(value)
+                        summary.write("\t%032x -> %032x %s\n" % (offset, value, flagged(is_storage_symbolic)))
+                        is_something_symbolic = is_something_symbolic or is_storage_symbolic
+
+                code = blockchain.get_code(account_address)
+                if len(code):
+                    summary.write("Code:\n")
+                    fcode = StringIO.StringIO(code)
+                    for chunk in iter(lambda: fcode.read(32), b''):
```
mossberg: chunking like this is nicer to humans visually, but makes it harder to copy/paste or parse from the file. i feel like we should just simply dump it

mossberg:
I tested `$ manticore demo.sol`, and the workspace only has 8 test_* file groups instead of 10. It looks like the 2 alive STOP states aren't having testcase files emitted.

```
pragma solidity ^0.4.15;

contract SymExExample {
    uint did_init = 0;

    // function id: 0x13371337
    function test_me(int input) {
        if (did_init == 0) {
            did_init = 1;
            return;
        }

        if (input < 42) {
            // safe
            return;
        } else {
            // overflow possibly!
            int could_overflow = input + 1;
        }

    }
}
```

manticore/seth.py
feliam: Yes I wanted to avoid calling
testcase_id = self._output.save_testcase(state, name, message)

Which in turns call stuff in workspace module that's too linux biased.
I think maybe workpsace should not do the formatting of any output.

mossberg: `output.save_testcase` generates:
- summary
- trace
- constraints
- input symbols
- state.pkl

all of these are fairly generic on not linux based, no? except summary, which currently has a hack for an evm summary. `save_testcase` then calls into the platform for the platform specific files.

feliam: constraints, input symbols, state.pkl yes.

'trace' should be the result of asking a plugin to generate the report for it.
for p in plugins:
   dump(p.generate_workspace_files())

Also that trace plugin wont work for EVM as a trace in evm needs a pair. ADDRESS, OFFSET (not state.platform.current.pc as it is now)

manticore/seth.py
```diff
@@ -885,188 +977,253 @@ def global_findings(self):
...
+
+
+        #Transactions
+        with testcase.open_stream('tx') as tx_summary:
+            is_something_symbolic = False
+            for tx in blockchain.transactions: #external transactions
+                tx_summary.write("Transactions Nr. %d\n" % blockchain.transactions.index(tx))
+
+                #The result if any RETURN or REVERT
+                tx_summary.write("Type: %s -> %s\n" % (tx.sort, 'OK' if tx.result == 'RETURN' else 'FAIL' ))
```
mossberg: 'STOP' is an acceptable result also, should not trigger a FAIL

manticore/__main__.py
```diff
@@ -103,75 +103,10 @@ def ethereum_cli(args):
...
-    #         # All accounts ever created by the script
-    #         # (may not all be alife in all states)
-    #         # (accounts created by contract code are not in this list )
-    # print "[+] Global coverage:"
-    # for address in seth.contract_accounts:
-    #     print address, seth.global_coverage(
-    #         address)  # coverage % for address in this state
+
+    m.finalize()
+    print "[+] Look for results in %s"% seth.workspace
```
mossberg: yikes, seth.workspace

feliam: What is the correct way to send ppl to the workspace folder?

mossberg:
I get this in my .tx file for a stop case, i don't think we should emit the `return:` for a STOP insn
```
Function call:
test_me(74470206411450415880257417809509085835044225432019813735180563832540208866468) -> STOP (*)
return: (96, 96, 96, 64, 82, 96, 0, 53, 124, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 144, 4, 99,
```

manticore/__main__.py
mossberg: `m.workspace` ;)

manticore/platforms/evm.py
```diff
@@ -2177,8 +2182,7 @@ def create_contract(self, origin=None, price=0, address=None, caller=None, balan
 
         self.storage[address]['storage'] = EVMMemory(self.constraints, 256, 256)
 
-
-        self._pending_transaction = ('Create', address, origin, price, '', origin, balance, init, header)
+        self._pending_transaction = ('Create', address, origin, price, '', origin, balance, ''.join(init), header)
```
mossberg: not for now, but TODO: use a namedtuple

manticore/core/workspace.py
```diff
@@ -415,6 +423,23 @@ def __init__(self, desc=None):
         self._id_gen = manager.Value('i', self._last_id)
         self._lock = manager.Condition(manager.RLock())
 
+    def testcase(self, prefix='test'):
```
mossberg: not for now, but TODO: this feels kind of structure feels awkward. the increasing prominence of APIs here and there that are used by only EVM stuff or only Linux is a sign there is deep restructuring we need to do.

manticore/platforms/evm.py
feliam: yes!

manticore/core/workspace.py
feliam: Yes!!! 
Please have only one of this testcase() in the restructuring !

manticore/platforms/evm.py
```diff
@@ -2237,21 +2244,74 @@ def transaction(self, address, origin=None, price=0, data='', caller=None, value
...
+
         self._pending_transaction=None
+
+
+        if ty == 'Create': 
+            data = bytecode
+ 
+        is_human_tx = ( self.depth == 0 )
+
+        if failed:
```
mossberg: formerly, we would raise `TerminateState("Not Enough Funds for transaction"` and generate a not enough funds final state. It seems we no longer generate these not enough funds states. it's not very obvious to me what this alternative code does, can you explain briefly?

manticore/platforms/evm.py
```diff
@@ -2237,21 +2244,74 @@ def transaction(self, address, origin=None, price=0, data='', caller=None, value
...
+        #Here we have enoug funds and room in the callstack
+
+        self.storage[address]['balance'] += value
+        self.storage[caller]['balance'] -= value
+
         new_vm = EVM(self._constraints, address, origin, price, data, caller, value, bytecode, header, global_storage=self.storage)
         self._push_vm(new_vm)
-        if self.depth == 1:
 
+        if is_human_tx:
```
mossberg: ❤️  this self documenting code

manticore/platforms/evm.py
feliam: A CALL issued by a contract with not enough funds just  need to push 0 on the caller VM and keep going. I think we lacked that before. 

A human trying to do a transaction with not enough funds is a fail tx.

manticore/seth.py
```diff
@@ -1082,71 +1239,22 @@ def global_coverage(self, account_address):
...
-                output += "   0x%04x %s\n"%(offset, i)
-            
-            total += 1
-            offset += i.size
-
-        output += "Total assembler lines: %d\n"% total
-        output += "Total assembler lines visited: %d\n"% count
-        output += "Coverage: %2.2f%%\n"%  (count*100.0/total)
-        return output
+    #Fixme. This is for avoinding  super._did_finish_run_callback
```
mossberg: TODO: find a better way to suppress execution of Manticore._did_finish_run_callback

Issue #630, opened 2017-12-05
Support EVM in CLI mode


mossberg:
The manticore cli mode should support Ethereum smart contracts, written in solidity.

todo
- [x] merge #629 
- [x] implement various analysis plugins corresponding to these issues: https://github.com/trailofbits/not-so-smart-contracts
  - [x] int overflow
  - ~[ ] reentrancy~
  - ~[ ] unchecked external call~
- [x] [UI](https://github.com/orgs/trailofbits/teams/manticore/discussions/3): accept solidity/evm files
- [x] [UI](https://github.com/orgs/trailofbits/teams/manticore/discussions/3): implement workspace files
- [ ] make workspace filenames compatible with ethersplay
- [ ] generate testcase when detectors detect a failure. right now that path will generate a testcase, but it will not specifically trigger the failure condition

mossberg:
currently, the evm cli has enough core functionality for the release. it is unclear if we will have reentrancy or uncheck external call analyses in time, but i think it's fine. since the functionality is there, the remaining required work is related to UI polish and cosmetics

- [ ] show human readable log messsage bytes in ascii (vs hexpair). because you can log a string to conveniently see if a certain contract path was hit
- [ ] clean up stdout output
  - the multiple "Starting 1 processes" thing looks odd. may not be possible to remove this without significant refactoring.
  - it would be nice to print out the statuses of the states as the analysis proceeds
  - we should attempt to make things look as similar as possible to how manticore looks for linux binaries

```
2017-12-15 16:18:53,989: [34451] m.manticore:INFO: Starting 1 processes.
2017-12-15 16:18:54,068: [34451] m.manticore:INFO: Starting 1 processes.
[+] Coverage after 1 transactions: 54%
[+] There are 4 reverted states now
[+] There are 1 alive states now
2017-12-15 16:19:09,978: [34451] m.manticore:INFO: Starting 1 processes.
[+] Coverage after 2 transactions: 100%
[+] There are 8 reverted states now
[+] There are 2 alive states now
[+] Look for results in /mnt/hgfs/code/manticore/examples/evm/mcore_548qV5
```

mossberg:
elaborating on stdout cleanup work:
- we should aim to use logging instead of `print '[+]'`. we can easily move [this code](https://github.com/trailofbits/manticore/blob/master/manticore/__main__.py#L78-L108) into `seth.py` as a method on `ManticoreEVM` in order to have access to the seth logger
[...]

PR #647, opened 2017-12-18
EVM logging cleanup

yan:
As described partly in #630

### DIFF
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -1,9 +1,17 @@
 import sys
+import logging
 import argparse
 
-from manticore import Manticore
+from . import Manticore
+from .utils import log
+
+# XXX(yan): This would normally be __name__, but then logger output will be pre-
+# pended by 'm.__main__: ', which is not very pleasing. hard-coding to 'main'
+logger = logging.getLogger('manticore.main')
+log.init_logging()
 
 sys.setrecursionlimit(10000)
+
 
 def parse_arguments():
     ###########################################################################
@@ -99,14 +107,13 @@ def ethereum_cli(args):
         last_coverage = new_coverage
         new_coverage = m.global_coverage(contract_account)
 
-        print "[+] Coverage after %d transactions: %d%%"%(tx_count, new_coverage)
-        print "[+] There are %d reverted states now"% len(m.terminated_state_ids)
-        print "[+] There are %d alive states now"% len(m.running_state_ids)
-
+        logger.info("Coverage after %d transactions: %d%%", tx_count, new_coverage)
+        logger.info("There are %d reverted states now", len(m.terminated_state_ids))
+        logger.info("There are %d alive states now", len(m.running_state_ids))
 
     m.finalize()
-    print "[+] Look for results in %s"% m.workspace
 
+    logger.info("Results in %s", m.workspace)
 
 
 def main():
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -8,7 +8,6 @@ from .utils import log
 # XXX(yan): This would normally be __name__, but then logger output will be pre-
 # pended by 'm.__main__: ', which is not very pleasing. hard-coding to 'main'
 logger = logging.getLogger('manticore.main')
-log.init_logging()
 
 sys.setrecursionlimit(10000)
 
@@ -74,6 +73,7 @@ def parse_arguments():
 
 def ethereum_cli(args):
     from seth import ManticoreEVM, IntegerOverflow, UnitializedStorage, UnitializedMemory
+    log.init_logging()
 
     m = ManticoreEVM(procs=args.procs)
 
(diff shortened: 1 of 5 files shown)


mossberg:
this output looks really good!

```
[I] mark ubuntu /m/h/c/m/e/evm (dev-evm-logging-cleanup) ❯ manticore demo.sol
2017-12-18 14:26:53,222: [64604] m.main:INFO: Starting with 1 processes
2017-12-18 14:26:53,553: [64604] m.seth:INFO: Generated testcase No. 0 - REVERT
2017-12-18 14:26:56,894: [64604] m.seth:INFO: Generated testcase No. 1 - REVERT
2017-12-18 14:27:00,600: [64604] m.seth:INFO: Generated testcase No. 2 - REVERT
2017-12-18 14:27:04,613: [64604] m.seth:INFO: Generated testcase No. 3 - REVERT
2017-12-18 14:27:08,180: [64604] m.seth:INFO: Coverage after 1 transactions.
2017-12-18 14:27:08,181: [64604] m.seth:INFO: There are 4 reverted states now
2017-12-18 14:27:08,181: [64604] m.seth:INFO: There are 1 alive states now
2017-12-18 14:27:08,575: [64604] m.seth:INFO: Generated testcase No. 4 - REVERT
2017-12-18 14:27:13,676: [64604] m.seth:INFO: Integer overflow at ADD instruction
2017-12-18 14:27:13,784: [64604] m.seth:INFO: Generated testcase No. 5 - REVERT
2017-12-18 14:27:19,795: [64604] m.seth:INFO: Generated testcase No. 6 - REVERT
2017-12-18 14:27:26,349: [64604] m.seth:INFO: Generated testcase No. 7 - REVERT
2017-12-18 14:27:31,731: [64604] m.seth:INFO: Coverage after 2 transactions.
2017-12-18 14:27:31,731: [64604] m.seth:INFO: There are 8 reverted states now
2017-12-18 14:27:31,731: [64604] m.seth:INFO: There are 2 alive states now
2017-12-18 14:27:31,890: [64604] m.seth:INFO: Generated testcase No. 8 - Still Running
2017-12-18 14:27:37,543: [64604] m.seth:INFO: Generated testcase No. 9 - Still Running
2017-12-18 14:27:43,500: [64604] m.seth:INFO: Look for results in /mnt/hgfs/code/manticore/examples/evm/mcore_zhDWEJ
```

manticore/__main__.py
```diff
@@ -1,10 +1,17 @@
 import sys
+import logging
 import argparse
 
-from manticore import Manticore
+from . import Manticore
+from .utils import log
```
mossberg: (mostly commenting to doc tech debt, not necessarily requiring a fix. i understand this is the first pass to get it to work)

ideally we shouldn't access internal modules in `__main__.py`, and only call `init_logging` once (i believe its called three times between `__main__.py:main`, `__main__.py:ethereum_cli`, and the top of `manticore.py`)

yan: Agree on ideally. I updated `init_logging()` to keep initialization in one place, to make porting easier once we make a more generic `main()`

manticore/manticore.py
```diff
@@ -404,10 +370,11 @@ def enqueue(self, state):
     ###########################################################################
     # Workers                                                                 #
     ###########################################################################
-    def _start_workers(self, num_processes, profiling=False):
+    def _start_workers(self, num_processes, profiling=False, verbose=True):
         assert num_processes > 0, "Must have more than 0 worker processes"
 
-        logger.info("Starting %d processes.", num_processes)
```
mossberg: I was questioning whether the "# processes" line even belongs in regular manticore output at the info level. i think we should demote this to a debug line. then we can remove the hack to suppress this line just for evm. i think we can remove that line with a general "beginning analysis" output at the info level

yan:
One issue with this change, `EVM` is technically a platform, so it gets exceptionally noise if you add `-v`.

manticore/__main__.py
```diff
@@ -95,21 +104,13 @@ def ethereum_cli(args):
                          data=symbolic_data,
                          value=symbolic_value )
 
-        tx_count += 1
         last_coverage = new_coverage
         new_coverage = m.global_coverage(contract_account)
 
-        print "[+] Coverage after %d transactions: %d%%"%(tx_count, new_coverage)
```
mossberg: we should still output the actual coverage number, let's use the `__main__` logger for the moment for this since we don't have access to it inside manticore. eventually this function will be moved inside manticore, and we can use the appropriate logger once we move it

mossberg: actually we can technically call `m.global_coverage` from within manticore to avoid logging it out here in main. hopefully it's not too slow since we're doing it twice

yan: Moved debug output to `transaction()` since we need access to the contract on which we're calculating coverage.

mossberg:
tech debt:

- some hacks with `log.init_logging` and calling it multiple times in `__main__.py`

PR #656, opened 2017-12-20
eth int overflow: tests, mul support

mossberg:
this pr improves eth int overflow detection to check for overflows due to the mul instruction. it adds a test file for eth tests, restructure the code to make it testable, and changes the CI scripts to install solc.


manticore/seth.py
```diff
@@ -54,13 +54,13 @@ class IntegerOverflow(Detector):
         Detects any it overflow on instructions ADD and SUB.
     '''
     def did_evm_execute_instruction_callback(self, state, instruction, arguments, result):
-        if instruction.semantics == 'ADD':
+        mnemonic = instruction.semantics
+        if mnemonic in ('ADD', 'MUL'):
```
yan: Should this support `ADDMOD` and `MULMOD` as well?

manticore/seth.py
```diff
@@ -823,6 +823,37 @@ def transaction(self, caller, address, value, data):
...
+            self.transaction(caller=attacker_account,
+                          address=contract_account,
+                          data=symbolic_data,
+                          value=symbolic_value )
+
+            prev_coverage = current_coverage
+            current_coverage = self.global_coverage(contract_account)
+            found_new_coverage = prev_coverage != current_coverage
+
+            if not found_new_coverage:
```
mossberg: @yan thoughts on this style of the intermediate variable? i wanted to try to make the code self documenting and avoid a comment which could get out of sync with the code.

manticore/seth.py
feliam: instruction.is_arithmetic == True may imply we could check for something.
div_by_zero is probably a finding too. There are a bunch of low hanging fruit there.

manticore/seth.py
yan: i like it! the only issue (error i made in my pseudocode too) is comparing floating point values with `==`/`!=` can produce false results if the order of operations changes; something like `found_new_coverage = prev_coverage > current_coverage` might be more accurate.

manticore/seth.py
yan: i think it's okay to not include addmod/mulmod if we note it somewhere. adding those checks is a little less straightforward.

mossberg: i added https://github.com/trailofbits/manticore/issues/657 to track it

Issue #658, opened 2017-12-20
evm: add documentation


mossberg:
- manticore evm cli: explicitly say that this requires solc in your path
- files must end in .sol to be recognized
- it is alpha quality, and not fit for production use

mossberg:
we should also document

- we have an initial api for evm dis/assembly
- we have an initial api for evm symex (api that powers the cli)
- expect both to change a lot, and be buggy
- we have some example usage of them in examples/evm

PR #659, opened 2017-12-20
0.1.6-related cleanups

### LINKED ISSUE #654: evm cli: use .bytecode extension in workspace
bytecode files should be compatible with https://github.com/trailofbits/ethersplay, which requires the `.bytecode` extension.

todo:
- `.runtime_bytecode` -> `_runtime.bytecode`. this is in seth.py

```
[I] mark ubuntu /m/h/c/m/e/e/mcore_oQlTF0 (dev-eth-tests) ❯ l
total 1304
-rw-r--r-- 1 501 dialout   1257 Dec 20 13:36 global.summary
-rw-r--r-- 1 501 dialout   3317 Dec 20 13:36 global_SymExExample.init_asm
-rw-r--r-- 1 501 dialout    492 Dec 20 13:36 global_SymExExample.init_bytecode
-rw-r--r-- 1 501 dialout     80 Dec 20 13:36 global_SymExExample.init_visited
-rw-r--r-- 1 501 dialout   2978 Dec 20 13:36 global_SymExExample.runtime_asm
-rw-r--r-- 1 501 dialout    458 Dec 20 13:36 global_SymExExample.runtime_bytecode
-rw-r--r-- 1 501 dialout    983 Dec 20 13:36 global_SymExExample.runtime_visited
-rw-r--r-- 1 501 dialout    504 Dec 20 13:36 global_SymExExample.sol
-rw-r--r-- 1 501 dialout   7363 Dec 20 13:35 test_00000000.constraints
```

yan:
Fixes #654 
Relates to #658

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![Documentation Status](https://readthedocs.org/projects/manticore/badge/?version=latest)](http://manticore.readthedocs.io/en/latest/?badge=latest)
 [![Bountysource](https://img.shields.io/bountysource/team/trailofbits/activity.svg)](https://www.bountysource.com/teams/trailofbits)
 
-Manticore is a prototyping tool for dynamic binary analysis, with support for symbolic execution, taint analysis, and binary instrumentation.
+Manticore is a prototyping tool for dynamic binary and smart contract analysis, with support for symbolic execution, taint analysis, and instrumentation.
 
 ## Features
 
@@ -19,7 +19,7 @@ Manticore supports binaries of the following formats, operating systems, and
 architectures. It has been primarily used on binaries compiled from C and C++.
 Examples of practical manticore usage are also [on github](https://github.com/trailofbits/manticore-examples).
 
-- OS/Formats: Linux ELF
+- OS/Formats: Linux ELF, [Solidity](https://solidity.readthedocs.io/).
 - Architectures: x86, x86_64, ARMv7, and Ethereum Virtual Machine (EVM)
 
 ## Requirements
@@ -123,6 +123,15 @@ def hook(state):
   m.terminate()  # tell Manticore to stop
 
 m.run()
+```
+
+### Ethereum
+
+Manticore includes experimental support for Ethereum smart contracts written in Solidity.
+To analyze a smart contract, make sure it has a `.sol` extension and simply pass it to Manticore:
+
+```
+$ manticore contract.sol   # symbolically explores a smart contract, creating a mcore_* directory with results
 ```
 
 Further documentation is available in several places:
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![Documentation Status](https://readthedocs.org/projects/manticore/badge/?version=latest)](http://manticore.readthedocs.io/en/latest/?badge=latest)
 [![Bountysource](https://img.shields.io/bountysource/team/trailofbits/activity.svg)](https://www.bountysource.com/teams/trailofbits)
 
-Manticore is a prototyping tool for dynamic binary and smart contract analysis, with support for symbolic execution, taint analysis, and instrumentation.
+Manticore is a prototyping tool for dynamic analysis of binaries and smart contracts, with support for symbolic execution, taint analysis, and instrumentation.
 
 ## Features
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![Documentation Status](https://readthedocs.org/projects/manticore/badge/?version=latest)](http://manticore.readthedocs.io/en/latest/?badge=latest)
 [![Bountysource](https://img.shields.io/bountysource/team/trailofbits/activity.svg)](https://www.bountysource.com/teams/trailofbits)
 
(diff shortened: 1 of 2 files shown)


manticore/seth.py
```diff
@@ -1189,7 +1189,7 @@ def finalize(self):
         for address, md in self.metadata.items():
             with self._output.save_stream('global_%s.sol'%md.name) as global_src :
                 global_src.write(md.source_code)
-            with self._output.save_stream('global_%s.runtime_bytecode'%md.name) as global_runtime_bytecode :
+            with self._output.save_stream('global_%s_runtime.bytecode'%md.name) as global_runtime_bytecode :
```
mossberg: sorry in #654 i was unclear, but we should also do this for the .init_bytecode file

yan: we don't want to keep this as `.bytecode` as well for symmetry?

mossberg: yes, both runtime and init bytecode should end in .bytecode

README.md
```diff
@@ -6,7 +6,7 @@
 [![Documentation Status](https://readthedocs.org/projects/manticore/badge/?version=latest)](http://manticore.readthedocs.io/en/latest/?badge=latest)
 [![Bountysource](https://img.shields.io/bountysource/team/trailofbits/activity.svg)](https://www.bountysource.com/teams/trailofbits)
 
-Manticore is a prototyping tool for dynamic binary analysis, with support for symbolic execution, taint analysis, and binary instrumentation.
+Manticore is a prototyping tool for dynamic binary and smart contract analysis, with support for symbolic execution, taint analysis, and instrumentation.
```
mossberg: makes it sound like we analyze "dynamic binaries" which we in fact perform quite poorly on

README.md
```diff
@@ -19,7 +19,7 @@ Manticore supports binaries of the following formats, operating systems, and
 architectures. It has been primarily used on binaries compiled from C and C++.
 Examples of practical manticore usage are also [on github](https://github.com/trailofbits/manticore-examples).
 
-- OS/Formats: Linux ELF
+- OS/Formats: Linux ELF, [Solidity](https://solidity.readthedocs.io/).
```
mossberg: we don't really analyze solidity, we analyze evm which is on the next line. i think we can omit this

README.md
```diff
@@ -125,6 +125,15 @@ def hook(state):
 m.run()
 ```
 
+### Ethereum
```
mossberg: i feel like the readme is getting kinda bloated and shouldn't be used for manticore's documentation. I think we should create a docs/ page for the cli and move this there. on the readme we can have a little teaser above alongside the `/path/to/binary` lines. let's also edit the below "API reference" line; that site is no longer just the API reference, it's manticore's documentation.

README.md
yan: Solidity is a 'format', evm is an 'architecture', no?

mossberg: i guess so. it's an odd situation. if we added support for c files, and just had manticore call out to gcc before running on the binary, would we put C on this line?

yan: We wouldn't, but then Manticore would no longer be a binary analysis toolkit. The current change was meant to be succinct while still mentioning EVM analysis as the first few paragraphs of the readme. I'm okay with removing the reference to Solidity here altogether as long as we leave the ref to smart contracts in the intro

mossberg:
i think the scope of this pr has grown larger than was initially anticipated, maybe we should do documentation/readme stuff in a separate PR since it is not related to the seth.py cleanup

manticore/seth.py
```diff
@@ -570,8 +582,8 @@ def _compile(source_code):
             return name, source_code, bytecode, runtime, srcmap, srcmap_runtime, hashes, abi, warnings
 
     def __init__(self, procs=1):
-        ''' A manticere EVM manager 
-            :param procs: number of workers to use in the exploration  
+        ''' A Manticere EVM manager
```
mossberg: "Manticere"

manticore/seth.py
```diff
@@ -729,51 +741,56 @@ def transactions(self, state_id=None):
     def solidity_create_contract(self, source_code, owner, balance=0, address=None, args=()):
         ''' Creates a solidity contract 
 
-            :param source_code: solidity source code
-            :type source_code: str
+            :param src source_code: solidity source code
```
mossberg: src str typo?

manticore/seth.py
```diff
@@ -729,51 +741,56 @@ def transactions(self, state_id=None):
...
             :type address: int or EVMAccount
-            :param args: constructor arguments
-            :type args: tuple
-            :return: an EVMAccount
+            :param tuple args: constructor arguments
+            :rtype: EVMAccount
         '''
-        name, source_code, init_bytecode, runtime_bytecode, metadata, metadata_runtime, hashes, abi, warnings = self._compile(source_code)
-        address = self.create_contract(owner=owner, address=address, balance=balance, init=tuple(init_bytecode)+tuple(ABI.make_function_arguments(*args)))
+        name, source_code, init_bytecode, runtime_bytecode, metadata, metadata_runtime, hashes, abi, warnings = \
```
mossberg: not sure this is more readable?

PR #665, opened 2017-12-21
Rename manticore.seth to manticore.ethereum

mossberg:
manticore.seth ostensibly stands for "symbolic ethereum" (along the lines of SLinux and SMemory) but this is not obvious. I think manticore.ethereum is a better fit for the module name


yan:
I like `manticore` over `seth`. There's still a lot of references to `seth` within the module though, from shared_context keys to local variables. I don't think that's a big problem as they're implementation details and can be changed later, but it could increase confusion.

PR #666, opened 2017-12-22
Rm unused evm workspace function

mossberg:
Since this func was added, _generate_testcase_callback was overrided in ManticoreEVM totally bypassing this function, which is now unused.

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -2459,18 +2459,3 @@ class EVMWorld(Platform):
             
         self.current._push(value)
         self.current.pc += self.current.instruction.size
-
-    def generate_workspace_files(self):
-        ret = {}
-        for i, tx in enumerate(self.transactions):
-            name = 'tx.{}'.format(i)
-            data = {
-                'to': tx.address,
-                'from': tx.caller,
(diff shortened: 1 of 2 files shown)


yan:
Document that `_generate_testcase_callback` supersedes that function in the comments and remove the outdated comments/`super()` call?

mossberg:
good point. the way `ManticoreEVM` overrides `_generate_testcase_callback` is actually kind of a hack though, that I don't think we should keep in the long term. so it doesn't truly supercede. It is a required hack for now because `ManticoreOutput.save_testcase` (called by the parent `_generate_testcase_callback`) is not general enough for EVM yet (https://github.com/trailofbits/manticore/pull/641#discussion_r157067892). `generate_workspace_files` is still how it works for Linux.



> remove the outdated comments/super() call

which comments/super() call, specifically?

mossberg:
https://github.com/trailofbits/manticore/blob/7875aaf5fd6c4f8fecbacef5bf58b4794733454f/manticore/ethereum.py#L1064-L1067

i think this comment still roughly holds true, and the super call should be there eventually. i added a todo comment clarifying a little more what should be done there

PR #667, opened 2017-12-22
Add --tx flag to control symbolic tx

mossberg:
Currently, the manticore ethereum cli uses coverage based logic for determining when to stop running symbolic transactions. This can be somewhat faulty currently due to bugs in coverage computation. It can be useful to manually be able to control the number of symbolic tx, for example, to limit unnecessary state exploration.

This adds a `--txlimit` flag to the CLI, whose argument must be a positive integer.

### DIFF
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -62,8 +62,8 @@ def parse_arguments():
                               "(default mcore_?????)"))
     parser.add_argument('--version', action='version', version='Manticore 0.1.5',
                          help='Show program version information')
-    parser.add_argument('--tx', type=check_positive,
-                        help='Number of symbolic transactions to run (positive integer) (Ethereum only)')
+    parser.add_argument('--txlimit', type=check_positive,
+                        help='Maximum number of symbolic transactions to run (positive integer) (Ethereum only)')
 
 
     parsed = parser.parse_args(sys.argv[1:])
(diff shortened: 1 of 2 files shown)


manticore/__main__.py
```diff
@@ -58,6 +62,9 @@ def parse_arguments():
                               "(default mcore_?????)"))
     parser.add_argument('--version', action='version', version='Manticore 0.1.5',
                          help='Show program version information')
+    parser.add_argument('--tx', type=check_positive,
```
yan: `--tx` is an ambiguous name for this parameter; on first glance it looks like it represents data to transmit. Maybe `--max_transactions`, `--max_tx_count`, `--max_tx`, or `--tx_limit`?

mossberg: great point, for some reason i have a mild dislike of underscores in cli flags like that though lol
`--maxtx` `--txlimit` `--txmax`?

yan: Out of those, I like `--txlimit`

mossberg:
the arg parse code doesn't check to see if `int(value)` has an exception, but it's kinda cool, since the function is named `positive`, argparse's default exception handler still makes sens

```
[I] mark ubuntu /m/h/c/m/e/evm (dev-tx-arg) ❯ manticore two_tx_ovf.sol  --txlimit zzz
usage: manticore [-h] [--coverage COVERAGE] [--data DATA] [--env ENV]
                 [--file FILES] [--policy POLICY] [--profile] [--procs PROCS]
                 [--timeout TIMEOUT] [-v] [--workspace WORKSPACE] [--version]
                 [--txlimit TXLIMIT]
                 argv [argv ...]
manticore: error: argument --txlimit: invalid positive value: 'zzz'

```

PR #668, opened 2017-12-22
Run linux examples in Travis

### LINKED ISSUE #581: CI should run scripts in examples/

yan:
Updates to how we generate and test our examples.

- Cleaned up the examples Makefile
- Execute all example scripts during Travis' testing that either have associated binaries or that are generic enough to run on simple ones.
 - Cleaned up general code hygiene of examples to make them easier to build.

Note: The example runner is the `run_examples` function in `test_travis.sh`. Ideally, this would be done in a separate script, or per-example script sh files.


scripts/travis_test.sh
```diff
@@ -1,13 +1,80 @@
...
+    done
 else
     echo "Failed to build Linux examples"
     RV=1
 fi
-cd ../..
+popd
+
+if [ "$RV" -eq "0" ]; then
+    echo "Successfully ran Linux examples"
```
mossberg: can we make the output messages a little clearer as to whether the example binaries or the examples scripts were successful?

scripts/travis_test.sh
```diff
@@ -1,13 +1,80 @@
...
+    python ./run_hook.py $HW $MAIN_ADDR
+    if [ $? -ne 0 ]; then
+        return 1
+    fi
+
+    # Straight from the header of state_control.py
+    gcc -static -g src/state_explore.c -o state_explore
+    SE_ADDR=0x$(objdump -S state_explore | grep -A 1 'value == 0x41' |
+               tail -n 1 | sed 's|^\s*||g' | cut -f1 -d:)
+    python ./state_control.py state_explore $SE_ADDR
```
mossberg: can we add output messages printing the names of the scripts as we run them?

yan: Is 7be2586 what you had in mind?

scripts/travis_test.sh
```diff
@@ -1,13 +1,94 @@
...
+    HW=../linux/helloworld
+    SA=../linux/simpleassert
+    END_OF_MAIN=$(objdump -d $SA|awk -v RS= '/^[[:xdigit:]].*<main>/'|grep ret|tr  -d ' ' | awk -F: '{print "0x" $1}')
+    python ./concolic.py $END_OF_MAIN
+    if [ $? -ne 0 ]; then
+        return 1
+    fi
+
+    echo "Running count_instructions.py..."
+    python ./count_instructions.py $HW |grep -q Executed
```
mossberg: if `count_instructions.py` crashed, but managed to output `Executed` before it crashed, the exit status would still be 0 since it's grep's exit status, not `count_instructions.py` 's. since the main reason for doing this is to verify all the scripts work we should make sure to detect failures of the scripts themselves :) . i think one way to easily implement the fail fast would be adding `set -e` at the top which activates bash's fail fast mode

mossberg: actually i was wrong, i don't think just adding `set -e` will fix it

yan: +1 for `set -e`, but printing `Executed` is the very last thing `count_instructions.py` does, so if `Executed` ever made it to stdout, it'd be during a successful cleanup of when it flushes its buffer.

mossberg: ah good point. but technically `m.context['count']` could crash, and `Executed` would still be printed

https://github.com/trailofbits/manticore/blob/7be258605649111b1f457744c3a6e2343b6d63da/examples/script/count_instructions.py#L30

```
[I] mark ubuntu /m/h/c/manticore (dev-test-examples) ❯ python -c "print 3, 3/0, 3" | grep  3; echo $status
3
  File "<string>", line 1, in <module>
ZeroDivisionError: integer division or modulo by zero
0

```

mossberg: that is a super nitpick though, i'm fine with leaving this line as is. i'm fine with merging this pr once we do the fail fast thing!

mossberg:
if i induce a compilation error, it still runs the tests. i think it's due to the `if make`

```
                 from indexhell.c:33:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:343:1: error: expect
 fread_unlocked (void *__restrict __ptr, size_t __size, size_t _
 ^
indexhell.c:36:1: warning: return type defaults to ‘int’ [-Wimpl
 main(){
 ^
indexhell.c: In function ‘main’:
indexhell.c:39:1: warning: implicit declaration of function ‘rea
 read(0, buffer, M);
 ^
Makefile:21: recipe for target 'indexhell' failed
make: *** [indexhell] Error 1
Failed to build Linux example binaries
/mnt/hgfs/code/manticore
Failed to run Linux example binaries
Failed to run example scripts
................................................................
................................................................
................................................................
................................................................
................................................................
................................................................
................................................................
```

mossberg:
the `set -e` will probably allow us to remove all that control flow regarding exit status!

PR #669, opened 2017-12-22
Update __main__.py

mossberg:


yan:
Update `description` in `setup.py` as well?

mossberg:
great catch! unfortunately i did it in dev-016

PR #675, opened 2017-12-27
Fix metadata setting race

### LINKED ISSUE #674: reentrancy_concrete.py example fails with AttributeError: 'EVMAccount' object has no attribute 'set_vulnerable_contract'
### OS / Environment

Distributor ID:	ArchLinux
Description:	ArchLinux
Release:	rolling-release

### Manticore version

Last revision (4c0f4bc6cf7509eb0f3399327fb494188b137671)

### Python version

Python 2.7.13

### Summary of the problem

The reentrancy_concrete.py example fails:
 
"AttributeError: 'EVMAccount' object has no attribute 'set_vulnerable_contract'"

```
$ python reentrancy_concrete.py                                                       
```

### Expected behavior

It shouldn't fail.

### Actual behavior

```
[+] Setup the exploit
  File "reentrancy_concrete.py", line 92, in <module>
    exploit_account.set_vulnerable_contract(contract_account)
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/ethereum.py", line 479, in __getattribute__
    return object.__getattribute__(self, name)            
AttributeError: 'EVMAccount' object has no attribute 'set_vulnerable_contract'
```

feliam:
Set Solitidty Metadata before EVMAccount constructor.
Make EVMAccount name solver lazy

This pull request fixes #674

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -2,7 +2,7 @@ import string
 
 from . import Manticore
 from .core.smtlib import ConstraintSet, Operators, solver, issymbolic, Array, Expression, Constant
-from .core.smtlib.visitors import arithmetic_simplifier
+from .core.smtlib.visitors import arithmetanticore/core/smtlib/solver.pyic_simplifier
 from .platforms import evm
 from .core.state import State
 import tempfile
@@ -763,7 +763,7 @@ class ManticoreEVM(Manticore):
         init_bytecode = compile_results[2]
 
         if address is None:
-            address = self.world._new_address()
+            address = self.world.new_address()
 
         #FIXME different states "could"(VERY UNLIKELY) have different contracts 
         # asociated with the same address
@@ -792,7 +792,7 @@ class ManticoreEVM(Manticore):
             assert context['_pending_transaction'] is None
         assert init is not None
         if address is None:
-            address = self.world._new_address()
+            address = self.world.new_address()
         with self.locked_context('seth') as context:
             context['_pending_transaction'] = ('CREATE_CONTRACT', owner, address, balance, init)
 
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -2,7 +2,7 @@ import string
 
 from . import Manticore
 from .core.smtlib import ConstraintSet, Operators, solver, issymbolic, Array, Expression, Constant
-from .core.smtlib.visitors import arithmetanticore/core/smtlib/solver.pyic_simplifier
+from .core.smtlib.visitors import arithmetic_simplifier
 from .platforms import evm
 from .core.state import State
 import tempfile
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -9,7 +9,6 @@ import tempfile
 from subprocess import Popen, PIPE
 import sha3
 import json
-import types
 import logging
 import StringIO
 import cPickle as pickle
@@ -443,6 +442,10 @@ class EVMAccount(object):
     def address(self):
         return self._address
 
+    def _null_func():
+        pass
+
+
     def _init_hashes(self):
         #initializes self._hashes lazy
         if self._hashes is None and self._seth is not None:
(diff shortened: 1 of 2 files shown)


manticore/ethereum.py
```diff
@@ -755,15 +762,18 @@ def solidity_create_contract(self, source_code, owner, balance=0, address=None,
         compile_results = self._compile(source_code)
         init_bytecode = compile_results[2]
 
+        if address is None:
+            address = self.world._new_address()
```
mossberg: let's remove the underscore prefix in `world._new_address` since it's a public func?

manticore/ethereum.py
```diff
@@ -448,6 +443,17 @@ def __str__(self):
...
+    def _init_hashes(self):
+        #initializes self._hashes lazy
+        if self._hashes is None and self._seth is not None:
+            self._hashes = {}
+            md = self._seth.get_metadata(self._address)
+            if md is not None:
+                for signature, func_id in md.hashes.items():
+                    func_name = str(signature.split('(')[0])
+                    self._hashes[func_name] = signature, func_id
+            self._init_hashes = types.MethodType(lambda *args: None, self)
```
mossberg: is this line necessary? this feels pretty hacky to me, and i think we could omit it?

mossberg:
can you briefly describe the metadata setting race problem being fixed here? the code seems good, but i don't understand the context.

manticore/ethereum.py
```diff
@@ -458,25 +464,27 @@ def __getattribute__(self, name):
...
-                                        value=value,
-                                        data=tx_data
-                                     )
-                self._caller = None
-                self._value = 0
-            return f
-        else:
-            return object.__getattribute__(self, name)            
+        if not name.startswith('_'):
+            self._init_hashes()
```
mossberg: won't this call always be a nop since, the ctor will execute `self._init_hashes()` before this?

manticore/ethereum.py
feliam: It's faster if we keep it and slower if we remove it.
I think there is an alternative without lambda/Methodtype. Maybe you prefer this: 
 
```
def _null_func():
    pass

self._init_hashes = self._null_func 
```
 ??

manticore/ethereum.py
feliam: The ctor will execute this before indeed. But _init_hashes may not had the expected effect in that first call.

manticore/ethereum.py
mossberg: is it a micro-optimization? it's a tradeoff between speed and simplicity/readability of the code. I think we should strive towards a cleaner/simpler codebase unless the optimization _really_ matters.

just wanted to comment though, i'll leave this to your judgement! :)

manticore/ethereum.py
mossberg: hm ok, this is part of the context i lack. can you elaborate?

ggrieco-tob: I don't know if it helps, but this pull request fix https://github.com/trailofbits/manticore/issues/674

manticore/ethereum.py
```diff
@@ -448,6 +443,17 @@ def __str__(self):
     def address(self):
         return self._address
 
+    def _init_hashes(self):
+        #initializes self._hashes lazy
+        if self._hashes is None and self._seth is not None:
+            self._hashes = {}
+            md = self._seth.get_metadata(self._address)
+            if md is not None:
```
feliam: _init_hashes will take effect only if metadata is defined. 
It may happen (happened before) that the account is created but the metadata is not yet set. So we need to delay the hashes initialization for later. ?

manticore/ethereum.py
mossberg: oh definitely, that should be noted in the pr description

manticore/ethereum.py
feliam: The most common complaint is about speed.
It's a pretty simple optimization that also captures the idea of it being a single execution.
I'll add some comment to improve readability and keep it.

manticore/ethereum.py
feliam: Ideally it should not be public. This qualifies as tech debt

manticore/ethereum.py
```diff
@@ -2,7 +2,7 @@
 
 from . import Manticore
 from .core.smtlib import ConstraintSet, Operators, solver, issymbolic, Array, Expression, Constant
-from .core.smtlib.visitors import arithmetic_simplifier
+from .core.smtlib.visitors import arithmetanticore/core/smtlib/solver.pyic_simplifier
```
mossberg: yikes lol

Issue #678, opened 2017-12-28
Datatype bytes32 is not supported


ggrieco-tob:
### OS / Environment

Distributor ID:	ArchLinux
Description:	ArchLinux
Release:	rolling-release

### Manticore version

Last revision (4c0f4bc6cf7509eb0f3399327fb494188b137671)

### Python version

Python 2.7.13

### Summary of the problem

Loading a simple contract triggers an NotImplemented error because bytes32 type is not supported.

### Step to reproduce the behavior

This is a small solidity contract to reproduce this issue:

```
contract C {
    function f() public returns(bytes32){
        return "abc";
    }
}
```

You can trigger this exception using the manticore cli:

```
$ manticore error.sol                                                                
```

### Expected behavior

I expect manticore to support the bytes32 type.

### Actual behavior

```
2017-12-28 13:04:24,028: [24111] m.main:INFO: Beginning analysis
2017-12-28 13:04:24,067: [24111] m.ethereum:INFO: Starting symbolic transaction: 1
2017-12-28 13:04:24,324: [24111] m.ethereum:INFO: Generated testcase No. 0 - REVERT
2017-12-28 13:04:25,787: [24111] m.ethereum:INFO: Generated testcase No. 1 - REVERT
2017-12-28 13:04:27,630: [24111] m.ethereum:INFO: Generated testcase No. 2 - REVERT
2017-12-28 13:04:29,534: [24111] m.ethereum:INFO: Generated testcase No. 3 - REVERT
2017-12-28 13:04:30,857: [24111] m.ethereum:INFO: Finished symbolic transaction: 1 | Code Coverage: 100% | Terminated States: 4 | Alive States: 1
2017-12-28 13:04:30,857: [24111] m.ethereum:INFO: Generated testcase No. 4 - RETURN
Traceback (most recent call last):
  File "/home/g/.local/bin/manticore", line 11, in <module>
    load_entry_point('manticore==0.1.6', 'console_scripts', 'manticore')()
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/__main__.py", line 104, in main
    ethereum_cli(args)
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/__main__.py", line 94, in ethereum_cli
    m.multi_tx_analysis(args.argv[0], args.txlimit)
[...]

PR #680, opened 2017-12-28
Adds support for unmarshaling bytes32 type

feliam:
Fixes https://github.com/trailofbits/manticore/issues/678

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -376,9 +376,8 @@ class ABI(object):
             dyn_offset = 4 + get_uint(256,offset)
             size = get_uint(256, dyn_offset)
             return data[dyn_offset+32:dyn_offset+32+size], offset+4
-        elif ty.startswith('bytes') and int(ty[5:]) in range(33):
+        elif ty.startswith('bytes') and 0 <= int(ty[5:]) <= 32:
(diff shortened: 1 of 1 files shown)


manticore/ethereum.py
```diff
@@ -376,6 +376,10 @@ def simplify(x):
             dyn_offset = 4 + get_uint(256,offset)
             size = get_uint(256, dyn_offset)
             return data[dyn_offset+32:dyn_offset+32+size], offset+4
+        elif ty.startswith('bytes') and int(ty[5:]) in range(33):
+            size = int(ty[5:])
+            assert size <= 32
```
mossberg: doesn't the condition guarantee this?

manticore/ethereum.py
```diff
@@ -376,6 +376,10 @@ def simplify(x):
             dyn_offset = 4 + get_uint(256,offset)
             size = get_uint(256, dyn_offset)
             return data[dyn_offset+32:dyn_offset+32+size], offset+4
+        elif ty.startswith('bytes') and int(ty[5:]) in range(33):
```
mossberg: also consider `0 <= int(ty[5:]) <= 32`

PR #697, opened 2018-01-08
Added multiprocessing support to generate testcases in finalize

ggrieco-tob:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/697)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -8,6 +8,7 @@ from .core.state import State
 import tempfile
 from subprocess import Popen, PIPE
 from multiprocessing import Process, Queue
+from Queue import Empty as EmptyQueue
 import sha3
 import json
 import logging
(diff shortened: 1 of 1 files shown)


feliam:
```
        from multiprocessing import Process, Queue
        N=10
        q = Queue()
        map(q.put, self.running_state_ids)
        def f(q):
            state_id = q.get_nowait()
            while state_id is not None:
                self.terminate_state_id(state_id)
                state_id = q.get_nowait()
        ps = [ Process(target=f)] * N
        ( p.start() for p in ps )
        ( p.join() for p in ps )              

```

ggrieco-tob:
Under certain circumstances (I'm not sure which ones), this code causes manticore to try to generate test case number `-1` (and it fails)

manticore/ethereum.py
```diff
@@ -1213,10 +1215,26 @@ def flagged(flag):
...
+        ps = []
 
+        for _ in range(self._config_procs):
+            p = Process(target=f, args=(q,))
+            p.start()
+            ps.append(p)
+
+        for p in ps:
+            p.join()            
+
```
feliam:
```
        q = Queue()
        map(q.put, self.running_state_ids)
        def f(q):
            try:
                while True:
                    state_id = q.get_nowait()
                    self.terminate_state_id(state_id)
            except EmptyQueue:
                pass

        ps = [ Process(target=f, args=(q,)) for _ in range(self._config_procs)]
        for p in ps:
            p.start()

        for p in ps:
            p.join()   
```
 ¯\_(ツ)_/¯

Issue #711, opened 2018-01-15
Manticore does not support (un)serialization of address[] parameters


ggrieco-tob:
### OS / Environment

Distributor ID:	ArchLinux
Description:	ArchLinux
Release:	rolling-release

### Manticore version

Last revision (465c926a2e39d5daf1405711dac61b6ddd0a2527)

### Python version

Python 2.7.13

### Summary of the problem

Manticore does not support (un)serialization of `address[]` parameters.

### Step to reproduce the behavior

This is a small solidity contract to reproduce this issue:

```
contract C {

  function f(address[] x) returns (address[]) {
    return x;
  }
}
```

You can trigger this issue using the manticore cli:

```
$ manticore error.sol                                                                
```

### Expected behavior

It shouldn't fail.

### Actual behavior

```
2018-01-15 07:46:30,270: [16648] m.main:INFO: Beginning analysis
2018-01-15 07:46:30,330: [16648] m.ethereum:INFO: Starting symbolic transaction: 1
2018-01-15 07:46:30,723: [16648] m.ethereum:INFO: Generated testcase No. 0 - REVERT
2018-01-15 07:46:31,632: [16648] m.c.executor:ERROR: Exception: address[]
Traceback (most recent call last):
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/core/executor.py", line 486, in run
    self.generate_testcase(current_state, str(e))
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/core/executor.py", line 354, in generate_testcase
    self._publish('will_generate_testcase', state, 'test', message)
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/utils/event.py", line 117, in _publish
    self._publish_impl(_name, *args, **kwargs)
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/utils/event.py", line 133, in _publish_impl
    sink._publish_impl(_name, *args, **kwargs)
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/utils/event.py", line 125, in _publish_impl
    callback(robj(), *args, **kwargs)
[...]

PR #715, opened 2018-01-17
Reporting something on free symbolic dynamic arguments

### LINKED ISSUE #673: AssertionError: No forking yet
### OS / Environment

Distributor ID:	ArchLinux
Description:	ArchLinux
Release:	rolling-release

### Manticore version

Last revision (4c0f4bc6cf7509eb0f3399327fb494188b137671)

### Python version

Python 2.7.13

### Summary of the problem

Loading a simple contract triggers an assertion error: No forking yet

This is a small solidity contract to reproduce this issue:

```
contract C {

    function C(address[] x) {
        throw;
    }
}
```

You can trigger this assertion using the manticore cli:

```
$ manticore error.sol                                                                
```

### Expected behavior

It shouldn't fail.

### Actual behavior

```
2017-12-27 12:36:06,314: [11125] m.main:INFO: Beginning analysis
2017-12-27 12:36:06,342: [11125] m.ethereum:INFO: Generated testcase No. 0 - REVERT
  File "/home/g/.local/bin/manticore", line 11, in <module>
    load_entry_point('manticore==0.1.6', 'console_scripts', 'manticore')()
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/__main__.py", line 104, in main
    ethereum_cli(args)
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/__main__.py", line 94, in ethereum_cli
    m.multi_tx_analysis(args.argv[0], args.txlimit)
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/ethereum.py", line 850, in multi_tx_analysis
    attacker_account = self.create_account(balance=1000)
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/ethereum.py", line 804, in create_account
    assert len(self.running_state_ids) == 1, "No forking yet"
AssertionError: No forking yet
```

feliam:
This fixes the reporting over symbolic dynamic arguments.

The problem with fully symbolic ABI encodings is the dynamic types.
On arguments or return strings you can have dynamic arguments like "bytes[]" of which you do not know the size in advance. So the ABI writes the size somewhere in the encoding and puts the whole dynamic argument someplace in the in the string. Normally at the end of it. And in place of the argument there is a pointer/offset to the actual argument in string.

A fully symbolic input or return string encoding with this ABI and with dynamic arguments will contain symbolic OFFSETS and SIZES. So a given dynamic argument could have any size and be actually at any offset of the string. This situation is basically unsolvable for any real input string size.

This patch is a start in the effort to concretize a ABI encoded symbolic string in a sensible way.
(Ideally we should be able to reuse the code here to pre-contrain the input string at the beginning of the analisys  if needed. So we do not have to explore all different sizes/offsets of each dynamic argument.)

https://github.com/trailofbits/manticore/issues/711

Fixes #673

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/715)


manticore/ethereum.py
```diff
@@ -891,6 +932,8 @@ def run_symbolic_tx():
 
             if not found_new_coverage:
                 break
+            if not self.running_state_ids:
```
mossberg: this is an optimization or a bug fix?

manticore/platforms/evm.py
```diff
@@ -1543,6 +1556,11 @@ def CALLDATALOAD(self, offset):
             #self._constraints.add(0 == offset%32)
         #    raise ConcretizeStack(3, expression=offset, policy='ALL')
 
+        #FIXME Document and make it configurable
```
mossberg: rm above comments?

manticore/ethereum.py
```diff
@@ -1078,6 +1123,38 @@ def global_findings(self):
     def workspace(self):
         return self._executor._workspace._store.uri
 
+    def _concretize_offsets_and_sizes(self, state, signature, data):
```
mossberg: i'm having a hard time understanding this function. is it necessary so we can do `state.solve_one` on the arguments returned from ABI.parse() later on?
is there relevant ABI documentation on encoding of dynamic arguments we can link to here?

manticore/ethereum.py
```diff
@@ -1078,6 +1123,38 @@ def global_findings(self):
...
+            if dynamic:
+                dyn_arguments.append(pos)
+
+        used = 4 + len(types)*32 
+        available = len(data) -used -len(dyn_arguments)*32
+
+        for i in range(0, len(dyn_arguments)):
+            #get, constraint and concretize dyn_offset to some reasonable value
+            dyn_offset = ABI.get_uint(data, 4 + dyn_arguments[i]*32, 256)
+            assert solver.can_be_true(state.constraints, dyn_offset == used)
```
mossberg: not sure this is the correct use of `assert`?

manticore/ethereum.py
feliam: Maybe this helps. Normally the ABI encoding looks like this

FUNCID, ARG0, ARG1, ARG2
sizeof(FUNCID) := 4
sizeof(ARG?) := 32

But if one of the arguments, let say ARG1, is a dynamic argument it uses more than 32 bytes. Then it'll be something like this 
```
FUNCID, ARG0, ARG1p, ARG2, ARG0N, ARG0.0,ARG00.1,,, ARG0.N
            \______________/\
sizeof(FUNCID) := 4
sizeof(ARG?) := 32
ARG1p is the offset where the actual ARG1 is in the input. Measured in bytes from the start of the arguments block. 
ARG0N is the actual size of ARG1 meassured in number of elements

ref. http://solidity.readthedocs.io/en/develop/abi-spec.html CTRL+F "Given the contract"
```

mossberg:
relevant link: http://solidity.readthedocs.io/en/develop/abi-spec.html#use-of-dynamic-types

manticore/ethereum.py
```diff
@@ -342,6 +342,22 @@ def check_bitsize(value, size):
         result.append(ABI.make_function_arguments(*args))
         return reduce(lambda x,y: x+y, result)
 
+    @staticmethod
+    def get_uint(data, offset, size):
```
mossberg: what is this function supposed to do? if it reads a uint out of the data @ offset, why is there a size param?

mossberg: oh, i see now, `uint` in the function name doesn't mean ethereum 256 uint

feliam: So. uint of different sizes. 
if 'address' -> uint160 -> size:=20.
We need to document it.

mossberg: i'm working on documentation/cleanup PR to help my understanding :)

mossberg: should we change the `size` parameter to be in bytes rather than bits? `size` should always be a multiple of 8, and if it is not, its an error.

feliam: Oh, size may be bits.

manticore/ethereum.py
```diff
@@ -1078,6 +1123,38 @@ def global_findings(self):
...
+                dyn_arguments.append(pos)
+
+        used = 4 + len(types)*32 
+        available = len(data) -used -len(dyn_arguments)*32
+
+        for i in range(0, len(dyn_arguments)):
+            #get, constraint and concretize dyn_offset to some reasonable value
+            dyn_offset = ABI.get_uint(data, 4 + dyn_arguments[i]*32, 256)
+            assert solver.can_be_true(state.constraints, dyn_offset == used)
+            state.constrain(dyn_offset == used)
```
mossberg: i think this is a bug. `dyn_offset` should be the offset `from the start of the value encoding (i.e. not counting the first four bytes containing the hash of the function signature)` and it does not count the first four bytes. however when we constraint it to `used`, `used` accounts for the first 4 bytes. so dyn_offset is too great by 4

http://solidity.readthedocs.io/en/develop/abi-spec.html#use-of-dynamic-types

feliam: Uhh! I think you are correct. Nice catch.

mossberg: we concretize dyn_offset by adding a constraint to a constant value, and we concretize `data[used:used+32]` (the size of the dynamic element) by directly assigning it.

shouldn't we prefer the former method? when we do the latter, we are erasing other data. that other data potentially could have been symbolic data with constraints on it?

manticore/ethereum.py
```diff
@@ -1078,6 +1123,38 @@ def global_findings(self):
...
+            data[used:used+32] = ("%064x"%(available/32/len(dyn_arguments))).decode('hex')
+
+            #get, constraint and concretize dyn_size to some reasonable value
+            dyn_size = ABI.get_uint(data, used, 256)
```
mossberg: i don't understand this block. i thought the above block concretizes the offset and the size, and then we're done for this argument?

feliam: First you add a constraint to the path constrraint. 
constraints.add(data[10:20] == 'AAAA..AA')
Then you overwrite the state (data) with the concrete value.
data[10:20] = 'AAAA..AA'
So next time you read data[10:20] you get a constant value and not a symbolic variable with a single solution. ?

manticore/ethereum.py
```diff
@@ -342,6 +342,22 @@ def check_bitsize(value, size):
...
+                return value.value
+            else:
+                return value
+
+        size = simplify(size)
+        offset = simplify(offset)
+        byte_size = size/8
+        padding = 32 - byte_size # for 160
+        value = arithmetic_simplifier(Operators.CONCAT(size, *map(Operators.ORD, data[offset+padding:offset+padding+byte_size])))
+        return simplify(value)
```
mossberg: does this simplify twice? redundant?

manticore/ethereum.py
```diff
@@ -342,6 +342,22 @@ def check_bitsize(value, size):
         result.append(ABI.make_function_arguments(*args))
         return reduce(lambda x,y: x+y, result)
 
+    @staticmethod
+    def get_uint(data, offset, size):
+        def simplify(x):
```
mossberg: we shouldn't copypaste this everywhere, let's just modify `arithmetic_simplifier`. while we're at it, rename `arithmetic_simplifier` to `arithmetic_simplify`? currently it sounds like it creates a new Simplifier object or something

manticore/ethereum.py
mossberg: let's also delete the internal `get_uint` used in `_consume_type`!

manticore/ethereum.py
mossberg: sorry for being a bit slow, trying to wrap my head around concretizing a variable :x 😬 

would it be correct to say that overwriting the state with the concrete value is an optimization? and technically all that is necessary for _correctness_ is to constrain the variable? not saying we shouldn't do it, but just checking

manticore/ethereum.py
mossberg: when would you want to read a size that is not divisible by 8?

manticore/ethereum.py
feliam: Yes, writing the concrete value to the state (after constraining it to a single value) is somehow redundant.

mossberg:
Review status: 0 of 4 files reviewed at latest revision, 15 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DTm2P6-B9UAzLd4xJ:-L4DTm2P6-B9UAzLd4xK:b1607w5) (no related file):*
test, second reviewable discussion

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

mossberg:
Review status: 0 of 4 files reviewed at latest revision, 15 unresolved discussions.

---

*[manticore/ethereum.py, line 1150 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-1150:-L4DYa9qc3bwkaKnOgrY:bb2cut2) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L1150)):*

Yes, writing the concrete value to the state (after constraining it to a single value) is somehow redundant. 
</blockquote></details>

I wouldn't say redundant, it's effect is to improve performance. Later on, when the variable is read, rather than it being symbolic (even though its constrained) and potentially making lots of other things symbolic, which is generally bad for perf, we have a nice clean concrete variable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

feliam:
Also and in general... when replacing a symbolic fully constrained value with its concrete counterpart we should also be careful not to loose any taint.

mossberg:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DTm2P6-B9UAzLd4xJ:-L4DZb0oaCyS4WOj6eE4:b32wtd) (no related file):*

test, second reviewable discussion
</blockquote></details>

ok

---

*[manticore/ethereum.py, line 1150 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-1150:-L4DZRXrpyOpJt7NXeUc:b-lcg782) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L1150)):*

I wouldn't say redundant, it's effect is to improve performance. Later on, when the variable is read, rather than it being symbolic (even though its constrained) and potentially making lots of other things symbolic, which is generally bad for perf, we have a nice clean concrete variable.
</blockquote></details>

ok, resolved though

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

mossberg:
Review status: 3 of 4 files reviewed at latest revision, 16 unresolved discussions.

---

*[manticore/ethereum.py, line 346 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-346:-L4IBhIwwQqK_8xLKFwE:b-k7sk4p) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L346)):*

when would you want to read a size that is not divisible by 8?
</blockquote></details>

*read a bits size

---

*[manticore/ethereum.py, line 1146 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-1146:-L4ID2xONe2tjCYxPe_A:b-z95xhw) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L1146)):*

Uhh! I think you are correct. Nice catch.
</blockquote></details>

Also, I think it should be 

```
state.constrain(dyn_offset == used)
data[4 + dyn_arguments[i]*32 : 4 + dyn_arguments[i]*32 + 32] = ... # not data[used:used+32]
```

right now we are constraining the offset, and then overwriting size. we should constrain the offset and overwrite the offset together

---

*[manticore/platforms/evm.py, line 1562 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4IDkpLPG8Zh2j_33wE:-L4IDkpLPG8Zh2j_33wF:b-ued5rz) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/platforms/evm.py#L1562)):*
> ```Python
>         #Do not explore offsets much bigger than actual data
>         if issymbolic(offset):
>             self.constraints.add(offset<len(self.data)+32)
> ```

why `+32`?  shouldn't the offset be within the bounds of the data?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

mossberg:
Review status: 3 of 4 files reviewed at latest revision, 15 unresolved discussions.

---

*[manticore/ethereum.py, line 1142 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-1142:-L4NNN1gOh_D6WAtmym6:brz56qc) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L1142)):*

`enumerate`?
</blockquote></details>

on second thought, why can't we simply iterate, `for index in dyn_arguments`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

mossberg:
Reviewed 1 of 4 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

manticore/ethereum.py
```diff
@@ -373,33 +398,57 @@ def simplify(x):
...
         elif ty.startswith('bytes') and 0 <= int(ty[5:]) <= 32:
             size = int(ty[5:])
             return data[offset:offset+size], offset+32
+        if ty == u'address[]':
+            dyn_offset = arithmetic_simplifier((get_uint(256,offset)))
+            size = arithmetic_simplifier(get_uint(256, dyn_offset))
+            result = []
+            for i in range(size):
+                x = get_uint(160, dyn_offset+32 + 32*i)
+                result.append(x)
```
disconnect3d: Or just:
```python
result = [get_uint(160, dyn_offset+32 + 32*i) for i in range(size)]
```

feliam:
*[manticore/ethereum.py, line 935 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-935:-L5nPyX773jlIWV3cv6C:b-g90y3e) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L935)):*

this is an optimization or a bug fix?
</blockquote></details>

I think it's a bugfix.  If all states got to an end then trying a new tx is a bug. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

feliam:
*[manticore/platforms/evm.py, line 1562 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4IDkpLPG8Zh2j_33wE:-L5pTPRQqrdtGnMjBoLq:bncprvo) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/platforms/evm.py#L1562)):*
>  #self._constraints.add(0 == offset%32)
No. It does not need to be within the bounds.  It'll just return a zero when outside the boundaries.  The 32 is an author choice so it receives some out of boundary zeros but not so many. This should be generalized better.  
In the real world it is actually constrained via gas consumption.
It's a FIXME, THINKME .

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

mossberg:
Reviewed 6 of 7 files at r3.
Review status: 5 of 8 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[manticore/platforms/evm.py, line 1562 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4IDkpLPG8Zh2j_33wE:-L5puZz4ROM6lHLJfTxL:b-dbli5) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/platforms/evm.py#L1562)):*

>  #self._constraints.add(0 == offset%32)
No. It does not need to be within the bounds.  It'll just return a zero when outside the boundaries.  The 32 is an author choice so it receives some out of boundary zeros but not so many. This should be generalized better.  
In the real world it is actually constrained via gas consumption.
It's a FIXME, THINKME .


</blockquote></details>

Ok, i created https://github.com/trailofbits/manticore/wiki/Analysis-Design-Choices to document stuff like this

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

feliam:
Review status: 5 of 8 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[manticore/ethereum.py, line 1142 at r1](https://beta.reviewable.io/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-1142:-L5srUbiAJQaE-cTKSca:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L1142)):*

on second thought, why can't we simply iterate, `for index in dyn_arguments`
</blockquote></details>

Done.

---

*[manticore/ethereum.py, line 1145 at r1](https://beta.reviewable.io/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-1145:-L5srVTf5KaXe6qrVCij:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L1145)):*

not sure this is the correct use of `assert`?
</blockquote></details>

Done.

---

*[manticore/ethereum.py, line 413 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r2-413:-L5sr1xS5U5i8g6OVb11:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/9a5c8355528625018fac38000c7e17e525125ef6/manticore/ethereum.py#L413)):*
<details><summary><i>Previously, disconnect3d (Disconnect3d) wrote…</i></summary><blockquote>

Or just:
```python
result = [get_uint(160, dyn_offset+32 + 32*i) for i in range(size)]
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/715)*

mossberg:
Reviewed 2 of 3 files at r4.
Review status: 7 of 8 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[manticore/ethereum.py, line 346 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L5tYIvzP8IoHN_qV8Wc:-L5tYIvzP8IoHN_qV8Wd:b-7uk0vs) ([raw file](https://github.com/trailofbits/manticore/blob/a1792f4077247ca800daefbe8de30f823bb1c2e6/manticore/ethereum.py#L346)):*
> ```Python
> 
>     @staticmethod
>     def get_uint(data, offset, byte_size):
> ```

can we use the variable name `nbytes`? we use it elsewhere and i think it's pretty clear

---

*[manticore/ethereum.py, line 353 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L5tXPpdOCAa-H6kZicc:-L5tXPpdOCAa-H6kZicd:b-d6cigf) ([raw file](https://github.com/trailofbits/manticore/blob/a1792f4077247ca800daefbe8de30f823bb1c2e6/manticore/ethereum.py#L353)):*
> ```Python
>         :param data: sliceable buffer; symbolic buffer of Eth ABI encoded data
>         :param offset: byte offset of 256 bit element
>         :param size: number of bits to read out of 256 bit element
> ```

great, we need to update this docstring also to not say "number of bits" -> "number of bytes to read out of 256 bit element, starting from least significant byte"

---

*[manticore/ethereum.py, line 386 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L5tYUx_MKGPGIekoMNm:-L5tYUx_MKGPGIekoMNn:bsl7jzv) ([raw file](https://github.com/trailofbits/manticore/blob/a1792f4077247ca800daefbe8de30f823bb1c2e6/manticore/ethereum.py#L386)):*
> ```Python
>             return data[offset:offset+size], offset+32
>         if ty == u'address[]':
>             dyn_offset = arithmetic_arithmetic_simplify((get_uint(256,offset)))
> ```

maybe a bad replace all here and around?

---

*[manticore/core/smtlib/visitors.py, line 477 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L5tZf60lcLE9JwehTzJ:-L5tZf60lcLE9JwehTzK:ba9p9br) ([raw file](https://github.com/trailofbits/manticore/blob/a1792f4077247ca800daefbe8de30f823bb1c2e6/manticore/core/smtlib/visitors.py#L477)):*
> ```Python
> 
> 
> def to_constant(expression):
> ```

i think this is a little misleading because this function does not necessarily return a constant. after simplifying, if it's not a constant, or it has taint, it will still return an expression. also i think some of the other refactors depended on this code being included in `arithmetic_simplify`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

feliam:
Review status: 7 of 8 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[manticore/ethereum.py, line 1146 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-1146:-L5ttWoPwjaqmBnpVpVG:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L1146)):*

Also, I think it should be 

```
state.constrain(dyn_offset == used)
data[4 + dyn_arguments[i]*32 : 4 + dyn_arguments[i]*32 + 32] = ... # not data[used:used+32]
```

right now we are constraining the offset, and then overwriting size. we should constrain the offset and overwrite the offset together
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

mossberg:
Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[manticore/ethereum.py, line 935 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-935:-L5uHU4goawvsnoQW05c:b-mx2ion) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L935)):*

I think it's a bugfix.  If all states got to an end then trying a new tx is a bug. 
</blockquote></details>

I feel like this should be handled in `ManticoreEVM.run()` rather than `multi_tx_analysis`, which I view as a very high level function that is basically client code and not really part of the Manticore core. If there are no running states, and an API used called `self.transaction`, we would also want to catch the error.

Can we modify `ManticoreEVM.run` to detect if there are no running states, and no initial state, and then just return if so? or maybe just have it put nothing in the executor queue, so the executor naturally does nothing? i think that would be the cleanest solution.

---

*[manticore/ethereum.py, line 1145 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-1145:-L5uSnc77AuBazSaJruM:b-1txchp) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L1145)):*

Done.
</blockquote></details>

i still think it would be good to check that this can be true, but i just didn't think AssertionError is the best way to handle the error. i'm not totally sure the best way though...we could completely bail, we could try a different offset, we could not attempt to concretize this data?

---

*[manticore/ethereum.py, line 1125 at r5](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L5uKCy6d6y842L4pTti:-L5uKCy6d6y842L4pTtj:b-u2lhjj) ([raw file](https://github.com/trailofbits/manticore/blob/fdef5ef6cf621cd7c071c4ac806cd3eaed5f2266/manticore/ethereum.py#L1125)):*
> ```Python
>         for index in dyn_arguments:
>             #get, constraint and concretize dyn_offset to some reasonable value
>             dyn_offset = ABI.get_uint(data, 4 + index*32, 256)
> ```

need to update the 256 here, and below

---

*[manticore/ethereum.py, line 1126 at r5](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L5uL5RBIANoFczQ2aBp:-L5uL5RBIANoFczQ2aBq:b-wqa2dc) ([raw file](https://github.com/trailofbits/manticore/blob/fdef5ef6cf621cd7c071c4ac806cd3eaed5f2266/manticore/ethereum.py#L1126)):*
> ```Python
>             #get, constraint and concretize dyn_offset to some reasonable value
>             dyn_offset = ABI.get_uint(data, 4 + index*32, 256)
>             state.constrain(dyn_offset+4 == free_pointer)
> ```

i think `dyn_offset + 4 == free_pointer` will produce a slightly more complex expression object than `dyn_offset == free_pointer - 4`, because of symbolic add. i think we should keep the expression objects as simple as possible, so consider the latter?

---

*[manticore/ethereum.py, line 1132 at r5](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L5uQyUySakODVwR4yj_:-L5uQyUySakODVwR4yja:bfe38w9) ([raw file](https://github.com/trailofbits/manticore/blob/fdef5ef6cf621cd7c071c4ac806cd3eaed5f2266/manticore/ethereum.py#L1132)):*
> ```Python
>             #get, constraint and concretize dyn_size to some reasonable value
>             dyn_size = ABI.get_uint(data, free_pointer, 256)
>             state.constrain(dyn_size == argument_size/32)
> ```

`argument_size/32` only applies to some types of dynamic arguments. for example, for `bytes` this field contains the number of bytes of the string. it seems to me like we need to also be aware of the type being encoded here to be correct?

http://solidity.readthedocs.io/en/develop/abi-spec.html#use-of-dynamic-types

---

*[manticore/ethereum.py, line 1137 at r5](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L5uSe8lzvh4PNxEp05g:-L5uSe8lzvh4PNxEp05h:b62ixkv) ([raw file](https://github.com/trailofbits/manticore/blob/fdef5ef6cf621cd7c071c4ac806cd3eaed5f2266/manticore/ethereum.py#L1137)):*
> ```Python
> 
>             free_pointer += 32 + argument_size
>             #free_pointer points to the first unused byte in the dinamic argument area
> ```

nit, sp "dinamic" "dynamic"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

mossberg:
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[manticore/ethereum.py, line 385 at r5](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L5ugXS4cLcfYfiwEOTv:-L5ugXS4cLcfYfiwEOTw:b-s6xlvj) ([raw file](https://github.com/trailofbits/manticore/blob/fdef5ef6cf621cd7c071c4ac806cd3eaed5f2266/manticore/ethereum.py#L385)):*
> ```Python
>             return data[offset:offset+size], offset+32
>         if ty == u'address[]':
>             dyn_offset = arithmetic_simplify((get_uint(256,offset)))
> ```

i think this needs to be ABI.get_uint, also watch out for the 256

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

manticore/ethereum.py
```diff
@@ -342,64 +342,78 @@ def check_bitsize(value, size):
...
-            return data[dyn_offset+32:dyn_offset+32+size], offset+4
+            dyn_offset = ABI.get_uint(data, 32,offset)  #256 bits
+            size = ABI.get_uint(data, 32, dyn_offset)  #256 bits
+            return data[dyn_offset+32:dyn_offset+32+size], offset+32
         elif ty.startswith('bytes') and 0 <= int(ty[5:]) <= 32:
             size = int(ty[5:])
             return data[offset:offset+size], offset+32
+        if ty == u'address[]':
+            dyn_offset = arithmetic_simplify((get_uint(256,offset)))
+            size = arithmetic_simplify(get_uint(256, dyn_offset))
```
ggrieco-tob: This also should use ABI.get_uint instead of get_uint?

manticore/ethereum.py
```diff
@@ -342,64 +342,78 @@ def check_bitsize(value, size):
...
-            size = get_uint(256, dyn_offset)
-            return data[dyn_offset+32:dyn_offset+32+size], offset+4
+            dyn_offset = ABI.get_uint(data, 32,offset)  #256 bits
+            size = ABI.get_uint(data, 32, dyn_offset)  #256 bits
+            return data[dyn_offset+32:dyn_offset+32+size], offset+32
         elif ty.startswith('bytes') and 0 <= int(ty[5:]) <= 32:
             size = int(ty[5:])
             return data[offset:offset+size], offset+32
+        if ty == u'address[]':
+            dyn_offset = arithmetic_simplify((get_uint(256,offset)))
```
ggrieco-tob: This should use ABI.get_uint instead of get_uint?

ggrieco-tob:
After fixing the conflicts and running the test case:

```
contract C {

  function f(address[] x) returns (address[]) {
    return x;
  }
}
```

manticore does not finish solving it in a reasonable time (more than 10 minutes at least).

mossberg:
I found some pretty serious bugs, all the callers of `ABI.get_uint` inside `ABI._consume_type` switch the `offset` and `nbytes` parameters :X 
maybe this is why the perf is so bad @ggrieco-tob!
i will start fixing things

mossberg:
I have been working on fixes etc for the this branch in #787

mossberg:
Review status: 0 of 8 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[manticore/ethereum.py, line 935 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/715#-L4DSRnNnGOsZ4a03AKM-r1-935:-L76WC6fSm6Z1kxptFIi:b-dsbmsn) ([raw file](https://github.com/trailofbits/manticore/blob/e8b02eabb900dca0c15804586472817fc28bf542/manticore/ethereum.py#L935)):*

I feel like this should be handled in `ManticoreEVM.run()` rather than `multi_tx_analysis`, which I view as a very high level function that is basically client code and not really part of the Manticore core. If there are no running states, and an API used called `self.transaction`, we would also want to catch the error.

Can we modify `ManticoreEVM.run` to detect if there are no running states, and no initial state, and then just return if so? or maybe just have it put nothing in the executor queue, so the executor naturally does nothing? i think that would be the cleanest solution.
</blockquote></details>

Since this issue has nothing to do with dynamic arguments, in my branch (#787) I am going to revert this change  and address in a different one.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/715)*

mossberg:
Ok, so #787 was merged which has a lot of initial fixes for this branch, but it's not complete yet. This branch will be mergable once manticore can run the above contract: https://github.com/trailofbits/manticore/pull/715#issuecomment-369381828

I've started debugging this, and ran into a couple issues
- fairly trivial crash where we pass `range()` a symbolic piece of data. easily fixed by calling `to_constant` on that sym data beforehand, since that data should be constrained to a single value by that point
- issue with CALLDATACOPY being executed with a symbolic size operand. it's a crash in _consume. we omitted the _consume symbolic gas support because it introduced an unacceptable performance hit. not sure how to handle this one. i comment it out, and kept going
- most recent issue is a hang a few lines later in CALLDATACOPY here https://github.com/trailofbits/manticore/blob/dev-evm-dynamicarguments/manticore/platforms/evm.py#L1602. it hangs in the solver trying to concretize some stack variable. i've tried changing to policy to minmax and one but this does not help the hang.

mossberg:
Looked into the aforementioned hang today. One major issue is that some really excessively big expression trees are being generated by the time the calldatacopy. i mostly looked was looking at what calldataload returns. when calldataload executes, `self.data` represents the tx data. this is an ArrayProxy object, that wraps over an already massive expression tree object composed of ArrayStores and ArraySelects. It is big due to various code like this https://github.com/trailofbits/manticore/blob/master/manticore/platforms/evm.py#L2290-L2294 which loops over an ArrayProxy performing assignments. every write builds another layer on the tree, and so our trees get hundreds of layers deep immediately, based on the length of data.

then when calldataload executes, the tree grows every bigger when the line https://github.com/trailofbits/manticore/blob/master/manticore/platforms/evm.py#L1586 slices the arrayproxy'd tree. i don't totally understand what this does yet. 

the excessive size of these trees is definitely a problem, and i suspect we could definitely be more intelligent in generating our trees. however, it's not totally clear to me that fixing this problem is actually what we need to do, as the deeper problem remains that the `total_space` argument below is symbolic and currently we do eager solving. a couple thoughts for more thorough solutions

- use lazy solving instead of eager solving, as i believe #771 will allow
- preconstrain the data, if we know the function prototypes ahead of time. we would basically be doing what _concretize_offsets_and_sizes does, but before launching the tx, not after.

fixing the smt expr trees size is probably easier than those two, and we should see if that helps the issue first.

---

also, here's a little pseudocode snippet of the main evm code that's causing issues:

```
offset = read_input(4)
nelements = read_input(offset + 4) 
total_space = 32 * nelements
copy(dst, src, total_space)
```

mossberg:
here are a couple slightly simpler test programs that use dynamic arguments; i'd be fine merging this branch once mcore can analyzer any of these. however, they all involve an identical section of difficult-to-analyze code with the symbolic calldatacopy.

```
[I] mark ubuntu /m/h/c/m/dyn ❯ cat simp.sol
contract C {

  function f(address[] x){
  }
}

[I] mark ubuntu /m/h/c/m/dyn ❯ cat lendyn.sol
contract C {

    function myfunc(address[] adds) returns (uint){
        if (adds.length > 1) {
            return 1;
        } else {
            return 0;
        }
    }
}

```

mossberg:
I've pushed my debugging branch for this to dbg-dynargs-expr-size if anyone wants to play around. it should not be merged. i tracked down one issue today, the ` bytes = list(self.data[offset:offset+32])` line is pretty wasteful. all we really need to do to cast those 32 bytes out of the data into a bitvec is concat all the individual arrayselects selected from self.data, however the `list(self.data[offset:offset+32])` does a lot of elaborate, but unnecessary(?) stuff so that we end up concating arrayselects that are selecting not from the original self.data array, but a new dynamically created array populated based on the contents of self.data (and ends up being while large due to all those arraystores).

mossberg:
i solved the unsat error, it was a bug in some of my own code, not existing code. current status is i've gotten things to a point where the size calldatacopy expression is simple enough for z3 to solve it in 'ONE' mode

mossberg:
i made some progress, but since this is not quite high priority anymore, and because I suspect #771 will have many fixes to get this working also, i'm going to pause work on this.

manticore/ethereum.py
```diff
@@ -352,73 +352,99 @@ def make_function_arguments(*args):
...
+        Read a `nbytes` bytes long big endian unsigned integer from `data` starting at `offset` 
+
+        :param data: sliceable buffer; symbolic buffer of Eth ABI encoded data
+        :param offset: byte offset
+        :param nbytes: number of bytes to read starting from least significant byte
+        :rtype: int or Expression
+        """
+        nbytes = arithmetic_simplify(nbytes)
+        offset = arithmetic_simplify(offset)
+        padding = 32 - nbytes # for 160
```
disconnect3d: Why is padding 32 and what does `for 160` mean? Might be good to describe it more so that it is clear for ppl who look over this code (I am not into the topic of ethereum, maybe this is something very obvious, but still...)

manticore/core/smtlib/expression.py
```diff
@@ -708,8 +711,14 @@ def __getitem__(self, index):
         if isinstance(index, slice):
             start, stop = self._fix_index(index)
             size = self._get_size(index)
-            new_array = ArrayVariable(self.index_bits, size, self.value_bits, name='%s_b%d_e%d'%(self.name, start, stop), taint=self.taint)
-            new_array = ArrayProxy(new_array)
+
+            if isinstance(start, Expression) or isinstance(stop,Expression):
```
disconnect3d: space after `stop,` :P

mossberg:
Revisiting this. So even though there is more work to be done to get Manticore to officially run on those various contracts above that use dynamic arguments, this branch contains a lot of very useful fixes for bugs all over the place, and those changes should get into master ASAP. #855 is a slice of this PR with all of the aforementioned fixes.

mossberg:
current status of this pr: #855 extracted and merged the extraneous good parts that were also included with this pr, so now this PR is solely focused on the dynargs stuff. next steps are to continue work to make those test contracts above run successfully

feliam:
I'm closing this. This needs a complete rework using new ABI api.

PR #719, opened 2018-01-31
Use the 'force' flag for bss zeroing

yan:
This PR updates the Linux loader to use the new `force` parameter for writing data to clear the BSS section, instead of setting/resetting memory permissions.

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -839,31 +839,13 @@ class Linux(Platform):
         entry = elf_entry
         real_elf_brk = elf_brk
 
-        # We need to explicitly zero any fractional pages
-        # after the data section (i.e. bss).  This would
-        # contain the junk from the file that should not
-        # be in memory
-        #TODO:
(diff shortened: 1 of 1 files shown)


manticore/platforms/linux.py
```diff
@@ -839,31 +839,13 @@ def load(self, filename):
...
-            cpu.memory[elf_bss:elf_brk] = '\x00'*(elf_brk-elf_bss)
-        else:
-            logger.warning("Failing to zerify the trailing: elf_brk-elf_bss")
+
+        # We need to explicitly clear bss, as fractional pages will have data from the file
+        bytes_to_clear = elf_brk - elf_bss
+        if bytes_to_clear > 0:
+            logger.debug("Zeroing main elf fractional pages. From bss(%x) to brk(%x), %d bytes.",
+                    elf_bss, elf_brk, bytes_to_clear)
+            cpu.write_bytes(elf_bss, chr(0) * bytes_to_clear, force=True)
```
mossberg: nit, `chr(0)` takes me .01344 more milliseconds to grok than `\x00`, i'd let let's just use the latter

PR #722, opened 2018-02-01
Emit events for exception raising evm instructions

### LINKED ISSUE #690: evm: trace does not record terminating instructions
evm records the execution trace via a handler for the `did_execute_instruction` event. that event is emitted after the instruction implementation function returns. for some instructions like REVERT, the impl function never returns, as an exception is raised. this means that instruction is missed in the trace.

https://github.com/trailofbits/manticore/blob/743d479cbf10f40e3079d56e3f26b2d1f4e34893/manticore/ethereum.py#L1033-L1035

https://github.com/trailofbits/manticore/blob/743d479cbf10f40e3079d56e3f26b2d1f4e34893/manticore/platforms/evm.py#L1301-L1317

https://github.com/trailofbits/manticore/blob/743d479cbf10f40e3079d56e3f26b2d1f4e34893/manticore/platforms/evm.py#L1799-L1801

mossberg:
Previously, `did_evm_execute_instruction` and other events did not fire for SELFDESTRUCT, RETURN, REVERT, etc, because these instructions raise exceptions as part of their implementation. They do now, which means our traces will be more complete, and detectors will be able to watch for these instructions also.

Fixes #690

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/722)

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -1305,14 +1305,13 @@ class EVM(Eventful):
         self._publish('will_execute_instruction', self.pc, current)
         self._publish('will_evm_execute_instruction', current, arguments)
 
+        last_pc = self.pc
+        result = None
+
         def emit_did_execute_signals():
             self._publish('did_evm_execute_instruction', current, arguments, result)
             self._publish('did_execute_instruction', last_pc, self.pc, current)
 
-        last_pc = self.pc
-        result = None
-
-        #Execute
         try:
             result = implementation(*arguments)
             emit_did_execute_signals()
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -1310,6 +1310,8 @@ class EVM(Eventful):
             self._publish('did_execute_instruction', last_pc, self.pc, current)
 
         last_pc = self.pc
+        result = None
+
         #Execute
         try:
             result = implementation(*arguments)
@@ -1326,8 +1328,7 @@ class EVM(Eventful):
         except EVMException as e:
             self.last_exception = e
             if isinstance(e, EVMInstructionException):
-                # FIXME(mark) this should also publish did_execute_instruction, but that causes some crashes
-                e.on_handled = lambda: self._publish('did_evm_execute_instruction', current, arguments, None)
+                emit_did_execute_signals()
             raise
 
         #Check result (push)
@@ -2147,7 +2148,12 @@ class EVMWorld(Platform):
         return new_address
 
     def execute(self):
-        def handle_evm_instruction_exception(ex):
+        self._process_pending_transaction()
+        try:
+            if self.current is None:
+                raise TerminateState("Trying to execute an empty transaction", testcase=False)
+            self.current.execute()
+        except EVMInstructionException as ex:
             if isinstance(ex, Create):
                 self.CREATE(ex.value, ex.data)
             elif isinstance(ex, Call):
@@ -2162,13 +2168,6 @@ class EVMWorld(Platform):
                 self.SELFDESTRUCT(ex.to)
             elif isinstance(ex, Sha3):
                 self.HASH(ex.data)
-
-        self._process_pending_transaction()
-        try:
-            self.current.execute()
-        except EVMInstructionException as iex:
-            iex.on_handled()
-            handle_evm_instruction_exception(iex)
         except EVMException as e:
             self.THROW()
         except Exception:
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -2167,10 +2167,8 @@ class EVMWorld(Platform):
         try:
             self.current.execute()
         except EVMInstructionException as iex:
-            try:
-                handle_evm_instruction_exception(iex)
-            finally:
-                iex.on_handled()
+            iex.on_handled()
+            handle_evm_instruction_exception(iex)
         except EVMException as e:
             self.THROW()
         except Exception:
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -2159,21 +2159,20 @@ class EVMWorld(Platform):
             if self.current is None:
                 raise TerminateState("Trying to execute an empty transaction", testcase=False)
             self.current.execute()
-        except EVMInstructionException as ex:
-            if isinstance(ex, Create):
-                self.CREATE(ex.value, ex.data)
-            elif isinstance(ex, Call):
-                self.CALL(ex.gas, ex.to, ex.value, ex.data)
-            elif isinstance(ex, Stop):
-                self.STOP()
-            elif isinstance(ex, Return):
-                self.RETURN(ex.data)
-            elif isinstance(ex, Revert):
-                self.REVERT(ex.data)
-            elif isinstance(ex, SelfDestruct):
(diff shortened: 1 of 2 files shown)


mossberg:
note to self: I tried emitting the `did_evm_execute_instruction` event completely after the respective EVMWorld handler function finished (self.CALL, self.REVERT, self.SELFDESTRUCT, etc), however this would cause `state.platform.current` to be `None`, which was problematic. I opted to emit the event right before the EVMException is raised back to the EVMWorld, which is slightly in violation of the semantics of the name "did_..._execute" but, was a pragmatic solution, which should work.

yan:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[manticore/platforms/evm.py, line 1328 at r3](https://reviewable.io:443/reviews/trailofbits/manticore/722#-L5KKYUtfnlsz6yJSpa-:-L5KKYUtfnlsz6yJSpa0:b-cmcc1z) ([raw file](https://github.com/trailofbits/manticore/blob/5bbfc8c6275611fc41e5d46830e08d57ff44db83/manticore/platforms/evm.py#L1328)):*
> ```Python
> 
>             # Technically, this is not the right place to emit these events because the
>             # instruction hasn't execute yet; it executes in the EVM platform class (EVMWorld).
> ```

nit, "executed"

---

*[manticore/platforms/evm.py, line 1333 at r3](https://reviewable.io:443/reviews/trailofbits/manticore/722#-L5KLYhGm0GqOVmcSmVb:-L5KLYhGm0GqOVmcSmVc:b-v7jifo) ([raw file](https://github.com/trailofbits/manticore/blob/5bbfc8c6275611fc41e5d46830e08d57ff44db83/manticore/platforms/evm.py#L1333)):*
> ```Python
>             # the event before technically executing the instruction.
>             if isinstance(e, EVMInstructionException):
>                 emit_did_execute_signals()
> ```

Why not create a separate `except`? 

---

*[manticore/platforms/evm.py, line 2138 at r3](https://reviewable.io:443/reviews/trailofbits/manticore/722#-L5KJxDXf7RtbtnoqzPB:-L5KJxDYOylUoCKx2Zu9:b-5vauwk) ([raw file](https://github.com/trailofbits/manticore/blob/5bbfc8c6275611fc41e5d46830e08d57ff44db83/manticore/platforms/evm.py#L2138)):*
> ```Python
>                 raise TerminateState("Trying to execute an empty transaction", testcase=False)
>             self.current.execute()
>         except EVMInstructionException as ex:
> ```

Can you help me understand how this way of handling is different from the old one?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/722)*

mossberg:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[manticore/platforms/evm.py, line 1328 at r3](https://reviewable.io:443/reviews/trailofbits/manticore/722#-L5KKYUtfnlsz6yJSpa-:-L5LC6Zifh5_zLh66vaB:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/5bbfc8c6275611fc41e5d46830e08d57ff44db83/manticore/platforms/evm.py#L1328)):*

nit, "executed"
</blockquote></details>

Done.

---

*[manticore/platforms/evm.py, line 1333 at r3](https://reviewable.io:443/reviews/trailofbits/manticore/722#-L5KLYhGm0GqOVmcSmVb:-L5LBCnP3zJaDhGxsR0q:bjq0tr6) ([raw file](https://github.com/trailofbits/manticore/blob/5bbfc8c6275611fc41e5d46830e08d57ff44db83/manticore/platforms/evm.py#L1333)):*

Why not create a separate `except`? 
</blockquote></details>

mostly to avoid copy-pasting the `self.last_exception = e` and `raise` lines which apply to both cases

---

*[manticore/platforms/evm.py, line 2138 at r3](https://reviewable.io:443/reviews/trailofbits/manticore/722#-L5KJxDXf7RtbtnoqzPB:-L5LBfcJzG_G0NDvYC6z:b-a1t4kq) ([raw file](https://github.com/trailofbits/manticore/blob/5bbfc8c6275611fc41e5d46830e08d57ff44db83/manticore/platforms/evm.py#L2138)):*

Can you help me understand how this way of handling is different from the old one?
</blockquote></details>

hm it really isn't different, this is an artifact of some of my previous attempts to emit the signal at this part of the code rather than the EVM cpu, and i never reverted it. i'll revert it, since it doesn't need to be changed, and it less pythonic this way

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/722)*

yan:
Reviewed 1 of 1 files at r4.

---

*[manticore/platforms/evm.py, line 1308 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/722#-L5nniLGHbuZgpfIGd7y:-L5nniLGHbuZgpfIGd7z:b-zci9z8) ([raw file](https://github.com/trailofbits/manticore/blob/17e41061c95ce2280d273cdeb4824b988bf2182d/manticore/platforms/evm.py#L1308)):*
> ```Python
>         result = None
> 
>         def emit_did_execute_signals():
> ```

Move this definition outside of this method? @pgoodman provided a change a few months ago that un-inlined this style of exception handling for performance reasons to abstractcpu.py (https://github.com/trailofbits/manticore/commit/d341bc94ff1cd015aa3fffc96c155aeab166a910). We should probably follow that style here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/722)*

yan:
:lgtm:

---

Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/722#-:-L5oZFNaHGZBxV_OY8uJ:bnfp4nl)*

mossberg:
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[manticore/platforms/evm.py, line 1308 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/722#-L5nniLGHbuZgpfIGd7y:-L5oaCrVVRPEUGnG6mpo:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/17e41061c95ce2280d273cdeb4824b988bf2182d/manticore/platforms/evm.py#L1308)):*

Move this definition outside of this method? @pgoodman provided a change a few months ago that un-inlined this style of exception handling for performance reasons to abstractcpu.py (https://github.com/trailofbits/manticore/commit/d341bc94ff1cd015aa3fffc96c155aeab166a910). We should probably follow that style here.
</blockquote></details>

Done.

---

*[tests/test_eth.py, line 24 at r6](https://reviewable.io:443/reviews/trailofbits/manticore/722#-L5oaDiiTAHDN1vIUwgl:-L5oaDiiTAHDN1vIUwgm:b6rgva0) ([raw file](https://github.com/trailofbits/manticore/blob/6d53e6092935afa4a9e37d9e4bf31bc4a84e8264/tests/test_eth.py#L24)):*
> ```Python
>         self.assertIn('overflow at MUL', all_findings)
> 
> class EthTests(unittest.TestCase):
> ```

@yan i also added this test, can you review it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/722)*

Issue #713, opened 2018-01-15
Manticore fails to generate a test case when a parameter in an event is declared with the indexed keyword.


ggrieco-tob:
### OS / Environment

Distributor ID:	ArchLinux
Description:	ArchLinux
Release:	rolling-release

### Manticore version

Last revision (465c926a2e39d5daf1405711dac61b6ddd0a2527)

### Python version

Python 2.7.13

### Summary of the problem

Manticore fails to generate a test case when a parameter in an event is declared with the [indexed keyword](https://ethereum.stackexchange.com/questions/8658/what-does-the-indexed-keyword-do).

### Step to reproduce the behavior

This is a small solidity contract to reproduce this issue:

```
contract C {
    event E (address indexed x);
    function f(address x) {
      E(x);
    }
}
```

You can trigger this issue using the manticore cli:

```
$ manticore error.sol                                                                
```

### Expected behavior

It shouldn't fail.

### Actual behavior

```
2018-01-15 14:27:46,208: [24901] m.main:INFO: Beginning analysis
2018-01-15 14:27:46,249: [24901] m.ethereum:INFO: Starting symbolic transaction: 1
2018-01-15 14:27:46,581: [24901] m.ethereum:INFO: Generated testcase No. 0 - REVERT
2018-01-15 14:27:48,840: [24901] m.ethereum:INFO: Generated testcase No. 1 - REVERT
2018-01-15 14:27:51,981: [24901] m.ethereum:INFO: Generated testcase No. 2 - REVERT
2018-01-15 14:27:54,994: [24901] m.ethereum:INFO: Generated testcase No. 3 - REVERT
2018-01-15 14:27:57,232: [24901] m.ethereum:INFO: Finished symbolic transaction: 1 | Code Coverage: 100% | Terminated States: 4 | Alive States: 1
2018-01-15 14:27:57,233: [24901] m.ethereum:INFO: Generated testcase No. 4 - STOP
Traceback (most recent call last):
  File "/home/gustavo/.local/bin/manticore", line 9, in <module>
    load_entry_point('manticore==0.1.6', 'console_scripts', 'manticore')()
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/__main__.py", line 104, in main
    ethereum_cli(args)
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/__main__.py", line 94, in ethereum_cli
[...]

Issue #45, opened 2017-03-02
Port to Python3


dguido:

mossberg:
i've been learning about py3 and porting recently and wanted to share some resources i've found useful:

http://python3porting.com/bookindex.html
https://docs.python.org/3/howto/pyporting.html
http://python-future.org/overview.html
https://www.youtube.com/watch?v=KPzDX5TX5HE

PR #739, opened 2018-02-06
Dev fix solver getvalue

feliam:
Make solver get value return expressions untouched when concrete.

related issue.  https://github.com/trailofbits/manticore/issues/713

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/739)


manticore/platforms/evm.py
```diff
@@ -1754,11 +1753,10 @@ def LOG(self, address, size, *topics):
     def read_buffer(self, offset, size):
         if size:
             self._allocate(offset+size)
-
         data = []
         for i in xrange(size):
-            data.append(Operators.CHR(self._load(offset+i)))
-
+            data.append(self._load(offset+i))
+        data = map(Operators.CHR, data)
```
disconnect3d: As we eventually want to go to Python 3 world (https://github.com/trailofbits/manticore/issues/45) this should be `list(map(...))`.

The `map` returns a generator-like object in Py3 instead of a `list`.

manticore/platforms/linux.py
```diff
@@ -2395,9 +2395,13 @@ def sys_open(self, buf, flags, mode):
 
     def generate_workspace_files(self):
         def solve_to_fd(data, fd):
```
disconnect3d: Wishful thinking: since this is a public function it would be nice to have a `"""..."""` docstring here.

manticore/platforms/evm.py
yan: Making that a generator object should be okay; it's iterated/transformed below. On a more structural note, the above can be expressed as:

```
data = [Operators.CHR(self._load(i)) for i in range(offset, offset+size)]
```

mossberg:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[manticore/ethereum.py, line 1179 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/739#-L5tJJtH5wUUe6B6BhWS:-L5tJJtH5wUUe6B6BhWT:b3g9na1) ([raw file](https://github.com/trailofbits/manticore/blob/9e2dd9281ee5759cb3f8b9cb7a78d97e5b5b51ca/manticore/ethereum.py#L1179)):*
> ```Python
>                 tx_summary.write("To: 0x%x %s\n" % (state.solve_one(tx.address), flagged(issymbolic(tx.address))))
>                 tx_summary.write("Value: %d %s\n"% (state.solve_one(tx.value), flagged(issymbolic(tx.value))))
>                 tx_data = ''.join(state.solve_one(tx.data))
> ```

why the `''.join`? `state.solve_one(tx.data)` should return a string right?

---

*[manticore/platforms/linux.py, line 2397 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/739#-L4h6g59luw8A4OLRQGk-r1-2397:-L5tUScDkWHMnJRwtzTV:b-l51yvy) ([raw file](https://github.com/trailofbits/manticore/blob/9e2dd9281ee5759cb3f8b9cb7a78d97e5b5b51ca/manticore/platforms/linux.py#L2397)):*
<details><summary><i>Previously, disconnect3d (Disconnect3d) wrote…</i></summary><blockquote>

Wishful thinking: since this is a public function it would be nice to have a `"""..."""` docstring here. 
</blockquote></details>

This is a public method on the State class, but not really a public API method. of course docstrings would be great everywhere, but i think the method name is fairly self documenting here also.

---

*[manticore/platforms/linux.py, line 2398 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/739#-L5tMKRyrBJneO5g9BiQ:-L5tMKRyrBJneO5g9BiR:b-5j8s8w) ([raw file](https://github.com/trailofbits/manticore/blob/9e2dd9281ee5759cb3f8b9cb7a78d97e5b5b51ca/manticore/platforms/linux.py#L2398)):*
> ```Python
>     def generate_workspace_files(self):
>         def solve_to_fd(data, fd):
>             def make_chr(c):
> ```

hm, this feels a bit hacky to me. i think we should make `solver.get_value` be consistent with the return type, rather than having function to handle it. shouldn't `solver.get_value` always return int types?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/739)*

feliam:
*[manticore/platforms/linux.py, line 2398 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/739#-L5tMKRyrBJneO5g9BiQ:-L5tpvhTGkrxUiF9o__C:be32y5k) ([raw file](https://github.com/trailofbits/manticore/blob/9e2dd9281ee5759cb3f8b9cb7a78d97e5b5b51ca/manticore/platforms/linux.py#L2398)):*

hm, this feels a bit hacky to me. i think we should make `solver.get_value` be consistent with the return type, rather than having function to handle it. shouldn't `solver.get_value` always return int types?
</blockquote></details>

We can do that. We would need to browse the expression and apply ord to any concrete char. 

solver.get_value('A') := ord('A') ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/739)*

mossberg:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[manticore/platforms/linux.py, line 2398 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/739#-L5tMKRyrBJneO5g9BiQ:-L5ua2Y4jyQ3x-RwwqxB:b-yafjuf) ([raw file](https://github.com/trailofbits/manticore/blob/9e2dd9281ee5759cb3f8b9cb7a78d97e5b5b51ca/manticore/platforms/linux.py#L2398)):*

We can do that. We would need to browse the expression and apply ord to any concrete char. 

solver.get_value('A') := ord('A') ?


</blockquote></details>

Yes, I think so. I think the solver interfaces should be based around integral types.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/739)*

feliam:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[manticore/platforms/linux.py, line 2398 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/739#-L5tMKRyrBJneO5g9BiQ:-L5xekECHSZLnvdQmxxf:b4p7a9m) ([raw file](https://github.com/trailofbits/manticore/blob/9e2dd9281ee5759cb3f8b9cb7a78d97e5b5b51ca/manticore/platforms/linux.py#L2398)):*

Yes, I think so. I think the solver interfaces should be based around integral types.
</blockquote></details>

So does the following look okay?
```
solver.get_value('AAA') := (0x41, 0x41, 0x41) 
```
and 
```
solver.get_value( (0x41, 0x41, 0x41) ) := (0x41, 0x41, 0x41) 
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/739)*

feliam:
*[manticore/ethereum.py, line 1179 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/739#-L5tJJtH5wUUe6B6BhWS:-L62i0qXKdsb4pQVCCGn:b-9rlciy) ([raw file](https://github.com/trailofbits/manticore/blob/9e2dd9281ee5759cb3f8b9cb7a78d97e5b5b51ca/manticore/ethereum.py#L1179)):*

why the `''.join`? `state.solve_one(tx.data)` should return a string right?
</blockquote></details>

We could make state.solve_one to be much more forgiving than solver.get_value if we want to.  Though that would be another refactor. We could call that branch  dev-fix-state-solveone 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/739)*

mossberg:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[manticore/ethereum.py, line 1179 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/739#-L5tJJtH5wUUe6B6BhWS:-L62kJGG-kkplnealYNl:b1x2b3q) ([raw file](https://github.com/trailofbits/manticore/blob/9e2dd9281ee5759cb3f8b9cb7a78d97e5b5b51ca/manticore/ethereum.py#L1179)):*

We could make state.solve_one to be much more forgiving than solver.get_value if we want to.  Though that would be another refactor. We could call that branch  dev-fix-state-solveone 
</blockquote></details>

I mean currently, `state.solve_one(tx.data)` returns a regular string right? So I don't understand why the `''.join` is necessary

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/739)*

feliam:
*[manticore/ethereum.py, line 1179 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/739#-L5tJJtH5wUUe6B6BhWS:-L62nR00xANBse2O7ax0:bxdg689) ([raw file](https://github.com/trailofbits/manticore/blob/9e2dd9281ee5759cb3f8b9cb7a78d97e5b5b51ca/manticore/ethereum.py#L1179)):*

I mean currently, `state.solve_one(tx.data)` returns a regular string right? So I don't understand why the `''.join` is necessary
</blockquote></details>

Oh if. tx.dta is a tuple of ints.  state.solve_one will return a tuple of ints. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/739)*

mossberg:
I included a small fix for the `--data` cli flag, which was broken/ignored.

mossberg:
:lgtm:

---

Reviewed 2 of 3 files at r2, 1 of 2 files at r3, 1 of 2 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/739#-:-L6TSrHt89fRvLgnXgI_:bnfp4nl)*

PR #749, opened 2018-02-13
Armv7 code refactor

disconnect3d:
Removes dead imports, makes code style more consistent and a bit more compliant with pep8 (yeah i seems we don't enforce any particular style, but this seems to be a reasonable default for me).

...lets see if it passes CI checks.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/749)

### DIFF
--- a/manticore/core/cpu/arm.py
+++ b/manticore/core/cpu/arm.py
@@ -18,15 +18,17 @@ OP_NAME_MAP = {
     'MOVW': 'MOV'
 }
 
+
 def HighBit(n):
     return Bit(n, 31)
+
 
 def instruction(body):
     @wraps(body)
     def instruction_implementation(cpu, *args, **kwargs):
         ret = None
 
-        should_execute = cpu.shouldExecuteConditional()
+        should_execute = cpu.should_execute_conditional()
 
         if cpu._at_symbolic_conditional:
             cpu._at_symbolic_conditional = False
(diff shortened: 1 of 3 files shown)


mossberg:
great, we'll get to reviewing this shortly!

disconnect3d:
@mossberg I have rebased this to the latest master so it can be reviewed :P.

manticore/core/cpu/arm.py
```diff
@@ -404,23 +408,23 @@ def setFlags(self, **flags):
...
         # XXX: capstone incorrectly sets .update_flags for adc
         if self.instruction.mnemonic == 'adc':
             return
         for flag, val in self._last_flags.iteritems():
             flag_name = 'APSR_{}'.format(flag)
             self.regfile.write(flag_name, val)
 
-    def _Shift(cpu, value, _type, amount, carry):
-        assert(_type > cs.arm.ARM_SFT_INVALID and _type <= cs.arm.ARM_SFT_RRX_REG)
+    def _shift(cpu, value, _type, amount, carry):
```
mossberg: The reason we originally had it named like this was to closely match the arm manual pseudocode; however I think this change is a good thing to make the codebase more pythonic. Can you please add a docstring which states "See Shift() and Shift_C() in the ARM manual"?

image

disconnect3d: @mossberg added

PR #752, opened 2018-02-15
Fixed Issue #723 - changed second _pop() to _pop_vm()

### LINKED ISSUE #723: self._pop() does not exist
https://github.com/trailofbits/manticore/blob/9f7b87d0b8e047d26f24ed9e5ddddb0068069ae9/manticore/platforms/evm.py#L2084

2 places in this file call `self._pop()`, which does not exist. i think it should be `self._pop_vm()`

arunjohnkuruvilla:
Fixes  #723 - changed second _pop() to _pop_vm(). All _pop() are now _pop_vm().

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/752)

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -1291,7 +1291,7 @@ class EVM(Eventful):
             arguments.append(current.operand)
 
         for _ in range(current.pops):
-            arguments.append(self._pop())
+            arguments.append(self._pop_vm())
 
         #simplify stack arguments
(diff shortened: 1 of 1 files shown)


mossberg:
ack, I apologize, my comment in #723 was wrong; that other spot I mentioned was actually correct, since `EVM._pop()` exists. so i'm going to close this PR, and that issue. sorry about that!

Issue #753, opened 2018-02-15
Infrastructure and skeleton classes for aarch64 support


mossberg:
This is the tracking issue for the first step in aarch64 support. 

Task: Implement support for aarch64 binary paths as the argument to the Manticore constructor/linux classmethod.

Calling `Manticore(‘aarch64_elf’, [‘arg1’, ‘arg2’], dict(ENV=’env’))` and `Manticore.linux(‘aarch64_elf’, [‘arg1’, ‘arg2], dict(ENV=’env’))` should successfully execute, having loaded the binary and prepared the initial execution State. In this example, ‘aarch64_elf’ is the path to an aarch64 Linux ELF binary.

Completing this will involve modifying files including manticore.py, linux.py, cpufactory.py, and creating a new file in manticore/core/cpu for all the aarch64 code (aarch64.py? Not sure of the best name). The changes will involve modifying mostly the plumbing between components, modifying some logic, and creating new empty, skeleton classes for the Aarch64/Armv8 (naming?) cpu, Armv8 function abi, and Armv8 syscall abi. This task will not involve implementing an ELF loader, just connecting to one that already exists.

Changes that should also be implemented:
- Improve the error message Manticore gives for an unsupported binary arch. Currently KeyError, let’s raise an Exception with a good error message.
- Emit a logger warning that clearly states that Aarch64 support is experimental. This should be clearly visible to all users who pass an aarch64 binary to Manticore.

Please use the #manticore channel in the Empire Hacking slack to communicate with the Manticore development team while completing this task! https://empireslacking.herokuapp.com

disconnect3d:
Hey,

I started some work to make it work properly.

By the way, did you mean that envs should be passed as a keyword argument like this:
```
Manticore(‘aarch64_elf’, [‘arg1’, ‘arg2’], env=dict(ENV=’env’))
```

(when I pass the last arg as `dict(ENV='env')` I get an exception of `Invalid workspace` - I am going to change the message to `Invalid workspace_url` so that it is a bit more verbose :P).

mossberg:
awesome, thanks for starting!

> By the way, did you mean that envs should be passed as a keyword argument like this:

yes, my mistake, sorry 😅 



> I am going to change the message to Invalid workspace_url so that it is a bit more verbose :P).

sounds good 👍 

> should we name it aarch64 or arm64 or maybe armv8 (as we have armv7 for arm x32)?

[...]

PR #754, opened 2018-02-16
AArch64 skeleton classes

disconnect3d:
~**NOTE: This is based upon https://github.com/trailofbits/manticore/pull/749 - so #749 should be merged first and then I can rebase this.**~

**NOTE: This is based upon https://github.com/trailofbits/manticore/pull/761 - so #761 should be merged first and then this will be rebased.**

This is an initial stuff to support aarch64 - https://github.com/trailofbits/manticore/issues/753. 

Current result is that `Manticore` loads the binary but probably nothing more works:
```bash
$ python -c "from manticore import Manticore; m = Manticore.linux('./release.stripped', ['arg1', 'arg2'], env=dict(ENV='env'))"
/home/dc/Projects/manticore/manticore/core/cpu/aarch64.py:92: UserWarning: Aarch64 support is experimental; it might not work well yet; feel free to help with that
  warnings.warn('Aarch64 support is experimental; it might not work well yet; feel free to help with that')

$ python -c "from manticore import Manticore; m = Manticore('./release.stripped', ['arg1', 'arg2'], env=dict(ENV='env'))"
/home/dc/Projects/manticore/manticore/core/cpu/aarch64.py:92: UserWarning: Aarch64 support is experimental; it might not work well yet; feel free to help with that
  warnings.warn('Aarch64 support is experimental; it might not work well yet; feel free to help with that')
```

Some of the changes:
* applied some code refactoring
* using `__future__` imports in `aarch64.py` so [it should be easier to support both Py2 and Py3 in the future](https://github.com/trailofbits/manticore/issues/45) (**NOTE: this changes some stuff like how division works**)
* using `open` instead of `file` when opening interpreter file ([this is the recommended way, supported by both Py2 and Py3](https://stackoverflow.com/questions/112970/python-when-to-use-file-vs-open))
* handling case when ELF interpreter path doesn't exist - the aarch64 binary I test it on is from Android and it has interpreter path set to `/system/bin/linker64` which doesn't exist on my system.

Obviously there are still things to improve like:
* [x] Add aarch64 cdecl abi (`Aarch64CdeclAbi`)
* [x] Add linux syscall abi (`Aarch64LinuxSyscallAbi`)
* [x] Set proper `uname_machine`
* [x] Add operand class (`Aarch64Operand`)
* [x] Add unicorn emulation
* [ ] Add missing registers
* [ ] Proper banking/aliasing registers
* [ ] Handle aarch32 and aarch64 states: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0801a/BABBDFIH.html - this will probably require us to keep a aarch32/aarch64 state flag and change banking/aliasing registers based on that (**as I believe those differs between the modes**)

If anyone wants to help with any of those, feel free to comment/catch me on slack.

Some questions:
* Merging against master is much easier (no painful conflicts solving later on) but maybe we should merge aarch64 related stuff to its own branch?

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/754)


manticore/core/cpu/arm.py
```diff
@@ -293,7 +293,7 @@ class Armv7LinuxSyscallAbi(SyscallAbi):
 
     # EABI standards:
     #  syscall # is in R7
-    #  arguments are passed in R0-R6
+    #  arguments are passed in R0-R5
```
disconnect3d: The code below (`for i in range(6): yield 'R'.format(i)` returns exactly 6 args while `R0-R6` would be 7. Anyway afaik syscalls have maximum of 6 args.

yan:
Reviewed 4 of 6 files at r1, 2 of 2 files at r2.

---

*[manticore/core/cpu/aarch64.py, line 43 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/754#-L5pE-ISR3MWyskll1v4:-L5pE-ISR3MWyskll1v5:b8vqaxi) ([raw file](https://github.com/trailofbits/manticore/blob/bf5a18db0455eeb1b43953300953c7ec0787447c/manticore/core/cpu/aarch64.py#L43)):*
> ```Python
>         # • A 32-bit general-purpose register named W0 to W30.
>         alias_regs.update(
>             ((r, r.replace('R', 'X')) for r in self._R_REGS)
> ```

I'm not sure what we gain by replacing prefix letters instead of just using a similar initialization as the way `_R_REGS` and `_V_REGS` are done.  Another option is to create a very basic helper, something like:

```python
def registers_named(prefix):
    return tuple('{}{}'.format(prefix, i) for i in range(31))
class Aarch64RegisterFile(RegisterFile):
    _R_REGS = registers_named('R')
    _V_REGS = registers_named('V')
    ...
```

---

*[manticore/core/cpu/aarch64.py, line 78 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/754#-L5pFE17nqD_yknan6Yg:-L5pFE17nqD_yknan6Yh:bfhlkqp) ([raw file](https://github.com/trailofbits/manticore/blob/bf5a18db0455eeb1b43953300953c7ec0787447c/manticore/core/cpu/aarch64.py#L78)):*
> ```Python
>     @property
>     def all_registers(self):
>         return super(Aarch64RegisterFile, self).all_registers + self._R_REGS + self._V_REGS
> ```

Cache this in an ivar, just like the recent patches?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/754)*

disconnect3d:
@yan Fixed, I have also spotted that there are V0-V31 which means the `range(31)` was wrong and should be `range(32)`, doh ;).

yan:
Reviewed 1 of 1 files at r3.

---

*[manticore/core/cpu/aarch64.py, line 80 at r3](https://reviewable.io:443/reviews/trailofbits/manticore/754#-L5ul3Yk3WdAWDzAPVq2:-L5ul3Yk3WdAWDzAPVq3:b-oqi4wi) ([raw file](https://github.com/trailofbits/manticore/blob/ae035dd3454b85ce77ea717f80c616eedabb90a6/manticore/core/cpu/aarch64.py#L80)):*
> ```Python
>     @property
>     def all_registers(self):
>         return self._all_registers
> ```

There's probably an opportunity to move properties like this one to an ARM-common base class. This is good as is for now though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/754)*

yan:
Reviewed 3 of 3 files at r4, 3 of 4 files at r5.
Review status: 9 of 10 files reviewed at latest revision, 5 unresolved discussions.

---

*[manticore/core/cpu/aarch64.py, line 21 at r5](https://reviewable.io:443/reviews/trailofbits/manticore/754#-L6Iv3cIrBqRCNbXiUC4:-L6Iv3cIrBqRCNbXiUC5:b-wev0ee) ([raw file](https://github.com/trailofbits/manticore/blob/f0e34b93817fea1b228578676f6cf6aad5b2bf78/manticore/core/cpu/aarch64.py#L21)):*
> ```Python
> # TODO / FIXME / REVIEWME: This is probably missing a lot of instructions
> # map different instructions to a single impl here
> INSTRUCTION_MAPPINGS = {
> ```

I'm not sure I like the all caps mapping convention. Don't have a great alternative but open to discuss. Options: just `instruction_mappings` or `__instruction_mappings__`.

---

*[manticore/platforms/linux.py, line 825 at r5](https://reviewable.io:443/reviews/trailofbits/manticore/754#-L6Iwg_FT5QBVBcrZUpO:-L6Iwg_G7RMGfKZhhhBf:b637htn) ([raw file](https://github.com/trailofbits/manticore/blob/f0e34b93817fea1b228578676f6cf6aad5b2bf78/manticore/platforms/linux.py#L825)):*
> ```Python
>                 interpreter = ELFFile(open(interpreter_filepath, 'rb'))
>             except IOError:
>                 logger.debug('No interpreter file: %s', interpreter_filepath)
> ```

Why not percolate the exception up, or wrap in custom. Not having the interpreter when it's called for should be an early exit, no?

---

*[manticore/platforms/linux.py, line 2313 at r5](https://reviewable.io:443/reviews/trailofbits/manticore/754#-L6IwM0O47hq7fb3M0DX:-L6IwM0O47hq7fb3M0DY:b-acrf6n) ([raw file](https://github.com/trailofbits/manticore/blob/f0e34b93817fea1b228578676f6cf6aad5b2bf78/manticore/platforms/linux.py#L2313)):*
> ```Python
>             self._init_arm_kernel_helpers()
>         elif self.arch == 'aarch64':
>             # TODO / FIXME / REVIEW PLS: is that okay? Seems some machines might have it set as armv8
> ```

Looks good for now.

---

*[manticore/utils/emulate.py, line 45 at r5](https://reviewable.io:443/reviews/trailofbits/manticore/754#-L6Ivnt-40RLgRwMj2W1:-L6Ivnt-40RLgRwMj2W2:b-clzvwb) ([raw file](https://github.com/trailofbits/manticore/blob/f0e34b93817fea1b228578676f6cf6aad5b2bf78/manticore/utils/emulate.py#L45)):*
> ```Python
>         if self._cpu.arch in (CS_ARCH_ARM, CS_ARCH_ARM64):
>             uc_arch = UC_ARCH_ARM if self._cpu.arch == CS_ARCH_ARM else UC_ARCH_ARM64
> ```

I'm not sure I like this style more than the simple if/elif branches. I think we should either create vars (`uc_arch` and `uc_mode`) depending on target environment then have a single place where we create `Uc`, or have the flat if/elif structure, not mix the two.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/754)*

disconnect3d:
@yan 

1. The convention for constants - most Python code I met used pep8 more or less and so I did here, see https://www.python.org/dev/peps/pep-0008/#constants
If we want to keep it private, we can go with `_SOME_VAR` or `_some_var`; the double underscores before and after a name are used mostly within special methods or attributes, so I wouldn't use that.

2. I am not sure about the interpreter. I was thinking that we should be okay to work without it as we check for it later on https://github.com/trailofbits/manticore/pull/754/files#diff-449d7ed09386871ab1de7bc978d81436R828 ; also I believe that angr works fine when interpreter is not present.

3. -

4. Yeah, I will create `uc_arch` and `uc_mode` and instantiate `Uc` in just one place.

yan:
Reviewed 2 of 3 files at r6, 1 of 4 files at r7, 6 of 6 files at r8.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[manticore/manticore.py, line 175 at r8](https://reviewable.io:443/reviews/trailofbits/manticore/754#-L70xRww9hM1ZWmnv0p2:-L70xRww9hM1ZWmnv0p3:bqew74r) ([raw file](https://github.com/trailofbits/manticore/blob/be8ec5e79e4a38e482cb4b4722fbe6a460c5bf34/manticore/manticore.py#L175)):*
> ```Python
> 
>         if isinstance(path_or_state, str):
>             assert os.path.isfile(path_or_state), "The file or state %s doesn't exist" % path_or_state
> ```

Raise exception here, instead of an assert? Also, prefer `.format` to `%`.

---

*[manticore/core/cpu/aarch64.py, line 121 at r8](https://reviewable.io:443/reviews/trailofbits/manticore/754#-L70yBTv7jHtaza4bfUP:-L70yBTv7jHtaza4bfUQ:buxd15b) ([raw file](https://github.com/trailofbits/manticore/blob/be8ec5e79e4a38e482cb4b4722fbe6a460c5bf34/manticore/core/cpu/aarch64.py#L121)):*
> ```Python
>         state = super(Aarch64Cpu, self).__getstate__()
>         state['_last_flags'] = self._last_flags
>         # TODO / FIXME / REVIEWME: do we need those in aarch64? [copied from armv7]
> ```

As far as I can tell, some instructions in aarch64 can be executed conditionally, so probably still relevant? Is it possible to capture symbolic status codes without these two state vars?

---

*[manticore/core/cpu/aarch64.py, line 231 at r8](https://reviewable.io:443/reviews/trailofbits/manticore/754#-L70yNZLrpQfJijHrXSC:-L70yNZLrpQfJijHrXSD:b-bl6hqa) ([raw file](https://github.com/trailofbits/manticore/blob/be8ec5e79e4a38e482cb4b4722fbe6a460c5bf34/manticore/core/cpu/aarch64.py#L231)):*
> ```Python
> 
> 
> class Aarch64Operand(Operand):
> ```

Would it make sense to unify large parts of `ArmOperand` and `Aarch64Operand`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/754)*

disconnect3d:
> Would it make sense to unify large parts of ArmOperand and Aarch64Operand?

Probably yes, however, I am still not yet sure. I have already tried to do this in an easy way (just inherit from ArmOperand and see what would happen) and there are some differences:

* The ARM operand has `subtracted` field https://github.com/aquynh/capstone/blob/master/include/arm.h#L242
* The AArch64 operand doesn't https://github.com/aquynh/capstone/blob/master/include/arm64.h#L358-L378

disconnect3d:
Bump.

I have changed the target of this branch to `dev-aarch64`.

I also merged it with `master` and removed/changed some code regarding setting CPU capstone mode as it seems there is no Thumb mode for aarch64/arm64 binaries (see https://stackoverflow.com/questions/46086329/can-i-use-thumb-instructions-in-an-arm64-binary).

Can we also merge this PR into the dev branch? As from its current state and comments its not that clear what is the state/etc. I would try to go with some smaller PRs. With that being said and if you agree then I think that this:

> As far as I can tell, some instructions in aarch64 can be executed conditionally, so probably still relevant? Is it possible to capture symbolic status codes without these two state vars?

will be seen/observed later when this thing will be needed for some tests/development.

mossberg:
@disconnect3d thanks for the bump on this, I apologize, I think we had some internal confusion on whether this was ready for review again or not; i think @yan will probably be unavailable for review of this until next week, maybe @f2xeb you could take a look within the next few days?

catenacyber:
@disconnected do you need help on this aarch64 stuff ?

disconnect3d:
@catenacyber Yeah, it would be nice. I haven't looked on it for some time.

As far as I remember I need to do small cleanup here (as I left some functions that are not needed) and it could be merged. Then we could proceed on implementing particular instructions and adding tests.

I will do the cleanup tomorrow or the day after tomorrow.

catenacyber:
Ok, I can take a look after your cleanup, and implement some particular instructions and tests.

Do you have a test sample and command line or script ?  (where we can look for unimplemented instructions)

disconnect3d:
@catenacyber I have removed some dead/unused code from aarch64 CPU.

So my testcase is a hello world program compiled for arm64/aarch64 which I launch as:
```python
manticore -vvv ./hello
```

(I installed manticore for development purposes by `python setup.py develop` after switching to my branch)

catenacyber:
@disconnect3d can you supply me your file `/lib/ld-linux-aarch64.so.1` as well ?

I checked out you aarch64/skeleton branch
And I get to 
`2018-06-18 09:33:44,470: [69825] m.c.c.abstractcpu:INFO: Unimplemented instruction: 0x0000000000400410:	20 02 1f d6	br	x17`

Does it look good to you ? (I mean good enough so that I can start working on it)

disconnect3d:
@catenacyber I have a new system so I have new `/lib/ld-linux-aarch64.so.1`.

I guess that if you want to run arm64 binaries e.g. on qemu, it is best to install package that will contain everything that is required (apart from interpreter you would also need specific arm64 libc.so and probably some more stuff).

For example, if you are using Ubuntu you can install `gcc-aarch64-linux-gnu` package from apt and then:

```
11:02:18 in ~/aarch_playground took 2s 
➜ aarch64-linux-gnu-gcc hello.c -o hello        

11:02:29 in ~/aarch_playground 
➜ qemu-aarch64 -L /usr/aarch64-linux-gnu ./hello
Hello world
```

The `-L ...` part is important - without it it will try to look for interpreter in `/lib/...` (I have used symlinks once for interpreter and it worked, but the more libs you use the more problematic things become).

Anyway, below you can find both interpreter and new hello world binary (I have found out the previous one might have used `scanf` before printing hello world, sorry for inconvenience).

http://re.disconnect3d.pl/arm64/hello
http://re.disconnect3d.pl/arm64/ld-linux-aarch64.so.1

catenacyber:
@disconnect3d thanks for the binaries and log.
Could you supply me your libc.so so that I am sure we use the same inputs ?

Then, I will try to go for implementing unimplemented instructions

disconnect3d:
@catenacyber 
I have zipped all arm64 directory, just in case: http://re.disconnect3d.pl/arm64/aarch64-linux-gnu.zip

```
$ sha256sum ./aarch64-linux-gnu.zip
b3745a350555d03201eb9839b1cd7dddb91ead74206fe6f856325220eb93de1d  ./aarch64-linux-gnu.zip
```

mossberg:
@disconnect3d @catenacyber 

hey all! i'm really excited to see some activity on this! i just wanted to mention that I think we should not really include instruction implementations in this pull request. I think it makes more sense to have a PR with the basic CPU/ABI/RegisterFile/etc infrastructure first; then we can have PRs which add instructions, which should make review easier.

Given this, if all the mentioned infrastructure is complete, I think we should merge this ASAP. it looks like @yan has already basically reviewed and effectively approved(?) it?

regarding merge strategy, we're now thinking of merging directly into master. long term intermediate staging branches haven't worked super well for us in the past since they introduce some amount of process overhead and also kind of let a big diff accumulate which becomes unwieldy. i'm going to change this PR back to merging into master.

mossberg:
we haven't forgotten about this! master is currently undergoing a freeze while we work on porting Manticore to Python 3 (abandoning python 2 support). once that is done we can make the hopefully small updates to update this to py3, and revisit merging to master. then we can look forward for future PRs to implement instructions, etc.

mossberg:
Our python3 freeze and port is now complete, so we can move forward with porting this branch to python3.

catenacyber:
ping @disconnect3d can I help with this ?

disconnect3d:
@catenacyber Yeah I think so. The branch needs to be rebased and checked if it works fine with Py3.

mossberg:
this has unfortunately gone super stale 😭 it has been neglected in our pr log for 8 months now, maybe we should just close it 😢also we don't entirely have resources to develop and maintain this right now. if no one objects, i'll close in a few days to clean things up and we can revisit /sometime/

Issue #91, opened 2017-03-21
Integrate linter into CI


offlinemark:

mossberg:
We should still do this at some point. One problem, is that typically the list of issues it outputs is immense, and we don't particularly care to eliminate all of them. A solution could be to, on each commit, only output the diff between the last lint output, and the current

mossberg:
In particular, and at the very least, I'm interesting in using them to help us check that our docstrings are consistent with the actual functions. i think they can do this

PR #772, opened 2018-02-24
Add codeclimate, tox, and pycodestyle configs

dguido:
I thought it might be a good idea to add some new developer tools to Manticore now that it's attracting more attention and to get the ball rolling on forward porting Manticore to Python3.

This PR includes:
- an autogenerated set of changes for pep8 conformance in manticore/*.py
- a very minimal codeclimate config and automated test coverage reporting from travis
- a very minimal tox config to make local testing of Python3 a little easier

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/772)


disconnect3d:
I am not sure about python-sonar and radon in codeclimate - the latter seems to be used for some not-really-needed metrics. For a security linter I've always used https://github.com/openstack/bandit so maybe this would be good to check.

Anyway I like this PR a lot :heart:

dguido:
Thanks, I’ll check out Bandit! This is still a work in progress. I started it while running out of batteries on a plane over the Atlantic :-). I’ll take a second look today or tomorrow after I get home.

manticore/core/cpu/x86.py
```diff
@@ -5020,12 +5020,15 @@ def _pcmpxstrx_polarity(self, res1, ctlbyte, arg2len):
...
             res2 = res1
+            pass
         if (Operators.EXTRACT(ctlbyte, 4, 2) == 1):
             res2 = ((1 << (128 / stepsize)) - 1) ^ res1
         if (Operators.EXTRACT(ctlbyte, 4, 2) == 2):
             res2 = res1
+            pass
         if (Operators.EXTRACT(ctlbyte, 4, 2) == 3):
             res2 = ((1 << arg2len) - 1) ^ res1
+            pass
```
disconnect3d: The `pass` here are redundant.

dguido: I didn't add that pass, it was there originally. autoflakes found a lot of redundant `pass` statements, which I removed and then promptly broke all our tests :-x. Safer to just leave these for now.

manticore/core/cpu/arm.py
```diff
@@ -752,7 +752,6 @@ def UXTB(cpu, dest, src):
     @instruction
     def PLD(cpu, addr, offset=None):
         """PLD instructs the cpu that the address at addr might be loaded soon."""
-        pass
```
disconnect3d: Heh, actually having `pass` in an empty functon or empty exception class is the way to go. I wonder why autoflake deletes it :(

yan: Way to go in what sense?

disconnect3d: I mean it is an explicit way to say "yeah I really did wanted this code block to be empty".

Some links:
* https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions - e.g. the examples for empty user-defined exceptions use `pass`
* https://docs.python.org/3.2/reference/simple_stmts.html#the-pass-statement

mossberg:
Cool! When refactoring a module's imports we have to be very careful for one thing: even if it looks like an import is unused in module A (e.g. `struct`), if module B does `from moduleA import *` and then does not import `struct` but uses `struct.pack` etc, then removing the "unused" `struct` import in module A is unsafe. I'm reviewing now, and will be checking for this 🔍

mossberg:
Reviewed 16 of 40 files at r1, 1 of 4 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 16 of 22 files at r5, 7 of 7 files at r6.

---

*[manticore/core/executor.py, line 377 at r6](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6OQuTLsMX8mvHGHGXp:-L6OQuTLsMX8mvHGHGXq:bdcrfh9) ([raw file](https://github.com/trailofbits/manticore/blob/2853ee1dd91b72d2dfa2965182353698ef9d514e/manticore/core/executor.py#L377)):*
> ```Python
> 
>         if setstate is None:
>             def setstate(x, y): return None
> ```

not sure if this is more readable

---

*[manticore/core/state.py, line 82 at r6](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6OTxNd-RNKSV44CHGg:-L6OTxNd-RNKSV44CHGh:bpea71w) ([raw file](https://github.com/trailofbits/manticore/blob/2853ee1dd91b72d2dfa2965182353698ef9d514e/manticore/core/state.py#L82)):*
> ```Python
>         self._child = None
>         self._context = dict()
>         # 33
> ```

if we're cleaning this up, we should remove the 33

---

*[manticore/core/cpu/x86.py, line 776 at r6](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6OV56rYwn7dMyMQo5R:-L6OV56rYwn7dMyMQo5S:bqu2yaz) ([raw file](https://github.com/trailofbits/manticore/blob/39370ccf3c9f74c3bb67eb7413b41993377a0948/manticore/core/cpu/x86.py#L776)):*
> ```Python
> 
>     ################################################3
>     # instruction implementation
> ```

i think this is probably a good structural comment to keep

---

*[manticore/core/cpu/x86.py, line 3175 at r6](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6OVSWfRnZo9XPW0eMU:-L6OVSWfRnZo9XPW0eMV:b-1c8dqd) ([raw file](https://github.com/trailofbits/manticore/blob/39370ccf3c9f74c3bb67eb7413b41993377a0948/manticore/core/cpu/x86.py#L3175)):*
> ```Python
>         :param target: destination operand.
>         '''
>         cpu.PC = Operators.ITEBV(cpu.address_bit_size, cpu.CF == True, target.read(), cpu.PC)
> ```

I think this is a bug? if `cpu.CF` is symbolic, `cpu.CF == True` would be an Eq Expression tree (i think?), but with the change, it will just be the `cpu.CF` symbolic variable

---

*[manticore/core/smtlib/solver.py, line 208 at r6](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6OaRV5YaHAxcFj40se:-L6OaRV5YaHAxcFj40sf:b-omtl78) ([raw file](https://github.com/trailofbits/manticore/blob/2853ee1dd91b72d2dfa2965182353698ef9d514e/manticore/core/smtlib/solver.py#L208)):*
> ```Python
>             self._proc.stdin.writelines(('(exit)\n',))
>             self._proc.wait()
>         except Exception as e:
> ```

can completely rm the `as e` too

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/772)*

manticore/core/cpu/arm.py
mossberg: I somewhat prefer having the pass, but it is redundant if the docstring is present.

mossberg: That said, I don't really think it's worth going through the work to add them all back in, if it's not as simple as a git revert.

yan:
Reviewed 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[manticore/core/executor.py, line 377 at r6](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6OQuTLsMX8mvHGHGXp:-L6TiEzcq5pOui5ZlAju:bia26tr) ([raw file](https://github.com/trailofbits/manticore/blob/2853ee1dd91b72d2dfa2965182353698ef9d514e/manticore/core/executor.py#L377)):*

not sure if this is more readable
</blockquote></details>

I don't think it's worse; I'm okay with this.

---

*[manticore/core/state.py, line 82 at r6](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6OTxNd-RNKSV44CHGg:-L6TiOD0-23My8gTtnNA:baxrrvc) ([raw file](https://github.com/trailofbits/manticore/blob/2853ee1dd91b72d2dfa2965182353698ef9d514e/manticore/core/state.py#L82)):*

if we're cleaning this up, we should remove the 33
</blockquote></details>

Agree. Ditto for the `__setstate__` comment below.

---

*[manticore/core/cpu/x86.py, line 3175 at r6](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6OVSWfRnZo9XPW0eMU:-L6TiZn0Nn9rYcOV9y_k:b-pasubq) ([raw file](https://github.com/trailofbits/manticore/blob/39370ccf3c9f74c3bb67eb7413b41993377a0948/manticore/core/cpu/x86.py#L3175)):*

I think this is a bug? if `cpu.CF` is symbolic, `cpu.CF == True` would be an Eq Expression tree (i think?), but with the change, it will just be the `cpu.CF` symbolic variable
</blockquote></details>

It's possible but unlikely to be a bug since `.CF` is usually assigned a boolean expression anyway[1], but we should still probably keep it like it was since it's arguably more clear.

[1]  Run to confirm:`grep -r 'cpu\.CF\s*=[^=]' manticore/core/cpu/x86.py`

---

*[manticore/core/cpu/x86.py, line 103 at r7](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6TjgQ2VlBrMIKm7ZrU:-L6TjgQ2VlBrMIKm7ZrV:b-rdjzng) ([raw file](https://github.com/trailofbits/manticore/blob/39370ccf3c9f74c3bb67eb7413b41993377a0948/manticore/core/cpu/x86.py#L103)):*
> ```Python
>                 #REPE
>                 if cs.x86.X86_PREFIX_REP in prefix:
>                     FLAG = Operators.AND(cpu.ZF == True, count != 0) #true FLAG means loop
> ```

Same issue as the `CF` discussion.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/772)*

mossberg:
Review status: 41 of 42 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[manticore/core/executor.py, line 377 at r6](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6OQuTLsMX8mvHGHGXp:-L6YN9YElXFW0nD9tS5a:bpzv7r6) ([raw file](https://github.com/trailofbits/manticore/blob/2853ee1dd91b72d2dfa2965182353698ef9d514e/manticore/core/executor.py#L377)):*

I don't think it's worse; I'm okay with this.
</blockquote></details>

i actually do think it is slightly less readable in the new version; in the former the assignment is a very obvious mutation of the setstate variable, however with the new form, i don't directly view the function definition as an assignment of the setstate variable.

---

*[manticore/core/cpu/x86.py, line 3175 at r6](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6OVSWfRnZo9XPW0eMU:-L6YNstYTcLM9DFKZ5Of:bskobes) ([raw file](https://github.com/trailofbits/manticore/blob/39370ccf3c9f74c3bb67eb7413b41993377a0948/manticore/core/cpu/x86.py#L3175)):*


After some more thought, I'm pretty confident that this is definitely a bug; even if it's unlikely for the bug to _manifest_ due to `cpu.CF` commonly being concrete, that doesn't mean the `JB` implementation is correct this way, as the construction of `Equal` constraint is not something we can just get rid of

confirmed that `<symbolic> == True` returns an `Equal` expression:

```
>>> b = BitVec(32)
>>> b == True
<Equal at 7fcd7d46e210>

```

also fwiw I my grep output, I see a bunch of opportunities for CF to become an expression, any time there is an assignment from another register (cpu.AF, or from some computation, or Operators call)

---

*[manticore/core/cpu/x86.py, line 103 at r7](https://reviewable.io:443/reviews/trailofbits/manticore/772#-L6TjgQ2VlBrMIKm7ZrU:-L6YNb8R4HC2T8Wmhhos:b-j92saq) ([raw file](https://github.com/trailofbits/manticore/blob/39370ccf3c9f74c3bb67eb7413b41993377a0948/manticore/core/cpu/x86.py#L103)):*

Same issue as the `CF` discussion.
</blockquote></details>

great catch!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/772)*

dguido:
Closes #91

yan:
Reviewed 1 of 7 files at r6, 1 of 1 files at r8, 1 of 1 files at r9, 8 of 8 files at r10, 1 of 1 files at r11.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/772)*

mossberg:
Reviewed 1 of 1 files at r9, 1 of 1 files at r11.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/772)*

mossberg:
now that the `cpu.FLAG == True` discussion is resolved (decision: maintain equivalent semantics as before this branch, revisit this in https://github.com/trailofbits/manticore/issues/796), there are only a few very small fixes outstanding, notes in the reviewable discussions.

mossberg:
notes on remaining work can be found in reviewable: https://reviewable.io/reviews/trailofbits/manticore/772

disconnect3d:
@mossberg 
> notes on remaining work can be found in reviewable
mm what exactly? just those two small comments that are also visible on github? Or is there something that is not visible by not logged in users?

mossberg:
@disconnect3d yeah, i think it actually just is that. everything on reviewable should be public to everyone, regardless of whether they are logged in

mossberg:
:lgtm:

---

Reviewed 2 of 2 files at r12.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/772#-:-L7QD6pVgK7mkiYlQT4F:bnfp4nl)*

dguido:
ermagerd

PR #774, opened 2018-02-26
Let the Python interpreter be specified externally

khorben:
This is a trivial change to help build the "linux" example without having to modify the Makefile. It is necessary if the Python interpreter has a different name (or place) than simply `python`.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/774)

### DIFF
--- a/examples/linux/Makefile
+++ b/examples/linux/Makefile
@@ -1,6 +1,7 @@
 CC=gcc
 CFLAGS=-O3 -static
 NOSTDLIBFLAGS=-fno-builtin -static -nostdlib -fomit-frame-pointer
+PYTHON=python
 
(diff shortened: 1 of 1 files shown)


disconnect3d:
@khorben Can't we just do `./crackme.py` and let the shebang do the magic?
(see https://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take)

yan:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/774#-:-L6NJitg4QnQwg8QGnm9:bnfp4nl)*

mossberg:
thanks a lot @khorben! 🎊

PR #775, opened 2018-02-26
Used a namedtuple for pending transaction. Issue #689

### LINKED ISSUE #689: Use a namedtuple for pending transaction
https://github.com/trailofbits/manticore/blob/743d479cbf10f40e3079d56e3f26b2d1f4e34893/manticore/platforms/evm.py#L2185

arunjohnkuruvilla:
Imported namedtuple from collections

Created a namedtuple PendingTransaction which is assigned to self._pending_transcation

Fixes #689

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/775)

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -10,6 +10,7 @@ from ..core.smtlib.visitors import pretty_print, arithmetic_simplifier, translat
 from ..core.state import Concretize,TerminateState
 import logging
 import sys, hashlib
+from collections import namedtuple
 if sys.version_info < (3, 6):
     import sha3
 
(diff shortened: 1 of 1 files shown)


manticore/platforms/evm.py
```diff
@@ -21,6 +22,8 @@
 TT255 = 2 ** 255
 TOOHIGHMEM = 0x1000
 
+PendingTransaction = namedtuple("PendingTransaction", ['sort', 'address', 'origin', 'price', 'data', 'caller', 'value', 'return_data', 'result'])
```
mossberg: i think some of these field names are not quite correct. if you look at the line where this tuple is destructured (https://github.com/trailofbits/manticore/blob/master/manticore/platforms/evm.py#L2306), we can see that `sort` -> `type`, `return_data` -> `bytecode`, `result` -> `header` would be more appropriate names :)

arunjohnkuruvilla: I followed the field names used in the definition of Transaction at https://github.com/trailofbits/manticore/blob/master/manticore/platforms/evm.py#L31

mossberg: that is a good idea! but the elements of a PendingTransaction do not exactly correspond to the constructor arguments of a Transaction :)

PR #777, opened 2018-02-28
Fix Cpu.execute's improper reporting of `Unimplemented instruction`

disconnect3d:
When particular instruction's implementation raised an `AttributeError` it was catched by `Cpu.execute` and we just saw "oh this instruction is not implemented" while it was, but its implementation was broken.

(Found that when playing with adding `MOV` for aarch64 :P)

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/777)

### DIFF
--- a/manticore/core/cpu/abstractcpu.py
+++ b/manticore/core/cpu/abstractcpu.py
@@ -816,19 +816,23 @@ class Cpu(Eventful):
 
         name = self.canonicalize_instruction_name(insn)
             
-        if logger.level == logging.DEBUG :
+        if logger.level == logging.DEBUG:
             logger.debug(self.render_instruction(insn))
             for l in self.render_registers():
                 register_logger.debug(l)
 
         try:
-            try:
-                getattr(self, name)(*insn.operands)
-            except AttributeError:
-                text_bytes = ' '.join('%02x'%x for x in insn.bytes)
+            implementation = getattr(self, name, None)
+
+            if implementation is not None:
+                implementation(*insn.operands)
+
+            else:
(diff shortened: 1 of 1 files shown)


yan:
There are a few other instances in Manticore when we are more general than we should be with exception types.  Thanks for starting to address that!

---

Reviewed 1 of 1 files at r1.

---

*[manticore/core/cpu/abstractcpu.py, line 825 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/777#-L6Sluvc-KciGaG1Tya5:-L6Sluvc-KciGaG1Tya6:b2aika8) ([raw file](https://github.com/trailofbits/manticore/blob/1ee81c82532c09c663b709fbc4a5ef43929f5c64/manticore/core/cpu/abstractcpu.py#L825)):*
> ```Python
> 
>         try:
>             instr = getattr(self, name, None)
> ```

`insn` is used above to for the Capstone instruction object; rename to `impl` or `implementation` to resolve any ambiguity?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/777#-:-L6Sm7Mve1tNn20EGYQ5:b-h5vc6k)*

disconnect3d:
@yan good catch, changed to `implementation` :+1:

yan:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/777#-:-L6XlXuE2qRQrTH9PQq5:bnfp4nl)*

PR #778, opened 2018-02-28
Add assertion msg when target file is not present

disconnect3d:
Its just better to have a clear message intead of pure `AssertionError` with code:
```
mant(angr) [dc@dc:~]$ manticore asd
  File "/home/dc/.virtualenvs/angr/bin/manticore", line 11, in <module>
    load_entry_point('manticore', 'console_scripts', 'manticore')()
  File "/home/dc/Projects/manticore/manticore/__main__.py", line 110, in main
    m = Manticore(args.argv[0], argv=args.argv[1:], env=env, workspace_url=args.workspace,  policy=args.policy, disasm=args.disasm)
  File "/home/dc/Projects/manticore/manticore/manticore.py", line 171, in __init__
    assert os.path.isfile(path_or_state), "The file or state %s doesn't exist" % path_or_state
AssertionError: The file or state asd doesn't exist
```

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/778)

### DIFF
--- a/manticore/manticore.py
+++ b/manticore/manticore.py
@@ -168,10 +168,13 @@ class Manticore(Eventful):
         self.forward_events_from(self._executor)
 
         if isinstance(path_or_state, str):
-            assert os.path.isfile(path_or_state)
+            if not os.path.isfile(path_or_state):
+                raise Exception('{} is not an existing regular file'.format(path_or_state))
             self._initial_state = make_initial_state(path_or_state, argv=argv, **kwargs)
         elif isinstance(path_or_state, State):
             self._initial_state = path_or_state
+        else:
+            raise TypeError('path_or_state must be either a str or State, not {}'.format(type(path_or_state).__name__))
 
         if not isinstance(self._initial_state, State):
             raise TypeError("Manticore must be intialized with either a State or a path to a binary")


mossberg:
This is excellent; the codebase tends to overuse assertions, even when they're not the best form of error handling, so it's great to see help on this 👍

manticore/manticore.py
```diff
@@ -168,7 +168,7 @@ def __init__(self, path_or_state, argv=None, workspace_url=None, policy='random'
         self.forward_events_from(self._executor)
 
         if isinstance(path_or_state, str):
-            assert os.path.isfile(path_or_state)
+            assert os.path.isfile(path_or_state), "The file or state %s doesn't exist" % path_or_state
```
mossberg: could we raise an `Exception` instead for this, with that error message? I think that would be a little more appropriate than an assertion failure, and a little more pythonic. ideally we wouldn't raise a plain `Exception`, but we haven't gotten around to creating a Manticore exception hierarchy, so for the time being, I think `Exception` will be fine?

disconnect3d: @mossberg Yeah, `Exception`s are much better. We can use `Exception` for now. Fixed.

manticore/manticore.py
```diff
@@ -168,10 +168,13 @@ def __init__(self, path_or_state, argv=None, workspace_url=None, policy='random'
         self.forward_events_from(self._executor)
 
         if isinstance(path_or_state, str):
-            assert os.path.isfile(path_or_state)
+            if not os.path.isfile(path_or_state):
+                raise Exception("The file %s doesn't exist" % path_or_state)
```
mossberg: can we please change the phrasing: "{} is not an existing regular file".format(path_or_state) ? it will be a little more accurate with the specification for `os.path.isfile`

i also think we should try to use the `str.format()` style of string interpolation; i think it is a little more modern :)

manticore/manticore.py
```diff
@@ -168,10 +168,13 @@ def __init__(self, path_or_state, argv=None, workspace_url=None, policy='random'
...
             self._initial_state = make_initial_state(path_or_state, argv=argv, **kwargs)
         elif isinstance(path_or_state, State):
             self._initial_state = path_or_state
+        else:
+            raise TypeError('path_or_state must be either str or State while it is %s' % type(path_or_state))
```
mossberg: cool, I like the idea of being more explicit here, rather than relying on the below TypeError which is a slightly more indirect error message. 

for phrasing, this is SO nitpicky, but can we pls match the phrasing used by the `Manticore.add_hook` typecheck? 
https://github.com/trailofbits/manticore/blob/master/manticore/manticore.py#L457
"path_or_state must be either a str or State, not {}".format(type(path_or_state).__name__)
(also note how I did the `__name__` at the end, which should be a little nicer output)
it's very nitpicky, but i think attention to detail like this will be good as we strive to make manticore a more mature project :)

disconnect3d:
@mossberg all fixed:
* rebased to master
* rephrased/reworded the error messages
* using `str.format` instead of clunky `str % (format,)`

> i also think we should try to use the str.format() style of string interpolation; i think it is a little more modern :)

Huh, I was fighting for `str.format` for a long time but then I stopped :P. Both ways are okay for me and I am fine to accept one over another o/.

mossberg:
whoops i was too quick, looks like a test is failing 😭

PR #779, opened 2018-02-28
Unicorn emulator: get uc_arch & uc_mode values just once

disconnect3d:
This probably also speeds up a bit execution when we use emulation.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/779)

### DIFF
--- a/manticore/utils/emulate.py
+++ b/manticore/utils/emulate.py
@@ -34,24 +34,33 @@ class UnicornEmulator(object):
         # instruction
         self._should_be_written = {}
 
-    def reset(self):
-        self._emu = self._unicorn()
-        self._to_raise = None
-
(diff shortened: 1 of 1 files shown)


yan:
First, thank you again for this pr and the recent cleanup work you've been submitting! Regarding this one specifically, we've thought a bit about the "inline nested conditional" pattern and decided that it was cleaner to just use explicit `if`/`elif`/`else` to build up values. This pattern is also more common in Python's stdlib. 

However, creating the `Uc` argument tuple separately and creating `Uc` in one place is still the way to go. If you agree, we can update this PR and merge, or just bundle this update with the AArch64 pr.

Thoughts?

disconnect3d:
Changed to if + mappings.

PR #787, opened 2018-03-02
Fixes for dynargs PR

mossberg:
This address PR review comments and fixes some bugs present in the original branch #715 :

- ABI._consume_type buggs regarding offset math with address[]
- ArrayProxy.__getitem__ incorrectly calls the constructor of ArrayVariable, resulting in TypeError

```
In [1]: from manticore.core.smtlib import *

In [2]: a = ArrayProxy(ArrayVariable(32, None, 8, 's'))

In [3]: a
Out[3]: <ArrayProxy at 7f5e6527a850>

In [4]: a[2:3]
TypeError                                 Traceback (most recent call last)
<ipython-input-4-a518c90fbe21> in <module>()
----> 1 a[2:3]

/mnt/hgfs/code/manticore/manticore/core/smtlib/expression.py in __getitem__(self, index)
    715                 name='%s_sliced_b%d_e%d'%(self.name, start, stop)
    716
--> 717             new_array_var = ArrayVariable(self.index_bits, size, name=name, taint=self.taint)
    718             new_array = ArrayProxy(new_array_var)
    719             for i in xrange(size):

TypeError: __init__() takes at least 5 arguments (4 given)


```

This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/787)


manticore/platforms/evm.py
```diff
@@ -6,7 +6,7 @@
     SolverException
 from ..core.state import ForkState, TerminateState
 from ..utils.event import Eventful
-from ..core.smtlib.visitors import pretty_print, arithmetic_simplifier, translate_to_smtlib
+from ..core.smtlib.visitors import pretty_print, arithmetic_simplify, translate_to_smtlib
```
feliam: Maybe we should use a simplifier with constant folding in the future. (so we remember) https://github.com/trailofbits/manticore/blob/dev-evm-arraystorage/manticore/core/smtlib/visitors.py#L483-L486

mossberg:
@feliam this is still WIP but i'd specifically like your thoughts on this commit: https://github.com/trailofbits/manticore/pull/787/commits/3c75b9f6b8035311ce3b156bb3087e142efc7f06

I tried to explain the bug I ran into in the commit description. `related_to()` is an optimization in that case, yes?

manticore/core/smtlib/solver.py
```diff
@@ -330,8 +330,7 @@ def can_be_true(self, constraints, expression):
 
         with constraints as temp_cs:
             temp_cs.add(expression)
-
-            self._reset(temp_cs.related_to(expression))
```
feliam: If the expression has no variables, then it should be simplificable to a Constant. That's True or False.
In the False case the other constrains do not matter and the check should return False.
In the True case we'd need to check if the previous constraints set is feasible.
Something similar is fixed here: https://github.com/trailofbits/manticore/blob/dev-evm-arraystorage/manticore/core/smtlib/solver.py#L321-L335

mossberg:
There is something in this branch that causes `test_eth_regression` to take significantly longer than it should, resulting in Travis timeouts :( (but yay our tests are working)

mossberg:
https://github.com/trailofbits/manticore/commit/3c75b9f6b8035311ce3b156bb3087e142efc7f06 removed an optimization that had some negative affect on perf, but that doesn't seem to be the main cause.

mossberg:
Some notes from my work on debugging this:

I have this local diff in place to see how long each of the regression tests take.
```
    def test_eth_regressions(self):
        contracts = [676, 678, 701, 714, 735, 760]
        for contract in contracts:
            # sys.stderr.write('~')  # output something to keep travis going
            sys.stderr.write('running {}'.format(contract))
            start = time.time()
            self._simple_cli_run('{}.sol'.format(contract))
            sys.stderr.write(' took {} seconds\n'.format(time.time() - start))

```

initial data


```
[I] mark ubuntu /m/h/c/manticore (master) ❯ time nosetests tests/test_binaries.py:IntegrationTest.test_eth_regressions                                                                                                             ❮ 08:20
running 676 took 4.91237092018 seconds
running 678 took 3.5155479908 seconds
running 701 took 31.8618030548 seconds
running 714 took 28.65097785 seconds
running 735 took 25.0499229431 seconds
running 760 took 9.4359369278 seconds
.
----------------------------------------------------------------------
Ran 1 test in 103.430s





[I] mark ubuntu /m/h/c/manticore (dev-dynarg-rev) ❯ time nosetests tests/test_binaries.py:IntegrationTest.test_eth_regressions                                                                                                     ❮ 08:22
running 676 took 12.1561341286 seconds
running 678 took 9.24524402618 seconds
running 701 took 308.486123085 seconds
running 714 took 28.6029031277 seconds
running 735 took 24.923730135 seconds
running 760 took 9.24973893166 seconds
.
----------------------------------------------------------------------
Ran 1 test in 392.666s
```

i've tested commenting out and rerunning a number of parts of this PR to see if they make a difference. like i said above, the removal of the smt optimization did contribute to some amount of slowdown, but definitely not the whole thing. my current suspicions are towards the changes in evm.py in functions like _allocate and _consume

to test how the `dev-evm-dynamicarguments` branch does, you'll need this patch

```
diff --git a/manticore/core/smtlib/expression.py b/manticore/core/smtlib/expression.py
index 2d26503..c79049c 100644
--- a/manticore/core/smtlib/expression.py
+++ b/manticore/core/smtlib/expression.py
@@ -714,7 +714,7 @@ class ArrayProxy(Array):
             else:
                 name='%s_sliced_b%d_e%d'%(self.name, start, stop)

-            new_array_var = ArrayVariable(self.index_bits, size, name=name, taint=self.taint)
+            new_array_var = ArrayVariable(self.index_bits, size, self.value_bits, name=name, taint=self.taint)

```

mossberg:
@feliam 
> If the expression has no variables, then it should be simplificable to a Constant. That's True or False.
In the False case the other constrains do not matter and the check should return False.
In the True case we'd need to check if the previous constraints set is feasible.
Something similar is fixed here: https://github.com/trailofbits/manticore/blob/dev-evm-arraystorage/manticore/core/smtlib/solver.py#L321-L335

The issue that is fixed in this PR is not fixed in dev-evm-arraystorage, see the below example which was run on dev-evm-arraystorage. the `can_be_true` call should return false, yet it returns true despite both there being a concrete `False` constraint in the ConstraintSet, and the `b == 0` expression also being false, since b is a `BitVecConstant`, which is not equal to 0.
 
this is because the `b == 0` expression object has no `Variable` nodes in the tree, and the former implementation of `can_be_true` generated empty (trivially true) SMT in this case.

```
[I] mark ubuntu /m/h/c/manticore (dev-evm-arraystorage) ❯ python
Python 2.7.12 (default, Nov 20 2017, 18:23:56)
[GCC 5.4.0 20160609] on linux2
>>> b = BitVecConstant(32, 32)
>>> cs = ConstraintSet()
>>> cs.add(1 == 0)
>>> solver.can_be_true(cs, b == 0)
True
```

manticore/core/smtlib/solver.py
mossberg: the changes mentioned in the other branch do not fix the same issue that this fixes, so i will keep these changes. for more see https://github.com/trailofbits/manticore/pull/787#issuecomment-371642163

mossberg:
At this point, the performance issue (identified by @yan as the changes to _consume and _allocate) are really the only thing left to be resolved before we can fully review and merge 👍

mossberg:
imo, since the _consume symbolic gas changes cause an unacceptable performance hit, but are still useful changes, we should put them behind some kind of feature flag that is default off. i'll get started on this

yan:
Reviewed 1 of 9 files at r4, 3 of 3 files at r5, 2 of 3 files at r6, 2 of 2 files at r10.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[manticore/ethereum.py, line 1185 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77D7AAOhXd3_YK3HF9:-L77D7AAOhXd3_YK3HFA:b-qibw45) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/ethereum.py#L1185)):*
> ```Python
> 
>         space_for_each_arg = (space_for_arg_data/number_dyn_arguments) & (~0x1f)
>         assert space_for_each_arg % 32 == 0
> ```

Will never be false due to the masking

---

*[manticore/ethereum.py, line 1203 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77DsvDEwu2TLBqqbTM:-L77DsvDEwu2TLBqqbTN:b-fey66q) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/ethereum.py#L1203)):*
> ```Python
>             # FIXME TODO (mark) i'm not sure that the encoding for bytes and string is identical, this could
>             # be buggy for string
>             typ = types[index]
> ```

It might be cleaner to just get `typ` via the outer loop, i.e.: `for index, typ in zip(dyn_arguments, types):`.

---

*[manticore/ethereum.py, line 1211 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77AqIm2H0K5D17UVXd:-L77AqIm2H0K5D17UVXe:b-h1cvec) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/ethereum.py#L1211)):*
> ```Python
>             concrete_number_of_elements_in_arg = space_for_each_arg/element_size
>             state.constrain(number_of_elements_in_arg == concrete_number_of_elements_in_arg)
>             data[free_pointer:free_pointer + 32]= ("%064x"%(concrete_number_of_elements_in_arg)).decode('hex')
> ```

Extract this to a utility function, and use above as well? (And use `.format` instead of `%`? (format string  `{:064x}`))

---

*[manticore/core/state.py, line 305 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77APapgfJNrEyDiz52:-L77APapgfJNrEyDiz53:bmtnvla) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/core/state.py#L305)):*
> ```Python
> 
>     def must_be_true(self, expr):
>         return not self._solver.can_be_true(self._constraints, expr == False)
> ```

Leave note in docstring that this is used for tests? (To prevent future accidental removal)

---

*[manticore/core/smtlib/expression.py, line 689 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77AUPQ6uMJHMkUzE75:-L77AUPQ6uMJHMkUzE76:b8p9qgf) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/core/smtlib/expression.py#L689)):*
> ```Python
>     def _fix_index(self, index):
>         """
>         :param slice index:
> ```

Either add doc description, or remove this?

---

*[manticore/platforms/evm.py, line 1175 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77EbqjawSrKCibIG07:-L77EbqjawSrKCibIG08:b-ojxg3o) ([raw file](https://github.com/trailofbits/manticore/blob/23d893ba3948ef13741fd817640b5572287016ce/manticore/platforms/evm.py#L1175)):*
> ```Python
>     #Memory related
>     def _allocate(self, address):
>         cond = address > self.memory._allocated
> ```

Give this a more descriptive name? i.e. `need_to_allocate` or `is_oob` or `is_out_of_bounds`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/787)*

manticore/ethereum.py
```diff
@@ -1131,44 +1142,76 @@ def global_findings(self):
...
+
+            # FIXME TODO (mark) i'm not sure that the encoding for bytes and string is identical, this could
+            # be buggy for string
+            typ = types[index]
+            if typ not in ('bytes', 'string'):
+                element_size = 32
+            else:
+                element_size = 1
 
+            concrete_number_of_elements_in_arg = space_for_each_arg/element_size
```
feliam: In a future? We should ask the solver for a plausible solution instead of forcing/attempting a specific size which may not be feasible.

mossberg:
Review status: 6 of 8 files reviewed at latest revision, 8 unresolved discussions.

---

*[manticore/ethereum.py, line 1185 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77D7AAOhXd3_YK3HF9:-L7AlTFWbf3wWY3DRHeL:b-3l8l7p) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/ethereum.py#L1185)):*

Will never be false due to the masking
</blockquote></details>

do you think we should remove it? isn't the point of assertions that they should never be false? this is a somewhat important invariant that the below code depends on, and the assert will help for future proofing 

---

*[manticore/ethereum.py, line 1203 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77DsvDEwu2TLBqqbTM:-L7Am5tyAqC0ML2rW1T6:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/ethereum.py#L1203)):*

It might be cleaner to just get `typ` via the outer loop, i.e.: `for index, typ in zip(dyn_arguments, types):`.
</blockquote></details>

Done.

---

*[manticore/ethereum.py, line 1209 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L6cDcIC5iBzbIYpFdKn-r10-1209:-L7Ar7zV61cbbGHqxpec:btkxgtr) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/ethereum.py#L1209)):*

In a future? We should ask the solver for a plausible solution instead of forcing/attempting a specific size which may not be feasible.
</blockquote></details>

Ok. I think i kind of understand, but will need to follow up with you on this. tracking in #800.

---

*[manticore/ethereum.py, line 1211 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77AqIm2H0K5D17UVXd:-L7Aq_YF725TCMMErnRe:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/ethereum.py#L1211)):*

Extract this to a utility function, and use above as well? (And use `.format` instead of `%`? (format string  `{:064x}`))
</blockquote></details>

Done.

---

*[manticore/core/state.py, line 305 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77APapgfJNrEyDiz52:-L7AmdNAaK15cxQR2CGg:b-exhlt0) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/core/state.py#L305)):*

Leave note in docstring that this is used for tests? (To prevent future accidental removal)
</blockquote></details>

I don't think that's quite necessary, people should grep the codebase before removing functions :P also tests will break if it's removed anyway

---

*[manticore/core/smtlib/expression.py, line 689 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77AUPQ6uMJHMkUzE75:-L7AmyYmtPeEcstBS34I:bkt4zqw) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/core/smtlib/expression.py#L689)):*

Either add doc description, or remove this?
</blockquote></details>

I mostly added it to doc that index is of slice type.

---

*[manticore/platforms/evm.py, line 9 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L6cDcIC5iBzbIYpFdKo-r4-9:-L7As7pKS91SAyyix5Fc:betmfq3) ([raw file](https://github.com/trailofbits/manticore/blob/23d893ba3948ef13741fd817640b5572287016ce/manticore/platforms/evm.py#L9)):*

Maybe we should use a simplifier with constant folding in the future. (so we remember) https://github.com/trailofbits/manticore/blob/dev-evm-arraystorage/manticore/core/smtlib/visitors.py#L483-L486
</blockquote></details>

tracked on 801.

---

*[manticore/platforms/evm.py, line 1175 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77EbqjawSrKCibIG07:-L7Ar5s_z7deaayCnWRu:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/23d893ba3948ef13741fd817640b5572287016ce/manticore/platforms/evm.py#L1175)):*

Give this a more descriptive name? i.e. `need_to_allocate` or `is_oob` or `is_out_of_bounds`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/787)*

mossberg:
Review status: 6 of 8 files reviewed at latest revision, 7 unresolved discussions.

---

*[manticore/ethereum.py, line 1185 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L77D7AAOhXd3_YK3HF9:-L7BU6ZpCelAM-5VVxNz:bkanpaf) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/ethereum.py#L1185)):*

do you think we should remove it? isn't the point of assertions that they should never be false? this is a somewhat important invariant that the below code depends on, and the assert will help for future proofing 
</blockquote></details>

Done. added the new explicit aligned var

---

*[manticore/ethereum.py, line 1209 at r10](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L6cDcIC5iBzbIYpFdKn-r10-1209:-L7BUD0xMmCVR2qePO3g:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/8f31ea6ccfaaf160e4d4e9c85f953055e4af11ce/manticore/ethereum.py#L1209)):*

Ok. I think i kind of understand, but will need to follow up with you on this. tracking in #800.
</blockquote></details>

Done.

---

*[manticore/platforms/evm.py, line 9 at r4](https://reviewable.io:443/reviews/trailofbits/manticore/787#-L6cDcIC5iBzbIYpFdKo-r4-9:-L7BUB7bQPvsSmT9XNfH:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/23d893ba3948ef13741fd817640b5572287016ce/manticore/platforms/evm.py#L9)):*

tracked on 801.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/787)*

yan:
:lgtm:

---

Review status: 6 of 8 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/787#-:-L7BYaPHRqsN1KfHmnQH:bnfp4nl)*

yan:
Reviewed 1 of 2 files at r11, 1 of 1 files at r12.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/787)*

Issue #782, opened 2018-03-01
state.generate_testcase does not use the filename provided


montyly:
### Manticore version
Version: 0.1.7 (48f88be7590c7b15ff27cfd8c2da883d71e98190)

### Summary of the problem
`state.generate_testcase` does not use the filename provided to generate the files.
This was tested only for smart contracts.

### Step to reproduce the behavior
```
# simple.py
from manticore.ethereum import ManticoreEVM

m = ManticoreEVM() # initiate the blockchain
source_code ='''
pragma solidity^0.4.18;
contract Simple {
    function f(uint a) payable {
        if (a == 0x41) {
            throw;
        }
    }
}
'''

# Initiate the accounts
user_account = m.create_account(balance=1000)
contract_account = m.solidity_create_contract(source_code, owner=user_account, balance=0)

# Call f(a), with a symbolic value
contract_account.f(m.SValue)

# Check if an execution ends with a REVERT or INVALID
for state in m.terminated_states: 
    last_tx = state.platform.transactions[-1]
    if last_tx.result in ['REVERT','INVALID']:
        print "Error found in f() execution (see %s)"%m.workspace

        # filename should be named SimpleBug_00000000.tx
        state.generate_testcase('SimpleBug')
        break
```
`python simple.py` should generate files named "SimpleBug_*", but "test_*" is used insted

### Expected behavior
Generation of mcore_*/SimpleBug* files

### Actual behavior
Generation of mcore_*/test* files

### Any relevant logs
If I use ``m._generate_testcase_callback(state, 'SimpleBug','message')`` and this fix: https://github.com/montyly/manticore/commit/4f3a7ee059f6225bbe51275594e3c14a836a0795
 the files are correclty generated.

disconnect3d:
@montyly hey, since you already found a fix: with such simple things, next time you can just create the Pull Request and describe the issue there :).

ggrieco-tob:
Fixed on #791.

PR #783, opened 2018-03-01
Use name provided by ManticoreEVM._generate_testcase_callback to generate the testcase name

disconnect3d:
Fixes https://github.com/trailofbits/manticore/issues/782

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/783)

PR #791, opened 2018-03-05
Added a function to generate evm testcases without using callbacks

ggrieco-tob:
Possible fix for #782

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/791)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -1113,7 +1113,7 @@ class ManticoreEVM(Manticore):
     def workspace(self):
         return self._executor._workspace._store.uri
 
-    def generate_testcase(self, state, name, message):
+    def generate_testcase(self, state, name, message=''):
         self._generate_testcase_callback(state, name, message)
     
     def _generate_testcase_callback(self, state, name, message):
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -1113,6 +1113,9 @@ class ManticoreEVM(Manticore):
     def workspace(self):
         return self._executor._workspace._store.uri
 
+    def generate_testcase(self, state, name, message):
+        self._generate_testcase_callback(state, name, message)
+    
     def _generate_testcase_callback(self, state, name, message):
         '''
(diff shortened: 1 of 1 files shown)


mossberg:
thanks @ggrieco-tob can you please elaborate on this change? it will help me review it more quickly. at first glance it seems like it is adding a new API rather than fixing `state.generate_testcase`?

also what is the relation between this PR and #783 which has a very similar change in `_generate_testcase_callback`?

ggrieco-tob:
> thanks @ggrieco-tob can you please elaborate on this change? it will help me review it more quickly. at first glance it seems like it is adding a new API rather than fixing state.generate_testcase?

Sure!. We discussed with @feliam regarding the use of callbacks to generate testcases from states when manticore is not running (as @montyly shows in his example in #782). 
IMHO, we should avoid the direct use callbacks in that case, because I don't think it is possible to forward the events from the states to executor without pass such object to them. The most straightforward solution is to define `generate_testcase` in the Manticore/ManticoreEVM object to able to call directly the function used in the generate_testcase callback.

feliam:
I think I'm in favor of removing generate_case as an event. And let it be just a special case of terminate_state_callback. So each platform can decide what to generate a testcase or.



-- discouraged --
Though you can also just forward events on ManticoreEVM.load(). 
```
    def load(self, state_id=None):
        ''' Load one of the running or final states.
            
            :param state_id: If None it assumes there is a single running state
            :type state_id: int or None
        '''
        state = None
        if state_id is None:
            #a single state was assumed
            if self.count_running_states() == 1:  
                #Get the ID of the single running state              
                state_id = self._running_state_ids[0]
            else:
                raise Exception("More than one state running.")

        if state_id == -1:
            state = self.initial_state
        else:
            state = self._executor._workspace.load_state(state_id, delete=False)
            #froward events from newly loaded object
            self._executor.forward_events_from(state, True)
        return state
```
This way the events will be forwarded even if the state is loaded by the user outside the mainloop.

PR #793, opened 2018-03-07
Quick fix for #792 (EVM: STOP instruction returns random values)

feliam:
Sliced from experimental branch. Manage carefully.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/793)

Issue #792, opened 2018-03-07
EVM: STOP instruction returns random values


montyly:
### Manticore version
f4118f4f0f0db9ebcb97c08b103b04224bc795ee

### Summary of the problem
The STOP instruction returns an array of random values instead of 0

### Step to reproduce the behavior
```
# contract.sol
pragma solidity ^0.4.20;

contract Test{
    function () payable{
        assembly{
            stop
        }
    }
}
```

$ manticore contract.sol
m.main:INFO: Beginning analysis
m.ethereum:INFO: Starting symbolic transaction: 1
m.ethereum:INFO: Finished symbolic transaction: 1 | Code Coverage: 100% | Terminated States: 1 | Alive States: 1
m.ethereum:INFO: Generated testcase No. 0 - STOP
[..]
cat mcore_XXX/test_00000000.tx
[..]

Function call:
{fallback}() -> STOP 
return: ( ... garbage values ...)

### Expected behavior
STOP should return (0,0) (see http://solidity.readthedocs.io/en/develop/assembly.html)

### Actual behavior
STOP returns an array of random values

ggrieco-tob:
Fixed in fbf0823cb61aa7f3b417839a4e2e7e3f58352f10

PR #794, opened 2018-03-07
Very quick fix for #792 based on PR #793

ggrieco-tob:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/794)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -1213,6 +1213,7 @@ class ManticoreEVM(Manticore):
                         signature = metadata.get_func_signature(function_id)
                         function_name, arguments = ABI.parse(signature, tx.data)
 
+                        return_data = None
                         if tx.result == 'RETURN':
                             ret_types = metadata.get_func_return_types(function_id)
                             return_data = ABI.parse(ret_types, tx.return_data) #function return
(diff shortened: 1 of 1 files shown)


manticore/ethereum.py
```diff
@@ -1224,13 +1225,15 @@ def flagged(flag):
...
                         is_something_symbolic = is_something_symbolic or is_argument_symbolic
                         tx_summary.write(') -> %s %s\n' % ( tx.result, flagged(is_argument_symbolic)))
-                        is_return_symbolic = any(map(issymbolic, return_data))
-                        return_values = tuple(map(state.solve_one, return_data))
-                        if len(return_values) == 1:
-                            return_values = return_values[0]
 
-                        tx_summary.write('return: %r %s\n' % ( return_values, flagged(is_return_symbolic)))
-                        is_something_symbolic = is_something_symbolic or is_return_symbolic
+                        if return_data is not None:
```
mossberg: can we restructure this so that there is only one if block? move the `if tx.result == 'RETURN'` below the block above (the block that emits the "Function call:" text etc)?

mossberg:
@ggrieco-tob would you mind filling in a more descriptive PR title? the information about what issues are fixed etc should go in the PR description :)

PR #803, opened 2018-03-10
Introduce concretized_args decorator

yan:
EVM bytecode implementations previously manually raised `ConcretizeStack` exceptions when a certain argument needed to be concretized. This can be problematic -- the index could be misaligned from the actual argument. The `concretized_args` decorator makes this less error-prone, with additional error checking to make sure concretized arguments are mentioned in the function prototype and derives their index automatically.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/803)


manticore/platforms/evm.py
```diff
@@ -1049,6 +1051,41 @@ def __reduce__(self):
...
+    def concretizer(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            spec = inspect.getargspec(func)
+            sys.stderr.write("args: {}, kwargs: {}".format(repr(args), repr(kwargs)))
+            for arg, policy in policies.items():
+                assert arg in spec.args, "Concretizer argument not found in wrapped function."
+                policy = policy or 'MINMAX'
+                # index is 0-indexed, but ConcretizeStack is 1-indexed. However, this is correct
+                # since implementation methods always take a first `self` parameter.
```
disconnect3d: Its called `bound method` because the `self` is bound ;D
```
In [11]: class Cls:
    ...:     def a(self):
    ...:         print('a')
    ...:         
    ...:     

In [12]: Cls.a, Cls().a
Out[12]: 
(<function __main__.Cls.a>,
 <bound method Cls.a of <__main__.Cls object at 0x7fabe5ef5a90>>)

In [13]: Cls.a(Cls())
a

In [14]: Cls().a()
a

```

manticore/platforms/evm.py
```diff
@@ -1049,6 +1051,41 @@ def __reduce__(self):
...
+
+        @concretized_args(size='ONE')
+        def LOG(self, address, size, *topics):
+            ...
+
+    The above will make sure that the |size| parameter to LOG is Concretized when symbolic
+    according to the 'ONE' policy.
+
+    :param policies: A kwargs list of argument names and their respective policies.
+                         Provide None or '' as policy to use default.
```
disconnect3d: I'd rather require just one value to make it default. `None`?

Then it could be `policy = policy if policy is not None else 'MINMAX'`

manticore/platforms/evm.py
```diff
@@ -1049,6 +1051,41 @@ def __reduce__(self):
...
+    :return: A function decorator
+    '''
+    def concretizer(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            spec = inspect.getargspec(func)
+            sys.stderr.write("args: {}, kwargs: {}".format(repr(args), repr(kwargs)))
```
disconnect3d: Or just `'args: {!r}, kwargs: {!r}'.format(args, kwargs)`

btw https://pyformat.info/

manticore/platforms/evm.py
```diff
@@ -1049,6 +1051,41 @@ def __reduce__(self):
         return (self.__class__, (self.data, ))
 
 
+def concretized_args(**policies):
+    '''
```
disconnect3d: It seems we don't care but maybe triple quotes as [pep257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) suggests

> For consistency, always use """triple double quotes""" around docstrings. Use r"""raw triple double quotes""" if you use any backslashes in your docstrings. For Unicode docstrings, use u"""Unicode triple-quoted strings""".

manticore/platforms/evm.py
yan: I thought briefly about this, and I ended up defaulting it if falsifiable values are provided too to make the call site slightly easier. (i.e. `size=''` is more succinct than `size=None`).

manticore/platforms/evm.py
yan: That was an accidentally-committed debug output, which is now removed

mossberg:
how did you go about testing this? could we add a unit test?

---

Reviewed 1 of 1 files at r5.

---

*[manticore/platforms/evm.py, line 1069 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/803#-L7CPJ2M-i3Ig_jzAUgd-r1-1069:-L7a-_18ZzVQPBa4Knb9:b-oby0p7) ([raw file](https://github.com/trailofbits/manticore/blob/32ec85d6b8349e066415038ebf711d733a09f36f/manticore/platforms/evm.py#L1069)):*

I thought briefly about this, and I ended up defaulting it if falsifiable values are provided too to make the call site slightly easier. (i.e. `size=''` is more succinct than `size=None`).
</blockquote></details>

I think `size=''` looks really confusing at the callsite. I think if you want to use a default you should just specify a regular argument, not a keyword argument. For example `@concretize_args(arg1, arg2='SAMPLED')` (arg1 uses default, and arg2 uses the specific one).

---

*[manticore/platforms/evm.py, line 1096 at r5](https://reviewable.io:443/reviews/trailofbits/manticore/803#-L7a1--nmnqK_jEMuhR9:-L7a1--nmnqK_jEMuhRA:bsgzvz6) ([raw file](https://github.com/trailofbits/manticore/blob/482aa17992d4b0208ecc72abd5aabfc4f5031d23/manticore/platforms/evm.py#L1096)):*
> ```Python
>             for arg, policy in policies.items():
>                 assert arg in spec.args, "Concretizer argument not found in wrapped function."
>                 policy = policy or 'MINMAX'
> ```

I don't think it's this function's place to decide on what the default policy is exactly, I think it should defer to the Concretize exception for that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/803#-:-L7a3SblxYPvSBk-Cbta:b-w07ly9)*

yan:

---

*[manticore/platforms/evm.py, line 1069 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/803#-L7CPJ2M-i3Ig_jzAUgd-r1-1069:-L7a4gmp-21WkX8OV3mr:b-ve2m8z) ([raw file](https://github.com/trailofbits/manticore/blob/32ec85d6b8349e066415038ebf711d733a09f36f/manticore/platforms/evm.py#L1069)):*

I think `size=''` looks really confusing at the callsite. I think if you want to use a default you should just specify a regular argument, not a keyword argument. For example `@concretize_args(arg1, arg2='SAMPLED')` (arg1 uses default, and arg2 uses the specific one).
</blockquote></details>

`arg1` isn't in the definition scope though, that won't be valid Python. I definitely agree that it looks awkward, the non-kw arg would have been my first choice if it was permissible though.

---

*[manticore/platforms/evm.py, line 1096 at r5](https://reviewable.io:443/reviews/trailofbits/manticore/803#-L7a1--nmnqK_jEMuhR9:-L7a57MTADs55yper3J_:b-oz8tyz) ([raw file](https://github.com/trailofbits/manticore/blob/482aa17992d4b0208ecc72abd5aabfc4f5031d23/manticore/platforms/evm.py#L1096)):*

I don't think it's this function's place to decide on what the default policy is exactly, I think it should defer to the Concretize exception for that.
</blockquote></details>

Good call!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/803)*

mossberg:
:lgtm:

---

Reviewed 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[manticore/platforms/evm.py, line 1069 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/803#-L7CPJ2M-i3Ig_jzAUgd-r1-1069:-L7aASVrGUNprZcawm04:bvy1mkh) ([raw file](https://github.com/trailofbits/manticore/blob/32ec85d6b8349e066415038ebf711d733a09f36f/manticore/platforms/evm.py#L1069)):*

`arg1` isn't in the definition scope though, that won't be valid Python. I definitely agree that it looks awkward, the non-kw arg would have been my first choice if it was permissible though.
</blockquote></details>

ok ah i see, whoops my mistake!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/803#-:-L7aBbcllf0sHAbHdl4C:bnfp4nl)*

PR #811, opened 2018-03-14
Added all the uint types supported by Solidity

ggrieco-tob:
According to the [documentation](http://solidity.readthedocs.io/en/develop/types.html#integers), Solidity support signed and unsigned integer variables "*uint8 to uint256 in steps of 8 (unsigned of 8 up to 256 bits)*". This PR enables all these types.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/811)


manticore/ethereum.py
```diff
@@ -389,9 +389,11 @@ def simplify(x):
             value = arithmetic_simplifier(Operators.CONCAT(size, *map(Operators.ORD, data[offset + padding:offset + padding + byte_size])))
             return simplify(value)
 
-        if ty == u'uint256':
-            return get_uint(256, offset), offset + 32
-        elif ty in (u'bool', u'uint8'):
+        if ty.startswith('uint') and 0 <= int(ty[4:]) <= 256:
+            size = int(ty[4:])
+            assert((size % 8) == 0)
```
mossberg: Could we try to raise an Exception with a helpful error message rather than using an assert?

manticore/core/smtlib/constraints.py
```diff
@@ -205,7 +205,7 @@ def new_bitvec(self, size, name='V', taint=frozenset()):
                          if not uniq a numeric nonce will be appended
             :return: a fresh BitVecVariable
         '''
-        assert size in (1, 4, 8, 16, 32, 64, 128, 160, 256)
+        assert (size == 1 or size % 8 == 0)
```
mossberg: Similar to below, since we're looking at this code anyway, could we port it over to use an Exception instead of an assert?

disconnect3d:
I have no knowledge about solidity [yet?] but probably there are also diferent `int` types, so maybe it is worth to do similar thing with `elif ty == u'int256':`.

mossberg:
great point @disconnect3d, that would make sense to do in this PR also!

manticore/ethereum.py
ggrieco-tob: Should I use EthereumError?

mossberg: yes please!

ggrieco-tob:
@mossberg can you re-review the additional commits in this branch?

manticore/ethereum.py
```diff
@@ -389,17 +389,25 @@ def simplify(x):
             value = arithmetic_simplifier(Operators.CONCAT(size, *map(Operators.ORD, data[offset + padding:offset + padding + byte_size])))
             return simplify(value)
 
-        if ty == u'uint256':
-            return get_uint(256, offset), offset + 32
-        elif ty in (u'bool', u'uint8'):
+        if ty.startswith('uint') and 0 <= int(ty[4:]) <= 256:
+            size = int(ty[4:])
```
disconnect3d: It might not matter that much but it might be good idea to calculate `int(ty[4:])` just once.

ggrieco-tob: Uhm. Any suggestion on how to do that? 
I think I will need to split the two checks: `ty.startswith('uint')` and  `0 <= int(ty[4:]) <= 256` to do that, so the code will be slighter more complicated.

disconnect3d: Not really, seems like we have to split those :(

manticore/ethereum.py
```diff
@@ -389,17 +389,25 @@ def simplify(x):
             value = arithmetic_simplifier(Operators.CONCAT(size, *map(Operators.ORD, data[offset + padding:offset + padding + byte_size])))
             return simplify(value)
 
-        if ty == u'uint256':
-            return get_uint(256, offset), offset + 32
-        elif ty in (u'bool', u'uint8'):
+        if ty.startswith('uint') and 0 <= int(ty[4:]) <= 256:
```
mossberg: i think we should move this `0 <= ... <= 256` range check inside the `if` body and add it to the check where we raise an EthereumError. if we are given something like `uint3000` that is an invalid uint size, but we would raise a `NotImplementedError`

manticore/ethereum.py
```diff
@@ -389,17 +389,25 @@ def simplify(x):
...
+            size = int(ty[4:])
+            if not ((size % 8) == 0):
+                raise EthereumError('Invalid uint size %s' % size)
```
mossberg: actually this will crash since EthereumError is not defined here, we need to import it

manticore/ethereum.py
mossberg: same below with int

tests/EVM/test_EVMABI.py
```diff
@@ -0,0 +1,40 @@
+
+import unittest
+from manticore.ethereum import ABI, EthereumError
+
+class EVMTest_ABI(unittest.TestCase):
```
mossberg: stylistically, i prefer avoiding underscores in test class names. maybe something like "EthereumAbiTests"?

tests/EVM/test_EVMABI.py
```diff
@@ -0,0 +1,40 @@
...
+            for i in range(8, 257, 1):
+                try:
+                    datatype = "int"+str(i)
+                    data = "\xFF"*32
+                    abi.parse(datatype, data)
+                    self.assertEqual(i % 8, 0)
+                except EthereumError:
+                    self.assertNotEqual(i % 8, 0)
+
+if __name__ == '__main__':
```
mossberg: we can omit these lines, we generally don't directly run these python files, but rather use something like nosetests or `python -m unittest` to run them.

tests/EVM/test_EVMABI.py
mossberg: also instead of creating a new file in tests/EVM, please move this test class into test_eth.py. tests/EVM is mostly for the auto generated evm instruction tests :)

f2xeb:
*[manticore/core/smtlib/constraints.py, line 208 at r5](https://beta.reviewable.io/reviews/trailofbits/manticore/811#-L9LH2ow0gzdrWdApQ0Z:-L9LH2ow0gzdrWdApQ0_:b-9kwn6w) ([raw file](https://github.com/trailofbits/manticore/blob/be00f7ec2c30deed6e1854a865aa4ebb6a14897e/manticore/core/smtlib/constraints.py#L208)):*
> ```Python
>             :return: a fresh BitVecVariable
>         '''
>         if not (size == 1 or size % 8 == 0):
> ```

is 4 a valid uint size?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/811)*

mossberg:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[manticore/core/smtlib/constraints.py, line 208 at r5](https://beta.reviewable.io/reviews/trailofbits/manticore/811#-L9LH2ow0gzdrWdApQ0Z:-L9LI5qF2yRBJ4ibIazV:bj1duoi) ([raw file](https://github.com/trailofbits/manticore/blob/be00f7ec2c30deed6e1854a865aa4ebb6a14897e/manticore/core/smtlib/constraints.py#L208)):*
<details><summary><i>Previously, f2xeb (feeb) wrote…</i></summary><blockquote>

is 4 a valid uint size?
</blockquote></details>

hm nope uint4/int4 is not a valid solidity type. i think we're fine with not allowing bitvecs of 4 bits to be created with this interface, i can't see any reason to do that really. sizes that are multiples of 8 are by far the main users of this i think

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/811)*

f2xeb:
Reviewed 1 of 2 files at r3.
Review status: 1 of 3 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/811)*

f2xeb:
:lgtm:

---

Review status: 1 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/811#-:-L9LQusf5B2jAJW7Q2vI:bnfp4nl)*

mossberg:
Reviewed 2 of 3 files at r5.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/811)*

mossberg:
i kind of want to wait until PR 855 is merged before merging this

PR #815, opened 2018-03-15
Graceful exit when no states to run

### LINKED ISSUE #795: Assertion failed: context['_saved_states'] or self.initial_state is not None
### OS / Environment

Distributor ID:	ArchLinux
Description:	ArchLinux
Release:	rolling-release

### Manticore version

Last revision (fbf0823cb61aa7f3b417839a4e2e7e3f58352f10)

### Python version

Python 2.7.14

### Summary of the problem

Manticore crashes with a failed assertion when there is no more live state to run.

This is a small python script to reproduce this issue:

```python

# initiate the blockchain
m = ManticoreEVM()
source_code = '''
contract Simple {
    function f(uint a) payable public {
        if (a == 65) {
            revert();
        }
    }
}
'''

# Initiate the accounts
user_account = m.create_account(balance=1000)
contract_account = m.solidity_create_contract(source_code, owner=user_account, balance=0)

contract_account.f(1)   # it works
contract_account.f(65) # it works
contract_account.f(m.SValue) # fails with assert context['_saved_states'] or self.initial_state is not None                                                  
```

### Expected behavior

It shouldn't crash. 

### Actual behavior

```
  File "work.py", line 22, in <module>
    contract_account.f(m.SValue)
  File "/home/gustavo/repos/manticore-vanilla/manticore/ethereum.py", line 508, in f
    data=tx_data
  File "/home/gustavo/repos/manticore-vanilla/manticore/ethereum.py", line 887, in transaction
    status = self.run(procs=self._config_procs)
  File "/home/gustavo/repos/manticore-vanilla/manticore/ethereum.py", line 940, in run
    assert context['_saved_states'] or self.initial_state is not None
AssertionError
```

### LINKED ISSUE #802: ManticoreEVM.transaction should safely return if there are no running states
This is in reference to this comment/diff: https://github.com/trailofbits/manticore/pull/715#discussion_r164883644

That diff which added the `if not self.running_state_ids` to `multi_tx_analysis` was removed from Pr #715 since that was somewhat irrelevant to the dynargs feature. A separate pr should be created to handle this situation in the Manticore core; not the the `multi_tx_analysis` function, which is a very high level function, that is essentially user code. Handling it in the core would mean users of the api that are manually calling `.transaction` would also benefit.

currently we raise an assertion error if there are no running states and no initial state
https://github.com/trailofbits/manticore/blob/master/manticore/ethereum.py#L940

mossberg:
Fixes #795 
Fixes #802 

Rather than assertion error'ing out when there are no states, to run, simply return.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/815)

### DIFF
--- a/tests/test_eth.py
+++ b/tests/test_eth.py
@@ -4,7 +4,7 @@ import os
 from manticore.core.smtlib import ConstraintSet, operators
 from manticore.core.smtlib.expression import BitVec
 from manticore.core.state import State
-from manticore.ethereum import ManticoreEVM, IntegerOverflow, Detector
+from manticore.ethereum import ManticoreEVM, IntegerOverflow, Detector, NoAliveStates
 from manticore.platforms.evm import EVMWorld, ConcretizeStack, concretized_args
 
 
@@ -114,7 +114,8 @@ class EthTests(unittest.TestCase):
 
         contract_account.f(1)  # it works
         contract_account.f(65)  # it works
-        contract_account.f(m.SValue)  # no alive states, but try to run a tx anyway
+        with self.assertRaises(NoAliveStates):
+            contract_account.f(m.SValue)  # no alive states, but try to run a tx anyway
 
     def test_can_create(self):
         mevm = ManticoreEVM()
--- a/tests/test_eth.py
+++ b/tests/test_eth.py
@@ -92,6 +92,30 @@ class EthTests(unittest.TestCase):
         self.assertIn('STOP', context)
         self.assertIn('REVERT', context)
 
+    def test_graceful_handle_no_alive_states(self):
+        """
+        If there are no alive states, or no initial states, we should not crash. issue #795
+        """
+        # initiate the blockchain
+        m = ManticoreEVM()
+        source_code = '''
+        contract Simple {
+            function f(uint a) payable public {
+                if (a == 65) {
+                    revert();
+                }
+            }
+        }
+        '''
+
+        # Initiate the accounts
+        user_account = m.create_account(balance=1000)
+        contract_account = m.solidity_create_contract(source_code, owner=user_account, balance=0)
+
+        contract_account.f(1)  # it works
+        contract_account.f(65)  # it works
+        contract_account.f(m.SValue)  # no alive states, but try to run a tx anyway
+
     def test_can_create(self):
         mevm = ManticoreEVM()
         source_code = """
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -939,7 +939,10 @@ class ManticoreEVM(Manticore):
         current_coverage = 0
 
         while current_coverage < 100:
-            run_symbolic_tx()
+            try:
+                run_symbolic_tx()
+            except NoAliveStates:
+                break
 
             if tx_limit is not None:
                 tx_limit -= 1
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -890,11 +890,17 @@ class ManticoreEVM(Manticore):
             :type value: int or SValue
             :param data: initial data
             :return: an EVMAccount
+            :raises NoAliveStates: if there are no alive states to execute
         '''
         if isinstance(address, EVMAccount):
             address = int(address)
         if isinstance(caller, EVMAccount):
             caller = int(caller)
+
+        with self.locked_context('seth') as context:
+            has_alive_states = context['_saved_states'] or self.initial_state is not None
+        if not has_alive_states:
+            raise NoAliveStates
 
         if isinstance(data, self.SByte):
             data = (None,) * data.size
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -27,8 +27,10 @@ logger = logging.getLogger(__name__)
 class EthereumError(ManticoreError):
     pass
 
+
 class NoAliveStates(EthereumError):
     pass
+
 
 class Detector(Plugin):
     @property
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -948,7 +948,7 @@ class ManticoreEVM(Manticore):
         with self.locked_context('seth') as context:
             assert context['_pending_transaction'] is not None
 
-            if not context['_saved_states'] or self.initial_state is None:
+            if not context['_saved_states'] and self.initial_state is None:
                 return
 
             # there is no states added to the executor queue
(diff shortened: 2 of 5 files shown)


tests/test_eth.py
```diff
@@ -92,6 +92,30 @@ def did_evm_execute_instruction_callback(self, state, instruction, arguments, re
...
+        }
+        '''
+
+        # Initiate the accounts
+        user_account = m.create_account(balance=1000)
+        contract_account = m.solidity_create_contract(source_code, owner=user_account, balance=0)
+
+        contract_account.f(1)  # it works
+        contract_account.f(65)  # it works
+        contract_account.f(m.SValue)  # no alive states, but try to run a tx anyway
```
feliam: I'd expect this to raise an exception. No?

mossberg: Hm, so i originally was thinking that it makes sense for ManticoreEVM.run() to simply return if there are no states to analyze. The run() function executes the various alive states. If there are no alive states, there is nothing to do, so just return!

However, the question of whether or not executing a transaction when there are no states is a higher level question, and a different one. I think you're right; trying to execute a transaction when there are no states should raise an Exception. it could be confusing for a user, if they try to execute transactions and it appears to 'work', but in reality manticore is doing nothing.

mossberg: I still like keeping the behavior of .run() mentioned above, but I will include a check at a different layer to implement the exception.

feliam: So running over no states is not an exception.

yan:
Reviewed 2 of 4 files at r1, 3 of 3 files at r2.

---

*[manticore/ethereum.py, line 3 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/815#-L7ju-Qqwwf0FiQipW4g:-L7ju-Qqwwf0FiQipW4h:bef3gwo) ([raw file](https://github.com/trailofbits/manticore/blob/6dc96ad50e4af69a71671059f7453e2b7ee574ab/manticore/ethereum.py#L3)):*
> ```Python
> import string
> ```

Extra \n?

---

*[manticore/ethereum.py, line 31 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/815#-L7ju1HFpeTgFQOtW_wj:-L7ju1HFpeTgFQOtW_wk:by4k9wf) ([raw file](https://github.com/trailofbits/manticore/blob/6dc96ad50e4af69a71671059f7453e2b7ee574ab/manticore/ethereum.py#L31)):*
> ```Python
> 
> 
> class NoAliveStates(EthereumError):
> ```

Would this ever be promoted to `MaticoreError` or moved elsewhere? Not having eligible states sounds like a more generic exception.

---

*[manticore/ethereum.py, line 904 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/815#-L7juBdixzohab1zI62A:-L7juBdixzohab1zI62B:b85mwv7) ([raw file](https://github.com/trailofbits/manticore/blob/6dc96ad50e4af69a71671059f7453e2b7ee574ab/manticore/ethereum.py#L904)):*
> ```Python
>         with self.locked_context('seth') as context:
>             has_alive_states = context['_saved_states'] or self.initial_state is not None
>         if not has_alive_states:
> ```

Even understanding Python's function scopes, this indentation feels weird. Raise from within the `with` block?

---

*[manticore/manticore.py, line 36 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/815#-L7juaKq85MRenpo7QCJ:-L7juaKq85MRenpo7QCK:bl63dsm) ([raw file](https://github.com/trailofbits/manticore/blob/6dc96ad50e4af69a71671059f7453e2b7ee574ab/manticore/manticore.py#L36)):*
> ```Python
> 
> 
> class ManticoreError(Exception):
> ```

Add a docstring summarizing this exception?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/815)*

mossberg:

---

*[manticore/ethereum.py, line 3 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/815#-L7ju-Qqwwf0FiQipW4g:-L7k9wb20bbx1sUxOJB8:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/6dc96ad50e4af69a71671059f7453e2b7ee574ab/manticore/ethereum.py#L3)):*

Extra \n?
</blockquote></details>

Done.

---

*[manticore/ethereum.py, line 31 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/815#-L7ju1HFpeTgFQOtW_wj:-L7kA1_g9pOzeyKYfHDY:b-ba160q) ([raw file](https://github.com/trailofbits/manticore/blob/6dc96ad50e4af69a71671059f7453e2b7ee574ab/manticore/ethereum.py#L31)):*

Would this ever be promoted to `MaticoreError` or moved elsewhere? Not having eligible states sounds like a more generic exception.
</blockquote></details>

Maybe. This exception is actually somewhat unique to ethereum. For linux usage, users do not really have a way to trigger a symbolic exec run when there are no alive states existing. This is because for linux .run() is guaranteed to have an initial state, and .run() is not really invoked more than one.  Things are different for ethereum where users can manually trigger .transaction() at will, even in cases when there are no states to be processed. so this exception is somewhat unique to ethereum. it's possible we could move it, but i think this is fine for now.

---

*[manticore/ethereum.py, line 904 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/815#-L7juBdixzohab1zI62A:-L7kBalyB8phs2RUloYL:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/6dc96ad50e4af69a71671059f7453e2b7ee574ab/manticore/ethereum.py#L904)):*

Even understanding Python's function scopes, this indentation feels weird. Raise from within the `with` block?
</blockquote></details>

Done.

---

*[manticore/manticore.py, line 36 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/815#-L7juaKq85MRenpo7QCJ:-L7k9vnPAJW76jli75u8:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/6dc96ad50e4af69a71671059f7453e2b7ee574ab/manticore/manticore.py#L36)):*

Add a docstring summarizing this exception?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/815)*

yan:
:lgtm:

---

Reviewed 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/815#-:-L7kDCHgKvUhMbHyJXWm:bnfp4nl)*

PR #817, opened 2018-03-15
evm: add per state trace file

mossberg:
previously, we would report the 'visited.txt' which is the complete coverage that all states achieved, but it's often desirable to see what trace a particular state took. this implements some changes to the workspace files. we now emit a `.init.trace` and `.rt.trace` for the init and runtime traces for the state.

open to aesthetic suggestions for the divider. the '---' helps separate the individual traces of the various tranasactions within the file.

the colon delimiter was specifically chosen as it is ethersplay compatible https://github.com/trailofbits/ethersplay/blob/8c8934a03b4dc50ec6145573f7c35583d1213918/ethersplay/coverage.py#L17

this pr was expanded to also correct bugs in the trace recording (it was missing the last instruction of a trace (return/stop/revert)), and adds a test to ensure the trace files include those last instructions

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/817)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -129,9 +129,8 @@ class UninitializedStorage(Detector):
         state.context.setdefault('seth.detectors.initialized_storage', set()).add(offset)
 
 
-def calculate_coverage(code, seen):
-    ''' Calculates what percentage of code has been seen '''
-    runtime_bytecode = code
+def calculate_coverage(runtime_bytecode, seen):
+    ''' Calculates what percentage of runtime_bytecode has been seen '''
     end = None
     if ''.join(runtime_bytecode[-44: -34]) == '\x00\xa1\x65\x62\x7a\x7a\x72\x30\x58\x20' \
             and ''.join(runtime_bytecode[-2:]) == '\x00\x29':
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -1272,6 +1272,12 @@ class ManticoreEVM(Manticore):
                 logger.debug("Using iterpickle to dump state")
                 statef.write(iterpickle.dumps(state, 2))
 
+        with testcase.open_stream('trace') as f:
+            for contract, pc in state.context['seth.trace']:
+                if pc == 0:
+                    f.write('---\n')
+                ln = '0x{:x} 0x{:x}\n'.format(contract, pc)
+                f.write(ln)
     def finalize(self):
         # move runnign states to final states list
         # and generate a testcase for each
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -1276,8 +1276,9 @@ class ManticoreEVM(Manticore):
             for contract, pc in state.context['seth.trace']:
                 if pc == 0:
                     f.write('---\n')
-                ln = '0x{:x} 0x{:x}\n'.format(contract, pc)
+                ln = '0x{:x}:0x{:x}\n'.format(contract, pc)
                 f.write(ln)
+
     def finalize(self):
         # move runnign states to final states list
         # and generate a testcase for each
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -1091,8 +1091,12 @@ class ManticoreEVM(Manticore):
             coverage.add((state.platform.current.address, state.platform.current.pc))
 
     def _did_execute_instruction_callback(self, state, prev_pc, pc, instruction):
-        ''' INTERNAL USE '''
-        state.context.setdefault('seth.trace', []).append((state.platform.current.address, prev_pc))
+        if 'Call' in str(type(state.platform.current.last_exception)):
+            trace_context_name = 'seth.rt.trace'
+        else:
+            trace_context_name = 'seth.init.trace'
+
+        state.context.setdefault(trace_context_name, []).append((state.platform.current.address, prev_pc))
 
     def _did_evm_read_code(self, state, offset, size):
         ''' INTERNAL USE '''
@@ -1144,7 +1148,7 @@ class ManticoreEVM(Manticore):
         with testcase.open_stream('summary') as summary:
             summary.write("Last exception: %s\n" % state.context['last_exception'])
 
-            address, offset = state.context['seth.trace'][-1]
+            address, offset = state.context['seth.rt.trace'][-1]
 
             # Last instruction
             metadata = self.get_metadata(blockchain.transactions[-1].address)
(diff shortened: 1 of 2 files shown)


manticore/ethereum.py
```diff
@@ -1272,6 +1272,13 @@ def flagged(flag):
                 logger.debug("Using iterpickle to dump state")
                 statef.write(iterpickle.dumps(state, 2))
 
+        with testcase.open_stream('trace') as f:
```
feliam: Search for "runtime_coverage" I think you'll need similar trick to separate init from runtime trace?

yan:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[manticore/ethereum.py, line 1096 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/817#-L7jvIZaQi7Vl9tLeA9Y:-L7jvIZaQi7Vl9tLeA9Z:b6th3p0) ([raw file](https://github.com/trailofbits/manticore/blob/74f91b38117acc027c7a135f699905bd9c13b45d/manticore/ethereum.py#L1096)):*
> ```Python
>         # TODO(mark): here and above, we need a cleaner way to determine, from this class,
>         # whether the EVM cpu is executing init or runtime bytecode
>         if 'Call' in str(type(state.platform.current.last_exception)):
> ```

Why not `isinstance`? What's stopping us from adding a platform flag such as `in_init` that is toggled?

---

*[manticore/ethereum.py, line 1184 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/817#-L7jvP7TjmNtosbb2KYj:-L7jvP7TjmNtosbb2KYk:b6wkvi0) ([raw file](https://github.com/trailofbits/manticore/blob/74f91b38117acc027c7a135f699905bd9c13b45d/manticore/ethereum.py#L1184)):*
> ```Python
> 
>                 runtime_code = blockchain.get_code(account_address)
>                 if len(runtime_code):
> ```

Why not just `if runtime_code:`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/817)*

mossberg:

---

*[manticore/ethereum.py, line 1275 at r1](https://reviewable.io:443/reviews/trailofbits/manticore/817#-L7g56YOZTFSZObHEtpL-r1-1275:-L7kEHZXSEJtf3faESxY:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/e6eab14bd8f1a2394f0adb4c323dc5843d9dc707/manticore/ethereum.py#L1275)):*

Search for "runtime_coverage" I think you'll need similar trick to separate init from runtime trace?
</blockquote></details>

Done.

---

*[manticore/ethereum.py, line 1096 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/817#-L7jvIZaQi7Vl9tLeA9Y:-L7kEIrXIBlEFZZWYBRG:b-gklpii) ([raw file](https://github.com/trailofbits/manticore/blob/74f91b38117acc027c7a135f699905bd9c13b45d/manticore/ethereum.py#L1096)):*

Why not `isinstance`? What's stopping us from adding a platform flag such as `in_init` that is toggled?
</blockquote></details>

i was just copypasting the code from above, but yeah let's just fix this now :) it turns out this is buggy anyway, and will improperly record the runtime trace. it will miss the final instruction, which will change `.last_exception`. i'm looking at it now

```
31 JUMPDEST 0x46 <class 'manticore.platforms.evm.Call'>
31 CALLVALUE 0x47 <class 'manticore.platforms.evm.Call'>
31 ISZERO 0x48 <class 'manticore.platforms.evm.Call'>
31 PUSH2 0x51 0x49 <class 'manticore.platforms.evm.Call'>
31 JUMPI 0x4c <class 'manticore.platforms.evm.Call'>
31 PUSH1 0x0 0x4d <class 'manticore.platforms.evm.Call'>
31 DUP1 0x4f <class 'manticore.platforms.evm.Call'>
31 REVERT 0x50 <class 'manticore.platforms.evm.Revert'> <<<<<<<<< will get added to init trace, even though it should be in rt trace
018-03-16 15:31:12,959: [24676] m.ethereum:INFO: Generated testcase No. 0 - REVERT
31 JUMPDEST 0x51 <class 'manticore.platforms.evm.Call'>

```

---

*[manticore/ethereum.py, line 1184 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/817#-L7jvP7TjmNtosbb2KYj:-L7kD9YDlZsYfwH-SjQF:bgr0cpf) ([raw file](https://github.com/trailofbits/manticore/blob/74f91b38117acc027c7a135f699905bd9c13b45d/manticore/ethereum.py#L1184)):*

Why not just `if runtime_code:`?
</blockquote></details>

not sure. done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/817)*

mossberg:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[manticore/ethereum.py, line 1096 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/817#-L7jvIZaQi7Vl9tLeA9Y:-L7kmaBlqU4r2mi_pY3T:b-d4lss2) ([raw file](https://github.com/trailofbits/manticore/blob/74f91b38117acc027c7a135f699905bd9c13b45d/manticore/ethereum.py#L1096)):*

i was just copypasting the code from above, but yeah let's just fix this now :) it turns out this is buggy anyway, and will improperly record the runtime trace. it will miss the final instruction, which will change `.last_exception`. i'm looking at it now

```
31 JUMPDEST 0x46 <class 'manticore.platforms.evm.Call'>
31 CALLVALUE 0x47 <class 'manticore.platforms.evm.Call'>
31 ISZERO 0x48 <class 'manticore.platforms.evm.Call'>
31 PUSH2 0x51 0x49 <class 'manticore.platforms.evm.Call'>
31 JUMPI 0x4c <class 'manticore.platforms.evm.Call'>
31 PUSH1 0x0 0x4d <class 'manticore.platforms.evm.Call'>
31 DUP1 0x4f <class 'manticore.platforms.evm.Call'>
31 REVERT 0x50 <class 'manticore.platforms.evm.Revert'> <<<<<<<<< will get added to init trace, even though it should be in rt trace
018-03-16 15:31:12,959: [24676] m.ethereum:INFO: Generated testcase No. 0 - REVERT
31 JUMPDEST 0x51 <class 'manticore.platforms.evm.Call'>

```
</blockquote></details>

Addressed this to use isinstance in my most recent commits. the change is logic is due to the fact that Create is a subclass of Call. I think i prefer this way to introducing another api in EVMWorld

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/817)*

manticore/ethereum.py
```diff
@@ -1175,14 +1172,14 @@ def flagged(flag):
                         summary.write("\t%032x -> %032x %s\n" % (offset, value, flagged(is_storage_symbolic)))
                         is_something_symbolic = is_something_symbolic or is_storage_symbolic
 
-                code = blockchain.get_code(account_address)
-                if len(code):
+                runtime_code = blockchain.get_code(account_address)
+                if runtime_code:
                     summary.write("Code:\n")
-                    fcode = StringIO.StringIO(code)
+                    fcode = StringIO.StringIO(runtime_code)
```
yan: Approving this PR, but we should make sure our uses of `str` vs `bytes` (and `StringIO.StringIO` vs `io.BytesIO`) are consistent as we continue our port to py3. (cc @f2xeb)

yan:
:lgtm:

---

Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/817#-:-L7ygt6jXp3dFZ0HIgq4:bnfp4nl)*

manticore/ethereum.py
f2xeb: yeah in general BytesIO is very likely what you want unless you know you're dealing with a unicode encoded string

PR #819, opened 2018-03-17
Added --txaccount parameter to control the caller of the symbolic exploration of smart contracts

ggrieco-tob:
This addition will enable casual users to control how to test contracts using the manticore cli.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/819)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -920,18 +920,25 @@ class ManticoreEVM(Manticore):
 
         return status
 
-    def multi_tx_analysis(self, solidity_filename, contract_name=None, tx_limit=None):
+    def multi_tx_analysis(self, solidity_filename, contract_name=None, tx_limit=None, tx_account="attacker"):
         with open(solidity_filename) as f:
             source_code = f.read()
 
-        user_account = self.create_account(balance=1000)
-        contract_account = self.solidity_create_contract(source_code, contract_name=contract_name, owner=user_account)
+        owner_account = self.create_account(balance=1000)
+        contract_account = self.solidity_create_contract(source_code, contract_name=contract_name, owner=owner_account)
         attacker_account = self.create_account(balance=1000)
+
+        if tx_account == "attacker":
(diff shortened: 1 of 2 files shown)


manticore/ethereum.py
```diff
@@ -920,18 +920,25 @@ def transaction(self, caller, address, value, data):
...
+        owner_account = self.create_account(balance=1000)
+        contract_account = self.solidity_create_contract(source_code, contract_name=contract_name, owner=owner_account)
         attacker_account = self.create_account(balance=1000)
 
+        if tx_account == "attacker":
+            tx_account = attacker_account
+        elif tx_account == "owner":
+            tx_account = owner_account
+        else:
+            raise Exception('The account to perform the symbolic exploration of the contract should be either "attacker" or "owner"')
```
mossberg: we have some fancy new custom exceptions to raise now! let's raise an `EthereumError`

mossberg:
hi @ggrieco-tob could you please elaborate a little more on how this flag will be used? i can see you can use it to cause transactions in cli mode to come from the attacker account or the owner account. previously we only did attacker transactions. what is a use case for sending transactions from the owner? i don't think there should really be a different between transactions coming from the owner vs attacker?

feliam:
It would be good to assign a high level name to account addresses too. 
For reporting purposes. Maybe a different PR/issue.

ggrieco-tob:
> hi @ggrieco-tob could you please elaborate a little more on how this flag will be used? i can see you can use it to cause transactions in cli mode to come from the attacker account or the owner account. previously we only did attacker transactions. 

Sure.

> what is a use case for sending transactions from the owner? i don't think there should really be a different between transactions coming from the owner vs attacker?

The contract constructor is executed using the contract_owner as the caller. It is common to initialize variables with the address of that user to allow only the owner to use certain code. Then, the use of the attacker or the owner can help to understand the capabilities of both type of users.

Issue #809, opened 2018-03-12
Port Ethereum tests to individual files


yan:
Some tests for Manticore's EVM implementation (in `tests/test_eth.py`) use inline contracts for regression tests, specifically `test_can_create` and `test_writebuffer_doesnt_raise`. This is unlike how other EVM tests are implemented, as standalone scripts that are invoked from `test_eth_regressions` in `tests/test_binaries.py` because the regression tests in the latter do not support multi-contract invocations.

This issue covers refactoring inline contracts into separate files (named by the issue that raised them, #807 in `test_writebuffer_doesnt_raise` case, and #799 for `test_can_create`) and adjusting the runner script to handle multi-contract test cases.

PR #820, opened 2018-03-18
port ethereum tests to individual files (issue #809)

mroll:
Hi! I am new to the codebase, and this is my first PR.

My proposed change removes two tests that use inline contracts from `tests/test_eth.py`, and rolls them into the `test_eth_regressions()` method in `tests/test_binaries.py`.

I added a `contract` keyword argument to the `_simple_cli_run()` method, so that `.sol` files with multiple contracts can be run.

I believe this fixes [issue 809](https://github.com/trailofbits/manticore/issues/809). Please let me know if there is more to be done, or if my PR should be improved, for that issue to be satisfied.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/820)

### DIFF
--- a/tests/binaries/799.sol
+++ b/tests/binaries/799.sol
@@ -1,0 +1,2 @@
+contract X { function X(address x) {} }
+contract C { function C(address x) { new X(x); } }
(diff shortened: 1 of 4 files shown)


yan:
@mroll Thanks a ton for this PR! It definitely fixes #809. Looking at it once more, it might have made sense to just specify contract name `C` for every call since all other examples also define a contract named `C`, but the way you have it is likely more flexible going forward.

PR #827, opened 2018-03-20
A more complete example to initialize and explore a smart contract with manticore

ggrieco-tob:
This example is more suitable for the blockchain tool blog post.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/827)

### DIFF
--- a/examples/evm/complete.py
+++ b/examples/evm/complete.py
@@ -1,0 +1,48 @@
+from manticore.ethereum import ManticoreEVM
+################ Script #######################
+
+m = ManticoreEVM()
+#And now make the contract account to analyze
+# cat  | solc --bin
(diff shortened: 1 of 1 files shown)


examples/evm/complete.py
```diff
@@ -0,0 +1,48 @@
...
+                data=symbolic_data,
+                value=symbolic_value )
+
+
+print "[+] Resulting balances are:"
+for state in m.running_states:
+    balance = state.platform.get_balance(user_account)
+    print state.solve_one(balance)
+
+#Let seth know we are not sending more transactions
```
mossberg: let's not use the word 'seth', no one knows what it means ;)

PR #784, opened 2018-03-02
python3 compat sub-PR 3/3

f2xeb:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/784)

PR #828, opened 2018-03-20
Test Suite Python3 Compatibility

f2xeb:
First pull-request from the larger #784

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/828)


disconnect3d:
We **really** need builtins imports to be standarized and enforced to be exactly the same in each source file:
```
from builtins import map, range
```
as we might end up having problems supporting py2 after we go to py3.

E.g. someone uses `map` expecting it to return a generator but forgets to `import map` and so in Py2 it will return a `list` object.

disconnect3d:
Similar to the other py3 PR, the faster it will be merged, the better... :s

mossberg:
thanks @disconnect3d, that point about imports is really important, that sounds like something that will definitely trip us up.

on the topic of merge strategy: 
#784 represents the sum of all work we (well, @f2xeb) has been putting into this! to ease merging, we're are merging incrementally. this PR represents a subset of the changes from #784, so we're reviewing it in chunks. we're planning to merge these smaller PRs _not_ into master, but into an intermediary branch, so that whole py3 port can land in master at once in one commit so we don't end up in a half-baked port situation.

f2xeb:
@disconnect3d for the imports i totally agree. the fullest form is `from builtins import (ascii, bytes, chr, dict, filter, hex, input, int, map, next, oct, open, pow, range, round, str, super, zip)` but we might as well just do `from builtins import *` on the top of each file. i'll add that in now and work on enforcement in the build process later, maybe add a linter rule for it.

mossberg:
i'm a little hesitant about the `from builtins import *` form, which i think can increase ambiguity 😐

f2xeb:
would you prefer the full line? my rationale is that we'd want to pull in all functions that have different behavior on python3 in case they show up. `from builtins import *` seemed less ambiguous to me since you know it's importing everything, but i understand how the `from x import *` syntax is kind of icky.

mossberg:
i think i just have an automatic aversion to `import *` which made me say that :) i was concerned about it potentially being unclear if a certain type was being overwritten or not (with `import *` you need to know or look up everything that `*` imports). however, it it nice to know with `import *` you're completely covered; there is no possibility of accidentally using an old/incorrect type because you forgot to import it. i also suppose aesthetically `import *` seems preferable and less error prone than manually importing all those things. so after thinking through more, i'm less hesitant, and fine with `import *`, sorry about that!

f2xeb:
@mossberg no worries! it makes sense to be adverse to it, but as a bonus i just caught a bug while adding them in :smile:

mossberg:
ok, this is looking good to me!

---

Reviewed 78 of 78 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[setup.py, line 20 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/828#-L89CaoJb094cg1ylIv9:-L89CaoJb094cg1ylIvA:bcjh55p) ([raw file](https://github.com/trailofbits/manticore/blob/396ffd4a9450fea4611db54f94267d9407776054/setup.py#L20)):*
> ```Python
>     version='0.1.7',
>     classifiers=[
>             "Programming Language :: Python :: 2.7",
> ```

i'm not very familiar with the classifiers thing, what exactly is it used for? will we eventually add a python 3 classifier?

---

*[tests/test_events.py, line 39 at r2](https://reviewable.io:443/reviews/trailofbits/manticore/828#-L89PBj8BIIF2rQr06DD:-L89PBj8BIIF2rQr06DE:bm2cjrb) ([raw file](https://github.com/trailofbits/manticore/blob/396ffd4a9450fea4611db54f94267d9407776054/tests/test_events.py#L39)):*
> ```Python
>     def test_weak_references(self):
>         a = A()
>         self.assertSequenceEqual([len(s) for s in (a._signals, a._forwards)], (0, 0))
> ```

hm, this imports map, but then removes all the maps from the code. do we not need to import map anymore? i wonder why it removes the maps calls

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/trailofbits/manticore/828#-:-L89W3KT5_HsOnP9EfGL:b-usgb7j)*

f2xeb:
*[setup.py, line 20 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/828#-L89CaoJb094cg1ylIv9:-L8A1wg7A9-XHPCePnBv:b-edqgkt) ([raw file](https://github.com/trailofbits/manticore/blob/396ffd4a9450fea4611db54f94267d9407776054/setup.py#L20)):*

i'm not very familiar with the classifiers thing, what exactly is it used for? will we eventually add a python 3 classifier?
</blockquote></details>

it's pretty superficial, just for letting end-users get some info about the project via tags

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/828)*

mossberg:
:lgtm: awesome! fine to merge this into python3-staging when tests pass 

---

Reviewed 23 of 23 files at r4.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/828#-:-L8A30Zo3C_gcUv7-WPu:beltnjx)*

PR #833, opened 2018-03-21
Fix duplicated armv7 test name

disconnect3d:
The `Armv7RF` class has 8 tests while only 7 were launched during test runs.

The problem was that we had two tests with the same name: `test_flag_wr`.
The first one can be found few lines above the second https://github.com/trailofbits/manticore/compare/master...disconnect3d:fix-duplicated-armv7-test-name?expand=1#diff-a4f83ac12d210376031ccbf8e84c0294R23

I have also refactored the file in 2nd commit (and removed unused imports).

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/833)


disconnect3d:
I've also added 3rd commit which changes certain test classes names - so that they end up with `Test`, as other test classes.

Actually there was a duplicate of test class name `ArmV7RF` but it seems that it isn't a problem at all for unittest framework.

PR #835, opened 2018-03-22
Explicitly override __hash__ to use object.__hash__

mossberg:
This fixes one of the issues that is causing test_armv7cpu.py to fail when tested under python3.5

In python3 there are changes to how you are supposed to handle
overrides of __eq__, as it relates to __hash__
(http://python3porting.com/preparing.html?highlight=hash#implementing-hash)

In python2 we changed __eq__, but did nothing for __hash__, so __hash__
kept using the default one provided by object. In python3 it automatically
sets __hash__ to None, so we need to explicitly emulate this.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/835)

### DIFF
--- a/manticore/core/smtlib/expression.py
+++ b/manticore/core/smtlib/expression.py
@@ -111,6 +111,9 @@ class Bool(Expression):
     def __eq__(self, other):
         return BoolEq(self, self.cast(other))
 
+    def __hash__(self):
+        return object.__hash__(self)
+
     def __ne__(self, other):
         return BoolNot(self == self.cast(other))
 
@@ -336,6 +339,9 @@ class BitVec(Expression):
 
     def __eq__(self, other):
         return Equal(self, self.cast(other))
+
+    def __hash__(self):
+        return object.__hash__(self)
 
     def __ne__(self, other):
         return BoolNot(Equal(self, self.cast(other)))


mossberg:
so I actually think this explicitly does _not_ (and indeed manticore has never) met the "hash property" of ""two objects that compare equal should also have the same hash-value." with respect to our expression objects.

under this PR:

```
[I] (mpy3) mark ubuntu /m/h/c/m/hashissues ❯ ipython
Python 3.5.2 (default, Nov 23 2017, 16:37:01)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from manticore.core.smtlib import *

In [2]: a = BitVec(32) + BitVec(32)

In [3]: b = BitVec(32) + BitVec(32)

In [4]: hash(a) == hash(b)
Out[4]: False
```

technically `a` and `b` are identical expression trees, and "should" compare equal, yet their hashes do not compare equal because the default hashing operation (`object.__hash__`) returns a unique integer for distinct objects. the fact that their hashes are not equal _would_ be a problem if we were ever comparing expression trees to each other (and using them interchangeably if they are equal), *but* we're not; we never do this!

the tricky thing about this whole situation is that our purpose of overriding `__eq__` is not to implement an equality comparison (as the python porting docs assume); manticore generally never checks, and does not care if two expression objects are equal. (i think)

when manticore does `a == b` that's not a comparison, it's a convenience interface for _object creation_. `a == b` is a little bit nicer to code than `BoolEq(a, b)`, but the two are semantically equal.

to summarize, it's ok for our expression objects that are technically equal trees to not have the same hash value because expression objects are never compared for equality. furthermore, they are never _substituted_ for each other (for example in a dict/set lookup) based on whether they "should" be equal or not, and that's really what this whole thing is getting at.

disclaimer! i'm not 100% sure of all of this, more like ~80%, and in any case, the changes of this PR are a direct port of how manticore behaves under py2, so at the very least, it's not introducing a bug that wasn't there from before the port.

PR #836, opened 2018-03-22
Open subproc in line buffered text mode

mossberg:
Py3 changes how subproc file objects are handled; by default they are byte (not str) oriented, and _are_ buffered (not unbuffered). This pr manually specifies that we want a line buffered, text stream to communicate with.

This makes test_armv7cpu.py pass all tests.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/836)

### DIFF
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -38,6 +38,9 @@ from future.utils import with_metaclass
 logger = logging.getLogger(__name__)
 
 
+LINEBUF = 1
+
+
 class Z3NotFoundError(EnvironmentError):
     pass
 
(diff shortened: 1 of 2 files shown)


f2xeb:
this makes sense to me! seems like a good fix, thanks for finding it.

i'd say merge this but i want to double check why the travis builds aren't passing...

---







*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/836#-:-L8EohA-49YgK2yGs_tP:b-9y04zx)*

mossberg:
oops merged, but the tests error in this branch is the same one that already exists in the python3 branch

PR #837, opened 2018-03-22
Update CALLDATACOPY semantics

yan:
Currently, `CALLDATACOPY` reads `offset` and `size` from the stack and tries to copy data using those offsets. In the case of symbolic `data_offset`, it is correctly constrained in the implementation before attempting to copy data. However, in the concrete case, there's nothing to set bounds on `[data_offset, data_offset+size]` causing an indexing exception in case of out-of-bounds index.

This PR makes sure to limit copy offsets to valid ranges. This is exactly how the official `go-ethereum` implementation handles it.[1][2]

It might make sense to create a similar helper and use it wherever we need to index into data.

[1] calldatacopy implementation: https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L447
[2] the getDataBig helper: https://github.com/ethereum/go-ethereum/blob/master/core/vm/common.go#L49

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -1645,6 +1645,10 @@ class EVM(Eventful):
         if issymbolic(size):
             raise ConcretizeStack(3, policy='ALL')
 
+        if not issymbolic(data_offset):
+            data_offset = min(data_offset, len(self.data))
+            size = min(len(self.data) - data_offset, size)
+
         for i in range(size):
             c = Operators.ITEBV(8, data_offset + i < len(self.data), Operators.ORD(self.data[data_offset + i]), 0)
             self._store(mem_offset + i, c)


manticore/platforms/evm.py
```diff
@@ -1645,6 +1645,10 @@ def CALLDATACOPY(self, mem_offset, data_offset, size):
         if issymbolic(size):
             raise ConcretizeStack(3, policy='ALL')
 
+        if not issymbolic(data_offset):
+            data_offset = min(data_offset, len(self.data))
+            size = min(len(self.data) - data_offset, size)
+
         for i in range(size):
             c = Operators.ITEBV(8, data_offset + i < len(self.data), Operators.ORD(self.data[data_offset + i]), 0)
```
yan: This looks redundant with the above checks, but in the symbolic case, we still want this bounds check.

tests/EVM/test_EVMCALLDATACOPY.py
```diff
@@ -0,0 +1,129 @@
...
+
+class EVMTest_CALLDATACOPY(unittest.TestCase):
+    _multiprocess_can_split_ = True
+    maxDiff=None 
+    def _execute(self, new_vm):
+        last_returned = None
+        last_exception = None
+        try:
+            new_vm.execute()
+        except evm.Stop, e:
```
disconnect3d: For py3 compatibility use `except evm.Stop as e:`

As in py3 `, e` is a syntax error:
```
In [1]: try:
   ...:     pass
   ...: except Exception, e:
  File "<ipython-input-1-53869dcec347>", line 3
    except Exception, e:
                    ^
SyntaxError: invalid syntax
```

mossberg:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.

---

*[manticore/platforms/evm.py, line 1650 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-L9lcJas1S8C24hN67-X:-L9lcJas1S8C24hN67-Y:bh0mive) ([raw file](https://github.com/trailofbits/manticore/blob/ee8b3d71c540c7f4fbca217d20724d8711cb7b1b/manticore/platforms/evm.py#L1650)):*
> ```Python
>         if not issymbolic(data_offset):
>             data_offset = min(data_offset, len(self.data))
>             size = min(len(self.data) - data_offset, size)
> ```

I'm not sure is the correct handling of `size`, this actually truncates the copy. i think we might just want to leave out this line. however, if we do, if size is really massive the `range(size)` will fail..

 the geth implementation seems to respect the original value passed in as the operand, using it as the final length for the padding operation (https://github.com/ethereum/go-ethereum/blob/2e247705cd27e919e806f29ff2adcd57c163b1ac/core/vm/common.go#L56)

---

*[tests/EVM/test_EVMCALLDATACOPY.py, line 65 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-L9lfX7X-GDC_9BQUjoW:-L9lfX7X-GDC_9BQUjoX:b-scg11h) ([raw file](https://github.com/trailofbits/manticore/blob/ee8b3d71c540c7f4fbca217d20724d8711cb7b1b/tests/EVM/test_EVMCALLDATACOPY.py#L65)):*
> ```Python
>             self.assertEqual(new_vm.pc, 1)
>             self.assertEqual(new_vm.stack, [])
>             self.assertEqual(new_vm.memory.read(0, 9), range(ord('A')+9, ord('A')+18))
> ```

the use of regular range expecting it to return a list isn't py3 compat? need to wrap in list()?

---

*[tests/EVM/test_EVMCALLDATACOPY.py, line 95 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-L9lgUra8NSZ3WOk7Wpt:-L9lgUra8NSZ3WOk7Wpu:b8o8ksd) ([raw file](https://github.com/trailofbits/manticore/blob/ee8b3d71c540c7f4fbca217d20724d8711cb7b1b/tests/EVM/test_EVMCALLDATACOPY.py#L95)):*
> ```Python
>             self.assertEqual(new_vm.pc, 1)
>             self.assertEqual(new_vm.stack, [])
>             # nothing should get written with huge offset
> ```


this checks that nothing was written to memory, but that doesn't necessary test that the data_offset is handled correctly, since the test uses a zero for size which will also guarantee that nothing is written

i think this test should use a nonzero value n for size and check that n zeroes were written to memory

---

*[tests/EVM/test_EVMCALLDATACOPY.py, line 126 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-L9lfhq3A5pFu7UYRkCH:-L9lfhq3A5pFu7UYRkCI:b-xabv3) ([raw file](https://github.com/trailofbits/manticore/blob/ee8b3d71c540c7f4fbca217d20724d8711cb7b1b/tests/EVM/test_EVMCALLDATACOPY.py#L126)):*
> ```Python
>             self.assertEqual(new_vm.pc, 1)
>             self.assertEqual(new_vm.stack, [])
>             self.assertEqual(new_vm.memory.read(0, 3), map(ord, data[-3:]))
> ```

same, need to wrap in list() for py3 compat

---

*[tests/EVM/test_EVMCALLDATACOPY.py, line 126 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-L9lfpz_6HFwdplMSz6u:-L9lfpz_6HFwdplMSz6v:bmgy3ia) ([raw file](https://github.com/trailofbits/manticore/blob/ee8b3d71c540c7f4fbca217d20724d8711cb7b1b/tests/EVM/test_EVMCALLDATACOPY.py#L126)):*
> ```Python
>             self.assertEqual(new_vm.pc, 1)
>             self.assertEqual(new_vm.stack, [])
>             self.assertEqual(new_vm.memory.read(0, 3), map(ord, data[-3:]))
> ```

i think this test should also check that >9000 zeros were written to memory?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/837)*

mossberg:
Reviewed 2 of 2 files at r3.

---

*[manticore/platforms/evm.py, line 1660 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-lD1RB8i_W9TAbe9e:-LA-lD1RB8i_W9TAbe9f:b-rjgwcv) ([raw file](https://github.com/trailofbits/manticore/blob/93e485b16633d659e8c6685f69c6d58cedd03ee1/manticore/platforms/evm.py#L1660)):*
> ```Python
> 
>         for i in range(size):
>             c = Operators.ITEBV(8, data_offset + i < len(self.data), safe_index_with_ord(self.data, data_offset + i), 0)
> ```

could we put the dereference in a try block instead? i feel like introducing a variable rather than doing the deref inline would make this line more readable also

---

*[tests/EVM/test_EVMCALLDATACOPY.py, line 138 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-n7UoE2_c5KF_WwrP:-LA-n7UoE2_c5KF_WwrQ:bqc8e1q) ([raw file](https://github.com/trailofbits/manticore/blob/93e485b16633d659e8c6685f69c6d58cedd03ee1/tests/EVM/test_EVMCALLDATACOPY.py#L138)):*
> ```Python
>             self.assertTrue(all(val == 0 for addr, val in new_vm.memory.items() if addr > remainder - 1))
>             # and make sure there's 9999-remainder of them
>             self.assertEqual(sum(1 for addr, _ in new_vm.memory.items() if addr > remainder - 1), 9999-remainder)
> ```

would it be cleaner to slice new_vm.memory.items() and check the len() rather than using sum?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/837)*

yan:

---

*[manticore/platforms/evm.py, line 1650 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-L9lcJas1S8C24hN67-X:-L9p3UWT2A1L86Ynm7pU:bnra81n) ([raw file](https://github.com/trailofbits/manticore/blob/ee8b3d71c540c7f4fbca217d20724d8711cb7b1b/manticore/platforms/evm.py#L1650)):*

I'm not sure is the correct handling of `size`, this actually truncates the copy. i think we might just want to leave out this line. however, if we do, if size is really massive the `range(size)` will fail..

 the geth implementation seems to respect the original value passed in as the operand, using it as the final length for the padding operation (https://github.com/ethereum/go-ethereum/blob/2e247705cd27e919e806f29ff2adcd57c163b1ac/core/vm/common.go#L56)
</blockquote></details>

:+1: Failing with a large size is how the official impl handles it also: https://github.com/ethereum/go-ethereum/blob/master/core/vm/memory.go#L35 (`offset` is also pulled from stack, which can panic when out of bounds)


Question is, if failing is the correct response -- we should probably just generate a test case.

---

*[tests/EVM/test_EVMCALLDATACOPY.py, line 138 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-n7UoE2_c5KF_WwrP:-LA-qo7QCYmEvhI3MFEP:brha31a) ([raw file](https://github.com/trailofbits/manticore/blob/93e485b16633d659e8c6685f69c6d58cedd03ee1/tests/EVM/test_EVMCALLDATACOPY.py#L138)):*

would it be cleaner to slice new_vm.memory.items() and check the len() rather than using sum?
</blockquote></details>

I don't think so. That's a filtering sub-expression that lazily computes the number of elements above the `remainder`; filtering through items and then using `len()` would create a new explicit list. I don't mind either way though, it's just test code.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/837)*

mossberg:
Reviewed 1 of 2 files at r4.
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[manticore/platforms/evm.py, line 1653 at r4](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-xrQe984e4jCvncXh:-LA-xrQe984e4jCvncXi:b83neni) ([raw file](https://github.com/trailofbits/manticore/blob/ea015904150342c6206d393e551d1efe9abee6d3/manticore/platforms/evm.py#L1653)):*
> ```Python
>             try:
>                 maybe_byte = Operators.ORD(self.data[data_offset+i])
>             except (TypeError, IndexError):
> ```

what is the TypeError supposed to catch?

---

*[tests/EVM/test_EVMCALLDATACOPY.py, line 138 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-n7UoE2_c5KF_WwrP:-LA-ujc0AyKAYWEtUIbI:b-i87149) ([raw file](https://github.com/trailofbits/manticore/blob/93e485b16633d659e8c6685f69c6d58cedd03ee1/tests/EVM/test_EVMCALLDATACOPY.py#L138)):*

I don't think so. That's a filtering sub-expression that lazily computes the number of elements above the `remainder`; filtering through items and then using `len()` would create a new explicit list. I don't mind either way though, it's just test code.
</blockquote></details>

imo `len()` more clearly expresses the intent to count number of elements than `sum()` does, so i would disagree and say it actually is cleaner :) . whether it's more memory efficient is another story; i completely agree creating a new list is less efficient, but i'd say that matters less than clarity in this case.

i think something along the lines of  `self.assertEqual(len(new_vm.memory.items()[remainder:]), 9999-remainder)` would work since addr is equivalent to index here, so no need to filter?

---

*[tests/EVM/test_EVMCALLDATACOPY.py, line 132 at r4](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-wY5S02GBrErrDDVG:-LA-wY5S02GBrErrDDVH:bxaxwbl) ([raw file](https://github.com/trailofbits/manticore/blob/ea015904150342c6206d393e551d1efe9abee6d3/tests/EVM/test_EVMCALLDATACOPY.py#L132)):*
> ```Python
>             self.assertEqual(new_vm.pc, 1)
>             self.assertEqual(new_vm.stack, [])
>             self.assertEqual(list(new_vm.memory.read(0, 3)), [ord(c) for c in data[-remainder:]])
> ```

EVMMemory.read() already returns a list, so this is redundant?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/837)*

yan:
Review status: 1 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[manticore/platforms/evm.py, line 1653 at r4](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-xrQe984e4jCvncXh:-LA-ynwYB7gALNWy_nEJ:b3ji3bj) ([raw file](https://github.com/trailofbits/manticore/blob/ea015904150342c6206d393e551d1efe9abee6d3/manticore/platforms/evm.py#L1653)):*

what is the TypeError supposed to catch?
</blockquote></details>

In case of a symbolic `data_offset`. It might not be needed however.

---

*[tests/EVM/test_EVMCALLDATACOPY.py, line 138 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-n7UoE2_c5KF_WwrP:-LA-zIUB2GR-iu3d-iSd:byyllvb) ([raw file](https://github.com/trailofbits/manticore/blob/93e485b16633d659e8c6685f69c6d58cedd03ee1/tests/EVM/test_EVMCALLDATACOPY.py#L138)):*

imo `len()` more clearly expresses the intent to count number of elements than `sum()` does, so i would disagree and say it actually is cleaner :) . whether it's more memory efficient is another story; i completely agree creating a new list is less efficient, but i'd say that matters less than clarity in this case.

i think something along the lines of  `self.assertEqual(len(new_vm.memory.items()[remainder:]), 9999-remainder)` would work since addr is equivalent to index here, so no need to filter?
</blockquote></details>

It might work, or it might not as `items`' ordering isn't guaranteed and shouldn't be relied on, so we need to filter anyway. If we're filtering, we can create an intermediate list or we can sum inline.  If you feel very strongly about this, i can create an intermediate list and test for its `len`

---

*[tests/EVM/test_EVMCALLDATACOPY.py, line 132 at r4](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-wY5S02GBrErrDDVG:-LA-yz082aVH42EpTPXP:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/ea015904150342c6206d393e551d1efe9abee6d3/tests/EVM/test_EVMCALLDATACOPY.py#L132)):*

EVMMemory.read() already returns a list, so this is redundant?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/837)*

feliam:
*[tests/EVM/test_EVMCALLDATACOPY.py, line 138 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-n7UoE2_c5KF_WwrP:-LA0RfR5EfMg-REbS3iF:bqy3okt) ([raw file](https://github.com/trailofbits/manticore/blob/93e485b16633d659e8c6685f69c6d58cedd03ee1/tests/EVM/test_EVMCALLDATACOPY.py#L138)):*

It might work, or it might not as `items`' ordering isn't guaranteed and shouldn't be relied on, so we need to filter anyway. If we're filtering, we can create an intermediate list or we can sum inline.  If you feel very strongly about this, i can create an intermediate list and test for its `len`
</blockquote></details>

may be off topic but I leave this here ..
Note that .items() is (probably) going away if we want memory to support symbolic indexing with free indexes
```
x = free_bitvec()
x = free_bitvec()
memory[x] = y

memory.items() ???
```
You need free symbolic indexing to solve stuff like this: https://capturetheether.com/challenges/math/mapping/

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/837)*

mossberg:
Reviewed 1 of 2 files at r4, 1 of 1 files at r5.

---

*[manticore/platforms/evm.py, line 1653 at r4](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-xrQe984e4jCvncXh:-LA0IAc80WkW6vectLRp:bm2wxhw) ([raw file](https://github.com/trailofbits/manticore/blob/ea015904150342c6206d393e551d1efe9abee6d3/manticore/platforms/evm.py#L1653)):*

In case of a symbolic `data_offset`. It might not be needed however.
</blockquote></details>

hm, if `data_offset + i` ends up being symbolic, then i don't think the correct action is to copy a 0 over to memory. i think the correct action is to either concretize that symbolic offset, or do an smtlib array select with the symbolic index (but that is probably not available without dev-evm-reservoir-of-awesomeness).

that said, the prior version of this would just crash on indexing self.data with an expression object anyway, so i don't think truly handling a symbolic data_offset is within the scope of this PR (we might want to just bail on an exception with a helpful message saying calldataload symbolic data_offset is not supported atm)

i was also trying to take a step back from this code, and i realized that this double bounds check thing strikes me as a bit messy (how we use a try block, and then have a computed bounds check down in the ITEBV). i'm not totally sure of a solution yet, but i think maybe we should just totally scrap the idea of supporting symbolic data_offset for now, until we can make an earnest effort to do the full concretize & fork thing, or array select. until then, only supporting a concrete data_offset would make this code a lot simpler 

---

*[tests/EVM/test_EVMCALLDATACOPY.py, line 138 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/837#-LA-n7UoE2_c5KF_WwrP:-LA0KwJs-lvBacu-UGHR:b4y212n) ([raw file](https://github.com/trailofbits/manticore/blob/93e485b16633d659e8c6685f69c6d58cedd03ee1/tests/EVM/test_EVMCALLDATACOPY.py#L138)):*

may be off topic but I leave this here ..
Note that .items() is (probably) going away if we want memory to support symbolic indexing with free indexes
```
x = free_bitvec()
x = free_bitvec()
memory[x] = y

memory.items() ???
```
You need free symbolic indexing to solve stuff like this: https://capturetheether.com/challenges/math/mapping/


</blockquote></details>

another advantage of using an intermediate list would be how we wouldn't need to do the two identical generator comprehensions on two lines; we could just create the list once, then in one assert check the contents, and in the other check the length!

it seems like the two generator comprehensions got slightly out of sync (one uses a `>=`, the other uses `< remainder - 1`); generating an intermediate list would make that impossible :)

i guess i feel fairly strongly that the cleanest and most maintainable form this code could take would use an intermediate list, reuse it in the "items are 0" and length checks, with len() used for the length check.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/837)*

manticore/platforms/evm.py
```diff
@@ -1648,14 +1648,22 @@ def CALLDATACOPY(self, mem_offset, data_offset, size):
...
+        if not issymbolic(data_offset):
+            data_offset = min(data_offset, len(self.data))
+
         for i in range(size):
-            c = Operators.ITEBV(8, data_offset + i < len(self.data), Operators.ORD(self.data[data_offset + i]), 0)
+            try:
+                maybe_byte = Operators.ORD(self.data[data_offset+i])
+            except IndexError:
+                maybe_byte = 0
+            except TypeError:
```
feliam: Is it assured that self.data is a str or that it will raise if indexed with a symbol ?

manticore/platforms/evm.py
```diff
@@ -1648,14 +1648,22 @@ def CALLDATACOPY(self, mem_offset, data_offset, size):
 
         # FIXME put zero if not enough data
         if issymbolic(size) or issymbolic(data_offset):
-            #self._constraints.add(Operators.ULE(data_offset, len(self.data)))
             self._constraints.add(Operators.ULE(size + data_offset, len(self.data) + (32 - len(self.data) % 32)))
 
         if issymbolic(size):
             raise ConcretizeStack(3, policy='ALL')
 
+        if not issymbolic(data_offset):
```
feliam: Is data_offset able to be symbolic. If not I think it is better to just concretize it ?

PR #849, opened 2018-03-27
Added --txnocoverage option to manticore cli

ggrieco-tob:
This new option allows manticore to explore the execution of contracts it finds no more live state or reaches the txlimit.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/849)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -918,7 +918,7 @@ class ManticoreEVM(Manticore):
 
         return status
 
-    def multi_tx_analysis(self, solidity_filename, contract_name=None, tx_limit=None, tx_stop_using_coverage=True, tx_account="attacker"):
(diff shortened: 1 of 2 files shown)


manticore/ethereum.py
```diff
@@ -918,7 +918,7 @@ def transaction(self, caller, address, value, data):
 
         return status
 
-    def multi_tx_analysis(self, solidity_filename, contract_name=None, tx_limit=None, tx_account="attacker"):
+    def multi_tx_analysis(self, solidity_filename, contract_name=None, tx_limit=None, tx_stop_using_coverage=True, tx_account="attacker"):
```
yan: This double negative is confusing. What if this was `use_coverage` and the check below was
`while current_coverage < 100 and use_coverage:`.

Is the default here to ignore coverage?

ggrieco-tob: Default is to use coverage to stop. I commit a few changes, let me know if it is easier to understand now..

Issue #536, opened 2017-10-26
expression.Array assertion failure if index_max is a long instead of an int


SeanHeelan:
### OS / Environment
Fedora 25

### Manticore version
0.1.4 

### Python version
2.7.14

### Dependencies
alabaster==0.7.10
Babel==2.5.1
backports.shutil-get-terminal-size==1.0.0
capstone==3.0.5rc2
certifi==2017.7.27.1
chardet==3.0.4
coverage==4.4.1
decorator==4.1.2
docutils==0.14
enum34==1.1.6
idna==2.6
imagesize==0.7.1
ipdb==0.10.3
ipython==5.5.0
ipython-genutils==0.2.0
Jinja2==2.9.6
keystone-engine==0.9.1.post3
-e git+https://github.com/trailofbits/manticore.git@e84c0513a2d850e3fe63bd23425d7ba67a88c7dc#egg=manticore
MarkupSafe==1.0
nose==1.3.7
pathlib2==2.3.0
pexpect==4.2.1
pickleshare==0.7.4
ply==3.10
prompt-toolkit==1.0.15
ptyprocess==0.5.2
pyelftools==0.24
Pygments==2.2.0
pytz==2017.2
redis==2.10.6
requests==2.18.4
scandir==1.6
simplegeneric==0.8.1
six==1.11.0
snowballstemmer==1.2.1
Sphinx==1.6.4
sphinxcontrib-websupport==1.0.1
traitlets==4.3.2
typing==3.6.2
unicorn==1.0.1
urllib3==1.22
wcwidth==0.1.7
z3-solver==4.5.1.0.post2

### Summary of the problem

If `state.new_symbolic_buffer(x)` is called and `x` is a long instead of an int then an assertion in `expression.Array` fails. 

### Step to reproduce the behavior

Utilise `state.invoke_model` to trigger a model of a function that takes a numeric argument. The argument reaches the model as a long, not an int. If the model then uses this argument in a call to `new_symbolic_buffer` the assertion failure occurs.

```
# Hook for function with the following signature 
# size_t read_data(const char *fname, char *dest, size_t count) 
def read_data(state, fname, dest, count):
    print "Introducing %d symbolic bytes at 0x%x" % (count, dest)
    buf = state.new_symbolic_buffer(count)
    state.cpu.write_bytes(dest, buf)
    with m.locked_context() as ctx:
        ctx['input_addr'] = dest

    return count

@m.hook(READ_DATA)
def read_data_model(state):
[...]

PR #850, opened 2018-03-27
Python3 compatibility, Python2.7 tests passing

f2xeb:
A large amount of futurization done for the entire codebase, which all tests passing on Python 2.7. Should be the majority of the work, with the remaining work being fixing edge cases and mechanical differences for Python 3.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/850)


mossberg:
Reviewed 88 of 95 files at r1, 1 of 2 files at r2, 5 of 5 files at r3.
Review status: 93 of 96 files reviewed at latest revision, all discussions resolved.

---

*[manticore/ethereum.py, line 25 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8rixwl1Tui3X-xb9Ua:-L8rixwl1Tui3X-xb9Ub:b4dnupt) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/manticore/ethereum.py#L25)):*
> ```Python
> from .core.plugin import Plugin
> from functools import reduce
> from utils.helpers import isstring
> ```

can we use the .utils relative import style here?

---

*[manticore/manticore.py, line 461 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8rjf8G0hDAxIiVzgTs:-L8rjf8G0hDAxIiVzgTt:b-b4s5d2) ([raw file](https://github.com/trailofbits/manticore/blob/38bb0cbab43b3a740911e0926d87802e755c1573/manticore/manticore.py#L461)):*
> ```Python
>         :param callable callback: Hook function
>         '''
>         if not (isinstance(pc, (int, long)) or pc is None):
> ```

i think we might need to keep the check for long to preserve py2 semantics? maybe we need something analogous to isstring for ints?

---

*[manticore/core/executor.py, line 18 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8rkjvx36-DWpcUsGY5:-L8rkjvx36-DWpcUsGY6:bd6lpmb) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/manticore/core/executor.py#L18)):*
> ```Python
> from contextlib import contextmanager
> 
> # This is the single global manager that will handle all shared memory among workers
> ```

we should move this dangling comment back to the code. although @yan is it accurate anymore, since it no longer appears to be a global manager?

---

*[manticore/core/workspace.py, line 22 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8rpM7H9iZzWUuziE7i:-L8rpM7H9iZzWUuziE7j:bgdeztv) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/manticore/core/workspace.py#L22)):*
> ```Python
> logger = logging.getLogger(__name__)
> 
> _manager = None
> ```

can you help me understand the reason for the changes to the global managers?

---

*[manticore/core/smtlib/constraints.py, line 85 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8rwZrsBcO2Zm3YfjCI:-L8rwZrsBcO2Zm3YfjCJ:b-fwe0vr) ([raw file](https://github.com/trailofbits/manticore/blob/38bb0cbab43b3a740911e0926d87802e755c1573/manticore/core/smtlib/constraints.py#L85)):*
> ```Python
>                 variables = get_variables(constraint)
>                 if related_variables & variables:
>                     remaining_constraints.remove(constraint)
> ```

are these changes related to the py3 port, or just general cleanup?

---

*[manticore/core/smtlib/expression.py, line 4 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8rxJ5x2uGm3PY2k9sh:-L8rxJ5x2uGm3PY2k9si:bvhs4d4) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/manticore/core/smtlib/expression.py#L4)):*
> ```Python
> from builtins import *
> from functools import reduce
> from manticore.utils.helpers import isstring
> ```

let's prefer relative import style here

---

*[manticore/platforms/evm.py, line 1025 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8s04ls5xIZSUXyx8ji:-L8s04ls5xIZSUXyx8jj:b-75ywm7) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/manticore/platforms/evm.py#L1025)):*
> ```Python
> 
> 
>     def __reduce__(self):
> ```

why is this needed?

---

*[manticore/platforms/platform.py, line 3 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8s15ZO2i-hOyHQCZYA:-L8s15ZO2i-hOyHQCZYB:b-u2pxix) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/manticore/platforms/platform.py#L3)):*
> ```Python
> from builtins import *
> from manticore.utils.event import Eventful
> from itertools import islice
> ```

these two are unused?

---

*[tests/test_memory.py, line 1295 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8s2Fo98fb22uaZOTI7:-L8s2Fo98fb22uaZOTI8:byy82ot) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/tests/test_memory.py#L1295)):*
> ```Python
>         #In the beggining the solver was 'sat' ...
>         self.assertTrue(solver.check(cs))
>         print('!')
> ```

could we rm these now?

---

*[tests/test_smtlibv2.py, line 485 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8s2HdT7MUTRBzqSeYf:-L8s2HdT7MUTRBzqSeYg:bd5v0kt) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/tests/test_smtlibv2.py#L485)):*
> ```Python
> 
>         with cs as cs_up:
>             print('!')
> ```

same, could we rm these?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

f2xeb:
*[manticore/manticore.py, line 461 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8rjf8G0hDAxIiVzgTs:-L8s4e4B7ju3tbax4wwG:btn3v88) ([raw file](https://github.com/trailofbits/manticore/blob/38bb0cbab43b3a740911e0926d87802e755c1573/manticore/manticore.py#L461)):*

i think we might need to keep the check for long to preserve py2 semantics? maybe we need something analogous to isstring for ints?
</blockquote></details>

i guess you're right. it hasn't caused issues yet but according to https://eev.ee/blog/2016/07/31/python-faq-how-do-i-port-to-python-3/ potentially could, i'll go through and add that in

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

mossberg:
Review status: 93 of 96 files reviewed at latest revision, 10 unresolved discussions.

---

*[manticore/manticore.py, line 461 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8rjf8G0hDAxIiVzgTs:-L8s5fUxDLgQU7gXgBlA:b-zaxbx3) ([raw file](https://github.com/trailofbits/manticore/blob/38bb0cbab43b3a740911e0926d87802e755c1573/manticore/manticore.py#L461)):*
<details><summary><i>Previously, f2xeb (feeb) wrote…</i></summary><blockquote>

i guess you're right. it hasn't caused issues yet but according to https://eev.ee/blog/2016/07/31/python-faq-how-do-i-port-to-python-3/ potentially could, i'll go through and add that in
</blockquote></details>

yeah, we've run into the int/long thing before, which is why we explicitly check everywhere for both https://github.com/trailofbits/manticore/issues/536

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

f2xeb:
*[manticore/core/smtlib/constraints.py, line 85 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8rwZrsBcO2Zm3YfjCI:-L8s61XkC4qTosAbmbBp:b-jlk8rc) ([raw file](https://github.com/trailofbits/manticore/blob/38bb0cbab43b3a740911e0926d87802e755c1573/manticore/core/smtlib/constraints.py#L85)):*

are these changes related to the py3 port, or just general cleanup?
</blockquote></details>

kind of both. the original loop modified the dictionary that it was iterating over, and python3 throws an exception when you do that. i probably wouldn't have touched it otherwise

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

f2xeb:
*[manticore/platforms/evm.py, line 1025 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8s04ls5xIZSUXyx8ji:-L8s8QNw-omuhciDEWdv:b-qhqu87) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/manticore/platforms/evm.py#L1025)):*

why is this needed?
</blockquote></details>

this was something that came about when trying to deal with bytestrings vs lists-of-bytestrings getting passed around.... this makes it able to use both in that setitem call. i think this is a pretty mediocre solution personally but doing it better would probably take some digging

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

mossberg:
Review status: 93 of 96 files reviewed at latest revision, 9 unresolved discussions.

---

*[manticore/platforms/evm.py, line 1025 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8s04ls5xIZSUXyx8ji:-L8sCjek7PBRI7hQXOmV:b-av89ae) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/manticore/platforms/evm.py#L1025)):*
<details><summary><i>Previously, f2xeb (feeb) wrote…</i></summary><blockquote>

this was something that came about when trying to deal with bytestrings vs lists-of-bytestrings getting passed around.... this makes it able to use both in that setitem call. i think this is a pretty mediocre solution personally but doing it better would probably take some digging
</blockquote></details>

huh ok, maybe we could add a comment?  it's totally nonobvious to me that this diff is related to what you described

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

f2xeb:
*[manticore/platforms/evm.py, line 1025 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8s04ls5xIZSUXyx8ji:-L8sDzSvDmL041cdHbR_:bm6ypza) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/manticore/platforms/evm.py#L1025)):*

huh ok, maybe we could add a comment?  it's totally nonobvious to me that this diff is related to what you described
</blockquote></details>

oh wait, i was totally looking at the wrong thing :x this was because of an unpickling issue. when pickling objects it'll call a __reduce__ method if present on the object. Call implements one of these but the arguments are totally different than the ones for the constructor here, so we needed to implement __reduce__ methods on classes that inherit from Call otherwise stuff broke. i'm probably explaining this poorly, more info here https://docs.python.org/2/library/pickle.html#object.__reduce__

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

f2xeb:
*[manticore/manticore.py, line 461 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8rjf8G0hDAxIiVzgTs:-L8scPcg6VcCQPiCspg6:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/38bb0cbab43b3a740911e0926d87802e755c1573/manticore/manticore.py#L461)):*

yeah, we've run into the int/long thing before, which is why we explicitly check everywhere for both https://github.com/trailofbits/manticore/issues/536
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

f2xeb:
*[manticore/core/smtlib/expression.py, line 4 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8rxJ5x2uGm3PY2k9sh:-L8smc5l77xEFptFnDt3:b-7ql1td) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/manticore/core/smtlib/expression.py#L4)):*

let's prefer relative import style here
</blockquote></details>

fixed

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

f2xeb:
*[manticore/platforms/platform.py, line 3 at r3](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L8s15ZO2i-hOyHQCZYA:-L8scyCM4b-N4AgDS145:b-7ql1td) ([raw file](https://github.com/trailofbits/manticore/blob/44b8e56bd13462dd4970ba9a17ecf5dcbaf41fd1/manticore/platforms/platform.py#L3)):*

these two are unused?
</blockquote></details>

fixed

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

mossberg:
Reviewed 18 of 19 files at r4.
Review status: 93 of 96 files reviewed at latest revision, 3 unresolved discussions.

---

*[manticore/core/cpu/binja.py, line 965 at r4](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L97b01w96wcvw4b8k4n:-L97b01w96wcvw4b8k4o:blzqjbs) ([raw file](https://github.com/trailofbits/manticore/blob/6aadf889c5d55ec65714f142b06860870bb32f35/manticore/core/cpu/binja.py#L965)):*
> ```Python
> 
>         quotient = Operators.SDIV(dividend, divisor)
>         if isinstance(divisor) and isint(dividend):
> ```

small bug here with `isinstance(divisor)`

---

*[manticore/platforms/evm.py, line 14 at r4](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L97b_Lz3uLlXeShr1Op:-L97b_Lz3uLlXeShr1Oq:b97kyk9) ([raw file](https://github.com/trailofbits/manticore/blob/6aadf889c5d55ec65714f142b06860870bb32f35/manticore/platforms/evm.py#L14)):*
> ```Python
> import logging
> import sys
> import inspect
> ```

unused import here?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

mossberg:
this is looking great!

---

Reviewed 1 of 95 files at r1, 1 of 2 files at r2, 1 of 19 files at r4.

---

*[manticore/core/memory.py, line 261 at r4](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L97gEAS8BezfrdSQ4e2:-L97gEAS8BezfrdSQ4e3:bs3mpnd) ([raw file](https://github.com/trailofbits/manticore/blob/6aadf889c5d55ec65714f142b06860870bb32f35/manticore/core/memory.py#L261)):*
> ```Python
>                 value = [Operators.ORD(n) for n in value]
>             for i in range(index.stop - index.start):
>                 self._data[index.start + i] = value[i]
> ```

rather than looping like this, would it be possible just to `self._data[index] = value`, now that `value` is guaranteed to have ints?

---

*[manticore/core/memory.py, line 564 at r4](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L97i7VY6KW6xhHv8cwc:-L97i7VY6KW6xhHv8cwd:b-hhvmw) ([raw file](https://github.com/trailofbits/manticore/blob/38bb0cbab43b3a740911e0926d87802e755c1573/manticore/core/memory.py#L564)):*
> ```Python
>         addr = self._search(size, addr)
> 
>         # It should not be allocated
> ```

any particular reason to rm this comment?

---

*[manticore/core/workspace.py, line 584 at r4](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L97nsnxAnz5y79Yy3sZ:-L97nsnxAnz5y79Yy3s_:botierk) ([raw file](https://github.com/trailofbits/manticore/blob/6aadf889c5d55ec65714f142b06860870bb32f35/manticore/core/workspace.py#L584)):*
> ```Python
>             try:
>                 for c in data:
>                     fd.write(bytes[solver.get_value(state.constraints, c)])
> ```

bug here with `bytes[]`? (but this code is never called anyway it seems)

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-:-L97pEK_CiF0pROFDNc6:b-zgtdrq)*

mossberg:
Reviewed 2 of 2 files at r5.

---

*[manticore/core/memory.py, line 260 at r5](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-L9ByPhB1IIkMIg42-mX:-L9ByPhB1IIkMIg42-mY:b1pr937) ([raw file](https://github.com/trailofbits/manticore/blob/a179b420c676025211abccfedbbe92d72095797c/manticore/core/memory.py#L260)):*
> ```Python
>             if not isint(value[0]):
>                 value = [Operators.ORD(n) for n in value]
>             #for i in range(index.stop - index.start):
> ```

we should probably rm these commented lines too :)

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850)*

mossberg:
:lgtm: yesss

---

Reviewed 3 of 3 files at r6.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/850#-:-L9CHkFM8oFJzyI45Zq-:b-7szb0o)*

PR #852, opened 2018-04-02
Sign extends x86 AND operand (capstone bug)

### LINKED ISSUE #560: Incorrect invalid memory access due to capstone bug
This is a follow up issue to #537.

The attached programs simply aims to cause a libc malloc error with a bad `free()`. Manticore should successfully execute to a clean exit, however it reports an invalid memory access in __libc_malloc.

```
2017-11-13 18:35:25,527: [67131] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000041dc20:  mov rax, qword ptr [rdx - 8]
2017-11-13 18:35:25,529: [67131] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000041dc24:  test    al, 2
2017-11-13 18:35:25,531: [67131] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000041dc26:  jne 0x41dc47
2017-11-13 18:35:25,532: [67131] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000041dc28:  test    al, 4
2017-11-13 18:35:25,534: [67131] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000041dc2a:  mov ecx, 0x6ca800
2017-11-13 18:35:25,535: [67131] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000041dc2f:  je  0x41dc3e
2017-11-13 18:35:25,537: [67131] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000041dc31:  lea rax, qword ptr [rdx - 0x10]
2017-11-13 18:35:25,539: [67131] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000041dc35:  and rax, 0xfc000000
2017-11-13 18:35:25,541: [67131] m.c.c.abstractcpu:DEBUG: INSTRUCTION: 0x000000000041dc3b:  mov rcx, qword ptr [rax]
2017-11-13 18:35:26,227: [67131] m.manticore:INFO: Generated testcase No. 0 - Invalid memory access (mode:r) <f8000000>
2017-11-13 18:35:26,248: [67131] m.manticore:INFO: Results in /mnt/hgfs/code/manticore/examples/linux/mcore_NzXjgF
2017-11-13 18:35:26,249: [67131] m.manticore:INFO: Total time: 149.902643919
```

[537.log](https://github.com/trailofbits/manticore/files/1469052/537.log)
[537.zip](https://github.com/trailofbits/manticore/files/1469047/537.zip)

catenacyber:
Fixes #560

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/852)

### DIFF
--- a/manticore/core/cpu/x86.py
+++ b/manticore/core/cpu/x86.py
@@ -889,7 +889,13 @@ class X86Cpu(Cpu):
         :param dest: destination operand.
         :param src: source operand.
         '''
-        res = dest.write(dest.read() & src.read())
+        # XXX bypass a capstone bug that incorrectly extends and computes operands sizes
+        # the bug has been fixed since capstone 4.0.alpha2 (commit de8dd26)
+        if src.size == 64 and src.type == 'immediate' and dest.size == 64:
+            arg1 = Operators.SEXTEND(src.read(), 32, 64)
+        else:
+            arg1 = src.read()
+        res = dest.write(dest.read() & arg1)
         # Defined Flags: szp
         cpu._calculate_logic_flags(dest.size, res)
 
--- a/tests/test_cpu_manual.py
+++ b/tests/test_cpu_manual.py
@@ -711,6 +711,42 @@ class SymCPUTest(unittest.TestCase):
 
         self.assertTrue(solver.check(cs))
 
+    # regression test for issue #560
+    def test_AND_1(self):
+        ''' Instruction AND
+            Groups:
+            0x7ffff7de390a:	and rax, 0xfc000000
+        '''
+        mem = Memory64()
+        cpu = AMD64Cpu(mem)
+        mem.mmap(0x7ffff7de3000, 0x1000, 'rwx')
+        mem[0x7ffff7de390a] = '\x48'
+        mem[0x7ffff7de390b] = '\x25'
+        mem[0x7ffff7de390c] = '\x00'
+        mem[0x7ffff7de390d] = '\x00'
+        mem[0x7ffff7de390e] = '\x00'
+        mem[0x7ffff7de390f] = '\xfc'
+        cpu.PF = True
+        cpu.RAX = 0x7ffff7ff7658
+        cpu.OF = False
+        cpu.ZF = False
+        cpu.CF = False
+        cpu.RIP = 0x7ffff7de390a
+        cpu.SF = False
+        cpu.execute()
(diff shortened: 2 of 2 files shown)


mossberg:
awesome :) thanks for the submitting the PR, we'll get to it soon!

catenacyber:
I agree that moving to a newer version than capstone is a better fix for this bug than using a workaround.

@mossberg I let you decide which solution you prefer

catenacyber:
Finally test is ok :-)

mossberg:
finally getting to this!



> Here is a try for a unit test.
> Did I choose the right file ? ie tests/test_cpu_automatic.py

@catenacyber we use test_cpu_automatic.py for automatically generated unit tests, it's not a big deal, but could you please move the test to test_cpu_manual.py




> Do I need to set and test the flags ?
yes, that would be nice for completeness

---




> If I can suggest something, instead of a workaround like this, I would vote for installing next capstone branch, which means putting below line in requirements.txt:

@disconnect3d we don't use a requirements.txt file, we have everything in setup.py, so i'm not sure this solution can work for us, since github links are not allowed in the setup.py `install_requires` parameter.

disconnect3d:
@mossberg 
https://stackoverflow.com/questions/12518499/pip-ignores-dependency-links-in-setup-py

https://stackoverflow.com/questions/3472430/how-can-i-make-setuptools-install-a-package-thats-not-on-pypi

catenacyber:
> @catenacyber we use test_cpu_automatic.py for automatically generated unit tests, it's not a big deal, but could you please move the test to test_cpu_manual.py?

Here you go.
The unit test should be good whatever the solution (upgrade capstone or workaround)

How do you automatically generate the unit tests ?

mossberg:
> Seems it is doable with dependency_links

Based on my reading, it seems this requires the `--process-dependency-links` flag to be passed with the pip install command for all users. I'm not a big fan of making everyone do that, so I'm a little reluctant to specify the dep like this.

catenacyber:
> So, if I understand correctly: that capstone bug does not only affect the AND instruction, it actually affect many more instructions that use immediates, right?

That is also my understanding

PR #859, opened 2018-04-05
Speed up solver get_all_values

mossberg:
This removes some code that was doing a lot of unnecessary, expensive solver resetting.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/859)


feliam:
ROA approved!
https://github.com/trailofbits/manticore/blob/dev-evm-reservoir-of-awesomeness/manticore/core/smtlib/solver.py#L368-L373

Issue #69, opened 2017-03-14
Support loading ELF shared objects


offlinemark:
use case: analyzing a function in a .so

catenacyber:
What should be done now about this ?
Pass arguments to that function ?

PR #861, opened 2018-04-06
Loads shared ELF to analyze a function

catenacyber:
See #69 

For now, this is rather a PoC to discuss about.
The ELF is loaded, and the function specified with `manticore --function` is the new entry point.

There remains questions :
Are the patches in the right place ? 
Do we want to do it from the command line ?
I am doing the stuff in `load`because that is where I can get the info about `load_addr`
What should be done if the user specifies a function which is not in the library ? Raise an exception ?
How do we set arguments for the function ? Should Abi class have a `set_arguments`method ?

Plus there is a try except patch so that I can go on analyzing a .so with my MacOS even if I do not have the interpreter on my computer

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/861)

### DIFF
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -65,7 +65,7 @@ def parse_arguments():
     parser.add_argument('--workspace', type=str, default=None,
                         help=("A folder name for temporaries and results."
                               "(default mcore_?????)"))
-    parser.add_argument('--version', action='version', version='Manticore 0.1.8',
+    parser.add_argument('--version', action='version', version='Manticore 0.1.9',
                         help='Show program version information')
     parser.add_argument('--txlimit', type=positive,
                         help='Maximum number of symbolic transactions to run (positive integer) (Ethereum only)')
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -36,6 +36,11 @@ def parse_arguments():
                         help=argparse.SUPPRESS)
     parser.add_argument('--env', type=str, nargs=1, default=[], action='append',
                         help='Specify symbolic environment variable VARNAME=++++++')
+    #TODO allow entry as an address
+    #parser.add_argument('--entry', type=str, default=None,
+    #                    help='address as entry point')
+    parser.add_argument('--entrysymbol', type=str, default=None,
+                        help='symbol as entry point')
     parser.add_argument('--file', type=str, default=[], action='append', dest='files',
                         help='Specify symbolic input file, \'+\' marks symbolic bytes')
     parser.add_argument('--names', type=str, default=None,
@@ -115,7 +120,7 @@ def main():
 
     env = {key: val for key, val in map(lambda env: env[0].split('='), args.env)}
 
-    m = Manticore(args.argv[0], argv=args.argv[1:], env=env, workspace_url=args.workspace,  policy=args.policy, disasm=args.disasm, concrete_start=args.data)
+    m = Manticore(args.argv[0], argv=args.argv[1:], env=env, entry_symbol=args.entrysymbol, workspace_url=args.workspace,  policy=args.policy, disasm=args.disasm, concrete_start=args.data)
 
     # Fixme(felipe) remove this, move to plugin
     m.coverage_file = args.coverage
--- a/manticore/manticore.py
+++ b/manticore/manticore.py
@@ -258,7 +258,7 @@ class Manticore(Eventful):
         plugin.manticore = None
 
     @classmethod
-    def linux(cls, path, argv=None, envp=None, symbolic_files=None, concrete_start='', **kwargs):
+    def linux(cls, path, argv=None, envp=None, entry_symbol=None, symbolic_files=None, concrete_start='', **kwargs):
         """
         Constructor for Linux binary analysis.
 
@@ -267,6 +267,8 @@ class Manticore(Eventful):
         :type argv: list[str]
         :param envp: Environment to provide to the binary
         :type envp: dict[str, str]
+        :param entry_symbol: Entry symbol to resolve to start execution
+        :type envp: str
         :param symbolic_files: Filenames to mark as having symbolic input
         :type symbolic_files: list[str]
         :param str concrete_start: Concrete stdin to use before symbolic inputt
@@ -275,7 +277,7 @@ class Manticore(Eventful):
         :rtype: Manticore
         """
         try:
-            return cls(make_linux(path, argv, envp, symbolic_files, concrete_start), **kwargs)
+            return cls(make_linux(path, argv, envp, entry_symbol, symbolic_files, concrete_start), **kwargs)
         except elftools.common.exceptions.ELFError:
             raise Exception('Invalid binary: {}'.format(path))
 
--- a/manticore/manticore.py
+++ b/manticore/manticore.py
@@ -76,7 +76,7 @@ def make_decree(program, concrete_start='', **kwargs):
     return initial_state
 
 
-def make_linux(program, argv=None, env=None, symbolic_files=None, concrete_start=''):
+def make_linux(program, argv=None, env=None, entry_symbol=None, symbolic_files=None, concrete_start=''):
     env = {} if env is None else env
     argv = [] if argv is None else argv
     env = ['%s=%s' % (k, v) for k, v in env.items()]
@@ -86,6 +86,15 @@ def make_linux(program, argv=None, env=None, symbolic_files=None, concrete_start
     constraints = ConstraintSet()
     platform = linux.SLinux(program, argv=argv, envp=env,
                             symbolic_files=symbolic_files)
+    if entry_symbol is not None:
+        entry_pc = platform._find_symbol(entry_symbol)
+        if entry_pc is None:
+            logger.error("No symbol for '%s' in %s", entry_symbol, program)
+            raise Exception("Symbol not found")
+        else:
+            logger.info("Found symbol '%s' (%x)", entry_symbol, entry_pc)
+            #TODO: use argv as arguments for function
+            platform.set_entry(entry_pc)
 
     initial_state = State(constraints, platform)
 
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -10,6 +10,8 @@ import socket
 
 # Remove in favor of binary.py
 from elftools.elf.elffile import ELFFile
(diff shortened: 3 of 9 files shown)


yan:
@catenacyber Whoa, thank you, this looks great! I'm not sure if we'll get to reviewing this PR today or Monday, but stand by we'll make sure to get to it soon.

catenacyber:
No worries, no hurry

manticore/__main__.py
```diff
@@ -36,6 +36,8 @@ def positive(value):
                         help=argparse.SUPPRESS)
     parser.add_argument('--env', type=str, nargs=1, default=[], action='append',
                         help='Specify symbolic environment variable VARNAME=++++++')
+    parser.add_argument('--function', type=str, default=None,
```
yan: I feel that `--entry` (and `entry` in general for variable naming, etc) is a more descriptive name than `function`. It more accurately represents what we're selecting, and is less confusing imho. However, since the current code dispatches based on name, I'd suggest adding `--entry` that takes an address and `--entry-symbol` which takes a string and resolves it correctly.

manticore/platforms/linux.py
```diff
@@ -894,6 +922,14 @@ def load(self, filename):
                 elf_brk = k
 
         elf_entry = elf.header.e_entry
+        if function is not None:
+            symbolPC = self._find_symbol(function)
+            if symbolPC is None:
+                logger.error("No symbol for %s in %s", function, program)
+                #TODO raise exception
```
yan: Actually raise here?

manticore/platforms/linux.py
```diff
@@ -826,7 +851,10 @@ def load(self, filename):
                 continue
             interpreter_filename = elf_segment.data()[:-1]
             logger.info('Interpreter filename: %s', interpreter_filename)
-            interpreter = ELFFile(file(interpreter_filename))
+            try:
+                interpreter = ELFFile(file(interpreter_filename))
+            except:
```
yan: Except on a specific exception?

yan:
Thanks again @catenacyber! I added a few comments above, we'd be happy to merge after they're addressed.

manticore/platforms/linux.py
```diff
@@ -447,20 +450,42 @@ def _init_cpu(self, arch):
         self._function_abi = CpuFactory.get_function_abi(cpu, 'linux', arch)
         self._syscall_abi = CpuFactory.get_syscall_abi(cpu, 'linux', arch)
 
-    def _execve(self, program, argv, envp):
+    def _find_symbol(self, name):
+        symbol_tables = [s for s in self.elf.iter_sections()
+                         if isinstance(s, SymbolTableSection)]
+
+        if not symbol_tables and self.elffile.num_sections() == 0:
+            return None
```
disconnect3d: I think this condition is redundant. If `symbol_tables` will be `[]` we will get into `return None` anyway.

manticore/platforms/linux.py
```diff
@@ -447,20 +450,42 @@ def _init_cpu(self, arch):
         self._function_abi = CpuFactory.get_function_abi(cpu, 'linux', arch)
         self._syscall_abi = CpuFactory.get_syscall_abi(cpu, 'linux', arch)
 
-    def _execve(self, program, argv, envp):
+    def _find_symbol(self, name):
+        symbol_tables = [s for s in self.elf.iter_sections()
```
disconnect3d: We can freely go with creating a generator here instead of a list (change `[...]` to `(...)`) as we do `return None` if we iterate over nothing anyway.

manticore/platforms/linux.py
```diff
@@ -447,20 +450,42 @@ def _init_cpu(self, arch):
...
+                         if isinstance(s, SymbolTableSection)]
+
+        if not symbol_tables and self.elffile.num_sections() == 0:
+            return None
+
+        for section in symbol_tables:
+            if not isinstance(section, SymbolTableSection):
```
disconnect3d: This `isinstance(s, SymbolTableSection)` is already considered when we create `symbol_tables`, so this condition can be removed.

manticore/__main__.py
catenacyber: Ok, doing entry-symbol first

manticore/platforms/linux.py
```diff
@@ -447,20 +450,42 @@ def _init_cpu(self, arch):
...
+                continue
+
+            if section['sh_entsize'] == 0:
+                continue
+
+            for nsym, symbol in enumerate(section.iter_symbols()):
```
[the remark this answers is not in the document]
catenacyber: Correct

manticore/platforms/linux.py
catenacyber: a normal Exception ? or is there a better subclass ?

manticore/platforms/linux.py
catenacyber: Ok for IOError

manticore/platforms/linux.py
catenacyber: indeed

manticore/platforms/linux.py
catenacyber: ok

manticore/platforms/linux.py
catenacyber: indeed

catenacyber:
I am pushing a new commit over the old one so that review can be easier. They should be squashed when merged into master branch.

manticore/__main__.py
```diff
@@ -36,6 +36,11 @@ def positive(value):
                         help=argparse.SUPPRESS)
     parser.add_argument('--env', type=str, nargs=1, default=[], action='append',
                         help='Specify symbolic environment variable VARNAME=++++++')
+    #TODO allow entry as an address
+    #parser.add_argument('--entry', type=str, default=None,
+    #                    help='entry point as an address')
+    parser.add_argument('--entrysymbol', type=str, default=None,
+                        help='entry point as a symbol')
```
yan: Think you mean "symbol as entry point"? (ditto for comment above)

manticore/__main__.py
```diff
@@ -112,7 +117,7 @@ def main():
 
     env = {key: val for key, val in map(lambda env: env[0].split('='), args.env)}
 
-    m = Manticore(args.argv[0], argv=args.argv[1:], env=env, workspace_url=args.workspace,  policy=args.policy, disasm=args.disasm, concrete_start=args.data)
+    m = Manticore(args.argv[0], argv=args.argv[1:], env=env, entry=args.entrysymbol, workspace_url=args.workspace,  policy=args.policy, disasm=args.disasm, concrete_start=args.data)
```
yan: There's a disconnect from command line argument naming (`entrysymbol`) to internal naming (`entry`), especially that the TODO indicates future support for entry.

manticore/manticore.py
```diff
@@ -76,15 +76,15 @@ def make_decree(program, concrete_start='', **kwargs):
...
+def make_linux(program, argv=None, env=None, entry=None, symbolic_files=None, concrete_start=''):
     env = {} if env is None else env
     argv = [] if argv is None else argv
     env = ['%s=%s' % (k, v) for k, v in env.items()]
 
     logger.info('Loading program %s', program)
 
     constraints = ConstraintSet()
-    platform = linux.SLinux(program, argv=argv, envp=env,
+    platform = linux.SLinux(program, argv=argv, envp=env, entry=entry,
```
yan: I've been looking over this code and trying to determine why this feels like it makes `SLinux` API feel heavy. I am not sure that modifying the loading process to add an entry is the cleanest way to go; otherwise if we add future support for other methods of initialization, it'll grow it further. After some thought, what if we do the following:

Leave the loading process as it was, and add a `set_entry()` method that takes a virtual address.[1] You can then use that and `_find_symbol()` below (or potentially renamed to `_resolve_symbol()`) to alter the entry point after initialization. This way, we also won't introduce ambiguous parameters (`entry`/`entrysymbol`/etc) to the constructor and our elf loading process more closely resembles the actual process in Linux. 

So new initialization will roughly look like:
```python
platform = linux.SLinux(platform, argv, etc...)
if args.entrysymbol and args.entry:
    # raise an exception and notify that they should be mutually exclusive
if args.entrysymbol:
    entry = platform._find_symbol(args.entrysymbol)
    platform.set_entry(entry)
if args.entry:
    platform.set_entry(args.entry)
initial_state = State(constraints, platform)
...
etc
```

What are your thoughts?

[1] We might be able to extend this API to other platforms.

manticore/platforms/linux.py
```diff
@@ -826,7 +845,10 @@ def load(self, filename):
                 continue
             interpreter_filename = elf_segment.data()[:-1]
             logger.info('Interpreter filename: %s', interpreter_filename)
-            interpreter = ELFFile(file(interpreter_filename))
+            try:
+                interpreter = ELFFile(file(interpreter_filename))
```
yan: Check with the `python3` branch; this might have been changed to use `open`.

manticore/__main__.py
catenacyber: Ok, sorry for my English

manticore/__main__.py
catenacyber: I meant that.
Argument parsing should only allow either one.
And we can pass it to subsequent functions which could tell the difference, based on the type (an address would be an unsigned int and a symbol would be a string).
What do you think of this ?

manticore/platforms/linux.py
catenacyber: Ok

manticore/manticore.py
catenacyber: Looks good. I just had to store the load_addr variable so that it an be used in `set_entry`

yan:
@catenacyber I took a few mins to cleanup your branch, it's almost ready to merge. Do you think you can update your fork to track our most recent master? I am able to push to `elfshare`, but i can't sync to upstream.

catenacyber:
> Do you think you can update your fork to track our most recent master?

@yan I just did : is it ok ?

yan:
@catenacyber I don't see most recent changes updated in your `master`. Check your fork, `This branch is 2 commits ahead, 13 commits behind trailofbits:master`, make sure master is updated and that `elfshare` is merged. I am afraid some of the current test failures we see are related to stale code.

catenacyber:
@yan just did the git pull rebase and push

Issue #110, opened 2017-03-27
Initial support for macOS


mossberg:
we've gotten a few requests from users to be able to have Manticore run on OS X. this shouldn't be too hard as nothing we do inherently ties us to linux. we'd need to

- [ ] eliminate all uses of /dev (i think there are a few)
- [x] make sure all python stdlib calls we make work uniformly on linux/osx
- [x] make sure python dependencies work on both
- [x] add features to allow users to specify locations for ld.so and libraries (#93)

chronoslynx:
What should be done to verify proper execution? I've some small fixes to the tests that ensure they work on Mac OS, and I'm working through the linux examples and they seem to work just fine.

mossberg:
The example binaries in `examples/linux` are configured to be statically linked, but dynamic binaries will not work on macOS. I also just added that to the comment above

catenacyber:
I am using MacOS and managing to run manticore, but maybe not all the features.
The only hack I did is the pull request above.

What is left to be done here ?

mossberg:
@catenacyber check out the top comment, primarily #93 is the big one :)

mossberg:
at this point since #93 is closed, i think the main thing is making sure all of those checkboxes in the top comment are completed, and doing testing of running manticore on mac. analyzing mach-o binaries is out of scope of this issue, we are simply looking to have mac users able to use manticore on their computers :)

catenacyber:
Ok so first one 

> eliminate all uses of /dev (i think there are a few)

doing `grep -r "/dev" manticore`
I only see one : ```manticore/platforms/decree.py:                data = file("/dev/urandom", "r").read(count)```

But I have my `/dev/urandom` on macOS as well

Is this ok to you  ?

mossberg:
ah right, macos is unixy 😅  that looks fine to me

catenacyber:
Second one 

> make sure all python stdlib calls we make work uniformly on linux/osx

Maybe #897 is part of this, what do you think @mossberg ?

mossberg:
yup, i'd say so

catenacyber:
Any ideas on how to test *all* stdlib calls ?

mossberg:
hm, i'm not totally sure. i think it might just be good enough to do a cursory review of the codebase and document any stdlib calls we make that are particularly suspect for non-portability.

catenacyber:
[...]

PR #862, opened 2018-04-06
MacOS support : hack around SIP to load libc

catenacyber:
see https://stackoverflow.com/questions/32905322/oserror-dlopenlibsystem-dylib-6-image-not-found

For #110

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/862)

### DIFF
--- a/manticore/utils/mappings.py
+++ b/manticore/utils/mappings.py
@@ -7,44 +7,44 @@ munmap_function = None
 
 # initialize global functions depending on platform
 osname = sys.platform.lower()
-if osname == "darwin" or osname.startswith("linux"):
 
-    if osname == "darwin":
-        try:
-            libc = ctypes.cdll.LoadLibrary("libc.dylib")
-        except:
-            libc = ctypes.cdll.LoadLibrary("/usr/lib/libc.dylib")
-    else:
-        libc = ctypes.cdll.LoadLibrary("libc.so.6")
-
-    # void* mmap(void* addr, size_t len, int prot, int flags, int fd, off_t offset)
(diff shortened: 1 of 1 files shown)


mossberg:
hi @catenacyber can you provide me with a little additional context around the PR? what was the situation you ran into that motivated this change? i have very little experience trying to run manticore on os x :)

catenacyber:
After my latest git pull, manticore did not start anymore and I had a python traceback.
Looking at it, it was pretty obvious that it failed trying to open libc.dylib
So I looked for it and found it in /usr/lib/
Maybe something smarter can be done to find the right path...

With this hack, I was able to run manticore again and to work on #861 

Manticore is not the only one with this problem, see for instance https://github.com/ffi/ffi/issues/461

manticore/utils/mappings.py
```diff
@@ -10,7 +10,10 @@
 if osname == "darwin" or osname.startswith("linux"):
 
     if osname == "darwin":
-        libc = ctypes.cdll.LoadLibrary("libc.dylib")
+        try:
```
disconnect3d: It looks like we can refactor above ifs into:
```
if osname == 'darwin':
    # ...
elif osname.startswith('linux'):
    # ...
```

yan:
@catenacyber , sorry for the late review; can you update your branch to current `master`? That file has significantly changed, this PR would likely need to be reworked or might not be necessary anymore.

catenacyber:
It looks like it is working now without my patch, so we should just close this pull request...

PR #863, opened 2018-04-06
Add dedicated files for ARGV/ENVP in report

### LINKED ISSUE #169: Add a dedicated file for ARGV/ENVP in the final report (generate_testcase)
stdin, stderr, stdout all have a dedicated file at the final state report in the workspace.
Lets add another like that for arguments and environment variables. As those are common inputs of the programs we deal with.

catenacyber:
Fixes #169

Is this exhaustive ?
Is `repr` ok for the reports ?
Do you need tests for this ?

Example 
`manticore -v --env "var1=value1" binary arg1 arg2`
Then
`cat mcore_FgaI4o/test_00000000.argv `gives`['arg1', 'arg2']`
`cat mcore_FgaI4o/test_00000000.env `gives`['lol=oui']`

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/863)

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -388,6 +388,8 @@ class Linux(Platform):
         # Many programs to support SLinux
         self.programs = program
         self.disasm = disasm
+        self.envp = envp
+        self.argv = argv
 
         # dict of [int -> (int, int)] where tuple is (soft, hard) limits
         self._rlimits = {
@@ -513,6 +515,8 @@ class Linux(Platform):
         state['twait'] = self.twait
         state['timers'] = self.timers
         state['syscall_trace'] = self.syscall_trace
+        state['argv'] = self.argv
+        state['envp'] = self.envp
         state['base'] = self.base
         state['elf_bss'] = self.elf_bss
         state['end_code'] = self.end_code
@@ -565,6 +569,8 @@ class Linux(Platform):
         self.clocks = state['clocks']
 
         self.syscall_trace = state['syscall_trace']
+        self.argv = state['argv']
+        self.envp = state['envp']
         self.base = state['base']
         self.elf_bss = state['elf_bss']
         self.end_code = state['end_code']
@@ -826,7 +832,10 @@ class Linux(Platform):
                 continue
             interpreter_filename = elf_segment.data()[:-1]
             logger.info('Interpreter filename: %s', interpreter_filename)
-            interpreter = ELFFile(file(interpreter_filename))
+            try:
+                interpreter = ELFFile(file(interpreter_filename))
+            except:
+                pass
             break
         if interpreter is not None:
             assert interpreter.get_machine_arch() == elf.get_machine_arch()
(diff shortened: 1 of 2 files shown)


manticore/platforms/linux.py
```diff
@@ -2623,6 +2629,8 @@ def solve_to_fd(data, fd):
 
         ret = {
             'syscalls': repr(self.syscall_trace),
+            'argv': repr(self.argv),
```
feliam: I _think_ you need to solve an example solution out of argv/envp here. Otherwise you'll get the repr for an Expression. You need to apply something like solve_to_fd() for each variable/argument.

catenacyber: Thanks, will look into that

manticore/platforms/linux.py
```diff
@@ -2621,8 +2629,34 @@ def solve_to_fd(data, fd):
...
+        argIO.write("[")
+        for a in self.argv:
+            if first:
+                first = False
+            else:
+                argIO.write(", ")
+            argIO.write("'")
+            solve_to_fd(a, argIO)
+            argIO.write("'")
+        argIO.write("]")
```
yan: Hey, a few comments about this specific implementation: This doesn't feel very pythonic. If summarizing like above, create an intermediate list with `solve_to_fd` results than just use `str.join` to create the final report. That, or even simpler, just produce a single argument per line since it's being produced to a separate file anyway.

yan: Also, if you can, add a unit test or two.

catenacyber: > That, or even simpler, just produce a single argument per line since it's being produced to a separate file anyway.

Not sure I got it.
Is the new patch better ?
How would you handle arguments with multiple lines ?

catenacyber:
Do you have a good example for a unit test ?
I do not see any for the sys calls output file for instance (except testing that it exists)

manticore/platforms/linux.py
```diff
@@ -826,7 +832,10 @@ def load(self, filename):
                 continue
             interpreter_filename = elf_segment.data()[:-1]
             logger.info('Interpreter filename: %s', interpreter_filename)
-            interpreter = ELFFile(file(interpreter_filename))
+            try:
+                interpreter = ELFFile(file(interpreter_filename))
+            except:
+                pass
```
yan: Why ignore the exception?

manticore/platforms/linux.py
```diff
@@ -2621,8 +2632,18 @@ def solve_to_fd(data, fd):
             if name in ('_receive', '_read') and fd == 0:
                 solve_to_fd(data, inn)
 
+        for a in self.argv:
+            solve_to_fd(a, argIO)
+            argIO.write("\n")
+
+        for e in self.envp:
+            solve_to_fd(e, envIO)
+            envIO.write("\n")
```
yan: Would this label the variables correctly in the output?

manticore/platforms/linux.py
catenacyber: My mistake : I will just use the code from python3 branch

manticore/platforms/linux.py
catenacyber: Depends on the use cases, as I asked previously : 
Are only values symbolic ?
Or do we want environment variable names to be symbolic as well ?

PR #865, opened 2018-04-08
Initial automatic library loading (now with some form of toposort)

feliam:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/865)


mossberg:
include a unit test with this also?

manticore/ethereum.py
```diff
@@ -27,6 +27,12 @@ class EthereumError(ManticoreError):
     pass
 
 
+class DependencyError(EthereumError):
+    def __init__(self, lib_names):
+        super(DependencyError, self).__init__("You must pre-load and provide libraries addresses{ libname:address, ...} for %r" % lib_names)
```
disconnect3d: space in wrong position: `libraries addresses{ libname:address, ...}` -> `libraries addresses {libname:address, ...}`

manticore/ethereum.py
```diff
@@ -655,17 +661,49 @@ def make_symbolic_value(self):
...
+    def compile(source_code, contract_name=None, libraries=None):
         ''' Get initialization bytecode from a Solidity source code '''
-        name, source_code, bytecode, runtime, srcmap, srcmap_runtime, hashes, abi, warnings = ManticoreEVM._compile(source_code, contract_name)
+        name, source_code, bytecode, runtime, srcmap, srcmap_runtime, hashes, abi, warnings = ManticoreEVM._compile(source_code, contract_name, libraries)
         return bytecode
 
     @staticmethod
-    def _compile(source_code, contract_name):
+    def _link(bytecode, libraries=None):
+        has_dependencies = '_' in bytecode
```
disconnect3d: Might be good to explain it somewhere why `_` in bytecode means it has dependencies (or give link to some source?)
(is it that obvious?)

feliam: A link to some doc should be good to have.
(bytecode should be hex.. I assume the _ will be in all the link placeholders)

manticore/ethereum.py
```diff
@@ -655,17 +661,49 @@ def make_symbolic_value(self):
...
+                    lib_name = lib_placeholder.split(':')[1].split('_')[0]
+                    deps.setdefault(lib_name, []).append(pos)
+                    pos += 40
+                else:
+                    pos += 2
+
+            if libraries is None:
+                raise DependencyError(deps.keys())
+            libraries = dict(libraries)
+            hex_contract_lst = list(hex_contract)
```
disconnect3d: We are changing it to list just because a string doesn't support item assignment, is that right?

Maybe we should use bytearray instead for that purpose? An example:
```
>>> a = bytearray('x'*10, 'utf8')
>>> a[2:4] = b'xy'
>>> a[5:6] = [1, 2]
>>> a[-1] = 123
>>> a
bytearray(b'xxxyx\x01\x02xxx{')
>>> bytes(a)
b'xxxyx\x01\x02xxx{'
```

manticore/ethereum.py
```diff
@@ -697,14 +735,15 @@ def _compile(source_code, contract_name):
 
             if contract['bin'] == '':
                 raise Exception('Solidity failed to compile your contract.')
-                
-            bytecode = contract['bin'].decode('hex')
+
+            bytecode = ManticoreEVM._link(contract['bin'], libraries).decode('hex')
```
disconnect3d: `str.decode` doesn't exist in Py3 anymore. Use `import codecs` and `codecs.decode`

manticore/ethereum.py
```diff
@@ -697,14 +735,15 @@ def _compile(source_code, contract_name):
...
             srcmap = contract['srcmap'].split(';')
             srcmap_runtime = contract['srcmap-runtime'].split(';')
             hashes = contract['hashes']
             abi = json.loads(contract['abi'])
-            runtime = contract['bin-runtime'].decode('hex')
+            runtime = ManticoreEVM._link(contract['bin-runtime'], libraries).decode('hex')
```
disconnect3d: `str.decode` doesn't exist in Py3 anymore. Use `import codecs` and `codecs.decode`

manticore/ethereum.py
mossberg: I think `bytes.fromhex` is what we've been using in the py3 port branch

manticore/ethereum.py
```diff
@@ -668,6 +669,16 @@ def compile(source_code, contract_name=None, libraries=None):
...
+        '''
+        Fill bytecode placeholders with library addresses.
+
+        From https://solidity.readthedocs.io/en/v0.4.21/contracts.html:
+        `If the addresses are not given as arguments to the compiler, the 
+        compiled hex code will contain placeholders of the form __Set______
+        (where Set is the name of the library). The address can be filled
+        manually by replacing all those 40 symbols by the hex encoding of 
+        the address of the library contract.`
+        '''
```
disconnect3d: For consistency: use `"""` :P

feliam:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[manticore/ethereum.py, line 671 at r2](https://beta.reviewable.io/reviews/trailofbits/manticore/865#-L9adUeMic-1rm0ifhjB-r2-671:-LAOGkX1Ai64nIDWhBq7:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/7062ecd498040d23d5b577e05cffcd8d8aed358e/manticore/ethereum.py#L671)):*

A link to some doc should be good to have.
(bytecode should be hex.. I assume the _ will be in all the link placeholders)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/trailofbits/manticore/865)*

tests/test_binaries.py
```diff
@@ -146,6 +146,7 @@ def test_eth_regressions(self):
             {'number': 795, 'contract': None},
             {'number': 799, 'contract': 'C'},
             {'number': 807, 'contract': 'C'},
+            {'number': 808, 'contract': 'C'},
```
mossberg: this test is not a regression test, so it doesn't quite belong here (the numbers here correspond to github issues). this seems more like an integration test, so i think that would fit better in a new function for integration tests?

tests/binaries/808.sol
```diff
@@ -0,0 +1,8 @@
+library X {
```
mossberg: can we rename this file to have a more descriptive name? the number scheme was just for contracts that correspond to github issues.

manticore/ethereum.py
```diff
@@ -866,7 +914,7 @@ def transactions(self, state_id=None):
         state = self.load(state_id)
         return state.platform.transactions
 
-    def solidity_create_contract(self, source_code, owner, contract_name=None, balance=0, address=None, args=()):
+    def solidity_create_contract(self, source_code, owner, contract_name=None, libraries=None, balance=0, address=None, args=()):
```
mossberg: pls document type of the libraries param here too?

manticore/ethereum.py
```diff
@@ -655,17 +662,57 @@ def make_symbolic_value(self):
...
+                    hex_contract[pos:pos + 40] = '%040x' % lib_address
+        return bytes(hex_contract)
+
+    @staticmethod
+    def _compile(source_code, contract_name, libraries=None):
         """ Compile a Solidity contract, used internally
 
             :param source_code: a solidity source code
             :param contract_name: a string with the name of the contract to analyze
+            :param libraries: an itemizable of pairs (library_name, address)
```
mossberg: itemizable? iterable?

manticore/ethereum.py
```diff
@@ -655,17 +662,57 @@ def make_symbolic_value(self):
...
+
+        From https://solidity.readthedocs.io/en/v0.4.21/contracts.html:
+        `If the addresses are not given as arguments to the compiler, the 
+        compiled hex code will contain placeholders of the form __Set______
+        (where Set is the name of the library). The address can be filled
+        manually by replacing all those 40 symbols by the hex encoding of 
+        the address of the library contract.`
+        """
+        has_dependencies = '_' in bytecode
+        hex_contract = bytearray(bytecode, 'utf8')
```
mossberg: is utf8 an encoding? not 'utf-8'?

manticore/ethereum.py
```diff
@@ -655,17 +662,57 @@ def make_symbolic_value(self):
...
-    def compile(source_code, contract_name=None):
+    def compile(source_code, contract_name=None, libraries=None):
         ''' Get initialization bytecode from a Solidity source code '''
-        name, source_code, bytecode, runtime, srcmap, srcmap_runtime, hashes, abi, warnings = ManticoreEVM._compile(source_code, contract_name)
+        name, source_code, bytecode, runtime, srcmap, srcmap_runtime, hashes, abi, warnings = ManticoreEVM._compile(source_code, contract_name, libraries)
         return bytecode
 
     @staticmethod
-    def _compile(source_code, contract_name):
+    def _link(bytecode, libraries=None):
```
mossberg: since this is a stateless staticmethod, can't we write some unit tests pretty easily?

manticore/ethereum.py
```diff
@@ -655,17 +662,57 @@ def make_symbolic_value(self):
...
+        From https://solidity.readthedocs.io/en/v0.4.21/contracts.html:
+        `If the addresses are not given as arguments to the compiler, the 
+        compiled hex code will contain placeholders of the form __Set______
+        (where Set is the name of the library). The address can be filled
+        manually by replacing all those 40 symbols by the hex encoding of 
+        the address of the library contract.`
+        """
+        has_dependencies = '_' in bytecode
+        hex_contract = bytearray(bytecode, 'utf8')
+        if has_dependencies:
```
mossberg: nit style-wise can we maybe invert this to return early, so the main code here doesn't have to be indented?

manticore/ethereum.py
```diff
@@ -655,17 +662,57 @@ def make_symbolic_value(self):
...
+            if libraries is None:
+                raise DependencyError(deps.keys())
+            libraries = dict(libraries)
+            for lib_name, pos_lst in deps.items():
+                try:
+                    lib_address = libraries[lib_name]
+                except KeyError:
+                    raise DependencyError([lib_name])
+                for pos in pos_lst:
+                    hex_contract[pos:pos + 40] = '%040x' % lib_address
```
mossberg: .format please? `'{:040x}'.format()`?

manticore/ethereum.py
```diff
@@ -27,6 +28,12 @@ class EthereumError(ManticoreError):
     pass
 
 
+class DependencyError(EthereumError):
+    def __init__(self, lib_names):
+        super(DependencyError, self).__init__("You must pre-load and provide libraries addresses {libname:address, ...} for %r" % lib_names)
```
mossberg: prefer format here too?

manticore/ethereum.py
```diff
@@ -655,17 +662,57 @@ def make_symbolic_value(self):
...
+        (where Set is the name of the library). The address can be filled
+        manually by replacing all those 40 symbols by the hex encoding of 
+        the address of the library contract.`
+        """
+        has_dependencies = '_' in bytecode
+        hex_contract = bytearray(bytecode, 'utf8')
+        if has_dependencies:
+            deps = {}
+            pos = 0
+            while pos < len(hex_contract):
```
mossberg: maybe we could split this body into a few functions for parsing out the deps/offsets, and patching them in?

PR #848, opened 2018-03-26
Fix check of solc metadata tail

feliam:
Fixes 
https://github.com/trailofbits/manticore/issues/846
https://github.com/trailofbits/manticore/issues/844

Creates the need for
https://github.com/trailofbits/manticore/issues/847

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/848)

PR #873, opened 2018-04-14
Support Solidity compilation of a file handle

### LINKED ISSUE #705: Manticore does not support contracts with imports
### OS / Environment

Distributor ID:	ArchLinux
Description:	ArchLinux
Release:	rolling-release

### Manticore version

Last revision (465c926a2e39d5daf1405711dac61b6ddd0a2527)

### Python version

Python 2.7.13

### Summary of the problem

Loading a contract with imports fails to compile.

Use these two small solidity contracts to reproduce this issue:

* C.sol:

```
contract C {
}
```

* imp.sol 

```
import "C.sol";
```

Manticore works correctly if we use C.sol directly:

```
2018-01-12 07:32:53,997: [7335] m.main:INFO: Beginning analysis
2018-01-12 07:32:54,018: [7335] m.ethereum:INFO: Starting symbolic transaction: 1
2018-01-12 07:32:54,149: [7335] m.ethereum:INFO: Generated testcase No. 0 - REVERT
2018-01-12 07:32:55,053: [7335] m.ethereum:INFO: Generated testcase No. 1 - REVERT
2018-01-12 07:32:56,828: [7335] m.ethereum:INFO: Finished symbolic transaction: 1 | Code Coverage: 100% | Terminated States: 2 | Alive States: 0
2018-01-12 07:32:56,848: [7335] m.ethereum:INFO: Results in /home/gustavo/projects/gemini-wallets/manticore/mcore_n5WljI
```

But it fails to compile if we use imp.sol.

### Expected behavior

Manticore should be able to handle contracts with imports.

### Actual behavior

```
2018-01-12 07:37:08,889: [7458] m.main:INFO: Beginning analysis
  File "/home/gustavo/.local/bin/manticore", line 9, in <module>
    load_entry_point('manticore==0.1.6', 'console_scripts', 'manticore')()
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/__main__.py", line 104, in main
    ethereum_cli(args)
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/__main__.py", line 94, in ethereum_cli
    m.multi_tx_analysis(args.argv[0], args.txlimit)
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/ethereum.py", line 872, in multi_tx_analysis
    contract_account = self.solidity_create_contract(source_code, owner=user_account)
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/ethereum.py", line 775, in solidity_create_contract
    compile_results = self._compile(source_code)
  File "/home/gustavo/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/ethereum.py", line 587, in _compile
    raise Exception('Solidity compilation error')
Exception: Solidity compilation error
```

dwhjames:
Issue #705 reports that Manticore doesn’t support Solidity code with imports.
This is because Manticore currently only supports loading source code via strings.
This PR adds support for compiling code via a file handle, which then allows the Solidity compiler to resolve the imports.

---

This preserves the current behavior of using a string value as Solidity
source code, but now also supports instances of the `file` type, and
will use the `name` property of the file object to pass to the
subprocess invocation of solc.

This implementation also now ensures resource cleanup for the stdout and
stderr pipe handles.

---

fixes #705

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/873)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -658,13 +658,12 @@ class ManticoreEVM(Manticore):
         return bytecode
 
     @staticmethod
-    def _compile(source_code, contract_name):
-        """ Compile a Solidity contract, used internally
+    def _run_solc(source_file):
+        ''' Compile a source file with the Solidity compiler
 
-            :param source_code: a solidity source code
-            :param contract_name: a string with the name of the contract to analyze
-            :return: name, source_code, bytecode, srcmap, srcmap_runtime, hashes
-        """
+            :param source_file: a file object for the source file
+            :return: output, warnings
+        '''
         solc = "solc"
 
         #check solc version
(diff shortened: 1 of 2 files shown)


manticore/ethereum.py
```diff
@@ -712,6 +712,11 @@ def _compile(source_code, contract_name):
                     break
 
         assert(name is not None)
+        # capture source code if file handle was passed as arg
+        if isinstance(source_code, file):
```
disconnect3d: This won't work with Py3 as there is no `file` constant there.

The proper way to check this in py3 is to do:
```python
>>> from io import IOBase
>>> isinstance(open('/tmp/demo2', 'w'), IOBase)
True
```

This however doesn't work in Py2 so we should probably do:
```
try:
    file_type = file  # Python 2
except NameError:
    file_type = IOBase  # Python 3

# ...
if isinstance(source_code, file_type):
```

manticore/ethereum.py
```diff
@@ -668,44 +688,47 @@ def _compile(source_code, contract_name):
             :param contract_name: a string with the name of the contract to analyze
             :return: name, source_code, bytecode, srcmap, srcmap_runtime, hashes
         """
-        solc = "solc"
-        with tempfile.NamedTemporaryFile() as temp:
-            temp.write(source_code)
-            temp.flush()
-            p = Popen([solc, '--combined-json', 'abi,srcmap,srcmap-runtime,bin,hashes,bin-runtime', '--allow-paths', '.', temp.name], stdout=PIPE, stderr=PIPE)
+        if isinstance(source_code, str):
```
disconnect3d: Change function docstring to say that `source_code` is either `str` or a file object

dwhjames: fixed in https://github.com/trailofbits/manticore/pull/873/commits/6cb4bfaf1ceded779c20349d5311cff4b9d9675d
I also updated the doc string for the tuple of return values

dwhjames:
Branch has been rebased to resolve the conflict with PR #848

PR #874, opened 2018-04-14
Support ABI parsing of specs that use function types

dwhjames:
See http://solidity.readthedocs.io/en/v0.4.21/types.html#function-types

This is a special case of `bytes24`, 24 bytes are right zero padded in a 32 byte word.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/874)


mossberg:
ty unit tests! 😍

PR #875, opened 2018-04-17
Helpers for serializing bytesN and function ABI types

dwhjames:
This is a followup to #874 and extends `ABI.serialize` to support statically sized bytes and external function references (`address` + `bytes4` function selector; special case of `bytes24`).

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/875)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -288,6 +288,76 @@ class ABI(object):
     SValue = None
 
     @staticmethod
+    def _tuple_bytes_to_long(tpl):
+        return reduce(lambda acc, b: (acc << 8) + ord(b), tpl, 0L)
+
+    class StaticBytes(object):
+        def __init__(self, value, size=32):
+            if not 0 < size <= 32:
+                raise ValueError('bytes<M>: binary type of M bytes, 0 < M <= 32.')
+            if isinstance(value, (int, long)):
+                if not 0 <= value < (2 ** (size*8)):
+                    raise Value("value '%s' exceeds size %s" % (hex(value), size))
+                self.sub_word = ABI.serialize_uint(value, size)
+            elif isinstance(value, str) and value[:2] == '0x':
+                if not 3 <= len(value) <= 42:
+                    raise ValueError("hex '%s' does not fix in 32-byte bounds" % value)
+                # len(value) - 2 is the hex length
+                # (n + 1) // 2 gives the ceiling (accounts for odd length hex)
+                byte_len = (len(value) - 1) // 2
+                self.sub_word = ABI.serialize_uint(long(value, 16), byte_len)
+            elif isinstance(value, tuple):
+                if not 0 < len(value) <= 32:
+                    raise ValueError("value %s does not fix in 32-byte bounds" % value)
+                self.sub_word = value
+            else:
+                raise TypeError('Expected int, long, str, or tuple')
+
+        def size(self):
+            return len(self.sub_word)
+
+        def serialize(self):
+            return self.sub_word + tuple('\0' * (32 - len(self.sub_word)))
+
+    class Address(object):
+        def __init__(self, value):
+            if isinstance(value, (int, long)):
+                if not 0 <= value < (2 ** 160):
+                    raise Value("value '%s' is too large to be an address" % hex(value))
+                self.value = value
+            elif isinstance(value, str) and value[:2] == '0x':
+                if len(value) != 42:
+                    raise ValueError("hex '%s' does not have 'address'-length" % value)
+                self.value = long(value, 16)
+            elif isinstance(value, tuple):
+                if len(value) != 20:
+                    raise ValueError("value %s does not have 'address'-length" % (value, size))
+                self.value = ABI._tuple_bytes_to_long(value)
+            else:
+                raise TypeError('Expected int, long, str, or tuple')
+
+        def serialize(self):
+            return ABI.serialize_uint(self.value)
+
+    class ExternalFunction(object):
+        def __init__(self, address, selector):
+            if not isinstance(address, ABI.Address):
+                raise TypeError("Expected ABI.Address, found %s" % type(address))
+            if not isinstance(selector, ABI.StaticBytes):
+                raise TypeError("Expected ABI.StaticBytes, found %s" % type(selector))
+            if selector.size() != 4:
+                raise ValueError("Expected selector size of 4, found %s", selector.size())
+            self.address = address
+            self.selector = selector
+
+        def serialize(self):
+            return (
+                ABI.serialize_uint(self.address.value, 20) +
+                self.selector.sub_word +
+                tuple('\0' * 8)
+            )
+
+    @staticmethod
     def serialize(value):
         '''
         Translates a Python object to its EVM ABI serialization.
@@ -298,6 +368,12 @@ class ABI(object):
             return ABI.serialize_array(value)
         if isinstance(value, (int, long)):
             return ABI.serialize_uint(value)
+        if isinstance(value, ABI.StaticBytes):
+            return value.serialize()
+        if isinstance(value, ABI.Address):
+            return value.serialize()
+        if isinstance(value, ABI.ExternalFunction):
+            return value.serialize()
         if isinstance(value, ABI.SByte):
             return ABI.serialize_uint(value.size) + (None,) * value.size + (('\x00',) * (32 - (value.size % 32)))
         if value is None:
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -296,7 +296,7 @@ class ABI(object):
             if not 0 < size <= 32:
                 raise ValueError('bytes<M>: binary type of M bytes, 0 < M <= 32.')
             if isinstance(value, (int, long)):
-                if not 0 <= value < (2 ** (size*8)):
+                if not 0 <= value < (2 ** (size * 8)):
                     raise Value("value '%s' exceeds size %s" % (hex(value), size))
                 self.sub_word = ABI.serialize_uint(value, size)
             elif isinstance(value, str) and value[:2] == '0x':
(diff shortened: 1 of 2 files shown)


manticore/ethereum.py
```diff
@@ -287,21 +287,100 @@ def __mul__(self, reps):
...
+        elif isinstance(value, ABI.Address):
+            result = value.serialize()
+        elif isinstance(value, ABI.ExternalFunction):
+            result = value.serialize()
+        elif isinstance(value, ABI.SByte):
+            result = ABI.serialize_uint(value.size) + (None,) * value.size + (('\x00',) * (32 - (value.size % 32)))
+        elif value is None:
+            result = (None,) * 32
+        else:
+            raise TypeError
```
mossberg: let's add a descriptive error message here expressing that the type of value is not supported

manticore/ethereum.py
```diff
@@ -287,21 +287,100 @@ def __mul__(self, reps):
...
+            else:
+                raise TypeError('Expected int, long, str, or tuple')
+
+        def size(self):
+            return len(self.sub_word)
+
+        def serialize(self):
+            return self.sub_word + tuple('\0' * (32 - len(self.sub_word)))
+
+    class Address(object):
```
mossberg: it seems this class is technically unnecessary b/c address is identical to uint160, as you noted in slack. in the name of minimizing the size of our codebase, i think i'd like to omit this class unless you feel strongly? i do understand it is used below in ExternalFunction, but it seems the main usage is in the type check, which i think could reasonably be left out

manticore/ethereum.py
```diff
@@ -287,21 +287,100 @@ def __mul__(self, reps):
...
 
+    @staticmethod
+    def _tuple_bytes_to_long(tpl):
+        return reduce(lambda acc, b: (acc << 8) + ord(b), tpl, 0L)
+
+    class StaticBytes(object):
+        def __init__(self, value, size=32):
+            if not 0 < size <= 32:
+                raise ValueError('bytes<M>: binary type of M bytes, 0 < M <= 32.')
+            if isinstance(value, (int, long)):
```
mossberg: do you think it's necessary for the ctor to be so "accommodating" and accept so many different types? if this not required, i think i'd prefer to choose one input type/format. 

alternatively, instead of creating a new class for StaticBytes, could we reuse the builtin `bytearray` type and check for that in the big type check in ABI.serialize?

manticore/ethereum.py
```diff
@@ -287,21 +287,100 @@ def __mul__(self, reps):
     SUINT = SByte(32)
     SValue = None
 
+    @staticmethod
+    def _tuple_bytes_to_long(tpl):
+        return reduce(lambda acc, b: (acc << 8) + ord(b), tpl, 0L)
```
disconnect3d: Py3 requires `from functools import reduce` and `0L` -> `0`

manticore/ethereum.py
```diff
@@ -287,21 +287,100 @@ def __mul__(self, reps):
...
-        if isinstance(value, (list)):
-            return ABI.serialize_array(value)
-        if isinstance(value, (int, long)):
-            return ABI.serialize_uint(value)
-        if isinstance(value, ABI.SByte):
-            return ABI.serialize_uint(value.size) + (None,) * value.size + (('\x00',) * (32 - (value.size % 32)))
-        if value is None:
-            return (None,) * 32
+            result = ABI.serialize_string(value)
+        elif isinstance(value, (list)):
```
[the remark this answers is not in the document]
dwhjames: fixed in https://github.com/trailofbits/manticore/pull/875/commits/63efc4fb384bdd36f3e54d614b1c5e5625098f14

manticore/ethereum.py
dwhjames: fixed in https://github.com/trailofbits/manticore/pull/875/commits/63efc4fb384bdd36f3e54d614b1c5e5625098f14

manticore/ethereum.py
dwhjames: fixed in https://github.com/trailofbits/manticore/pull/875/commits/63efc4fb384bdd36f3e54d614b1c5e5625098f14

dwhjames:
@mossberg, inlining your comments from

- https://github.com/trailofbits/manticore/pull/875#discussion_r183689389

> it seems this class is technically unnecessary b/c address is identical to uint160, as you noted in slack. in the name of minimizing the size of our codebase, i think i'd like to omit this class unless you feel strongly? i do understand it is used below in ExternalFunction, but it seems the main usage is in the type check, which i think could reasonably be left out

- https://github.com/trailofbits/manticore/pull/875#discussion_r183691397

> do you think it's necessary for the ctor to be so "accommodating" and accept so many different types? if this not required, i think i'd prefer to choose one input type/format.
>
> alternatively, instead of creating a new class for StaticBytes, could we reuse the builtin bytearray type and check for that in the big type check in ABI.serialize?

---

And extracting a conversation on May 1st 2018 with @ggrieco-tob from Empire Hacking _#manticore_ channel.

@ggrieco-tob was helping someone else with understanding an error with symbolic values and dynamically sized types.

---

**gustavo.grieco**
(https://github.com/trailofbits/manticore/tree/dev-evm-dynamicarguments)

**dwhjames**
**gustavo.grieco**, looking at the impl. of `ABI.make_function_arguments` it doesn’t look like serialization of something like `myFunc(address[2])` is handled correctly
or at least to make it do the right thing, one would have to call it as though it’s actually `myFunc(address,address)`

**gustavo.grieco**
Uhm.

**dwhjames**
and on the other side array support is also partial in `ABI._consume_type`, i.e., only `address[]` is handled?
the tests in https://github.com/trailofbits/manticore/pull/715
are only for `address[]`

**gustavo.grieco**
I created an issue to request the implementation of dynamic arguments and provide an small example contract using `address[]` as example. But the implementation should be easily extended to other types of arrays.
Tomorrow I check the serialization of `address[2]` and submit a bug report if necessary.

**dwhjames**
```python
from manticore.ethereum import ABI
spec1 = 'myFunc(uint256[2])'
calldata1 = ABI.make_function_call(spec1, [42, 1337])
ABI.parse(spec1, calldata1)
```

1. `calldata1` has a dynamic encoding of the array, not a static one
2. `ABI.parse` raises an error because it doesn’t know about arrays for `uint256` and tries to convert `256[2]` to an `int`

```python
from manticore.ethereum import ABI
spec2 = 'myFunc(address[2])'
calldata2 = ABI.make_function_call(spec1, 0x1, 0x2)
ABI.parse(spec2, calldata2)
```

1. by exploding the two-array and treating it as a two-param function `calldata2` is correct, according to the ABI spec
2. but `ABI.parse` raises a `NotImplementedError`
`ABI.make_function_call` has the ABI spec, but it delegates to `ABI.make_function_arguments`, which does not consult the spec, and instead dispatches off of the Python types, so in the current design there is no way to handle both static and dynamic arrays
`ABI._consume_type` is simply missing the cases to the handle any array type other than a dynamic array of addresses

---

So, with all that context, I thought I’d respond to your questions in the scope of `ABI.make_function_call` and `ABI.make_function_arguments`—the primary way I believe that most Manticore users would be interacting with the behavior of `ABI.serialize`.

Right now, the subset of Solidity types supported are approximated with Python types. `ABI.serialize` certainly not a surjective function, as not all Solidity types are supported, but it’s not really even a function in the mathematical sense, as a Python long is mapped into all sorts of types.

We can get away with long for (u)int8–256, address, and bytes32. And if one does the required left-shift, then a long will even suffice for types bytes1–31.

Static arrays, as I pointed out on the Slack channel run into a brick wall with the inference-by-Python-type approach. And the `function` type as another sort of compound type is also problematic.

So, my thoughts thus far:

1. I think there needs to be a `ABI.make_function_arguments_with_spec(spec, *args)` that `ABI.make_function_call` uses.
2. Having a Python long overloaded for all sorts of uses is rather convenient and succinct, but that approximation also causes problems when one wants precise control. That’s why I started introducing nested classes to enumerate each Solidity type (`StaticBytes`, `Address`, `ExternalFunction`, etc.). I can see how this becomes un-Pythonic, as my understanding of that philosophy is that there should be a canonical way of doing everything. And I may have gone overboard with the convenience constructors that would be accepting of non-canonical representations. Nonetheless, I think Manticore ultimately needs a representation of all Solidity types as new nominal types, rather than overloading built-in Python types.

mossberg:
@dwhjames thank you for the very clear post with all that background context! :) 

> I think there needs to be a ABI.make_function_arguments_with_spec(spec, *args) that ABI.make_function_call uses

agreed! i think that would be a good solution for allowing the user to more precisely control the solidity types that are interpreted. i in general would prefer this kind of solution to creating many new classes corresponding to Solidity types which feels a little heavy to me. i think that as long as it's adequately documented what python types correspond to to solidity types, re-using built in types would not be a _terrible_ solution. for the Solidity types that don't have an obvious python type (ExternalFunction), I think it's fine to introduce a new type, but I think we should strive to minimize the classes that we add.

mossberg:
hi @dwhjames, I wanted to revisit this. We recently merged a large refactor of the ABI implementation (https://github.com/trailofbits/manticore/pull/926), which notably features an `ABI.serialize` interface that takes a type spec, which can be used to more correctly serialize arguments without having to guess purely based on the python type passed in.

It does not currently support fixed sized bytes well (https://github.com/trailofbits/manticore/issues/952), which I understand was one of the main points of this PR, but it should not be a hard fix (we would also love help with this 😅 !)

It should support serializing external functions via `ABI.serialize('function', (address, selector))` (e.g `ABI.serialize('function', (0xfB6916095ca1df60bB79Ce92cE3Ea74c37c5d359, '\xaa\xbb\xcc\xdd'))`).

I think this should largely address the issues you were dealing with in a relatively clean way :) Do you have any other thoughts?

Once we fix #952 I think we should be able to close this PR.

PR #878, opened 2018-04-23
Add LD_LIBRARY_PATH type feature

### LINKED ISSUE #93: Add LD_LIBRARY_PATH type feature
Something that ties us to Linux right now is fetching libc and the dynamic linker, for example, directly from the filesystem on the machine. To allow mcore to run on more platforms like OS X, add a way to specify directories for mcore to look in for things like ld-linux.so, so you could analyze a dynamically linked ELF on OS X.

catenacyber:
Use of environment variable MANTICORE_LD_PATH
instead of LD_LIBRARY_PATH which is protected on MacOS

Fixes #93

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/878)

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -830,7 +830,7 @@ class Linux(Platform):
                 interpreter = ELFFile(open(interpreter_filename))
             elif 'MANTICORE_LD_PATH' in os.environ:
                 for mpath in os.environ['MANTICORE_LD_PATH'].split(":"):
-                    interpreter_path_filename = os.path.join(mpath,os.path.basename(interpreter_filename))
+                    interpreter_path_filename = os.path.join(mpath, os.path.basename(interpreter_filename))
                     logger.info("looking for interpreter %s", interpreter_path_filename)
                     if os.path.exists(interpreter_filename):
                         interpreter = ELFFile(open(interpreter_path_filename))
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -826,7 +826,14 @@ class Linux(Platform):
                 continue
             interpreter_filename = elf_segment.data()[:-1]
             logger.info('Interpreter filename: %s', interpreter_filename)
-            interpreter = ELFFile(file(interpreter_filename))
+            if os.path.exists(interpreter_filename):
+                interpreter = ELFFile(file(interpreter_filename))
+            elif 'MANTICORE_LD_PATH' in os.environ:
+                for mpath in os.environ['MANTICORE_LD_PATH'].split(":"):
+                    logger.info("looking for interpreter %s", mpath+'/'+os.path.basename(interpreter_filename))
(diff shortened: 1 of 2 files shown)


manticore/platforms/linux.py
```diff
@@ -826,7 +826,14 @@ def load(self, filename):
                 continue
             interpreter_filename = elf_segment.data()[:-1]
             logger.info('Interpreter filename: %s', interpreter_filename)
-            interpreter = ELFFile(file(interpreter_filename))
+            if os.path.exists(interpreter_filename):
+                interpreter = ELFFile(file(interpreter_filename))
+            elif 'MANTICORE_LD_PATH' in os.environ:
+                for mpath in os.environ['MANTICORE_LD_PATH'].split(":"):
+                    logger.info("looking for interpreter %s", mpath+'/'+os.path.basename(interpreter_filename))
```
disconnect3d: Use `os.path.join` to concatenate paths

disconnect3d: Also use a variable for the interpreter path.

catenacyber:
Thanks @disconnect3d 
Here come the changes

yan:
Thanks again @catenacyber. I am thinking of what the best way to incorporate changes from this. I have a few concerns:

1. We don't use environment variables to pass other options to Manticore; this would be slightly bucking the trend.
2. In this case, `MANTICORE_LD_PATH` only controls the ELF interpreter loading path, not ld path in general. We can extend it to cover lib load path as well, that'd just require setting `LD_LIBRARY_PATH`.

catenacyber:
> We don't use environment variables to pass other options to Manticore; this would be slightly bucking the trend.

Ok, so I need to convert MANTICORE_LD_PATH  environment variable to argument ld_path.
Or is there another way ? such as config file...

> We can extend it to cover lib load path as well, that'd just require setting LD_LIBRARY_PATH

@yan I am not sure what you mean.
On MacOS, `LD_LIBRARY_PATH` is protected by so-called system integrity protection : you cannot set it.

yan:
> On MacOS, LD_LIBRARY_PATH is protected by so-called system integrity protection : you cannot set it.

What I meant was setting `MANTICORE_LD_PATH` won't actually affect the dynamic linker, so it's a misnomer. Further, macOS is not currently supported, we only support Linux guests.

What I meant above is, currently `MANTICORE_LD_PATH` only controls the ELF interpreter path; we should either name it as such (although I still believe a command line argument is potentially more appropriate), or forward along the library path to the emulated binary so that works as expected. Does that make sense?

catenacyber:
> Further, macOS is not currently supported, we only support Linux guests.

The point of this pull request is to progress towards macOS support

> either name it as such
> or forward along the library path to the emulated binary so that works as expected

The latter looks the best option to me.
What do you think ?

catenacyber:
thanks @yan you were faster than me

PR #882, opened 2018-04-24
Dockerfile updates

lazzarello:
* Add solc binary to Dockerfile
* Run container as root
* Document interactive usage of container

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/882)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,9 @@ RUN apt-get install -y build-essential software-properties-common && \
     apt-get update && \
     apt-get install -y solc ethereum
 
-WORKDIR /root
-ENV HOME /root
+RUN useradd -m manticore
+USER manticore
+WORKDIR /home/manticore
+ENV HOME /home/manticore
 
 CMD ["/bin/bash"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,9 +7,12 @@ RUN apt-get -y update && \
 	cd manticore && \
 	pip install .
 
-RUN useradd -m manticore
-USER manticore
-WORKDIR /home/manticore
-ENV HOME /home/manticore
+RUN apt-get install -y build-essential software-properties-common && \
+    add-apt-repository -y ppa:ethereum/ethereum && \
+    apt-get update && \
+    apt-get install -y solc ethereum
+
+WORKDIR /root
+ENV HOME /root
 
 CMD ["/bin/bash"]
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -119,6 +119,33 @@ def hook(state):
...
+`docker build -t manticore .`
+
+From there it can be run interactively with examples mounted with the following
+
+``docker run -it -v `pwd`/examples:/root/examples manticore``
+
+This drops you into a shell in the container. The following is a run of the Getting Started example.
+
+```
+2blockchains:manticore leeazzarello$ docker run -it -v `pwd`/examples:/root/examples manticore
```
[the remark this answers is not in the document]
lazzarello: sure...copy pasta...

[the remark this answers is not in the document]
lazzarello: pushed

mossberg:
thanks for the PR, looking at it now (sorry for taking so long!). thanks also for the docker instructions; i don't have a ton of docker experience so they're super helpful :)

Dockerfile
```diff
@@ -7,9 +7,12 @@ RUN apt-get -y update && \
...
-RUN useradd -m manticore
-USER manticore
-WORKDIR /home/manticore
-ENV HOME /home/manticore
+RUN apt-get install -y build-essential software-properties-common && \
+    add-apt-repository -y ppa:ethereum/ethereum && \
+    apt-get update && \
+    apt-get install -y solc ethereum
+
+WORKDIR /root
```
mossberg: any particular reason to change from the manticore user to the root user?

mossberg:
I just added a commit with some superficial, stylistic README changes. Good to merge once we figure out the root/manticore user thing :)

Dockerfile
```diff
@@ -7,9 +7,12 @@ RUN apt-get -y update && \
...
 	pip install .
 
-RUN useradd -m manticore
-USER manticore
-WORKDIR /home/manticore
-ENV HOME /home/manticore
+RUN apt-get install -y build-essential software-properties-common && \
+    add-apt-repository -y ppa:ethereum/ethereum && \
+    apt-get update && \
+    apt-get install -y solc ethereum
```
mossberg: what does installing the `ethereum` package get us?

mossberg: i noticed that it does add some time to the docker build stage, but then i realized docker hub will generally be doing that, not our users directly, so i don't think it's a big deal, nvm.

mossberg:
I ended up reverting the changes related to root user, and reverted back to the manticore user, but we can have future discussion for a later PR if need be :)

Issue #880, opened 2018-04-23
inline function definition in potentially hot code


mossberg:
we define an inline function in FileMap.__getitem__ which might potentially get called often, we might want to move this into the class to help with performance

https://github.com/trailofbits/manticore/blob/4a98110f4a74916305231a0a2fabc076e6218462/manticore/core/memory.py#L315

mossberg:
same thing with Z3Solver._recv.readline

PR #883, opened 2018-04-24
Moved code outside from inline function get_byte_at_offset() and removed function get_byte_by_offset() - Fixes Issue #880

arunjohnkuruvilla:


mossberg:
hi @arunjohnkuruvilla thanks for this! 

so one thing i'm curious with this Pr is whether it actually improves performance or not; i was actually sure when i filed the original issue. would you be able to run some simple experiments and report some rough data on how long manticore takes with and without the pr?

also in the future, could you include the "fixes #880" type text in the pr description? thanks :)

arunjohnkuruvilla:
hi @mossberg 

Thank you for your suggestions. I will make the changes in future prs.

I will try my best in testing on the performance difference. However, I am not sure if I will be able to test it accurately. Do you have any suggestions on any particular binaries that I can use that might give an accurate result?

mossberg:
To start let's try something like simpleassert in the examples directory,
I'm a fan of that one :)

manticore/core/memory.py
```diff
@@ -312,22 +312,26 @@ def __setitem__(self, index, value):
             self._overlay[index] = value
 
     def __getitem__(self, index):
-        def get_byte_at_offset(offset):
```
mossberg: i think we probably should still keep this as a function, rather than copy/pasting the code. rather than being an inline function , it can be a private method.

PR #885, opened 2018-04-30
Added optional message parameter in ConcretizeMemory

ggrieco-tob:
In some files (e.g. manticore/core/cpu/abstractcpu.py), ConcretizeMemory is invoked with a message but it is interpreted as the "policy" value now, causing a crash. This commit should fix that issue.

### DIFF
--- a/manticore/core/memory.py
+++ b/manticore/core/memory.py
@@ -34,8 +34,10 @@ class ConcretizeMemory(MemoryException):
     Raised when a symbolic memory cell needs to be concretized.
     '''
 
-    def __init__(self, mem, address, size, policy='MINMAX'):
-        self.message = "Concretizing memory address {} size {}".format(address, size)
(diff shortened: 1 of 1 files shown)


disconnect3d:
Example of the calls: https://github.com/trailofbits/manticore/blob/master/manticore/core/cpu/abstractcpu.py#L360-L364

Btw shouldn't we pass that by keyword argument there? I mean `message=msg`.

manticore/core/memory.py
```diff
@@ -34,8 +34,10 @@ class ConcretizeMemory(MemoryException):
     Raised when a symbolic memory cell needs to be concretized.
     '''
 
-    def __init__(self, mem, address, size, policy='MINMAX'):
-        self.message = "Concretizing memory address {} size {}".format(address, size)
+    def __init__(self, mem, address, size, message=None, policy='MINMAX'):
+        self.message = message
```
yan: move this as the `else` branch below?

PR #886, opened 2018-05-03
Typo in ManticoreEVM#last_return

dwhjames:
Encountered the following trace:

```
  …, in <module>
    m.last_return()
  File "…/manticore/ethereum.py", line 906, in last_return
    return state.platform.last_return
AttributeError: 'EVMWorld' object has no attribute 'last_return'
```

It appears this bug was introduced in commit ed29a22f

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/886)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -903,7 +903,7 @@ class ManticoreEVM(Manticore):
     def last_return(self, state_id=None):
         ''' Last returned buffer for state `state_id` '''
         state = self.load(state_id)
-        return state.platform.last_return
+        return state.platform.last_return_data
 
     def transactions(self, state_id=None):
         ''' Transactions list for state `state_id` '''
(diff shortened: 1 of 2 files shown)


dwhjames:
`ManticoreEVM#last_return` attempts to call `EVMWorld#last_return`
https://github.com/trailofbits/manticore/blob/9307475af5f52d641e97b8208ac53ea4812e5940/manticore/ethereum.py#L903-L906
but the property is actually
https://github.com/trailofbits/manticore/blob/9307475af5f52d641e97b8208ac53ea4812e5940/manticore/platforms/evm.py#L2049-L2051

yan:
@dwhjames thanks for this pr! It looks like a similar issue exists in [evm.py](https://github.com/trailofbits/manticore/blob/master/manticore/platforms/evm.py#L2243). Update that too and we'll merge?

dwhjames:
So there is, @yan!

I’ve updated that and squashed both changes into a new commit https://github.com/trailofbits/manticore/pull/886/commits/8752b48f4fbf638ce5d9aa235d491c1fe3b03e1a

PR #887, opened 2018-05-03
Update concolic.py

yan:
Previous `concolic.py` depended on an out-of-bounds provided address of the last instruction in `main`; this removes that dependency. We subscribe to terminate state event and capture context there. This also makes trace collection a bit more resilient.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/887)

### DIFF
--- a/examples/script/concolic.py
+++ b/examples/script/concolic.py
@@ -28,7 +28,6 @@ import copy
 from manticore.core.smtlib.expression import *
 
 prog = '../linux/simpleassert'
-main_end = 0x400ae9
 VERBOSITY = 0
 
 def _partition(pred, iterable):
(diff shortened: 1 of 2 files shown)


examples/script/concolic.py
```diff
@@ -164,15 +163,16 @@ def symbolic_run_get_cons(trace):
...
             readdata = []
-            for name, fd, data in s.platform.syscall_trace:
+            for name, fd, data in state.platform.syscall_trace:
                 if name in ('_receive', '_read') and fd == 0:
                     readdata.append(data)
             ctx['readdata'] = readdata
-            ctx['constraints'] = list(s.constraints.constraints)
+            ctx['constraints'] = list(state.constraints.constraints)
+
+    m2.subscribe('will_generate_testcase', on_gen_testcase)
```
mossberg: is will_generate_testcase the best event? maybe some kind of terminate_state event would be better?

PR #889, opened 2018-05-08
Rollback to old mini assembler only example

feliam:
Refactor report generator to handle bytecode only contracts (no metadata) again. RFC/WIP

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/889)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -1422,41 +1422,6 @@ class ManticoreEVM(Manticore):
         Terminate and generate testcases for all currently alive states (contract states that cleanly executed
         to a STOP or RETURN in the last symbolic transaction).
         """
-        # move running states to final states list
-        # and generate a testcase for each
-        q = Queue()
-        map(q.put, self.running_state_ids)
-
-        def f(q):
-            try:
-                while True:
-                    state_id = q.get_nowait()
-                    self.terminate_state_id(state_id)
-            except EmptyQueue:
-                pass
-
-        ps = []
-
-        for _ in range(self._config_procs):
-            p = Process(target=f, args=(q,))
-            p.start()
-            ps.append(p)
-
-        for p in ps:
-            p.join()
-
-        # delete actual streams from storage
-        for state_id in self.all_state_ids:
-            # state_id -1 is always only on memory
-            if state_id != -1:
-                self._executor._workspace.rm_state(state_id)
-
-        # clean up lists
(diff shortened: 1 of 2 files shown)


manticore/ethereum.py
```diff
@@ -1423,41 +1423,6 @@ def finalize(self):
         to a STOP or RETURN in the last symbolic transaction).
         """
 
-        # move runnign states to final states list
```
mossberg: what was this section moved to the bottom?

feliam: Ugh...
The problem of not having source/metadata was at the reporting of "global_XXXX" stuff. Reordering this just made it fail faster for debugging.

manticore/ethereum.py
```diff
@@ -1549,13 +1549,13 @@ def global_coverage(self, account_address):
...
             world = self.get_world(state_id)
-            if account_address in world.storage:
+            if account_address in world.contract_accounts:
                 break
 
         with self.locked_context('runtime_coverage') as coverage:
             seen = coverage
-        #runtime_bytecode = world.storage[account_address]['code']
-        runtime_bytecode = self.get_metadata(account_address).runtime_bytecode
+        runtime_bytecode = world.get_code(account_address)[:-44]
```
mossberg: let's not use the magic number -44 here, either add a comment or a variable or something?

manticore/ethereum.py
```diff
@@ -1549,13 +1548,14 @@ def global_coverage(self, account_address):
...
-            if account_address in world.storage:
+            if account_address in world.contract_accounts:
                 break
 
         with self.locked_context('runtime_coverage') as coverage:
             seen = coverage
-        #runtime_bytecode = world.storage[account_address]['code']
-        runtime_bytecode = self.get_metadata(account_address).runtime_bytecode
+
+        #fixme use this https://github.com/trailofbits/manticore/blob/a98902c81adfa59a86b35e116869609ad1774cdc/manticore/ethereum.py#L173-L179
```
mossberg: wait, SolidityMetadata._without_metadata exists here though!

manticore/ethereum.py
feliam: Update: This section is moved to the bottom so there is at least one state alive from where we can take the runtime bytecode (then assume that is the same in every state) and calculate the coverage over it.

PR #891, opened 2018-05-08
Partially implement chroot(2)

khorben:
Really this return EPERM (permission denied), which seems to be absolutely fine
since a non-privileged user is currently assumed. This is what would normally
be returned in this scenario.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/891)


yan:
@khorben thanks for this pr! I slightly updated it to include some error conditions for which you don't actually need to have a working `fork` for and moved the method out of `SLinux`/

PR #892, opened 2018-05-08
Partially implement fork(2)

khorben:
Really this return ENOSYS (not implemented), which allows sub-processes to
detect the condition and continue to run (or crash on their own if they did not
expect fork() to fail).

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/892)


yan:
Thanks @khorben!

PR #893, opened 2018-05-09
Add support for NetBSD as a host OS

khorben:
With these changes, I am able to run Manticore on Linux (static)
binaries from a NetBSD host.

Tested with NetBSD/amd64 (7.1_STABLE XEN3_DOM0).

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/893)

### DIFF
--- a/manticore/utils/mappings.py
+++ b/manticore/utils/mappings.py
@@ -43,5 +43,38 @@ if osname == "darwin" or osname.startswith("linux"):
     mmap = _mmap
     munmap = _munmap
 
+elif osname.startswith("netbsd"):
+
+    libc = ctypes.cdll.LoadLibrary("libc.so.12")
+
+    # void* mmap(void* addr, size_t len, int prot, int flags, int fd, off_t offset)
+    mmap_function = libc.mmap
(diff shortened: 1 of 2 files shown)


khorben:
The only change in comparison to macOS and Linux hosts is the filename for the libc. It should be easy to add support for more Operating Systems as hosts (e.g. FreeBSD, OpenBSD...) It should also be possible to come up with simpler way to cater for their respective differences; I will think about it.

manticore/utils/mappings.py
```diff
@@ -43,5 +43,38 @@ def _mmap(fd, offset, size):
     mmap = _mmap
     munmap = _munmap
 
+elif osname.startswith("netbsd"):
+
+    libc = ctypes.cdll.LoadLibrary("libc.so.12")
```
krytarowski: why not `libc.so`?

krytarowski: On some OSes (GNU/Linux?) we require `libc.so.$VERSION`, on NetBSD we prefer `libc.so`

PR #897, opened 2018-05-21
Inode number is encoded with 64 bits

### LINKED ISSUE #824: m.c.executor:ERROR: Exception: an integer is required
### OS / Environment
Ubuntu 16.04.4 LTS

### Manticore version
0.1.7

### Python version
2.7.12

### Dependencies
capstone==3.0.5rc2
ply==3.11
pyelftools==0.24
pysha3==1.0.2
unicorn==1.0.1
z3-solver==4.5.1.0.post2

### Summary of the problem
Exception in the title is raised during symbolic solve.

Run this script with the [attached binary](https://github.com/trailofbits/manticore/files/1822688/activate.zip) (a problem from angstrom CTF):
```
#!/usr/bin/env python2

m = Manticore.linux('./activate', concrete_start="Artemis Tosini\nartemis.tosini@example.com\nABCD-EFGH-IIJK-LMNO-PQRS-TUVW\n")
m.context['key'] = ""

print("hello")
@m.hook(0x0040112a) # "Windows has been activated" in main
def hook(state):
    print("in hook")
    cpu = state.cpu
    transform_base = cpu.RBP - 0x30 # the product key
    print("Email: ")
    print(cpu.RBP - 0x60)
    print("Name: ")
    print(cpu.RBP - 0x90)
    for i in range(29): # 6*4 + 5 dashes
        solved = state.solve_one(cpu.read_int(transform_base + i, 8))
        print(solved)
        m.context['key'] += chr(solved)
    print(m.context['key'])
    m.terminate()

m.run()
```

### Expected behavior
No exception, logging from the hook function is printed.

### Actual behavior
"hello" is printed then every few minutes this exception is raised:
```
[13694] m.c.executor:ERROR: Exception: an integer is required
  File "/usr/local/lib/python2.7/dist-packages/manticore/core/executor.py", line 463, in run
  File "/usr/local/lib/python2.7/dist-packages/manticore/core/state.py", line 126, in execute
  File "/usr/local/lib/python2.7/dist-packages/manticore/platforms/linux.py", line 2133, in execute
    self.syscall()
  File "/usr/local/lib/python2.7/dist-packages/manticore/platforms/linux.py", line 1978, in syscall
  File "/usr/local/lib/python2.7/dist-packages/manticore/core/cpu/abstractcpu.py", line 384, in invoke
    ret = super(SyscallAbi, self).invoke(model, prefix_args)
  File "/usr/local/lib/python2.7/dist-packages/manticore/core/cpu/abstractcpu.py", line 337, in invoke
    result = model(*arguments)
  File "/usr/local/lib/python2.7/dist-packages/manticore/platforms/linux.py", line 1165, in sys_lseek
    signed_offset = self._to_signed_dword(offset)
  File "/usr/local/lib/python2.7/dist-packages/manticore/platforms/linux.py", line 1024, in _to_signed_dword
    sdword = ctypes.c_int64(dword).value
TypeError: an integer is required
```
None of the other print statements are hit.

catenacyber:
Fixes #824 

Inode number from stat is either encoded on 32 or 64 bits
Since this is function `sys_fstat64`, I suggest that we use 64 bits.

Do you see any other way to get `sizeof(ino_t)` ?

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/897)

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -2337,22 +2337,22 @@ class Linux(Platform):
             return struct.pack('<LL', int(ts), int(ts % 1 * 1e9))
 
         bufstat = add(8, stat.st_dev)        # unsigned long long      st_dev;
-        bufstat += add(4, 0)                  # unsigned char   __pad0[4];
-        bufstat += add(8, stat.st_ino)        # unsigned long   __st_ino;
+        bufstat += add(8, stat.st_ino)        # unsigned long long   __st_ino;
         bufstat += add(4, stat.st_mode)       # unsigned int    st_mode;
         bufstat += add(4, stat.st_nlink)      # unsigned int    st_nlink;
         bufstat += add(4, stat.st_uid)        # unsigned long   st_uid;
         bufstat += add(4, stat.st_gid)        # unsigned long   st_gid;
         bufstat += add(8, stat.st_rdev)       # unsigned long long st_rdev;
-        bufstat += add(4, 0)                  # unsigned char   __pad3[4];
-        bufstat += add(4, 0)                  # unsigned char   __pad3[4];
+        bufstat += add(8, 0)                  # unsigned long long __pad1;
         bufstat += add(8, stat.st_size)       # long long       st_size;
-        bufstat += add(8, stat.st_blksize)    # unsigned long   st_blksize;
+        bufstat += add(4, stat.st_blksize)    # int   st_blksize;
+        bufstat += add(4, 0)                  # int   __pad2;
         bufstat += add(8, stat.st_blocks)     # unsigned long long st_blocks;
         bufstat += to_timespec(stat.st_atime)  # unsigned long   st_atime;
         bufstat += to_timespec(stat.st_mtime)  # unsigned long   st_mtime;
         bufstat += to_timespec(stat.st_ctime)  # unsigned long   st_ctime;
-        bufstat += add(8, stat.st_ino)        # unsigned long long      st_ino;
+        bufstat += add(4, 0)                   # unsigned int __unused4;
+        bufstat += add(4, 0)                   # unsigned int __unused5;
 
         self.current.write_bytes(buf, bufstat)
         return 0
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -2338,7 +2338,7 @@ class Linux(Platform):
 
         bufstat = add(8, stat.st_dev)        # unsigned long long      st_dev;
         bufstat += add(4, 0)                  # unsigned char   __pad0[4];
-        bufstat += add(4, stat.st_ino)        # unsigned long   __st_ino;
+        bufstat += add(8, stat.st_ino)        # unsigned long   __st_ino;
         bufstat += add(4, stat.st_mode)       # unsigned int    st_mode;
         bufstat += add(4, stat.st_nlink)      # unsigned int    st_nlink;
         bufstat += add(4, stat.st_uid)        # unsigned long   st_uid;
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -868,7 +868,7 @@ class Linux(Platform):
                 for mpath in env['LD_LIBRARY_PATH'].split(":"):
                     interpreter_path_filename = os.path.join(mpath, os.path.basename(interpreter_filename))
                     logger.info("looking for interpreter %s", interpreter_path_filename)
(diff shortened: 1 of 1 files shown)


yan:
@catenacyber thanks again for your contributions! 

Can you refer to the stat64 declaration you used to get at that field size? From what I recall, the sizes in that syscall implementation were pulled from kernel headers. I tried running your changes with the binary provided in #824, and it didn't seem to have fixed it. Are you seeing different results?

catenacyber:
Here is a ref 
https://www.tcm.phy.cam.ac.uk/sw/inodes64.html

Different kernels have different sizes for inodes. Maybe there is a better fix than the one proposed.

@yan yes I reproduced the bug and noticed that it was because function add
```
        def add(width, val):
            fformat = {2: 'H', 4: 'L', 8: 'Q'}[width]
            return struct.pack('<' + fformat, val)
```
was called with a width of 4 with a 64 bit integer

This change fixed it and I no longer had the exception

catenacyber:
Here is the traceback I get before my patch (commit b5d9d1da278ed281bbda7e8c0dac2c7010fc5e57)
```
2018-05-25 08:34:56,900: [3853] m.c.executor:ERROR: Exception: 'L' format requires 0 <= number <= 4294967295
  File "manticore/manticore/core/executor.py", line 459, in run
  File "manticore/manticore/core/state.py", line 126, in execute
  File "manticore/manticore/platforms/linux.py", line 2219, in execute
    self.syscall()
  File "manticore/manticore/platforms/linux.py", line 2065, in syscall
  File "manticore/manticore/core/cpu/abstractcpu.py", line 400, in invoke
    ret = super(SyscallAbi, self).invoke(model, prefix_args)
  File "manticore/manticore/core/cpu/abstractcpu.py", line 351, in invoke
    result = model(*arguments)
  File "manticore/manticore/platforms/linux.py", line 2365, in sys_newstat
    return self.sys_stat64(fd, buf)
  File "manticore/manticore/platforms/linux.py", line 2375, in sys_stat64
    return self._stat(path, buf, True)
  File "manticore/manticore/platforms/linux.py", line 2383, in _stat
    ret = self.sys_fstat64(fd, buf)
  File "manticore/manticore/platforms/linux.py", line 2341, in sys_fstat64
    bufstat += add(4, stat.st_ino)        # unsigned long   __st_ino;
  File "manticore/manticore/platforms/linux.py", line 2334, in add
    return struct.pack('<' + fformat, val)
error: 'L' format requires 0 <= number <= 4294967295
```

And with the patch + dumb printing nodes with `print "inode", stat.st_ino`
```
> python inode.py 
hello
inode 8597146798
inode 8590579382
```

yan:
Ah I see what you meant now, it's Python's stat inode type. I believe the problem is dispatching to the correct fstat implementation from `sys_newstat`. Do you get correct results with your commit if you emulate a sample binary that calls `fstat()`?

catenacyber:
> I believe the problem is dispatching to the correct fstat implementation from sys_newstat

Maybe : `sys_newfstat` seems to allow different inode type lengths :
```
        # From linux/arch/x86/include/uapi/asm/stat.h
        # Numerous fields are native width-wide
        nw = self.current.address_bit_size / 8

        bufstat = add(nw, stat.st_dev)     # long st_dev
        bufstat += add(nw, stat.st_ino)     # long st_ino
```

But I do not see where it is called.
I see call to `sys_fstat64`(which uses uint32_t inode type)
```
        if is64bit:
            ret = self.sys_fstat64(fd, buf)
        else:
            ret = self.sys_fstat(fd, buf)
```

yan:
@catenacyber I had a chance to look at this PR again and it looks like this might not be the correct solution. For instance, if expanding inode width, that'd have influence on the padding immediate following it. The various stat implementations[0] should be consulted.

While also looking at this, I discovered a related issue of syscall invocation #940, which might have influenced this one.

[0] https://code.woboq.org/linux/linux/fs/stat.c.html

catenacyber:
@yan I pushed another commit, copying the `struct stat64` defined in
https://code.woboq.org/linux/linux/include/uapi/asm-generic/stat.h.html
and used in https://code.woboq.org/linux/linux/fs/stat.c.html
especially by 
`SYSCALL_DEFINE2(fstat64, unsigned long, fd, struct stat64 __user *, statbuf)`

From what I gather, there should also be a try catch surrounding `bufstat += add(4, stat.st_ino)`in `sys_fstat`which should return `EOVERFLOW`in case the inode number overflows the 32bit field

yan:
Thanks again!

PR #904, opened 2018-05-30
Incremental Py3: Switch exception syntax

reaperhulk:
Hi everyone! It looks like #784 might have gotten a bit bogged down due to the size of the change and general nervousness about how much it touches. I'd like to suggest that instead of doing it all at once we instead commit the safe changes in a few PRs and then we can focus review effort on the trickier conversion pieces. Accordingly, I'd like to propose a set of PRs:

* Convert the exception syntax. This is just changing `except Exception,e` to `except Exception as e` and not worrying about whether the code actually uses the `as e`. That's this PR!
* Strip the long literals (`38888L` becomes `38888`).
* Convert print statement to a function
* Convert to relative imports

These PRs will cover over 90% of the diff lines but represent no change in actual execution behavior. While we're doing this development can continue to ignore Python 3 (since we will have no testing for it and it would fail even if we did).

From there we can move into the more significant work (much of which is already covered in #784). An incomplete list:
* changes to metaclasses
* various stdlib import moves
* lazy iterators (e.g. map no longer returns a list and that means you can't index into it unless you exhaust the iterator first)
* Overriding `__eq__` causing objects to become unhashable
* Changes to hasattr
* Closure scope changes
* assertItemEqual -> assertCountEqual
* Change of division operator
* And, of course, the dreaded str -> bytes.

Once we get there we can decide if we want to handle this via `futurize` (the approach in #784) or tackle it via `six`. There are tradeoffs in each approach.

Another thing that will make this conversion process easier and help prevent regressions while the work is ongoing would be adoption of a linter/formatter. This isn't strictly necessary, but would make the code more consistent and allow automated gating to prevent various small regressions while we're working towards adding a py3 job. `flake8` is the standard here, but a true formatter like [black](https://github.com/ambv/black) might be a good choice if there's interest. (Quick note: A black run over the existing codebase suggests that if we want to go that path the commit for that will be...large. `118 files changed, 197019 insertions(+), 169227 deletions(-)`)

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/904)

### DIFF
--- a/scripts/gdb.py
+++ b/scripts/gdb.py
@@ -89,7 +89,7 @@ def stepi():
 def getM(m):
     try:
         return long(correspond('x/xg %s\n'%m).strip().split('\t')[-1], 0)
-    except Exception,e:
+    except Exception as e:
         raise e
         return 0
 def getPid():
--- a/scripts/stats.py
+++ b/scripts/stats.py
@@ -39,7 +39,7 @@ for filename in saved_states:
                 db[pc] = db.setdefault(pc,0) + 1
                 edges.setdefault(lastpc,[]).append(pc)
                 lastpc=pc
-    except Exception,e:
+    except Exception as e:
         print "#Failed to load saved state %s (%s)"%(filename,e)
 
 if args.pcfreq :
--- a/tests/EVM/test_EVMADD.py
+++ b/tests/EVM/test_EVMADD.py
@@ -16,7 +16,7 @@ class EVMTest_ADD(unittest.TestCase):
         last_exception = None
         try:
             new_vm.execute()
-        except evm.Stop, e:
+        except evm.Stop as e:
             last_exception = "STOP"
         except evm.NotEnoughGas:
             last_exception = "OOG"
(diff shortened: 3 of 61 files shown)


reaperhulk:
Current CI failure appears to be a networking issue inside Travis CI. Hopefully that'll clear up soon so we can see this pass 😁

dguido:
To follow up on the comment re: code formatters, I'm absolutely a fan of adopting one! I setup CodeClimate in anticipation that we would adopt one in the near future.

We'll want to update the [Hacking on Manticore](https://github.com/trailofbits/manticore/wiki/Hacking-on-Manticore) developer instructions to include setup for `flake8` or `black`, then modify the CodeClimate config to enforce it unless there is a better way to ensure we're all on the same page.

reaperhulk:
@dguido I'll take a look at the codeclimate configs to see how we can potentially ease ourselves into this. I like the opinionated nature of `black`, but since it is a true code formatter (in the vein of `go fmt`) to use it we'd need to reformat the entire codebase. That would touch nearly every line in the whole project and break every single PR. Incrementally ratcheting up to "flake8 with no warnings turned off" might be an easier path, at least initially.

PR #905, opened 2018-05-30
py3 incremental: remove long literals

reaperhulk:
Continuing the process from #904 

After this we can convert relative imports to absolute.

I'm probably missing a few small locations so expect a few (much smaller) PRs that clean up that stuff later 😄

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/905)


reaperhulk:
This build failure appears to be because the z3 solver is assumed to be present but never actually installed. Other builds work because it *is* present in the cache, but the cache was not successfully downloaded this time. I'll go ahead and add z3-solver to the travis scripts in this PR but I can move it to a separate PR once I confirm that fixes it.

PR #843, opened 2018-03-23
[WIP] Large EVM refactor

feliam:
WIP
It make storage and memory be raw smtlib arrays.
*All but one of the tests.zip now pass. 
*This enables to have unconstrained symbolic indexes. (previously we solved symbolic indexes eagerly)
*EVM internal api  evm_cpu vs evm_world improved
*EVM basic block to smtlib. Arbitrary reasoning over EVM traces.
*Initial symbolic gas support. Gas restricts memory access on a memory read over a free symbolic index. Example: With a gas budget of 99,  MREAD(free_symbolic_index) can not really read any memory offset but jjust the ones that it has enough gas for. 

Fixes #736

```
$ python asm_to_smtlib.py 
CODE:
	0 PUSH1 0x60
	2 PUSH1 0x40
	4 MSTORE
	5 PUSH1 0x0
	7 DUP1
	8 PUSH1 0x14
	10 PUSH2 0x100
	13 EXP
	14 DUP2
	15 SLOAD
	16 DUP2
	17 PUSH1 0xff
	19 MUL
	20 NOT
	21 AND
	22 SWAP1
	23 DUP4
	24 ISZERO
	25 ISZERO
	26 MUL
	27 OR
	28 SWAP1
	29 SSTORE
	30 POP
	31 CALLVALUE
	32 ISZERO
	33 PUSH2 0x29
	36 JUMPI
STORAGE = (store STORAGE #x0000000000000000000000000000000000000000000000000000000000000000 (bvand (select STORAGE #x0000000000000000000000000000000000000000000000000000000000000000) #xffffffffffffffffffffff00ffffffffffffffffffffffffffffffffffffffff))
[...]

PR #906, opened 2018-05-30
install z3-solver in our install script so it's available

reaperhulk:
The current test suite passes because at some point in the past the z3 solver was installed into the environment and the result was cached by travis. If, for whatever reason, travis fails to load the cache, tests fail since the current install scripts do not install z3.

This adds z3 to the installer to fix that issue. I would have added it to the actual dependencies of manticore itself, but since z3 is obtained via subprocess presumably we don't want to define a specific way that users have it installed and rather just want to document that it must be available on `PATH`.

Extracted from #905

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/906)

### DIFF
--- a/scripts/travis_install.sh
+++ b/scripts/travis_install.sh
@@ -8,5 +8,6 @@ function install_solc {
 install_solc
 
 pip install -U pip
+pip install -U z3-solver
 pip uninstall -y Manticore || echo "Manticore not cached"  # uninstall any old, cached Manticore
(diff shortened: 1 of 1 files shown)


mossberg:
Actually the reason z3-solver wasn't in the dependencies was purely due to a mistake in #843! Just for context, we generally have always had it in the dependencies so manticore can be a single pip install for our users. I submitted a fix in #923

PR #908, opened 2018-05-30
incremental py3: print statements, octals, and more

reaperhulk:
print statements, octal literals, and various small changes to support identical calling between py2 and py3. Plus a few missed exception syntax, etc. There will surely be more...

refs #904 

(This may not pass until it gets rebased against #907 but if it does then it's good to review/merge)

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/908)

### DIFF
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function
 ###############################################################################
 # Solver
 # A solver maintains a companion smtlib capable process connected via stdio.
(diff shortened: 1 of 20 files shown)


manticore/core/smtlib/solver.py
```diff
@@ -170,7 +170,7 @@ def _start_proc(self):
         try:
             self._proc = Popen(self._command.split(' '), stdin=PIPE, stdout=PIPE)
         except OSError as e:
-            print e, "Probably too  much cached expressions? visitors._cache..."
+            print(e, "Probably too  much cached expressions? visitors._cache...")
```
alex: Without a `from __future__ import print_function` this is a beahvior change. Applies to several files here.

reaperhulk: *shakes fist at automated conversion tool* I'll look at resolving this when I wake up.

reaperhulk: Should be resolved

PR #910, opened 2018-05-30
Dev evm INVALID detector

feliam:
* Some Detector improvements and rearrangements.
* Command line arguments for enabling detectors
```
  --detect-io           Enable integer overflow detection (Ethereum only)
  --detect-invalid      Enable INVALID instruction detection (Ethereum only)
  --detect-uninitialized-memory
                        Enable detection of uninitialized memory usage
                        (Ethereum only)
  --detect-uninitialized-storage
                        Enable detection of uninitialized storage usage
                        (Ethereum only)
  --detect-all          Enable all detector heuristics (Ethereum only)

```

* A simple Detector to report on INVALID instruction even in internal tx.
* Fix reporting so each state has a local Detector report. If a detector found something on a specific state it will now print this out to the test_000000.summary file. (Note there is global list of findings at global.summary too)

```
$evm-analyzer-benchmark-suite/benchmarks $ python -m manticore --detect-invalid assert_minimal.sol 
2018-05-30 19:16:54,188: [6022] m.main:INFO: Beginning analysis
2018-05-30 19:16:54,335: [6022] m.ethereum:INFO: Starting symbolic transaction: 1
2018-05-30 19:16:54,922: [6022] m.ethereum:WARNING: INVALID intruction
2018-05-30 19:16:55,258: [6022] m.ethereum:INFO: Finished symbolic transaction: 1 | Code Coverage: 89% | Terminated States: 4 | Alive States: 0
2018-05-30 19:16:55,499: [6093] m.ethereum:INFO: Generated testcase No. 0 - RETURN
2018-05-30 19:16:56,437: [6093] m.ethereum:INFO: Generated testcase No. 1 - THROW
2018-05-30 19:16:57,805: [6093] m.ethereum:INFO: Generated testcase No. 2 - REVERT
2018-05-30 19:16:59,324: [6093] m.ethereum:INFO: Generated testcase No. 3 - REVERT
2018-05-30 19:17:01,046: [6022] m.ethereum:INFO: Results in benchmarks/mcore_0D03sE
$evm-analyzer-benchmark-suite/benchmarks $ cat benchmarks/mcore_0D03sE/test_00000001.summary 
Message: 
Last exception: THROW

Findings:
INVALID intruction
	Contract: 0x67518339e369ab3d591d3569ab0a0d83b2ff5198
	EVM Program counter: 95
	Snippet:
		    6  contract Benchmark {
		    7      function run() public {
		    8          assert(false);
		    9      }
		    10  }
		

Last instruction at contract 67518339e369ab3d591d3569ab0a0d83b2ff5198 offset 5e
    6  contract Benchmark {
    7      function run() public {
    8          assert(false);
    9      }
    10  }

3 accounts.
Address: 0x67518339e369ab3d591d3569ab0a0d83b2ff5198 
Balance: 0 (*)
Storage: STORAGE_6
Code:
	606060405260043610603f576000357c01000000000000000000000000000000
	00000000000000000000000000900463ffffffff168063c0406226146044575b
	600080fd5b3415604e57600080fd5b60546056565b005b60001515605f57fe5b
	5600a165627a7a72305820a8651cc94cc6f988dee33272d6726010fd92092087
	da0e3a7a39847850ef7dac0029
Coverage 75% (on this state)

Address: 0xd30a286ec6737b8b2a6a7b5fbb5d75b895f62956 
Balance: 1000 
Storage: STORAGE_3

Address: 0x1bfa530d5d685155e98cd7d9dd23f7b6a801cfef 
Balance: 1000 (*)
Storage: STORAGE_4



(*) Example solution given. Value is symbolic and may take other values
```

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/910)


manticore/ethereum.py
```diff
@@ -356,24 +356,26 @@ def __build_source_map(self, bytecode, srcmap):
...
-            new_srcmap[asm_offset] = (s, l, f, j)
+            pos_to_offset[asm_pos] = asm_offset
             asm_pos += 1
             asm_offset += i.size
 
+        for asm_pos, md in enumerate(srcmap):
+            if len(md):
+                d = {}
+                for p, k in enumerate(md.split(':')):
+                    if len(k):
```
disconnect3d: `if k` would also work

manticore/ethereum.py
```diff
@@ -356,24 +356,26 @@ def __build_source_map(self, bytecode, srcmap):
...
+                        d[p] = k
+
+                s = int(d.get(0, s))
+                l = int(d.get(1, l))
+                f = int(d.get(2, f))
+                j = d.get(3, j)
```
disconnect3d: can we rename those to sth more explicit? ;/

feliam: Yes captain!

manticore/__main__.py
```diff
@@ -93,20 +111,40 @@ def positive(value):
...
+        filter_nohuman_constants = FilterFunctions(regexp=r".*", depth='human', mutability='constant', include=False)
+        self.register_plugin(filter_nohuman_constants)
 
     logger.info("Beginning analysis")
 
     m.multi_tx_analysis(args.argv[0], args.contract, args.txlimit, not args.txnocoverage, args.txaccount)
 
+    #TODO unregister all plugins
+
+    m.finalize()
```
disconnect3d: If there needs to be some 'destruction part' maybe we should implement `__enter__` and `__exit__` and use `with statement` e.g. `with Manticore(...) as m:`

tests/test_eth_benchmark.py
```diff
@@ -0,0 +1,65 @@
...
+    def _test_assert(self, name, should_find):
+        """
+        Tests DetectInvalid over the consensys benchmark suit
+        """
+        mevm = self.mevm
+
+        d = DetectInvalid()
+        mevm.register_detector(d)
+
+        filename = os.path.join(THIS_DIR, 'binaries', 'benchmark', '{}.sol'.format(name))
```
disconnect3d: An interesting approach to static paths like this: https://github.com/pwndbg/pwndbg/blob/dev/tests/binaries/__init__.py

Then it can be used as: https://github.com/pwndbg/pwndbg/blob/dev/tests/test_loads.py#L63-L64

manticore/__main__.py
feliam: It may be the case you do not want to call that .finalize().
Maybe the naming is off. That will go through all the states in the queues and generate a report for each and remove each from the queues. 
Sometimes you want to do that browsing by hand and report only on some states (as opposed to all of them).

But maybe... we could have the enter/exit AND also have something to clean the lists so at __exit__ it does not do anything on those special cases. Need to think about it.

manticore/__main__.py
```diff
@@ -93,20 +111,40 @@ def positive(value):
...
+    from .ethereum import ManticoreEVM, DetectInvalid, DetectIntegerOverflow, DetectUninitializedStorage, DetectUninitializedMemory, FilterFunctions
     log.init_logging()
 
     m = ManticoreEVM(procs=args.procs)
 
-    ################ Default? Detectors #######################
-    m.register_detector(IntegerOverflow())
-    m.register_detector(UninitializedStorage())
-    m.register_detector(UninitializedMemory())
+    if args.detect_all:
```
yan: Is setting `args` attributes an accepted way of doing this? instead of creating aliases or just checking for `detect_all` flags below?

manticore/__main__.py
```diff
@@ -93,20 +111,40 @@ def positive(value):
...
+    if args.detect_io:
+        m.register_detector(DetectIntegerOverflow())
+    if args.detect_uninitialized_storage:
+        m.register_detector(DetectUninitializedStorage())
+    if args.detect_uninitialized_memory:
+        m.register_detector(DetectUninitializedMemory())
+
+    if args.avoid_constant:
+        #avoid all human level tx that has no effect on the storage
+        print "avoiding contants"
```
yan: old-style print; should this be a `logger.info`?

manticore/ethereum.py
```diff
@@ -41,22 +39,48 @@ class NoAliveStates(EthereumError):
...
 
+################ Detectors ####################
 class Detector(Plugin):
     @property
     def name(self):
         return self.__class__.__name__.split('.')[-1]
 
     def get_findings(self, state):
-        return state.context.setdefault('seth.findings.%s' % self.name, set())
+        return state.context.setdefault('%s.findings' % self.name, set())
```
yan: old style format; prefer `.format`?

manticore/ethereum.py
```diff
@@ -68,12 +92,115 @@ def report(self, state):
...
+
+            #Get metadata if any for the targe addreess of current tx
+            md = self.manticore.get_metadata(tx.address)
+            if md is None:
+                return
+            #Lets compile  the list of interesting hashes
+            selected_functions = []
+
+            for func_hsh in md.hashes:
+                if func_hsh == '00000000':
```
yan: should this be a `bytes` literal?

manticore/ethereum.py
```diff
@@ -68,12 +92,115 @@ def report(self, state):
...
+                if not re.match(self._regexp, func_name):
+                    continue
+                selected_functions.append(func_hsh)
+
+            if self._fallback:
+                selected_functions.append('00000000')
+
+            if self._include:
+                # constraint the input so it can take only the interesting values
+                constraint = reduce(Operators.OR, map(lambda x: tx.data[:4] == x.decode('hex'), selected_functions))
```
yan: `.decode('hex') -> binascii.unhexlify` (Ditto below)

manticore/ethereum.py
```diff
@@ -1447,6 +1572,26 @@ def flagged(flag):
...
+        local_findings = set()
+        for detector in self.detectors.values():
+            for address, pc, finding, at_init in detector.get_findings(state):
+                if (address, pc, finding, at_init) not in local_findings:
+                    local_findings.add((address, pc, finding, at_init))
+
+        if len(local_findings):
+            with testcase.open_stream('findings') as findings:
+                for address, pc, finding, at_init in local_findings:
+                    findings.write('Finding: %s\n' % finding)
```
yan: This func, move to `.format`?

manticore/ethereum.py
```diff
@@ -1447,6 +1572,26 @@ def flagged(flag):
...
+                    findings.write('Contract: 0x%x\n' % address)
+                    findings.write('EVM Program counter: %s%s\n' % (pc, at_init and " (at constructor)" or ""))
+                    md = self.get_metadata(address)
+                    if md is not None:
+                        src = md.get_source_for(pc, runtime=not at_init)
+                        findings.write('Snippet:\n')
+                        findings.write('\n'.join((' ' + x for x in src.split('\n'))))
```
yan: Use `textwrap.TextWrapper` here?

manticore/ethereum.py
feliam: idk. That's the hex representation of the 4 byte binary. It'll always be ascii.

Issue #514, opened 2017-10-02
ARM v7 Support - Invalid Instruction


mikedefrancis:
### OS / Environment
Ubuntu 16.04, Raspberry Pi 2

### Manticore version
Latest  as of 2017-09-22

### Python version
2.7

### Dependencies
All installed

### Summary of the problem
Manticore is not decoding ARMv7 instructions correctly.

Additional:
Manticore cannot run a simple "hello world" (or any other) program on the Raspberry Pi 2, because some instructions are unable to be decoded. This is likely an issue with the ARMv7 support within Manticore. 
I am trying to use manticore to create a program that compares the functional equivalence of two .elf binary programs using symbolic execution. I would like to port this program to ARM, but when running manticore (basic install) on my raspberry pi, I get the following error. I am assuming this is a genuine bug since the docs say that manticore supports ARMv7 (e.g., raspberry pi 2).  I am using Ubuntu 16.04.

### Step to reproduce the behavior
Build and run a " hello world" on a raspberry pi and run using manticore

### Expected behavior
ARMv7 instructions need to be decoded correctly in order for manticore to be used for various applications, including efforts being undertaken by large organizations who would like to spread the use of the tool.

### Actual behavior
"Error decoding instruction" thrown 

### Any relevant logs
![test](https://user-images.githubusercontent.com/9902708/31087566-81450060-a76b-11e7-82a4-e8ba84a8ae11.png)

yan:
@mikedefrancis can you provide the binary you were using?

catenacyber:
Here is a binary 
https://github.com/catenacyber/manticore/blob/rasp/tests/binaries/hellorasp

compiled from
```
#include <stdio.h>

int main()
{
    printf("Hello World!\n");
    return 0;
}
```

on 
`Linux raspberrypi 4.14.34-v7+ #1110 SMP Mon Apr 16 15:18:51 BST 2018 armv7l GNU/Linux`

I do not get the same error.
But I am getting 
```
2018-06-01 11:20:29,456: [36900] m.manticore:INFO: Generated testcase No. 0 - Invalid memory access (mode:r) <0>
```
instead of having "Hello world" printed

Failure happens at
`0x00000000bffc7b70:	ldrb	r2, [r1], #1`
where R1 register is 0, hence the invalid memory access
It is coming from `0x00000000bffbed70:	ldr	r1, [r1, #4]` where R1 value was 0x00000000bffde568

This code is in file `ld-linux-armhf.so.3`

[...]

PR #914, opened 2018-06-04
Raspberry hello world

catenacyber:
See #514 

Adds ARM UQSUB8 instruction, and a binary example

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/914)


manticore/core/cpu/arm.py
```diff
@@ -1258,3 +1258,15 @@ def DMB(cpu, *operands):
     @instruction
     def LDCL(cpu, *operands):
         """Occasionally used in glibc (longjmp in ld.so). Nop under our execution model."""
+
+    @instruction
+    def UQSUB8(cpu, dest, op1, op2):
+        src1 = op1.read()
+        src2 = op2.read()
+        result = 0
+        for i in (range(op1.size / 8)):
```
disconnect3d: We should do it as `op1.size // 8` so that we are compatible with both PY2 and Py3.
We could also add `from __future__ import division` at the top.

catenacyber: Ok, here you go

yan:
@catenacyber thanks for this pr! i took the liberty to slightly rewrite it to make it compatible with symbolic values, and added a few tests. Will merge as soon as tests are complete.

yan:
It's not being used because we'd still hit the capstone bug (check the last few comments on #514)

mossberg:
Hmm. Well, if it's not going to be used anytime soon, maybe we shouldn't add it to the repo and create clutter?

PR #922, opened 2018-06-07
[WIP] Add a lazy symbolic memory implementation 

yan:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/922)


manticore/core/memory.py
```diff
@@ -1045,6 +1085,74 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+            found = None
+            for m in self._maps:
+                if self._deref_can_succeed(m, address, 1):
+                    if found:
+                        raise InvalidMemoryAccess(address, 'r')
+                    found = m
+            return found
+
+    def read(self, address, size, force=False):
+        m = self.map_containing(address)
```
feliam: What if the size is big enough to read outside the map initial map?

manticore/core/memory.py
```diff
@@ -1045,6 +1085,74 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+        if addr is not None:
+            addr = self._floor(addr)
+
+        size = self._ceil(size)
+
+        addr = self._search(size, addr)
+
+        # It should not be allocated
+        for i in xrange(self._page(addr), self._page(addr + size)):
+            assert i not in self._page2map, 'Map already used'
```
mossberg: we should use an exception, not an assert here probably, in case we run w/ asserts disabled

manticore/core/memory.py
```diff
@@ -1045,6 +1085,74 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+            for m in self._maps:
+                if self._deref_can_succeed(m, address, 1):
+                    if found:
+                        raise InvalidMemoryAccess(address, 'r')
+                    found = m
+            return found
+
+    def read(self, address, size, force=False):
+        m = self.map_containing(address)
+        if isinstance(m, ArrayMap):
```
mossberg: could we simplify this by assuming all maps in the LazySMemory will be ArrayMaps?

yan: I am using `FileMap`s for things loaded by the elf loader; we should probably back them with initialized `ArrayMap`s

feliam: Should not you upgrade the map to symbolic when writting a symbolic value?

manticore/__main__.py
```diff
@@ -66,8 +66,11 @@ def positive(value):
     parser.add_argument('--workspace', type=str, default=None,
                         help=("A folder name for temporaries and results."
                               "(default mcore_?????)"))
+    parser.add_argument('--pure-symbolic', action='store_true',
```
feliam: We should find a way to have different configs for diff argv[0]

PR #935, opened 2018-06-18
Add UTXH to arm instructions

catenacyber:
Same as utxb but with halfword instead of byte

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/935)


yan:
@catenacyber thanks again! as with last few PRs, I refactored a bit and added a test.

catenacyber:
Yes, I should think about adding tests for the next PRs

PR #941, opened 2018-06-21
Manticore 0.1.10

mossberg:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/941)


manticore/__main__.py
```diff
@@ -66,7 +66,7 @@ def positive(value):
     parser.add_argument('--workspace', type=str, default=None,
                         help=("A folder name for temporaries and results."
                               "(default mcore_?????)"))
-    parser.add_argument('--version', action='version', version='Manticore 0.1.9',
+    parser.add_argument('--version', action='version', version='Manticore 0.1.10',
```
feliam: So this and setup.py... is where the "version gets bumped".

Issue #954, opened 2018-06-27
manticore-examples throw ValueError


devtty1er:
### OS / Environment
Travis CI

### Manticore version
Since cca85ef84749db8081f469988ead03b79a505d71

### Summary of the problem
https://travis-ci.org/trailofbits/manticore-examples/builds/393183166

PR #955, opened 2018-06-27
Add uncaught exception

devtty1er:
Fixes trailofbits/manticore#954

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/955)

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -68,8 +68,9 @@ class File(object):
         state = {}
         state['name'] = self.name
         state['mode'] = self.mode
+        state['closed'] = self.closed
         try:
-            state['pos'] = self.tell()
+            state['pos'] = None if self.closed else self.tell()
         except IOError:
             # This is to handle special files like /dev/tty
             state['pos'] = None
@@ -78,8 +79,15 @@ class File(object):
     def __setstate__(self, state):
         name = state['name']
         mode = state['mode']
+        closed = state['closed']
         pos = state['pos']
-        self.file = open(name, mode)
+        try:
+            self.file = open(name, mode)
+            if closed:
+                self.file.close()
+        except IOError:
+            # If the file can't be opened anymore, should not typically happen
+            self.file = None
         if pos is not None:
             self.seek(pos)
 
@@ -90,6 +98,10 @@ class File(object):
     @property
     def mode(self):
         return self.file.mode
+
+    @property
+    def closed(self):
+        return self.file.closed
 
     def stat(self):
         return os.fstat(self.fileno())


mossberg:
awesome! thanks so much for helping us with this :) we'll get to this soon. fyi we are in somewhat of a feature freeze on master right now while we work on porting things to python3, so it might be a little bit before we can actually merge this.

devtty1er:
It looks like it was just the pickle serializer recursing down `__getattr__`'s from state to platform, which includes `_closed_files`. I added a comment to clarify. While I don't think it is necessary, we could
```python
try:
    state['pos'] = self.tell()
except IOError:
    # This is to handle special files like /dev/tty 
    state['pos'] = None 
except ValueError: 
    if state.file.closed: 
        state['pos'] = None 
    else: 
        raise 
return state 
```
just to be safe.

mossberg:
ok just spent some time looking it, and confirmed that you're totally right. we try to serialize ._closed_files and end up calling .tell() on a file object that has been closed.

this made me realize something else. in File.__setstate__ we probably should not open the underlying file if the file has been closed. maybe we should include some kind of instance variable in File to track the current open/closed status. we update that variable in the constructor/.close(), and can check it in __setstate__ to know whether to open the file or not.

manticore/platforms/linux.py
```diff
@@ -67,16 +67,18 @@ def __getstate__(self):
...
+            # This is to handle special files like /dev/tty
             state['pos'] = None
         return state
 
     def __setstate__(self, state):
         name = state['name']
         mode = state['mode']
+        closed = state['closed']
         pos = state['pos']
         self.file = file(name, mode)
```
mossberg: the original idea was to now use `closed` here to be able to check whether to call `file()` or not. but i'm now realizing that if we don't call `file()`, and just assign `self.file = None` or something, that will break most of the properties in this class and cause them to AttributeError or something. 

a solution to this might be to always try to call `file()`, but then if the file is closed, immediately call `.close()`. this way, we don't unnecessarily have a file open, but the various informational attributes like `.name` will still exist.

manticore/platforms/linux.py
```diff
@@ -77,8 +78,15 @@ def __getstate__(self):
     def __setstate__(self, state):
         name = state['name']
         mode = state['mode']
+        closed = state['closed']
         pos = state['pos']
-        self.file = file(name, mode)
+        try:
+            self.file = file(name, mode)
```
mossberg: let's use `open()` instead of `file()` while we're making changes here

mossberg:
Ok! To move forward with this, now we need to get it up to date with current master (which has now shifted to a Python3 _only_ codebase), and do any porting to python3 necessary. Could you please either merge master into this branch, or rebase? I don't actually expect we'll really need to change the code much for py3 porting.

PR #959, opened 2018-06-29
Fix Array Slice and test

feliam:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/959)


tests/test_smtlibv2.py
```diff
@@ -242,6 +242,14 @@ def testBasicArrayConcatSlice(self):
...
         self.assertTrue(self.solver.must_be_true(cs, array.read(0,1) + bytearray('ello ') + array.read(6,5) + bytearray('!') == hw))
 
+        self.assertTrue(len(array[1:2]) == 1)
+
+        self.assertTrue(len(array[0:12]) == 12)
+
+        results = []
+        for c in array[6:11]:
+            results.append(c)
+        self.assertTrue(len(results) == 5)
```
disconnect3d: Just `len(array[6:11]) == 5` ?

feliam: __len__ vs. IndexError ?

disconnect3d: What do you mean? `array[6:11]` won't get us an `IndexError`

```
In [1]: [][6:11]
Out[1]: []

In [2]: [0,1,2,3,4,5][6:11]
Out[2]: []

In [3]: [0,1,2,3,4,5,6][6:11]
Out[3]: [6]

In [4]: [0,1,2,3,4,5,6,7,8][6:11]
Out[4]: [6, 7, 8]

In [5]: [0,1,2,3,4,5,6,7,8,9,10,11,12,13][6:11]
Out[5]: [6, 7, 8, 9, 10]
```

feliam: Oh. How does the `for c in ARR` know where to stop? I assumed IndexError

disconnect3d: It stops on `StopIteration`, e.g.:

```
In [5]: next(iter([][6:11]))
StopIteration                             Traceback (most recent call last)
<ipython-input-5-4e5aab60a3af> in <module>()
----> 1 next(iter([][6:11]))

StopIteration: 
```

To elaborate more, `for c in array[6:11]:` does more or less the same as code below:
```python
iterator = iter(array[6:11])

while True:
    try:
        c = next(iterator)
    except StopIteration:
        break

    # do sth with c
    # ...
```

The thing here is how slicing works in Python: if there are no items, it will just return an empty list.

TLDR: The `result` array is redundant, as `array[6:11]` returns the same content.

feliam: Do you know if this line is tested if you remove the redundant `result` ? https://github.com/trailofbits/manticore/blob/yolo_master/manticore/core/smtlib/expression.py#L654

PR #961, opened 2018-06-29
add test for "arguments_linux_armv7" binary

### LINKED ISSUE #554: tests: add test for "arguments_linux_armv7" binary
The armv7 binary "arguments_linux_armv7" exists in the repo, but appears to be entirely unused. We should use it in test_binaries.py

We should be able to nearly copy/paste the existing "testArgumentsAssertions" test for this.

What needs to be done:
- Fix #553
- add a new function to test_binaries.py called "testArgumentsAssertionsArmv7" or something
- run `manticore arguments_linux_armv7 +++++++++`
- take the visited.txt from the workspace of that run, add it to the repo in the references/ dir
- copy paste the "testArgumentsAssertions" function and slightly change to make it work for the arm binary

catenacyber:
Fixes #554

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/961)


tests/test_binaries.py
```diff
@@ -120,6 +120,25 @@ def testArgumentsAssertions(self):
         expected = self._loadVisitedSet(os.path.join(dirname, 'reference', 'arguments_linux_amd64_visited.txt'))
         self.assertGreaterEqual(actual, expected)
 
+    def testArgumentsAssertionsArmv7(self):
+        dirname = os.path.dirname(__file__)
```
disconnect3d: Can we move this to a global variable e.g. `DIR_PATH = ...` ? some other tests might want to use that too :P

tests/test_binaries.py
```diff
@@ -120,6 +120,25 @@ def testArgumentsAssertions(self):
...
+        filename = os.path.abspath(os.path.join(dirname, 'binaries/arguments_linux_armv7'))
+        self.assertTrue(filename.startswith(os.getcwd()))
+        filename = filename[len(os.getcwd())+1:]
+        workspace = '%s/workspace'%self.test_dir
+        assertions = '%s/assertions.txt'%self.test_dir
+        file(assertions,'w').write('0x0000000000401003 ZF == 1')
+        with open(os.path.join(os.pardir, '%s/output.log'%self.test_dir), "w") as output:
```
disconnect3d: Just `os.path.join(os.pardir, self.test_dir, output.log)`

tests/test_binaries.py
```diff
@@ -120,6 +120,25 @@ def testArgumentsAssertions(self):
...
+            subprocess.check_call(['python', '-m', 'manticore',
+                                   '--workspace', workspace,
+                                   '--proc', '4',
+                                   '--assertions', assertions,
+                                   filename,
+                                   '+++++++++'], stdout=output)
+        actual = self._loadVisitedSet(os.path.join(dirname, '%s/visited.txt'%workspace))
```
disconnect3d: same as above: no need to do string formatting, just `os.path.join(dirname, workspace, 'visited.txt')`

tests/test_binaries.py
```diff
@@ -120,6 +120,25 @@ def testArgumentsAssertions(self):
...
         self.assertGreaterEqual(actual, expected)
 
+    def testArgumentsAssertionsArmv7(self):
+        dirname = os.path.dirname(__file__)
+        filename = os.path.abspath(os.path.join(dirname, 'binaries/arguments_linux_armv7'))
+        self.assertTrue(filename.startswith(os.getcwd()))
+        filename = filename[len(os.getcwd())+1:]
+        workspace = '%s/workspace'%self.test_dir
+        assertions = '%s/assertions.txt'%self.test_dir
+        file(assertions,'w').write('0x0000000000401003 ZF == 1')
```
disconnect3d: There is no `file` in Py3, this should be at least `open`, and the best would be `with open(assertions 'w') as f: f.write(...)`

disconnect3d: Since this is a test, it might be good to think of a named temporary file (see https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile)

tests/test_binaries.py
disconnect3d: E.g. the next test also uses this line:
```
     def testDecree(self):
         dirname = os.path.dirname(__file__)```

mossberg:
thanks for the review disconnect3d :)

tests/test_binaries.py
catenacyber: ok

tests/test_binaries.py
catenacyber: ok

tests/test_binaries.py
catenacyber: ok

catenacyber: As for the temporary file, we are already inside a temporary directory, is not that enough ?

tests/test_binaries.py
catenacyber: ok

catenacyber:
I have also been factoring code together between `testArgumentsAssertions` and `testArgumentsAssertionsArmv7`

tests/test_binaries.py
disconnect3d: > we are already inside a temporary directory, is not that enough ?

Yeah I think it is. The adventage of `NamedTemporaryFile` is that we don't have to remember to remove it at all as it has the `delete=True` default:
> If delete is true (the default), the file is deleted as soon as it is closed

catenacyber: the temp directory is deleted after execution of one test with `shutil.rmtree(self.test_dir)`

disconnect3d:
I have rebased this branch, refactored the code a bit and added some more assertions.
I still need to fix the list of addresses the program step on

The branch is https://github.com/trailofbits/manticore/tree/rebased-testarmv7

catenacyber:
Looks good to me even if it goes beyond fixing #554

disconnect3d:
Closing this issue in favor of #1216, gonna fix the last issue there and merge it.

Issue #964, opened 2018-07-04
install issue


lambdax01:
### OS / Environment
mac os 10.13.5

### Python version
2.7.10

### Dependencies
altgraph==0.10.2
bdist-mpkg==0.5.0
bonjour-py==0.3
capstone==3.0.5rc2
cytoolz==0.9.0.1
eth-hash==0.1.4
eth-utils==1.0.3
functools32==3.2.3.post2
macholib==1.5.1
matplotlib==1.3.1
modulegraph==0.10.4
numpy==1.8.0rc1
ply==3.11
py2app==0.7.3
pyelftools==0.24
pyobjc-core==2.5.1
pyobjc-framework-Accounts==2.5.1
pyobjc-framework-AddressBook==2.5.1
pyobjc-framework-AppleScriptKit==2.5.1
pyobjc-framework-AppleScriptObjC==2.5.1
pyobjc-framework-Automator==2.5.1
pyobjc-framework-CFNetwork==2.5.1
pyobjc-framework-Cocoa==2.5.1
pyobjc-framework-Collaboration==2.5.1
pyobjc-framework-CoreData==2.5.1
pyobjc-framework-CoreLocation==2.5.1
pyobjc-framework-CoreText==2.5.1
pyobjc-framework-DictionaryServices==2.5.1
pyobjc-framework-EventKit==2.5.1
pyobjc-framework-ExceptionHandling==2.5.1
pyobjc-framework-FSEvents==2.5.1
pyobjc-framework-InputMethodKit==2.5.1
pyobjc-framework-InstallerPlugins==2.5.1
pyobjc-framework-InstantMessage==2.5.1
pyobjc-framework-LatentSemanticMapping==2.5.1
pyobjc-framework-LaunchServices==2.5.1
pyobjc-framework-Message==2.5.1
pyobjc-framework-OpenDirectory==2.5.1
pyobjc-framework-PreferencePanes==2.5.1
pyobjc-framework-PubSub==2.5.1
pyobjc-framework-QTKit==2.5.1
pyobjc-framework-Quartz==2.5.1
pyobjc-framework-ScreenSaver==2.5.1
pyobjc-framework-ScriptingBridge==2.5.1
pyobjc-framework-SearchKit==2.5.1
pyobjc-framework-ServiceManagement==2.5.1
pyobjc-framework-Social==2.5.1
pyobjc-framework-SyncServices==2.5.1
pyobjc-framework-SystemConfiguration==2.5.1
[...]

PR #965, opened 2018-07-05
Simplify installation instructions to recommend install manticore only for the current user

ggrieco-tob:
This will avoid issues like #964

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/965)

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Install and try Manticore in a few shell commands (see an [asciinema](https://as
 sudo apt-get update && sudo apt-get install python-pip -y
 
 # Install manticore and its dependencies
-sudo pip2 install manticore
+pip2 install manticore --user
 
 # Download the examples
 git clone https://github.com/trailofbits/manticore.git && cd manticore/examples/linux


disconnect3d:
The users may still have problems if they install something that has the same dependencies with different versions.

This problem can be solved in many ways:
- we can bundle manticore along with dependencies using e.g. pyinstaller - this lets us create single binary file containing manticore and dependencies; however this is not that good as it doesn't bundle e.g. libc. Because of that a package built on latest ubuntu may not work on old ubuntu because of glibc version mismatch (afaiu it is forward compatible but not backward compatible)
- virtualenv - this was recommended for a long time in Python communities and so a lot of ppl know it [along with virtualenvwrapper which makes it super easy to use]
- [pipenv](https://github.com/pypa/pipenv) - this is a way of handling dependencies for projects, so I am not sure if this is suitable here;
- some virtualenv bundler - e.g. https://github.com/pantsbuild/pex - someone recommended me this, but I haven't used it

For example on one of my machines I use virtualenv + virtualenvwrapper and I had manticore for both py2 and py3 and so I can do this (unfortunately py3 stopped working for some reason :P):
```
16:54:34 in ~ 
➜ workon manticore2

16:54:38 in ~ via manticore2 
➜ manticore --version
Manticore 0.1.8

16:54:40 in ~ via manticore2 
➜ workon manticore3  

16:54:42 in ~ via manticore3 
➜ manticore --version
  File "/home/dc/.venv/manticore3/bin/manticore", line 7, in <module>
    from manticore.__main__ import main
  File "/home/dc/.venv/manticore3/lib/python3.6/site-packages/manticore/__init__.py", line 1, in <module>
    from .manticore import Manticore
  File "/home/dc/.venv/manticore3/lib/python3.6/site-packages/manticore/manticore.py", line 19, in <module>
    from .core.executor import Executor
  File "/home/dc/.venv/manticore3/lib/python3.6/site-packages/manticore/core/executor.py", line 8, in <module>
    from .smtlib import solver, Expression, SolverException
  File "/home/dc/.venv/manticore3/lib/python3.6/site-packages/manticore/core/smtlib/__init__.py", line 1, in <module>
    from expression import Expression, Bool, BitVec, Array
ModuleNotFoundError: No module named 'expression'
```

README.md
```diff
@@ -35,7 +35,7 @@ Install and try Manticore in a few shell commands (see an [asciinema](https://as
 sudo apt-get update && sudo apt-get install python-pip -y
 
 # Install manticore and its dependencies
-sudo pip2 install manticore
```
mossberg: We want these commands in the quick start to work for people 99.9999% of the time :)

the `--user` command requires ~/.local in your path for things to work, which not everybody has, so that's why in this section we recommend just using sudo, even if it may not be the exact best.

mossberg:
going to close this for now, feel free to open!

PR #968, opened 2018-07-10
Python 3

defunctio:
A new branch targeting only Python 3 (not 2/3).

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/968)


feliam:
___
*[manticore/ethereum.py, line 1586 at r2](https://reviewable.io/reviews/trailofbits/manticore/968#-LHi9qUGCXidBGXYhXIB:-LHi9qUGCXidBGXYhXIC:b-t1chj9) ([raw file](https://github.com/trailofbits/manticore/blob/b311c51f9fb9612c934bc34bac72c851bc59d532/manticore/ethereum.py#L1586)):*
> ```Python
>             raise EthereumError("A concrete address is needed")
>         assert address is not None
>         if address in list(map(int, list(self.accounts.values()))):
> ```

Do we need any/both `list` here?

feliam:
___
*[manticore/core/workspace.py, line 444 at r4](https://reviewable.io/reviews/trailofbits/manticore/968#-LHkURt_65jtfcw32CUD:-LHkURt_65jtfcw32CUE:bocpcsp) ([raw file](https://github.com/trailofbits/manticore/blob/4983a230cf9ba17e832250f4723dd2b256dc3ea1/manticore/core/workspace.py#L444)):*
:( another manager
Does _lock need to be a conditional variable? Is there a wait?
> ```
> _lock
> ```

mossberg:
to revisit:

- int type checking: int vs numbers.Integral (use int)
- gap in test cov here: https://github.com/trailofbits/manticore/pull/968#pullrequestreview-139925440

feliam:
___
*[manticore/core/parser/parser.py, line 170 at r5](https://reviewable.io/reviews/trailofbits/manticore/968#-LI7SHhyBTGpUfQ_vGnA:-LIDP90uCwMt6m_D2l6i:b26p53k) ([raw file](https://github.com/trailofbits/manticore/blob/8070233441de98380b9a0d9670c5af3c1f913665/manticore/core/parser/parser.py#L170)):*
<details><summary><i>Previously, defunctio (defunct) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

that for an intel assembler expression calculator like "DWORD PTR [ebx+4]*10 /eax == ecx" or similar. I guess then it is integer div

feliam:
___
*[manticore/ethereum.py, line 923 at r2](https://reviewable.io/reviews/trailofbits/manticore/968#-LHi9HlKChjPT2IMPS-0:-LIDXbwVD4xGBRnKCOzD:bhma5ym) ([raw file](https://github.com/trailofbits/manticore/blob/b311c51f9fb9612c934bc34bac72c851bc59d532/manticore/ethereum.py#L923)):*
<details><summary><i>Previously, defunctio (defunct) wrote…</i></summary><blockquote>

seems only numbers.Integral, EVMAccount have equality checks. Otherwise it will just be compared against object.__eq__, so nope.
</blockquote></details>

I suppose I'll add it again if I need to compare an EVMAccount against a BitVec.

feliam:
___
*[manticore/core/state.py, line 138 at r2](https://reviewable.io/reviews/trailofbits/manticore/968#-LHkS7S42H86LGH-USDN:-LIDYKeo6IOxVLoA4tOC:b-cscpej) ([raw file](https://github.com/trailofbits/manticore/blob/b311c51f9fb9612c934bc34bac72c851bc59d532/manticore/core/state.py#L138)):*
<details><summary><i>Previously, defunctio (defunct) wrote…</i></summary><blockquote>

python 3 GCs the assignment of `e` in an exception clause unless copied to a local. 

See:
https://docs.python.org/3/reference/compound_stmts.html#the-try-statement
> When an exception has been assigned using as target, it is cleared at the end of the except clause. 


</blockquote></details>

Also this "ConcretizeRegister/ConcretizeMemory" is platform specific (as in it wont work for EVM) so it should go to ._platform.execute

PR #990, opened 2018-07-25
Python 3; optimization / stylization pass

defunctio:
* cleanup list() from automation tools
  * style; use dict comprehensions
  * style; use set literals

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/990)


mossberg:
re: #968

mossberg:
in this pr, let's also move from `super(class, self)` -> `super()`

PR #1009, opened 2018-07-31
evm: aggressively check & migrate expressions into current ConstraintSet in case they are global/external

feliam:
this pr introduces advanced functionality (state.migrate_expression, and friends) for migration of potentially external/global ("foreign") expression objects into a particular ConstraintSet. with recent developments in the Ethereum API specifically (global expressions via `ManticoreEVM.make_....1), we expect foreign expressions to be passed into these interfaces fairly frequently now, so that PR "aggressively" adds a call to this function in nearly every interface that accepts an expression object, in order to be as safe as possible

```
from manticore.core.smtlib import Operators
from manticore.core.smtlib import solver

m = ManticoreEVM()

contract_src='''
contract Overflow {
  uint public sellerBalance=0;

  function add(uint value)public  returns (bool){
      sellerBalance += value;
  }
}
'''

owner_account=m.create_account(balance=1000)
attacker_account=m.create_account(balance=1000)
contract_account=m.solidity_create_contract(contract_src,
                                            owner=owner_account,
                                            balance=0)

#Some global expression `sym_add1` 
sym_add1 = m.make_symbolic_value(name='sym_add1')
#Let's constraint it on the global fake constraintset
m.constrain(sym_add1>0)
m.constrain(sym_add1<10)
#Symb tx 1
contract_account.add(sym_add1, caller=attacker_account)

# A new!? global expression 
sym_add2 = m.make_symbolic_value(name='sym_add2')
#constraints involve old expression.  Some states may get invalidated by this. Should this be accepted?
m.constrain(sym_add1 > sym_add2)
#Symb tx 2
contract_account.add(sym_add2, caller=attacker_account)

#random concrete tx
contract_account.sellerBalance(caller=attacker_account)

#anooother constraining on the global constraintset. Yet more running states could get unfeasible by this.
m.constrain(sym_add1 > 8)



for state_num, state in enumerate(m.all_states):

    if state.is_feasible():
        #todo merge with some expression got froms state
        print ('feasible state found')
        print ('sym_add1', state.solve_one(sym_add1))
        print ('sym_add2', state.solve_one(sym_add2))
        m.generate_testcase(state, 'test{}'.format(state_num))
```

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1009)


feliam:
I guess so. The `name=` is then a suggestion. User should do a `cs.get_variable(var.name) -> var`

mossberg:
I see. that's very non-obvious :)

why do we need to do the `_get_new_name` thing? could we throw an error if a name is passed that already exists? or have an optional flag to do the new name thing?

from a user perspective it would be very nice to just pick a name for the variable and be able to use that throughout.

feliam:
If it does not exist the name the user picked would be the same as in the returned var.name. 
Throwing when already exists will pull in a lot of `try:except:`
How would the new signature of new_bitvec?  We can also change `name` for `suggestedname` :)
pyz3 returns the old variable with the same name on that :woman_shrugging:

mossberg:
> If it does not exist the name the user picked would be the same as in the returned var.name

it looks to me like we unconditionally append digits and change the name

https://github.com/trailofbits/manticore/blob/59a5ff0ba335d3c0409dbac7dfed3a6a225943bf/manticore/core/smtlib/constraints.py#L245



> How would the new signature of new_bitvec?

very roughly, here's an idea

```
# ignoring self, size, taint
def new_bitvec(name, avoid_name_collision=False)
    if avoid_name_collision:
        if name exists:
            name = get_name_name(name)
    else:
        if name exists:
            raise Exception
    return BitVecVariable(name)


```

manticore/ethereum.py
```diff
@@ -631,7 +631,7 @@ def get_source_for(self, asm_offset, runtime=True):
 
     @property
     def signatures(self):
-        return dict(((b, a) for (a, b) in self._hashes.items()))
+        return dict(((b.encode(), a) for (a, b) in self._hashes.items()))
```
feliam: if we want to ref them as bytes should probably just store them as bytes rather than decoding every item on each access to signatures.

since its a hex string, should probably just be str no? this is where python static typing stuff would be handy

feliam:
rename=True/False instead of avoid_name_collision=True/False ?
Also if "avoid_name_collision=False" should it raise or return the old known version of the variable  (In which case we'll need to remove the "new_" from the name) ?

mossberg:
> rename=True/False instead of avoid_name_collision=True/False ?

i do think avoid_name_collision is pretty clunky, i didn't put much thought into it :) . but i'm not sure if `rename` is expressive enough. maybe `strict=`? `avoid_collision`

alternatively what if we have 

`new_bitvec(name=None, prefix=None)`.

if you specify name, it _must_ not exist, otherwise an error will be thrown.
if you specify prefix, you don't care, and things will be appended to it to make it unique.
it is an error to specify both.
if none are specified, and random name default name/prefix will be chosen.



> Also if "avoid_name_collision=False" should it raise or return the old known version of the variable (In which case we'll need to remove the "new_" from the name) ?

I think it should raise there. trying to a create a "new" one with an existing name is an error condition imo. users should probably check to see if a name exists before trying to make a new one. i guess they would use `get_variable()` for this check?

mossberg:
my understanding is that there two main users of the `new_bitvec` etc class of functions. 

1. external users
2. the manticore core

the `name=` interface is geared for the external users, who have a specific variable in mind they want to create, and want to attach a name to it so they can find it later.

the `prefix=` interface is meant for the manticore core, for example places like here

https://github.com/trailofbits/manticore/blob/24cb4bd3f9581cb21dd1881f859492abad2f88de/manticore/core/smtlib/solver.py#L406

where the name is somewhat auto generated, and doesn't particularly matter to anyway.

this could all be overcomplicating things though. i'm fine with the `name=, bool_flag=` interface, and i think either `rename=` or `avoid_collisions=` would be good. i'm happy with either, @feliam i'll let you choose :)

feliam:
___
*[manticore/core/smtlib/constraints.py, line 231 at r4](https://reviewable.io/reviews/trailofbits/manticore/1009#-LJ_xD0g3N82qSrW0JwM:-LJnUPw76-wWP0Sms2xK:b-oherbh) ([raw file](https://github.com/trailofbits/manticore/blob/5b522cca36b0bdadfcd3f1b9d926316c9d3a6f30/manticore/core/smtlib/constraints.py#L231)):*

i think we should choose a better name for fat_bindings, it's not very clear what it is based on the name. it's the actual map from "alient" expression object, to the newly created native object. maybe something like foreign_to_native_expression_map? but that is pretty verbose. i think throughout this code could be made more readable by deciding on a convention like "foreign"/"native" or alien/local etc etc and using those in the variable names throughout. i can submit a PR with what i think could be more readable
</blockquote></details>

I made some name refactoring. You could try a PR with more/diff changes and we can discuss more there. I do not have strong ideas about this but here there are some:

It sounds to me `foreign_to_native == alien_to_local == migrate` after a little while.
Note that the name of the variables composing the expression to migrate may o may not be "alien/foreign" The expression could be composed with all king of variables locals and aliens from different constraint sets. Lookout because it would be "potentially_alien_var_name".

feliam:
___
*[manticore/core/smtlib/constraints.py, line 275 at r10](https://reviewable.io/reviews/trailofbits/manticore/1009#-LK2LH0u9mOZKpPSGu59:-LK3dYEt7ens1tMjWyO-:b-3jz2px) ([raw file](https://github.com/trailofbits/manticore/blob/afbb43028b60782b8097b1824803e86c90bf6359/manticore/core/smtlib/constraints.py#L275)):*

Let's make an issue maybe?
</blockquote></details>

https://github.com/trailofbits/manticore/issues/1059
done!

PR #1023, opened 2018-08-03
Change how we query for version

### LINKED ISSUE #1021: Error in solver version parsing
### Manticore version
0d843457a9d00127607d5adb06395a280b643662

### Summary of the problem

There is a bug in solver.py Z3Solver._solver_version() where it incorrectly parses the z3 version.
### Actual behavior

```
[I] (py3) vagrant ubuntu-bionic /v/m/symbolic_file ❯ python
Python 3.6.5 (default, Apr  1 2018, 05:46:30)
[GCC 7.3.0] on linux
>>> solver.version
Version(major=0, minor=0, patch=0)
>>>
[I] (py3) vagrant ubuntu-bionic /v/m/symbolic_file ❯ z3 --version
Z3 version 4.5.1 - 64 bit
[I] (py3) vagrant ubuntu-bionic /v/m/symbolic_file ❯


```

yan:
Fixes #1021

This also should decrease how many times we invoke z3. (The instance used to query version should stick around)

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1023)

### DIFF
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -20,6 +20,7 @@ from .expression import *
 from .constraints import *
 import logging
 import re
+import shlex
 import time
 from .visitors import *
 from ...utils.helpers import issymbolic, istainted, taint_with, get_taints, memoized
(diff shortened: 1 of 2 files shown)


manticore/core/smtlib/solver.py
```diff
@@ -164,16 +166,12 @@ def _solver_version():
 
         '''
         their_version = Version(0, 0, 0)
```
mossberg: this variable is no longer used

PR #1024, opened 2018-08-03
fix sys_write logger output

### LINKED ISSUE #940: Incorrect argument width being used with some syscalls
### Summary of the problem
For some system calls, Manticore uses the incorrect bitwidth for its arguments. More specifically, openat(2) interprets its first argument as a 64 bit signed number, where as it gets passed a 32 bit signed number.

Emulate a binary that invokes openat() syscall, which is usually used for libc's `open()` implementation, especially when opening files in the cwd. Note the value of fd that gets passed to `sys_openat`.

### Any relevant logs

Example syscall trace:

```
2018-06-21 15:19:43,302: [42011] m.p.platform:DEBUG: sys_access(0x4a52fa (/etc/ld.so.nohwcap), 0x0) -> -1
2018-06-21 15:19:45,565: [42011] m.p.platform:DEBUG: sys_newuname(0x7ffffffffc10 (Linux)) -> 0
2018-06-21 15:19:45,892: [42011] m.p.platform:DEBUG: sys_newfstat(0x1, 0x7ffffffff3c0) -> 0
2018-06-21 15:19:45,935: [42011] m.p.platform:DEBUG: sys_ioctl(0x1, 0x5401, 0x7ffffffff320) -> -22
2018-06-21 15:19:47,445: [42011] m.p.platform:DEBUG: sys_openat(0xffffff9c, 0x4926cc (LICENSE), 0x0, 0x0) -> -9
2018-06-21 15:19:47,469: [42011] m.p.linux:INFO: Calling fd: 4294967295
2018-06-21 15:19:47,469: [42011] m.p.linux:INFO: Calling newfstat with invalid fd
2018-06-21 15:19:47,470: [42011] m.p.platform:DEBUG: sys_newfstat(0xffffffff, 0x7ffffffffb80) -> -9
```

Note that sys_openat() gets called with 0xffffff9c, and not -100 as it should be.

### LINKED ISSUE #1020: Odd logger output for linux logger
### Manticore version
0d843457a9d00127607d5adb06395a280b643662

### Summary of the problem
there's a bug in the logger printing in linux.py, that text should say "greater" rather than the output with all the bs. it's probably something related to bytes/str

`manticore -v x`
[x.zip](https://github.com/trailofbits/manticore/files/2258374/x.zip)

### Actual behavior

```
2018-08-03 17:40:44,922: [2056] m.p.platform:DEBUG: sys_newfstat(0x1, ()) -> 0
2018-08-03 17:40:44,966: [2056] m.p.platform:DEBUG: sys_ioctl(0x1, 0x5401, 0x7ffffffffb90) -> -22
2018-08-03 17:40:45,467: [2056] m.p.linux:DEBUG: sys_close(3)
2018-08-03 17:40:45,468: [2056] m.p.platform:DEBUG: sys_close(0x3) -> 0
2018-08-03 17:40:45,726: [2056] m.p.linux:DEBUG: WRITE(1, 0x006cf3b0, 8) -> <"b'g'b'r'b'e'b'a'b't'b'e'b'r'b'\\n'">
2018-08-03 17:40:45,727: [2056] m.p.platform:DEBUG: sys_write(0x1, (greater), 0x8) -> 8
2018-08-03 17:40:46,172: [2056] m.manticore:INFO: Generated testcase No. 0 - Program finished with exit status: 0
```
specifically this line

```
2018-08-03 17:40:45,726: [2056] m.p.linux:DEBUG: WRITE(1, 0x006cf3b0, 8) -> <"b'g'b'r'b'e'b'a'b't'b'e'b'r'b'\\n'">


```

defunctio:
- write/writev/read bugfixes
- add typing annotations
- resolves #1020 
- resolves #940

This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1024)

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -9,8 +9,8 @@ import ctypes
 import socket
 import binascii
 
-# Remove in favor of binary.py
 from typing import Union, List, TypeVar, Iterable, ByteString, Sequence, Generic, cast, Collection
+# Remove in favor of binary.py
 from elftools.elf.elffile import ELFFile
 from elftools.elf.sections import SymbolTableSection
 from elftools.elf.descriptions import describe_symbol_type
@@ -122,8 +122,7 @@ class File(object):
         return self.file.seek(*args)
 
     def write(self, buf):
-        for c in buf:
-            self.file.write(c)
+        return self.file.write(buf)
 
     def read(self, *args):
         return self.file.read(*args)
@@ -1310,7 +1309,7 @@ class Linux(Platform):
                              fd,
                              buf,
                              count,
-                             line.decode('latin-1'))
+                             line.decode('latin-1'))  # latin-1 encoding will happily decode any byte (0x00-0xff)
             self.syscall_trace.append(("_write", fd, data))
             self.signal_transmit(fd)
 
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -1459,7 +1459,7 @@ class Linux(Platform):
         '''
 
         filename = self.current.read_string(buf)
-        dirfd = self._to_signed_dword(dirfd)
+        dirfd = ctypes.c_int32(dirfd).value
 
         if os.path.isabs(filename) or dirfd == self.FCNTL_FDCWD:
             return self.sys_open(buf, flags, mode)
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -10,10 +10,12 @@ import socket
 import binascii
 
 # Remove in favor of binary.py
+from typing import Union, List, TypeVar, Iterable, ByteString, Sequence, Generic, cast, Collection
 from elftools.elf.elffile import ELFFile
 from elftools.elf.sections import SymbolTableSection
 from elftools.elf.descriptions import describe_symbol_type
 
+from manticore.core.smtlib import Expression
 from ..core.cpu.abstractcpu import Interruption, Syscall, ConcretizeArgument
 from ..core.cpu.cpufactory import CpuFactory
 from ..core.memory import SMemory32, SMemory64, Memory32, Memory64
@@ -25,6 +27,9 @@ from ..utils.helpers import issymbolic, is_binja_disassembler
 from . import linux_syscalls
 
 logger = logging.getLogger(__name__)
+
+T = TypeVar('T')
+MixedSymbolicBuffer = Union[List[Union[bytes, Expression]], bytes]
 
 
 class RestartSyscall(Exception):
@@ -54,7 +59,7 @@ def perms_from_protflags(prot_flags):
 
 
 def mode_from_flags(file_flags):
-    return {os.O_RDWR: 'r+', os.O_RDONLY: 'r', os.O_WRONLY: 'w'}[file_flags & 7]
+    return {os.O_RDWR: 'rb+', os.O_RDONLY: 'rb', os.O_WRONLY: 'wb'}[file_flags & 7]
 
 
 class File(object):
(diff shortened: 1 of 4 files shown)


manticore/platforms/linux.py
```diff
@@ -1292,12 +1292,12 @@ def sys_write(self, fd, buf, count):
...
             write_fd.write(data)
 
-            for line in ''.join([str(x) for x in data]).split('\n'):
+            for line in b''.join([x for x in data]).split(b'\n'):
                 logger.debug("WRITE(%d, 0x%08x, %d) -> <%.48r>",
                              fd,
                              buf,
                              count,
-                             line)
+                             line.decode('latin-1'))
```
mossberg: what's the reason for latin-1?

defunctio: latin-1 will decode 0x00-0xff

mossberg: regular `line.decode()` seems to work fine for me. tbh i don't know a whole lot about encodings, is there any reason we shouldn't just use utf-8?

defunctio: utf-8 it would raise with `UnicodeDecodeError` on any byte > 0x7f and in this case what is being written is just binary data. would be nice to replace this at a later date with something similar of `xxd` output. but that might be too verbose

mossberg:
```
>>> b'sdkjfhk\x20\x8f'.decode()
  File "<input>", line 1, in <module>
    b'sdkjfhk\x20\x8f'.decode()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8f in position 8: invalid start byte
>>> b'sdkjfhk\x20\x8f'.decode('latin-1')
'sdkjfhk \x8f'
>>>
```

manticore/platforms/linux.py
```diff
@@ -1292,12 +1292,12 @@ def sys_write(self, fd, buf, count):
             data = self._transform_write_data(data)
             write_fd.write(data)
 
-            for line in ''.join([str(x) for x in data]).split('\n'):
+            for line in b''.join([x for x in data]).split(b'\n'):
```
mossberg: data can be a list of bytes, but it can also contain Expression objects (ArraySelect) instances, which i believe is why the str() call was there. so we i think we still need to do something like `str(x).encode()`

one way to easily set this up:
```
import manticore
m = manticore.Manticore('/bin/ls')
s = m.initial_state
cpu = s.cpu
arr = s.new_symbolic_buffer(4)
cpu.push_bytes(arr)
cpu.read_bytes(cpu.RSP, 4)
b = cpu.read_bytes(cpu.RSP, 4)
```

defunctio: won't `data = self._transform_write_data(data)` line 1292 directly above only return concrete data?

Also, seems there is a bug in `SLinux._transform_write_data` where `concrete_data` should be bytes()

mossberg: > won't data = self._transform_write_data(data) line 1292 directly above only return concrete data

ah you're totally right, sorry about that.

> Also, seems there is a bug in SLinux._transform_write_data where concrete_data should be bytes()

perhaps it should be bytes, i think these functions were all generally written to take/return lists. for example when `write_fd.write(data)` happens, File.write is written to loop over the argument, rather than just passing it into file.write().

mossberg: since data is guaranteed to be concrete, instead of `b''.join([x for x in data])` i think we can do `b''.join(data)`

mossberg: > data should already guaranteed to be bytes we can just use data.split(b'\n')

i think data is a list of bytes, not bytes




> I should probably review / test all the read/write syscalls to see if anything else shows up

this would be amazing 🙏

defunctio: > i think data is a list of bytes, not bytes

Ah... local changes I already made I've switched this to using `bytes`. Any objection to that? There are very few references that use it internally and this just seems to make more sense.

defunctio: updated and tested read/readv/write/writev, everything appears to be working as expected

PR #1047, opened 2018-08-11
Dev account address provider master merged

feliam:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1047)


feliam:
ESultanik I made a PR from basically the same branch but against master so the diff is more tractable: https://github.com/trailofbits/manticore/pull/1054 

Let's keep the work there  
I'm closing this.

PR #1048, opened 2018-08-13
Serialization cleanup

defunctio:
serialization cleanup and more robust recursion failure handling

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1048)


manticore/core/smtlib/constraints.py
```diff
@@ -165,13 +167,14 @@ def declarations(self):
...
             try:
                 declarations.visit(a)
-            except BaseException:
-                # there recursion limit exceeded problem,
-                # try a slower, iterative solution
-                #logger.info('WARNING: using iterpickle to dump recursive expression')
-                #from utils import iterpickle
-                #file('recursive.pkl', 'w').write(iterpickle.dumps(a))
-                raise
+            except RuntimeError:
```
defunctio: note; this PR includes handling that is missing in 0.2.0 for expression declarations that exceed the max recursion

manticore/core/smtlib/constraints.py
```diff
@@ -165,13 +167,14 @@ def declarations(self):
...
+                # TODO: (defunct) move recursion management out of PickleSerializer
+                if sys.getrecursionlimit() >= PickleSerializer.MAX_RECURSION:
```
yan: Since these constants are being used for smtlib visitors and pickle serialization, should we move them outside of `PickleSerializer`? This might tie with the global config PR.

manticore/utils/helpers.py
```diff
@@ -137,3 +142,41 @@ def flush(self):
...
+        sys.setrecursionlimit(PickleSerializer.DEFAULT_RECURSION)
+
+    def serialize(self, state, f):
+        try:
+            f.write(pickle.dumps(state, 2))
+        except RuntimeError:
+            if sys.getrecursionlimit() >= PickleSerializer.MAX_RECURSION:
+                raise Exception(f'PickleSerializer recursion limit surpassed {PickleSerializer.MAX_RECURSION}, aborting')
+            logger.info(f'Recursion maximum {sys.getrecursionlimit()} hit, increasing')
+            sys.setrecursionlimit(sys.getrecursionlimit() + PickleSerializer.DEFAULT_RECURSION)
```
yan: Should this scale exponentially, not linearly? i.e. increase by a factor of 2 each time we hit it; like cpp's vector/strings grow in size.

yan:
😍 at all the red diffs

manticore/core/smtlib/constraints.py
defunctio: :100: , global config seems like the place for them to live. i'll open an issue to move them once global config is implemented.

re;#1058

PR #1054, opened 2018-08-14
Make account address provider more yellow

feliam:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1054)


manticore/platforms/evm.py
```diff
@@ -2103,6 +2154,13 @@ def _process_pending_transaction(self):
         if failed:
             self._close_transaction('TXERROR', rollback=True)
 
+        if caller not in self._world_state:
+            self._world_state[caller] = {}
+        if 'nonce' not in self._world_state[caller] or self._world_state[caller]['nonce'] is None:
+            self._world_state[caller]['nonce'] = 1
+        else:
+            self._world_state[caller]['nonce'] += 1
```
disconnect3d: Since we already access `[caller]['nonce']` then maybe:
```python
self._world_state[caller]['nonce'] = self._world_state[caller].get('nonce', 0) + 1

Comment (disconnect3d) [reply manticore/platforms/evm.py]: Is `caller not in self._world_state` the only situation when `'nonce'` key is not there?

If so we can do:
```python
        if caller not in self._world_state:
            self._world_state[caller] = {'nonce': 1}
        else:
            self._world_state[caller]['nonce'] += 1
```

manticore/utils/rlp.py
```diff
@@ -0,0 +1,83 @@
...
+
+def int_to_bytes(n):
+    return n.to_bytes((n.bit_length() + 7) // 8, 'big')
+
+def rlp_encode(item):
+    if item is None or item == 0:
+        return b'\x80'
+    elif isinstance(item, str):
+        return rlp_encode(item.encode('utf8'))
+    elif isinstance(item, bytearray) or isinstance(item, bytes):
```
disconnect3d: It can be just `elif isinstance(item (bytearray, bytes)):`

manticore/utils/rlp.py
```diff
@@ -0,0 +1,83 @@
...
+        if len(item) == 1 and item[0] < 0x80:
+            return item
+        else:
+            return encode_length(len(item), 0x80) + item
+    elif isinstance(item, collections.abc.Sequence):
+        output = b''.join(map(rlp_encode, item))
+        return encode_length(len(output), 0xC0) + output
+    elif isinstance(item, int) or isinstance(item, long):
```
disconnect3d: `elif isinstance(item, (int, long)):`

manticore/utils/rlp.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+
+import collections.abc
+
+def int_to_bytes(n):
+    return n.to_bytes((n.bit_length() + 7) // 8, 'big')
+
+def rlp_encode(item):
+    if item is None or item == 0:
```
disconnect3d: Can we get some docstring what is a `rlp` encoding?

manticore/utils/rlp.py
```diff
@@ -0,0 +1,83 @@
...
+            else:
+                ret.append(c)
+        return ret
+    def to_int(exp):
+        if isinstance(exp, str):
+            return ord(exp)
+        else:
+            return exp
+    LOREM_IPSUM = "Lorem ipsum dolor sit amet, consectetur adipisicing elit"
+    for item, expected in (
```
disconnect3d: Those tests could be written using [doctest](https://docs.python.org/3/library/doctest.html)

manticore/utils/rlp.py
```diff
@@ -0,0 +1,83 @@
...
+    elif isinstance(item, collections.abc.Sequence):
+        output = b''.join(map(rlp_encode, item))
+        return encode_length(len(output), 0xC0) + output
+    elif isinstance(item, int) or isinstance(item, long):
+        return rlp_encode(int_to_bytes(item))
+    else:
+        raise Exception("Cannot encode object of type %s" % type(item).__name__)
+
+def encode_length(length, offset):
+    if length < 56:
```
disconnect3d: Some docstring e.g. why the magic number 56?

ESultanik: It's all from the Ethereum RLP specification. I'll add a link to it.

manticore/platforms/evm.py
```diff
@@ -1977,7 +2025,11 @@ def create_contract(self, price=0, address=None, caller=None, balance=0, init=No
         This is done when the byte code in the init byte array is actually run
         on the network.
         '''
-        address = self.create_account(address)
+        expected_address = self.create_account(self.new_address(sender=caller))
+        if address is None:
+            address = expected_address
+        elif caller is not None and address != expected_address:
+            raise EVMException("Error: contract created from address %x with nonce %d was expected to be at address %x, but create_contract was called with address=%x" % (caller, self.get_nonce(caller), expected_address, address))
```
mossberg: nit, EthereumError here, instead of EVMException? since this is actually an error

manticore/ethereum/__init__.py
```diff
@@ -672,6 +672,24 @@ def solidity_create_contract(self, source_code, owner, name=None, contract_name=
             return None
         return contract_account
 
+    def get_nonce(self, address):
+        # type forgiveness:
+        address = int(address)
+        # get all states containing this address:
+        states = tuple(state for state in self.running_states if address in state.platform)
```
feliam: Note that this for read a state from storage and write them back in each iteration. So it is better if we can collapse this and the next for_all iteration thing

manticore/ethereum/__init__.py
```diff
@@ -672,6 +672,24 @@ def solidity_create_contract(self, source_code, owner, name=None, contract_name=
...
+            raise NoAliveStates("There are no alive states containing address %x" % address)
+        if len(states) == 1:
+            return states[0].platform.get_nonce(address)
+        # if there are multiple states with this address, ensure that they all currently have the same nonce:
+        existing_nonces = set(state.platform.get_nonce(address) for state in states)
+        if len(existing_nonces) != 1:
+            raise EthereumError("Cannot increase the nonce of address %x because it exists in multiple states with different nonces" % address)
+        ret = None
+        for state in states:
+            ret = state.platform.get_nonce(address)
```
disconnect3d: Is that fine that we iterate over many states but only take a result of last state?

Maybe we should just do `return states[-1].platform.get_nonce(address)`?
Or maybe we should have it as:
```python
for state in states:
    nonce = state.platform.get_nonce(address)
    if nonce is not None:
        return ret
```

manticore/ethereum/__init__.py
```diff
@@ -672,6 +672,24 @@ def solidity_create_contract(self, source_code, owner, name=None, contract_name=
...
+        return ret
```
feliam: Maybe just put the return inside the previous `for` (or a break) ?

manticore/ethereum/__init__.py
```diff
@@ -721,12 +739,20 @@ def _get_uniq_name(self, stem):
         assert name not in self.accounts
         return name
 
+    def _all_addresses(self):
+        """ Returns all addresses in all running states """
+        ret = set()
+        for state in self.running_states:
+            ret |= set(state.platform.accounts)
+        return ret
```
disconnect3d: Heh
```
In [9]: set().union([1,2,3],[4,5,6])
Out[9]: {1, 2, 3, 4, 5, 6}

In [10]: set().union(*(i for i in [[1,2,3],[4,5,6]]))
Out[10]: {1, 2, 3, 4, 5, 6}
```

It could be written as `return set().union(*(state.platform.accounts for state in self.running_states))`

manticore/ethereum/__init__.py
```diff
@@ -721,12 +739,20 @@ def _get_uniq_name(self, stem):
         assert name not in self.accounts
         return name
 
+    def _all_addresses(self):
+        """ Returns all addresses in all running states """
+        ret = set()
+        for state in self.running_states:
```
feliam: Just checking. Do we not care about addresses in states that ended?

manticore/platforms/evm.py
```diff
@@ -1648,6 +1649,7 @@ def _close_transaction(self, result, data=None, rollback=False):
             self._logs = logs
             self.send_funds(tx.address, tx.caller, tx.value)
 
+        self.increase_nonce(tx.caller)
```
feliam: Do nonces get incremented in rollbacked transactions? 
if not we need an `else:` there

manticore/platforms/evm.py
ESultanik: Up until my second-to-last-commit, this file wasn't even importing `EthereumError`, yet it was using it in a few places. It's currently that way in master! Our tests just never triggered those throws, so we never experienced the `EthereumError is not defined` error. But you're right, it should really be an `EthereumError`.

manticore/platforms/evm.py
ESultanik: In the real world, I think the transaction still gets mined even if a TXERROR occurs, right? If so, then the nonce should increment. I will double-check.

feliam: Failed Internal tx? Like internal tx with no enough funds ? Would those increment nonce?

ESultanik: Oh, a _contract's_ nonce is only ever incremented if it _creates_ another contract. Internal transactions should not increment the nonce.

feliam: Unless the internal tx create a contract I presume. 
What about account creation via CALL/SELFDESTRUCT to non existing accounts?
Ughh nonces are problematic.

manticore/ethereum/__init__.py
```diff
@@ -905,6 +927,10 @@ def _transaction(self, sort, caller, value=0, address=None, data=None, gaslimit=
             if '_pending_transaction' in state.context:
                 raise EthereumError("This is bad. It should not be a pending transaction")
 
+            # Choose an address here, because it will be dependent on the caller's nonce in this state
+            if address is None:
+                address = world.new_address(caller)
```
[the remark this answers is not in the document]
ESultanik: You're right. I think I can just move it after the call to `self._migrate_tx_expressions`, right?

manticore/ethereum/__init__.py
```diff
@@ -905,6 +927,10 @@ def _transaction(self, sort, caller, value=0, address=None, data=None, gaslimit=
...
+
             # Migrate any expression to state specific constraint set
             caller, address, value, data = self._migrate_tx_expressions(state, caller, address, value, data)
```
ESultanik: I just noticed that this line overwrites the variables `caller`, `address`, `value,`, and `data`. Won't that screw up the call to `self._migrate_tx_expressions` during the next iteration? This is code currently in `master`; I didn't write it. @feliam, should I change the variable names to preserve the originals for each iteration?

manticore/ethereum/__init__.py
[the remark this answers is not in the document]
feliam: Ouch I think not. Migrate only really rename the expressions just in case.

manticore/ethereum/__init__.py
feliam: OHhhhh yeeeees 
```caller_migrated, address_migrated, value_migrated, data_migrated = self._migrate_tx_expressions(state, caller, address, value, data)```

File: manticore/platforms/evm.py [hunk 16]
```diff
@@ -1921,12 +1933,26 @@ def tx_gasprice(self):
     def depth(self):
         return len(self._callstack)
 
-    def new_address(self):
+    def new_address(self, sender=None, nonce=None):
```
feliam: What if we get rid of nonce as an argument? A quick check did not get me any calls to this method with `nonce != None` (except from itself maybe)

manticore/platforms/evm.py
```diff
@@ -1968,7 +2012,11 @@ def create_account(self, address=None, balance=0, code='', storage=None, nonce=0
         return address
 
     def create_contract(self, price=0, address=None, caller=None, balance=0, init=None, gas=2300):
-        '''
+        ''' Create a contract account. Sends a transaction to initialize the contract
```
feliam: ~Does this even work with `caller = None`? ~ 
Caller must be defined in accounts. So I guess this is always called with a caller. (https://github.com/trailofbits/manticore/blob/2a4d62a2cc634a27c452841f420eb4e329532c32/manticore/platforms/evm.py#L2245-L2246)

So the question is why to have an `address` parameter if it is deterministic calculated using the caller.

manticore/platforms/evm.py
```diff
@@ -2103,6 +2154,12 @@ def _process_pending_transaction(self):
         if failed:
             self._close_transaction('TXERROR', rollback=True)
 
+        if caller not in self._world_state:
```
feliam: Got a little lost in the diff but the hypothesis is that this never happens, right?

manticore/platforms/evm.py
```diff
@@ -2070,7 +2121,7 @@ def _process_pending_transaction(self):
         self._pending_transaction_concretize_address()
         self._pending_transaction_concretize_caller()
         if caller not in self.accounts:
-            raise EVMException('Caller account does not exist')
+            raise EVMException("Caller account %x does not exist; valid accounts: %s" % (caller, list(map(hex, self.accounts))))
```
feliam: We are trying to move away from %x. Do that f-magic  `f"Caller account {hex(caller)} does not exist; valid accounts: {list(map(hex, self.accounts))}" `

manticore/platforms/evm.py
```diff
@@ -2103,6 +2154,12 @@ def _process_pending_transaction(self):
         if failed:
             self._close_transaction('TXERROR', rollback=True)
 
+        if caller not in self._world_state:
+            self._world_state[caller] = { 'nonce' : 1 }
+        elif sort == 'CREATE' or not self._world_state[caller]['code']:
```
feliam: I think you may want to move this to _open_transaction https://github.com/trailofbits/manticore/blob/2a4d62a2cc634a27c452841f420eb4e329532c32/manticore/platforms/evm.py#L1780

We need to do something to save the old nonce for the caller. Probably here https://github.com/trailofbits/manticore/blob/2a4d62a2cc634a27c452841f420eb4e329532c32/manticore/platforms/evm.py#L1803

So we can then rollback if the creation failed/reverted/throwed ?
https://github.com/trailofbits/manticore/blob/2a4d62a2cc634a27c452841f420eb4e329532c32/manticore/platforms/evm.py#L1815

manticore/utils/nointerrupt.py
```diff
@@ -11,7 +11,10 @@ def __init__(self, callback):
     def __enter__(self):
         self.signal_received = 0
         self.old_handler = signal.getsignal(signal.SIGINT)
-        signal.signal(signal.SIGINT, self.handler)
+        try:
+            signal.signal(signal.SIGINT, self.handler)
+        except ValueError as e:
```
feliam: Still necessary?

manticore/platforms/evm.py
ESultanik: A human in a script might call it, and they could choose to use `nonce=None` as a convenience.

manticore/platforms/evm.py
ESultanik: This is resolved by commit 25a2a2a307bb9cbd0e37582bd8fbdf46053cc377

manticore/platforms/evm.py
ESultanik: I am not sure about the rest of the downstream function calls, but `create_contract` itself should be fine if `caller` is `None`. Anyway, we have to keep `caller=None` in the function signature at this point because we have default arguments before, so the only way to remove the default value from `caller` would be to move its argument position earlier, which would break the API for any callers that call it with positional arguments.

manticore/ethereum/__init__.py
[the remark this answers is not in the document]
stephan-tolksdorf: The `_transaction` function has type checks and asserts that ensure that `address` can never be `None` here. Has maybe something been lost during a merge?

PR #1058, opened 2018-08-15
[WIP] Add basic global configuration support

### LINKED ISSUE #372: Feature: Add global configuration mechanism
Feature request.
Add a way for the different modules to register global configurations items like "solver query timeout".
This configs should be settable via a .ini file or via a command line argument like KLEE.

This .ini should be included in the analysis output so it's reproducible. It supersedes the "command.sh" at the current output.

### LINKED ISSUE #1015: Solver timeout interface
Manticore should have an interface for controlling the timeout used when submitting solver queries. This is distinct from manticore's overall timeout.

Right now, it's hardcoded in solver.py with the `-t` flag.

https://github.com/trailofbits/manticore/blob/90fdbb8753daf8aa84a22ef3789fe3933733ca51/manticore/core/smtlib/solver.py#L150

For my particular use case an API would be fine, but we should probably also expose a cli flag for it. Need some brainstorming on what kind of API.

defunctio:
Implement basic config for setting global variables.
- resolves #1015 
- resolves #372

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1058)

### DIFF
--- a/manticore/config.py
+++ b/manticore/config.py
@@ -1,19 +1,44 @@
 import configparser
 import os
+import copy
+
+#  These are not stored in __main__.py as they are used outside of the CLI
+defaults = {
+    'env': [],
+    'files': [],
+    'offset': 16,
+    'policy': 'random',
+    'procs': 1,
+    'timeout': 0,
+    'v': 1,
+    'txaccount': 'attacker',
+    'solver_timeout': 240000,
+    'cache_dict_max_size': 150000,
+    'cache_dict_flush_perc': 25
+}
 
 
-def _config(path=None) -> configparser.ConfigParser:
+def _config(path: str = None) -> configparser.ConfigParser:
+    """
+    Initialize the defaults and override with settings from configuration.
+    order of precedence;
+        path kwarg
+        ./.manticorerc
+        ${HOME}/.manticorerc
+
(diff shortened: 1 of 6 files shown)


manticore/config.py
```diff
@@ -0,0 +1,19 @@
+import configparser
+import os
+
+
+def _config(path=None) -> configparser.ConfigParser:
```
feliam: Mini docstring documenting the precedence?
This will read the the configuration from a .ini file in `path` or else the `./.manticorerc` or else `~/.manticorerc`

manticore/core/smtlib/solver.py
```diff
@@ -129,7 +130,8 @@ def __init__(self):
         super().__init__()
         self._proc = None
 
-        self._command = 'z3 -t:240000 -memory:16384 -smt2 -in'
+        self._command = \
+            f'z3 -t:{settings.getint("settings", "solver_timeout", fallback=240000)} -memory:16384 -smt2 -in'
```
feliam: If settings.solver_timeout was not defined and the fallback was used .. maybe it should be recorded in the .ini that will go to the workspace.

defunctio:
i'll rewrite it, i was just trying to be lazy

Issue #1008, opened 2018-07-31
Printing of the "Function call" field in the transaction files is failing


ggrieco-tob:
### OS / Environment

Distributor ID:	ArchLinux
Description:	ArchLinux
Release:	rolling-release

### Manticore version

Last revision (59a5ff0ba335d3c0409dbac7dfed3a6a225943bf)

### Python version

Python 3.6.6

### Summary of the problem

Manticore fails to serialize the Function call field in the  .tx files

### Step to reproduce the behavior

This is a solidity contract to reproduce this issue:

```solidity
contract C {
  function f(uint x) returns (uint) {
    return x;
  } 
}
```

You need to run manticore:

```
$ manticore --txlimit 1 test.sol
```

### Expected behavior

It should print correctly the names and arguments in the "Function call" field.

### Actual behavior

```
Transactions Nr. 0
Type: CREATE (0)
From: owner(0x5da83cffb5bab1cd888417a5ecefe37b9e250d67) 
To: contract0(0xca2410fda28718e5623a7a755531ae6dd30a28d2) 
Value: 0 
Gas used: 2300 
Data: b'608060405234801561001057600080fd5b5060b88061001f6000396000f300608060405260043610603f576000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff168063b3de648b146044575b600080fd5b348015604f57600080fd5b50606c600480360381019080803590602001909291905050506082565b6040518082815260200191505060405180910390f35b60008190509190505600a165627a7a7230582072a45d1e204f9caff202f38471ed656edf9a44967d1cd84f62a7640ca89da3c80029' 
Return_data: b'608060405260043610603f576000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff168063b3de648b146044575b600080fd5b348015604f57600080fd5b50606c600480360381019080803590602001909291905050506082565b6040518082815260200191505060405180910390f35b60008190509190505600a165627a7a7230582072a45d1e204f9caff202f38471ed656edf9a44967d1cd84f62a7640ca89da3c80029' (*)


Transactions Nr. 1
Type: CALL (0)
From: attacker(0xc6737b8b2a6a7b5fbb5d75b895f628f2922bae14) 
To: contract0(0xca2410fda28718e5623a7a755531ae6dd30a28d2) 
Value: 0 (*)
Gas used: 2300 
[...]

PR #1063, opened 2018-08-20
Fix for #1008

ggrieco-tob:
This is a very quick fix for the invalid printing of arguments in a function call.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1063)


yan:
Added a test to test this update, which passes, but calculations on cc don't get updated. Going ahead with squashing and merging.

PR #1064, opened 2018-08-21
Update the README

yan:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1064)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ $ manticore ./path/to/binary ++ ++  # use two symbolic strings of length two as 
 
 ### API
 
-Manticore has a Python programming interface which can be used to implement custom analyses and is capable of detailed verification of arbitrary properties of smart contracts via its Python API. Set starting conditions, identify symbolic transactions, then review discovered states to ensure invariants for your contract hold.
+Manticore has a Python programming interface which can be used to implement custom analyses. For Ethereum smart contracts, can be used for detailed verification of arbitrary contract properties. Set starting conditions, execute symbolic transactions, then review discovered states to ensure invariants for your contract hold.
 
 ```python
 from manticore.ethereum import ManticoreEVM
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ $ manticore ./path/to/binary ++ ++  # use two symbolic strings of length two as 
 
 Manticore has a Python programming interface which can be used to implement custom analyses.
 
-For Ethereum smart contracts, can be used for detailed verification of arbitrary contract properties. Set starting conditions, execute symbolic transactions, then review discovered states to ensure invariants for your contract hold.
+For Ethereum smart contracts, it can be used for detailed verification of arbitrary contract properties. Set starting conditions, execute symbolic transactions, then review discovered states to ensure invariants for your contract hold.
 
 ```python
 from manticore.ethereum import ManticoreEVM
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Manticore can analyze the following types of programs:
 
 Manticore has a command line interface which can be used to easily symbolically execute a supported program or smart contract. Analysis results will be placed into a new directory beginning with `mcore_`.
 
-Use the CLI to explore possible states in Ethereum smart contracts. Manticore includes _detectors_ which flag certain conditions, including known vulnerable code, as it explores possible states. Solidity smart contracts must have a `.sol` extension for analysis by Manticore. See a [demo](https://asciinema.org/a/154012).
+Use the CLI to explore possible states in Ethereum smart contracts. Manticore includes _detectors_ that flag potentially vulnerable code in discovered states. Solidity smart contracts must have a `.sol` extension for analysis by Manticore. See a [demo](https://asciinema.org/a/154012).
 
 ```
 $ manticore ./path/to/contract.sol  # runs, and creates a mcore_* directory with analysis results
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,14 +25,26 @@ Manticore is a symbolic execution tool for analysis of binaries and smart contra
 
 Manticore can analyze the following types of programs:
 
-- Linux ELF binaries (x86, x86_64 and ARMv7)
 - Ethereum smart contracts (EVM bytecode)
+- Linux ELF binaries (x86, x86_64 and ARMv7)
+
+Manticore includes a symbolic Ethereum Virtual Machine (EVM) and a convenient interface for automated compilation and analysis of Solidity. It integrates with [Ethersplay](https://github.com/trailofbits/ethersplay), Trail of Bits’ visual disassembler for EVM bytecode, for analysis visualization.
```
mossberg: I feel like this is a little too much detail for this part of the readme, but i'm not sure where it would go. maybe try to incorporate the content into the below sections somehow?

"Use the CLI to conveniently explore possible states in Solidity smart contracts using Manticore's symbolic Ethereum Virtual Machine." <- that would take care of the first sentence. not sure where we would put the second one about ethersplay (or if it really even belongs here). i feel like it is a better fit for a wiki page or something, and not the readme.

dguido:
I think Getting Help maybe could get merged with Documentation at the bottom?

For the Ethereum API example, could we use a real-er example? It's hard for me to understand the value of tracking foo and bar in a dummy contract. Maybe the same for the Linux API example.

README.md
```diff
@@ -42,7 +52,36 @@ $ manticore ./path/to/binary ++ ++  # use two symbolic strings of length two as
...
+user_account = m.create_account(balance=1000)
+contract_account = m.solidity_create_contract(contract_src,
+                                              owner=user_account,
+                                              balance=0)
+value = m.make_symbolic_value()
+
+contract_account.incremented(value)
+
+for state in m.running_states:
+    print(f"can value be 1? {state.can_be_true(value == 1)}")
```
mossberg: i think we might want to not use a f-str here. imo this gets into hard to read territory due how it's a complex expression, rather than just a simple variable name. also i expect many users will not exactly be familiar with this new python feature, which can slightly interfere with comprehension.

README.md
```diff
@@ -42,7 +52,36 @@ $ manticore ./path/to/binary ++ ++  # use two symbolic strings of length two as
 
 ### API
 
-Manticore has a Python programming interface which can be used to implement custom analyses.
+Manticore has a Python programming interface which can be used to implement custom analyses and is capable of detailed verification of arbitrary properties of smart contracts via its Python API. Set starting conditions, identify symbolic transactions, then review discovered states to ensure invariants for your contract hold.
+
+```python
+from manticore.ethereum import ManticoreEVM
```
mossberg: i also think we should change the contract we use here. i think the contract here should, as concisely as possible, showcase the features to

- execute contract functions with symbolic input
- execute arbitrary sequences of contract functions
- peruse the discovered states, looking for states that match a certain criteria (e.g. had an exception)

I think a good candidate for this would be a contract based on the umd example, but slightly simplified, as this example is a bit long. what i like about this example is that it's not immediately obvious from a human perspective how to cause the assertion to fail, but it's trivial to use manticore to find out the answer. i think we should come up with a slightly similar example in the same vein and use it here. https://github.com/trailofbits/manticore/blob/master/examples/evm/umd_example.sol

just putting these notes here; this is pretty low priority work and we should probably focus on other things for the moment

dguido: I don't like the UMD example because it's too "academic." I don't want a demo of features, I want to see how Manticore helps me solve a problem or achieve a goal.

README.md
```diff
@@ -42,7 +52,36 @@ $ manticore ./path/to/binary ++ ++  # use two symbolic strings of length two as
 
 ### API
 
-Manticore has a Python programming interface which can be used to implement custom analyses.
+Manticore has a Python programming interface which can be used to implement custom analyses and is capable of detailed verification of arbitrary properties of smart contracts via its Python API. Set starting conditions, identify symbolic transactions, then review discovered states to ensure invariants for your contract hold.
```
mossberg: this is a runon sentence, and also has redundancy w.r.t "Python programming interface" / "Python API". i'd like to keep "Manticore has a Python programming interface which can be used to implement custom analyses." as a standalone platform-agnostic first sentence of the intro

maybe something like this?

```
Manticore has a Python programming interface which can be used to implement custom analyses.	

For Ethereum smart contracts, it can be used for detailed verification of arbitrary contract properties. Set starting conditions, *execute* symbolic transactions, then review discovered states to ensure invariants for your contract hold.
```

note: changed identify to execute since i don't understand what "identify symbolic transactions" means.

README.md
```diff
@@ -25,14 +25,24 @@ Manticore is a symbolic execution tool for analysis of binaries and smart contra
...
+- Linux ELF binaries (x86, x86_64 and ARMv7)
 
 ## Usage
 
 ### CLI
 
-Manticore has a command line interface which can be used to easily symbolically execute a supported program. Analysis results will be placed into a new directory beginning with `mcore_`.
+Manticore has a command line interface which can be used to easily symbolically execute a supported program or smart contract. Analysis results will be placed into a new directory beginning with `mcore_`.
+
+Use the CLI to explore possible states in Ethereum smart contracts. Manticore includes _detectors_ which flag certain conditions, including known vulnerable code, as it explores possible states. Solidity smart contracts must have a `.sol` extension for analysis by Manticore. See a [demo](https://asciinema.org/a/154012).
```
mossberg: i know was just copied and pasted, but while we're here... i think "which flag certain conditions, including known vulnerable code" sounds kind of weird ( i think i wrote it :x ).

maybe this?

```
Manticore includes _detectors_ which flag potentially vulnerable code that is detected in an explored state.
```

README.md
```diff
@@ -42,7 +52,36 @@ $ manticore ./path/to/binary ++ ++  # use two symbolic strings of length two as
 
 ### API
 
-Manticore has a Python programming interface which can be used to implement custom analyses.
+Manticore has a Python programming interface which can be used to implement custom analyses. For Ethereum smart contracts, can be used for detailed verification of arbitrary contract properties. Set starting conditions, execute symbolic transactions, then review discovered states to ensure invariants for your contract hold.
```
mossberg: paragraph break at the end of first sentence?

README.md
```diff
@@ -44,6 +54,37 @@ $ manticore ./path/to/binary ++ ++  # use two symbolic strings of length two as
 
 Manticore has a Python programming interface which can be used to implement custom analyses.
 
+For Ethereum smart contracts, can be used for detailed verification of arbitrary contract properties. Set starting conditions, execute symbolic transactions, then review discovered states to ensure invariants for your contract hold.
```
mossberg: "can be used"

PR #1077, opened 2018-08-24
eth: add ether leak detector

mossberg:
this adds a detector ether leaks, and slightly refactors the detector tests

- make the LoopDepthLimiter plugin configurable (change the default to 5! the some for reason etherleak_pos_argument1.sol test needs 5, not 2 to work...)

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1077)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -253,6 +253,8 @@ class DetectEtherLeak(Detector):
                 if state.can_be_true(msg_sender == dest_address):
                     self.add_finding_here(state, "Reachable ether leak to sender via argument")
                 else:
+                    # This might be a false positive if the dest_address can't actually be solved to anything
+                    # useful/exploitable
                     self.add_finding_here(state, "Reachable ether leak to user controlled address via argument")
             else:
                 if msg_sender == dest_address:
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -249,17 +249,14 @@ class DetectEtherLeak(Detector):
                 return
 
             if issymbolic(dest_address):
-                if state.can_be_true(msg_sender == dest_address)
-                    self.add_finding_here(state, "Reachable ether leak to sender")
+                # We assume dest_address is symbolic because it came from symbolic tx data (user input argument)
+                if state.can_be_true(msg_sender == dest_address):
+                    self.add_finding_here(state, "Reachable ether leak to sender via argument")
                 else:
-                    # If dest_address is symbolic, but can't be set to the msg sender, we assume it's
-                    # symbolic because it came from the tx data (user input)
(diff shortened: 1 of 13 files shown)


manticore/ethereum.py
```diff
@@ -238,6 +242,30 @@ def will_evm_execute_instruction_callback(self, state, instruction, arguments):
...
+        if instruction.semantics == 'CALL':
+            dest_address = arguments[1]
+            sent_value = arguments[2]
+            msg_sender = state.platform.current_vm.caller
+
+            # If nothing can be transferred out, we don't care
+            if not state.can_be_true(sent_value > 0):
+                return
+
+            if issymbolic(dest_address):
```
feliam: This is good enough. Just as fyi what I do is to check that the symbolic value effectivelly has at least 2 solutions. slower.
```
            if issymbolic(address):
                possible_addresses = state.solve_n(address, 2)
                if len(possible_addresses) > 1:
                    self.add_finding_here(state, "Dellegatecall to user controlled address")
```
The question is: What's user controlled?

manticore/ethereum.py
```diff
@@ -238,6 +242,30 @@ def will_evm_execute_instruction_callback(self, state, instruction, arguments):
             self.add_finding_here(state, 'Reachable SELFDESTRUCT')
 
 
+class DetectEtherLeak(Detector):
+    def will_evm_execute_instruction_callback(self, state, instruction, arguments):
```
feliam: Should this be .. will _open_transaction..or should this include CREATE/DELEGATECALL?

manticore/ethereum.py
mossberg: it's slower, but this is not hot code, so it should be fine. i like it.

for the purposes here i think we can define user controlled as : symbolic address

manticore/ethereum.py
mossberg: i had detecting `address.transfer()`s in mind when writing this. those always use calls. however i would guess that you're probably right that attackers could alternatively use other calls to steal ether. i think delegatecall could be plausible, less sure for create haha, that sounds pretty funky

manticore/ethereum.py
feliam: `
x = input()
require (x == concrete)
do_stuff(x)   <<<<  x is concrete or symbolic/ x is controlled or not?
`

mossberg: uh i see. you're right, in this case, i could consider it to be effectively concrete. reworking

mossberg:
since this is good enough, i will merge this after tests pass. i've created a new branch https://github.com/trailofbits/manticore/tree/dev-ether-leak-improvements where we can do a second iteration on 

- robustness if the symbolic destination is effectively concrete
- considering create/delegatecall also

PR #1078, opened 2018-08-24
Added --txnoether option to avoid sending ether to contracts

ggrieco-tob:
The majority of the contracts we analyze everyday do not receive any ether (e.g. vanilla ERC20 ones). If Manticore tries to send ether to them, it will only generate useless test cases and slow down the analysis. That's why I believe it is useful to add an optional flag to avoid manticore to send ether when it is performing the symbolic execution of the transactions.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1078)

### DIFF
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -70,6 +70,9 @@ def parse_arguments():
     parser.add_argument('--txnocoverage', action='store_true',
                         help='Do not use coverage as stopping criteria (Ethereum only)')
 
+    parser.add_argument('--txnoether', action='store_true',
+                        help='Do not send ether to contract (Ethereum only)')
+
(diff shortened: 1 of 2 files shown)


manticore/__main__.py
```diff
@@ -70,6 +70,9 @@ def positive(value):
     parser.add_argument('--txnocoverage', action='store_true',
                         help='Do not use coverage as stopping criteria (Ethereum only)')
 
+    parser.add_argument('--txnoether', action='store_true',
+                        help='Do not send ether to contract (Ethereum only)')
```
feliam: "Do not attempt to send ether to contract (Ethereum only)" ?

PR #1082, opened 2018-08-24
eth: new/alternative reentrancy detector

mossberg:


manticore/ethereum.py
```diff
@@ -288,6 +288,47 @@ def will_evm_execute_instruction_callback(self, state, instruction, arguments):
...
+            pc = state.platform.current_vm.pc
+
+            if not state.can_be_true(gas > 2300):
+                return
+
+            # flag any external call that's going to a symbolic (potentially user controlled)?
+            # potentially bc, it could solve to only 1 address, it could solve to nothing userful
+            # or concretely the sender's address
+            if issymbolic(dest_address):
+                state.context.get(self.LOCS, []).append(pc)
```
feliam: I think we should also save "gas >2300" with the pc.
`... append((pc, gas>2300)) `

manticore/ethereum.py
```diff
@@ -288,6 +288,47 @@ def will_evm_execute_instruction_callback(self, state, instruction, arguments):
...
+            else:
+                if msg_sender == dest_address:
+                    state.context.get(self.LOCS, []).append(pc)
+
+    def did_evm_write_storage_callback(self, state, address, offset, value):
+        locs = state.context.get(self.LOCS, [])
+
+        # if we're here and locs has stuff in it. by definition this state has
+        # encountered a dangerous call and is now at a write.
+        for callpc in locs:
```
feliam: Here you should import the gas cond

manticore/ethereum.py
```diff
@@ -288,6 +288,47 @@ def will_evm_execute_instruction_callback(self, state, instruction, arguments):
...
+            # flag any external call that's going to a symbolic (potentially user controlled)?
+            # potentially bc, it could solve to only 1 address, it could solve to nothing userful
+            # or concretely the sender's address
+            if issymbolic(dest_address):
+                state.context.get(self.LOCS, []).append(pc)
+            else:
+                if msg_sender == dest_address:
+                    state.context.get(self.LOCS, []).append(pc)
+
+    def did_evm_write_storage_callback(self, state, address, offset, value):
```
feliam: What if the stored value is then never used? (this thing happened in tha benchmark)

manticore/ethereum.py
```diff
@@ -288,6 +288,47 @@ def will_evm_execute_instruction_callback(self, state, instruction, arguments):
...
+            state.context[self.LOCS] = []
+
+    def will_evm_execute_instruction_callback(self, state, instruction, arguments):
+        if instruction.semantics == 'CALL':
+            gas = arguments[0]
+            dest_address = arguments[1]
+            msg_sender = state.platform.current_vm.caller
+            pc = state.platform.current_vm.pc
+
+            if not state.can_be_true(gas > 2300):
```
feliam: Can we check the dest is a contract?
Can we check there is enough ether for the call ?

feliam:
___
*[manticore/ethereum.py, line 309 at r1](https://reviewable.io/reviews/trailofbits/manticore/1082#-LKhPZGnQOuU_ianRfBB-r1-309:-LL1CewW7KgkNLVMwqvD:bkv8bc7) ([raw file](https://github.com/trailofbits/manticore/blob/3c764a0e8cbaa6feba47b7aa2d2f356e15489d6d/manticore/ethereum.py#L309)):*

> Can we check there is enough ether for the call ?

doesn't the 2300 check do that?


</blockquote></details>

That's for the gas. I mean that the sender has enough ether value this tx is trying to send.

feliam:
___
*[manticore/ethereum.py, line 309 at r1](https://reviewable.io/reviews/trailofbits/manticore/1082#-LKhPZGnQOuU_ianRfBB-r1-309:-LL6jp2J6-5kv2hnlYKU:b-3h86cc) ([raw file](https://github.com/trailofbits/manticore/blob/3c764a0e8cbaa6feba47b7aa2d2f356e15489d6d/manticore/ethereum.py#L309)):*

after thinking about it a little more, i'm not sure it's worth doing those additional checks here. checking if the dest is a contract would be in an effort to reduce false positives, but i think it might be too strict. the cli's default configuration has the attacker who is sending tx created as a "normal" account, so it could cause false negatives with the cli.

also implementing these checks is complicated by the fact that the destination address can be symbolic, which makes it hard to tell if the destination is a contract, and if there is enough ether.

i might just be being lazy, but i think this detector is 'good enough' for the moment and gives a practical best effort at detecting reentrancy-looking situations.
</blockquote></details>

pfff for sure.
Note that the address in did_execute is concrete.

mossberg:
should this new detector replace the old one as the default used in the cli? i feel like the older one has limited usefulness since it requires a concrete list of bad/attacker addresses. 

also on this line, is `and not world.get_code(tx.address)` a bug? it's checking if there is not code at the address, i.e.g that it's a human? shouldn't it be checking it's a contract?

https://github.com/trailofbits/manticore/blob/15b2257b97234680700353ffb9ec6895c70833d1/manticore/ethereum.py#L323

PR #1084, opened 2018-08-24
Multiple send detector

yan:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1084)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -570,6 +570,42 @@ class DetectUninitializedStorage(Detector):
         state.context.setdefault('{:s}.initialized_storage'.format(self.name), set()).add((address, offset))
 
 
+class DetectMultipleSends(Detector):
+    '''
+    Detects multiple sends to the same address within a transaction
+    '''
+    def did_evm_execute_instruction_callback(self, state, instruction, arguments, result_ref):
+        mnemonic = instruction.semantics
+        #if mnemonic == 'CALL':
+            #calls = state.context.get(f'{self.name}.calls', {})
+            #gas, dest, *rest = arguments
+            #if dest in calls:
+                #pass
+                ##print(f"!!! {id(state)} Multiple calls to same addr!")
+            #else:
+                #state.context.setdefault(f'{self.name}.calls', {})[dest] = rest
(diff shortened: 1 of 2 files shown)


manticore/ethereum.py
```diff
@@ -631,6 +631,26 @@ def did_evm_write_storage_callback(self, state, address, offset, value):
         state.context.setdefault('{:s}.initialized_storage'.format(self.name), set()).add((address, offset))
 
 
+class DetectMultipleSends(Detector):
+    '''
+    Detects multiple sends to the same address within a transaction
+    '''
+    def will_evm_execute_instruction_callback(self, state, instruction, arguments):
```
feliam: Should we do this after the fact and only if the transaction was successful?
Also not sure if it is important for the current goal but you can send value with CREATE, and SELFDESTRUCT :) and DELEGATECALL 
all open_transactions + delegatecall

manticore/ethereum.py
```diff
@@ -631,6 +631,26 @@ def did_evm_write_storage_callback(self, state, address, offset, value):
         state.context.setdefault('{:s}.initialized_storage'.format(self.name), set()).add((address, offset))
 
 
+class DetectMultipleSends(Detector):
+    '''
+    Detects multiple sends to the same address within a transaction
```
mossberg: I actually don't think that it's particularly a problem if the transactions are to the same address. My understanding is that it's just a general not-great practice to have more than one transaction initiated (.send, .transfer, .call, ...) within a function

mossberg:
I am a little mixed on if we should ultimately include this detector 🤔 . Our detectors can be seen as detecting findings on spectrum from "maybe bug" to "just a code quality issue", and i think this falls very far on the code quality issue side of things, maybe almost to the point where we don't need it. i don't really think manticore is the tool for finding these kinds of code quality issues, i think slither would probably be a better fit, anyway?

mossberg:
confirmed that slither can provide these capabilities if desired. also confirmed that we probably don't want to warn on this general multiple sends scenario in all cases, since it is not directly representative of a finding or issue. going to close

PR #1100, opened 2018-08-30
Update FPSE-like memory model

yan:


disconnect3d:
___
*[.travis.yml, line 18 at r1](https://reviewable.io/reviews/trailofbits/manticore/1100#-LLaJum05gasb00v0HH5:-LLaJum05gasb00v0HH6:b-lgkq6w) ([raw file](https://github.com/trailofbits/manticore/blob/5a5325a128edd3438e80ec84bb8cd9ed842aca46/.travis.yml#L18)):*
> ```
> PYTHONWARNINGS="ignore::ResourceWarning" # Ignore unclosed file warnings
> ```
We should rather focus on closing them instead of ignoring the warning :(

yan:
Going to merge this into fpse-mem and just use that as the canonical branch; no reason to have them both exist now that this one has matured.

PR #1117, opened 2018-09-02
Fix b

yeti-detective:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1117)

### DIFF
--- a/manticore/ethereum.py
+++ b/manticore/ethereum.py
@@ -2570,6 +2570,12 @@ class ManticoreEVM(Manticore):
                 summary.write('\n\n(*) Example solution given. Value is symbolic and may take other values\n')
 
         # Transactions
+        def format_for_b(data):
+            if (data[0:2] == "b'"):
+                return data[2:len(data)]
+            else:
+                return data
+
         with testcase.open_stream('tx') as tx_summary:
             is_something_symbolic = False
             for tx in blockchain.human_transactions:  # external transactions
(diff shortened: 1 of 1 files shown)


disconnect3d:
@yeti-detective What is the case in here?

yan:
@yeti-detective Thank you for working on this issue! However, I'm not sure that editing the representation of that data is the right approach here. This is likely as simple as calling `.decode('utf-8')` on the output of `hexlify`, and potentially prepending a `0x`.

Would be glad to merge once it's implemented that way.

P.S., and this is totally optional, once here, it might make sense to rewrite `%`-style string expansion to Python3's `.`format`.

yeti-detective:
Hi, thanks for the feedback. I'll update this by tonight.

yeti-detective:
I've added the call to .decode('utf-8') and changed the string formatting to the new `"{0}".format('syntax')`

Also I've only made changes with regard to writing to tx files, since that was the original issue. I can go through the ethereum.py file & update other writes also.

manticore/ethereum.py
```diff
@@ -2586,10 +2587,10 @@ def flagged(flag):
                 tx_summary.write("Value: %d %s\n" % (state.solve_one(tx.value), flagged(issymbolic(tx.value))))
                 tx_summary.write("Gas used: %d %s\n" % (state.solve_one(tx.gas), flagged(issymbolic(tx.gas))))
                 tx_data = state.solve_one(tx.data)
-                tx_summary.write("Data: %s %s\n" % (binascii.hexlify(tx_data), flagged(issymbolic(tx.data))))
+                tx_summary.write("Data: {0}{1} {2}\n".format('0x', binascii.hexlify(tx_data).decode('utf-8'), flagged(issymbolic(tx.data))))
```
disconnect3d: The `'0x'` can be moved into docstring.

yeti-detective: This should be addressed by commit [3b16a06](https://github.com/trailofbits/manticore/pull/1117/commits/3b16a0613ecc3d925810f654fa28aca81824055f).

PR #1119, opened 2018-09-04
[WIP] Check that a contract exists before calling it

dguido:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1119)


dguido:
Probably call it "nonexistantcall"?

Issue #1123, opened 2018-09-05
Manticore may fail to work when solc was installed from snap


disconnect3d:
I create this issue so other people won't waste as much time as I did and so I can link to here from the README.

---

## The problem

When `solc` has been installed through [`snap`](https://snapcraft.io/) universal Linux packages, it is sandboxed by default and so won't work with files located in `/tmp`. This makes Manticore API not working properly as it may launch `solc` over a temporarily created file.

So, e.g. when being in `/tmp` we get something like this:
```bash
$ cat test.sol
contract C {}
$ solc test.sol
""test.sol"" is not found
```

While the same thing will work in other 'standard' directories.

### Solutions

There are three solutions to solve this issue:

1. Remove `solc` snap package by `snap disable solc && snap remove solc` and install it e.g. from source.
2. Install `solc` under snap developer mode **but you do it on your own risk**: `snap install --devmode solc`
3. Configure `solc` snap package so it can use `/tmp` directory. **If someone would do that, reply here, so we will add info how to do that to this issue.**

### Some interesting read

* https://forum.snapcraft.io/t/security-policy-and-sandboxing/554

disconnect3d:
With the #1124 changes we detect that the compilation might have been broken because of `/tmp` and `snap` installation and inform user about that.

PR #1124, opened 2018-09-05
Don't recommend snap solc install in README

disconnect3d:
Due to the issues described in #1123.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1124)


yan:
This is likely too low level to include in the README as is. One possible solution: update the error message with instructions on how to re-install or fix the installation.

disconnect3d:
@yan I have changed it a bit so pls revisit.

yan:
@disconnect3d I updated the phrasing a bit; let me know if this works. Good to merge if so.

PR #1126, opened 2018-09-06
Extra verbose debug trace generator plugin

feliam:
Small debug plugin that prints a verbose trace for each state.

### DIFF
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -121,6 +121,9 @@ def parse_arguments():
     parser.add_argument('--no-testcases', action='store_true',
                         help='Do not generate testcases for discovered states when analysis finishes (Ethereum only)')
 
+    parser.add_argument('--verbose-trace', action='store_true',
+                        help='Dump an extra vervose trace for each state (Ethereum only)')
+
     parsed = parser.parse_args(sys.argv[1:])
     if parsed.procs <= 0:
         parsed.procs = 1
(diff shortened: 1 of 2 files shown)


manticore/ethereum.py
```diff
@@ -2455,6 +2469,14 @@ def _generate_testcase_callback(self, state, name, message=''):
         def flagged(flag):
             return '(*)' if flag else ''
         testcase = self._output.testcase(name.replace(' ', '_'))
+
+        # call on_finalize for each plugin
+        for plugin in self.plugins:
+            try:
+                plugin.on_finalize(state, testcase)
+            except AttributeError:
```
disconnect3d: We shouldn't use a try except here as if the exception is thrown inside `on_finalize` we may not notice it and loose time debugging.

We should either assign method to variable and have try except over that (and call the method after the try except blocks) or use hasattr or getattr (with an empty function as a default value).

manticore/ethereum.py
```diff
@@ -238,6 +238,20 @@ def _get_src(self, address, pc):
         return self.manticore.get_metadata(address).get_source_for(pc)
 
 
+class VerboseTrace(Plugin):
+    def will_evm_execute_instruction_callback(self, state, instruction, arguments):
+        world = state.platform
+        mnemonic = instruction.semantics
```
disconnect3d: There are redundant vars here

manticore/ethereum.py
```diff
@@ -238,6 +238,20 @@ def _get_src(self, address, pc):
...
+        current_vm = world.current_vm
+        str_trace = state.context.get('str_trace',[])
+        str_trace.append(str(current_vm))
+        state.context['str_trace'] = str_trace
+
+    def on_finalize(self, state, testcase):
+        with testcase.open_stream('str_trace') as str_trace_f:
+            str_trace_f.write( '\n'.join(state.context.get('str_trace',[])))
```
disconnect3d: Formatting (no space before `'\n'` and a space before `[]`)?

manticore/__main__.py
```diff
@@ -121,6 +121,9 @@ def positive(value):
     parser.add_argument('--no-testcases', action='store_true',
                         help='Do not generate testcases for discovered states when analysis finishes (Ethereum only)')
 
+    parser.add_argument('--verbose-trace', action='store_true',
+                        help='Dump an extra vervose trace for each state (Ethereum only)')
```
feliam: typo

manticore/ethereum.py
```diff
@@ -238,6 +238,19 @@ def _get_src(self, address, pc):
         return self.manticore.get_metadata(address).get_source_for(pc)
 
 
+class VerboseTrace(Plugin):
+    def will_evm_execute_instruction_callback(self, state, instruction, arguments):
+        world = state.platform
+        current_vm = world.current_vm
+        str_trace = state.context.get('str_trace', [])
+        str_trace.append(str(current_vm))
+        state.context['str_trace'] = str_trace
```
disconnect3d: Those 3 lines can be changed to: `state.setdefault('str_trace', []).append(str(current_vm))`

PR #1127, opened 2018-09-06
Implement FPSE-like memory model

yan:
This PR tracks the implementation of the fpse-like memory model.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1127)


manticore/core/cpu/abstractcpu.py
```diff
@@ -10,10 +10,15 @@
...
 from .disasm import init_disassembler
 from ..smtlib import BitVec, Operators, Constant
-from ..memory import ConcretizeMemory, InvalidMemoryAccess
+from ..memory import ConcretizeMemory, InvalidMemoryAccess, LazySMemory
 from ...utils.helpers import issymbolic
 from ...utils.emulate import UnicornEmulator
 from ...utils.event import Eventful
+from ..smtlib import visitors
+
+# tmp imports
```
disconnect3d: Hmm? :P

manticore/core/cpu/abstractcpu.py
```diff
@@ -582,7 +587,7 @@ def write_int(self, where, expression, size=None, force=False):
         '''
         if size is None:
             size = self.address_bit_size
-        assert size in SANE_SIZES
+        # assert size in SANE_SIZES
```
disconnect3d: Either remove or bring back

manticore/core/cpu/abstractcpu.py
```diff
@@ -602,7 +607,7 @@ def read_int(self, where, size=None, force=False):
...
```
disconnect3d: Either remove or bring back

manticore/core/cpu/abstractcpu.py
```diff
@@ -759,18 +764,32 @@ def decode_instruction(self, pc):
             if not self.memory.access_ok(address, 'x'):
                 break
 
+            # import ipdb
+            # ipdb.set_trace()
```
disconnect3d: To be removed

manticore/core/memory.py
```diff
@@ -158,6 +160,13 @@ def __repr__(self):
         '''
         return '<%s 0x%016x-0x%016x %s>' % (self.__class__.__name__, self.start, self.end, self.perms)
 
+    def __iter__(self):
+        """
+        Iterate all valid addresses
+        :return:
```
disconnect3d: `:return: iterator` ? or just remove?

manticore/core/memory.py
```diff
@@ -881,6 +936,17 @@ def __getitem__(self, index):
             result = self.read(index, 1)[0]
         return result
 
+    def __iter__(self):
+        """
+        Iterate all valid addresses
+        :return:
```
disconnect3d: `:return: iterator` or just remove this line?

manticore/core/memory.py
```diff
@@ -1071,6 +1137,210 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+
+    def __setstate__(self, state):
+        self.backing_array = state['backing_array']
+        self.backed_by_symbolic_store = state['backed_by_symbolic_store']
+
+    def mmapFile(self, addr, size, perms, filename, offset=0):
+        '''
+        Creates a new file mapping in the memory address space.
+
+        :param addr: the starting address (took as hint). If C{addr} is C{0} the first big enough
```
disconnect3d: What do we mean by `C{addr}` or `C{0}` and so on?

manticore/core/memory.py
```diff
@@ -1071,6 +1137,210 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+        fdata = fdata.ljust(size, b'\0')
+
+        for i in range(size):
+            Memory.write(self, addr + i, chr(fdata[i]), force=True)
+
+        logger.debug('New file-memory map @%x size:%x', addr, size)
+
+        return addr
+
+    def _deref_can_succeed(self, map, address, size):
```
disconnect3d: shadows builtin `map`, might be good to change to `map_`

manticore/core/memory.py
```diff
@@ -1071,6 +1137,210 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+        if not issymbolic(address):
+            return address >= map.start and address + size < map.end
+        else:
+            # FIXME is address + size < map.end a bug? address + size is first oob of the access. address + size <= map.end?
```
disconnect3d: What is the bug here?

manticore/core/memory.py
```diff
@@ -1071,6 +1137,210 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+
+                self.backing_array[addr] = Memory.read(self, addr, 1)[0]
+                self.backed_by_symbolic_store.add(addr)
+
+    def _map_deref_expr(self, map, address, size):
+        return Operators.AND(
+            # address >= map.start,
+            # address + size < map.end)
+            Operators.UGE(address, map.start),
+            Operators.ULT(address, map.end))
```
disconnect3d: Shouldn't it be `address + size`? We don't use `size` argument at all

manticore/core/memory.py
```diff
@@ -1071,6 +1137,210 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+        valid = functools.reduce(Operators.OR, expressions)
+
+        return valid
+
+    def invalid_ptr(self, address):
+        return Operators.NOT(self.valid_ptr(address))
+
+    def read(self, address, size, force=False):
+
+        # if address is symbolic, self.valid_ptr(address) ? ?
```
disconnect3d: ? Remove or more explanation?

manticore/core/memory.py
```diff
@@ -1071,6 +1137,210 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+        access_min, access_max = self._reachable_range(address, size)
+
+        if issymbolic(address):
+            self._import_concrete_memory(access_min, access_max)
+
+        retvals = []
+        addrs_to_access = [address + i for i in range(size)]
+        for addr in addrs_to_access:
```
disconnect3d: Or just:
```python
# accessing / reading memory
for addr in range(address, address+size):
```

manticore/core/memory.py
```diff
@@ -1071,6 +1137,210 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+
+            retvals.append(val)
+
+        return retvals
+
+    def write(self, address, value, force=False):
+
+        size = len(value)
+
+        addrs_to_access = [address + i for i in range(size)]
```
disconnect3d: Or just `list(range(address, address+size))`

manticore/core/memory.py
```diff
@@ -1071,6 +1137,210 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+        # if address is symbolic, self.valid_ptr(address) ? ?
+        access_min, access_max = self._reachable_range(address, size)
+
+        if issymbolic(address):
+            self._import_concrete_memory(access_min, access_max)
+
+        retvals = []
+        addrs_to_access = [address + i for i in range(size)]
+        for addr in addrs_to_access:
+            from_array = False
```
disconnect3d: This means whether memory (address) is backed by a an array of symbolic values? Or something different?

The `if ... elif` is a bit unclear and it might be good to add some explanation here as a comment, e.g. why do we care if given memory has write permission?

manticore/core/memory.py
```diff
@@ -1071,6 +1137,210 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+        """
+
+        :param bytes data_to_find:
+        :return:
+        """
+
+        # TODO: for the moment we just treat symbolic bytes as bytes that don't match.
+        # for our simple test cases right now, the bytes we're interested in scanning
+        # for will all just be there concretely
+        # TODO: Can probably do something smarter here like BM, but unnecessary if we're
```
disconnect3d: What is BM?

manticore/core/memory.py
```diff
@@ -1094,3 +1364,13 @@ class SMemory32L(SMemory):
...
     page_bit_size = 12
+
+
+class LazySMemory32(LazySMemory):
+    memory_bit_size = 32
+    page_bit_size = 12
+
+
+class LazySMemory64(LazySMemory):
+    memory_bit_size = 64
```
disconnect3d: aren't we missing `page_bit_size` here?

manticore/core/smtlib/expression.py
```diff
@@ -907,8 +908,8 @@ def store(self, index, value):
             value = self.cast_value(value)
         from manticore.core.smtlib.visitors import simplify
         index = simplify(index)
-        if isinstance(index, Constant):
-            self._concrete_cache[index.value] = value
+        # if isinstance(index, Constant):
```
disconnect3d: :F

manticore/core/smtlib/solver.py
```diff
@@ -235,6 +235,7 @@ def _send(self, cmd):
             :param cmd: a SMTLIBv2 command (ex. (check-sat))
         '''
         logger.debug('>%s', cmd)
+        # print(cmd)
```
disconnect3d: to be removed

manticore/core/smtlib/solver.py
```diff
@@ -260,6 +261,7 @@ def readline():
             right += r
         buf = ''.join(bufl).strip()
         logger.debug('<%s', buf)
+        # print(buf)
```
disconnect3d: to be removed

manticore/core/memory.py
yan: I have to fight the urge to refactor here also :) The bounded `range` won't work with symbolic addresses, it needs to be done using `__add__`.

disconnect3d: Oh right. Can we add a comment about it? :P

feliam:
This may be mergeable with current memory, so the fully symbolic option gets in when symbolic indexing is too aggressive. And otherwise the optimized option is in place. The threshold in which memory behave as one or the other may be configurable via new awesome config.py.

manticore/__main__.py
```diff
@@ -54,6 +54,8 @@ def positive(value):
     parser.add_argument('--workspace', type=str, default=None,
                         help=("A folder name for temporaries and results."
                               "(default mcore_?????)"))
+    parser.add_argument('--pure-symbolic', action='store_true',
+                        help='Treat all writable memory as symbolic')
```
disconnect3d: Should rather be moved to `bin_flags` and we might add a `(linux only)` as it won't work with decree targets.

manticore/core/cpu/abstractcpu.py
```diff
@@ -762,15 +765,22 @@ def decode_instruction(self, pc):
...
+                        c = bytes([vals[0]])
+                    except visitors.ArraySelectSimplifier.ExpressionNotSimple:
+                        c = struct.pack('B', solver.get_value(self.memory.constraints, c))
                 else:
-                    logger.error('Concretize executable memory %r %r', c, text)
-                    raise ConcretizeMemory(self.memory,
-                                           address=pc,
-                                           size=8 * self.max_instr_width,
-                                           policy='INSTRUCTION')
+                    if isinstance(c, Constant):
```
disconnect3d: Maybe
```
elif ...:

else:
```

manticore/core/cpu/abstractcpu.py
```diff
@@ -762,15 +765,22 @@ def decode_instruction(self, pc):
...
+                        c = bytes([c.value])
+                    else:
+                        logger.error('Concretize executable memory %r %r', c, text)
+                        raise ConcretizeMemory(self.memory,
+                                               address=pc,
+                                               size=8 * self.max_instr_width,
```
disconnect3d: Is this correct for 32-bit targets?

manticore/core/memory.py
```diff
@@ -283,8 +295,50 @@ def __setitem__(self, index, value):
     def __getitem__(self, index):
         index = self._get_offset(index)
         if isinstance(index, slice):
-            return [bytes([i]) for i in self._data[index]]
-        return bytes([self._data[index]])
+            # return [bytes([i]) for i in self._data[index]]
```
disconnect3d: Do we need tihs comment?

manticore/core/memory.py
```diff
@@ -283,8 +295,50 @@ def __setitem__(self, index, value):
...
+    def __setitem__(self, key, value):
+        self._array[key] = value
+
+    def __getitem__(self, key):
+        return self._array[key]
+
+    def split(self, address):
+        if address <= self.start:
+            return None, self
+        if address >= self.end:
```
disconnect3d: Is this logic right?

Having a memory page of size 0x100 that starts at address A, the last address we can use is A+0x100-1.
So I assume making a split on `self.end` can't split the addresses.

But making a split on start could in theory return `A[0], A[1:]`.

manticore/core/memory.py
```diff
@@ -1071,6 +1135,211 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+            if span is None:
+                continue
+
+            start, stop = span
+
+            for addr in range(start, stop):
+                if addr in self.backed_by_symbolic_store:
+                    continue
+
+                self.backing_array[addr] = Memory.read(self, addr, 1)[0]
```
disconnect3d: Wouldn't it be better to read until stop or until it is backed by symbolic store?

manticore/core/memory.py
```diff
@@ -1071,6 +1135,211 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+                self.backed_by_symbolic_store.add(addr)
+
+    def _map_deref_expr(self, map, address):
+        return Operators.AND(
+            Operators.UGE(address, map.start),
+            Operators.ULT(address, map.end))
+
+    def _reachable_range(self, sym_address, size):
+        addr_min, addr_max = solver.minmax(self.constraints, sym_address)
+        return addr_min, addr_max + size - 1
```
disconnect3d: Is adding `+size-1` here correct?

manticore/core/memory.py
```diff
@@ -1071,6 +1135,211 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
+            for addr, byte in zip(addrs_to_access, value):
+                self.backing_array[addr] = Operators.ORD(byte)
+        else:
+            self.backed_by_symbolic_store -= set(addrs_to_access)
+            Memory.write(self, address, value)
+
+    def scan_mem(self, data_to_find):
+        """
+
+        :param bytes data_to_find:
```
disconnect3d: Either fill or remove

tests/test_lazy_memory.py
```diff
@@ -0,0 +1,257 @@
...
+        ret = mem._deref_can_succeed(m, valid, 1)
+        self.assertTrue(ret)
+
+        ret = mem._deref_can_succeed(m, invalid, 1)
+        self.assertFalse(ret)
+
+        ret = mem._deref_can_succeed(m, 0x1000, 1)
+        self.assertTrue(ret)
+
+        ret = mem._deref_can_succeed(m, 0x800, 1)
```
disconnect3d: We lack testcases when `size != 1` :P

manticore/__main__.py
yan: Good call; I originally thought evm stuff used `Memory` as well, but it doesn't. Will move.

manticore/core/cpu/abstractcpu.py
yan: This was written this way because previous implementation of this relied on them being separate; currently it's identical. I'll update.

manticore/core/memory.py
yan: Not sure how; the semantics of `split` is that the address is the start of the newly created "rhs" map. So `split(start)` should return (None, original_map). I also don't think split is used very frequently.

manticore/core/memory.py
yan: I believe so, because we're trying to check if the last byte of the range is reachable, not the byte immediately following it.

manticore/core/memory.py
disconnect3d: > address is the start of the newly created "rhs" map
The 'rhs' is the key here. I kinda assumed it shall be 'lhs'.

So it's fine; might be good to add a docstring that its about rhs :P.

pgoodman: The logic is that you can only split on an internal address, not a boundary address. Hence the later assertion on line 329.

manticore/core/smtlib/expression.py
```diff
@@ -609,11 +609,12 @@ def cast_index(self, index):
         return index
 
     def cast_value(self, value):
-        if isinstance(value, str) and len(value) == 1:
+        if isinstance(value, (str, bytes)) and len(value) == 1:
             value = ord(value)
```
disconnect3d: Funfact/TIL: `ord` works for unicode chars too in Py3

```
In [4]: ord('ĺ')
Out[4]: 314
```

PR #1139, opened 2018-09-17
Introduce a configuration system

### LINKED ISSUE #372: Feature: Add global configuration mechanism

yan:
This PR introduces a config/globals management system that addresses a few concerns about manticore:

1. The lack of a method to update certain configuration constants from Manticore's environment (such as solver timeout).
2. The ability to maintain a configuration file with a set of persistent flags and settings to simplify manticore invocation. (i.e. always enable all EVM detectors)
3. Make a central global place for configuration constants.

## Implementation

This PR adds the `manticore.utils.config` module. It implements a way to create and set configuration groups. Manticore's `main` tries loading all groups/values from yaml files in `$PWD` (file can be `.manticore.yml`, `manticore.yml`, `.mcore.yml`, or `mcore.yml`). A `manticore.yml` is also now produced in workspace directory upon a run's completion.

Variable values take the following precedence, lowest to highest:

1. The default value declared in the top of a module.
2. The configuration file read
3. The CLI if the variable is also a command line argument. (All command line arguments can be set in the yml under the key `cli`)

New flag, `--config` has been added to specify a config file if it's not one of the files that are automatically discovered. `--config-print` dumps all declared constants that can be set.

Declaring a configuration group is pretty straightforward. The following is a snippet from the top of `solver.py`:
```
consts = config.get_group('smt')
consts.add('timeout', default=240, description='Timeout, in seconds, for each Z3 invocation')
consts.add('memory', default=16384, description='Max memory for Z3 to use (in Megabytes)')
consts.add('maxsolutions', default=10000, description='Maximum solutions to provide when solving for all values')
```

Then using it later is as simple as referring to `consts.memory` or `consts.timeout`.

Fixes #372 	

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1139)


manticore/core/smtlib/solver.py
```diff
@@ -129,7 +134,7 @@ def __init__(self):
         super().__init__()
         self._proc = None
 
-        self._command = 'z3 -t:240000 -memory:16384 -smt2 -in'
+        self._command = f'z3 -t:{consts.timeout*1000} -memory:{consts.memory} -smt2 -in'
```
disconnect3d: We could add config parameter for z3 path and default it to `z3` aka fetch from $PATH.

manticore/core/smtlib/solver.py
```diff
@@ -359,13 +364,16 @@ def can_be_true(self, constraints, expression):
 
     # get-all-values min max minmax
     #@memoized
```
disconnect3d: Can we remove this line? Seems we are not going to memoize it, at least for now ( ? ).

manticore/core/workspace.py
```diff
@@ -9,12 +9,16 @@
...
 
+from manticore.utils import config
 from manticore.utils.helpers import PickleSerializer
 from .smtlib import solver
 from .state import State
 
 logger = logging.getLogger(__name__)
 
+consts = config.get_group('workspace')
+consts.add('prefix', default='mcore_', description="The prefix to use for output and worksapce directories")
```
disconnect3d: typo `worksapce` -> `workspace`

manticore/core/workspace.py
```diff
@@ -168,7 +172,7 @@ def __init__(self, uri=None):
         :param uri: The path to on-disk workspace, or None.
         """
         if not uri:
-            uri = os.path.abspath(tempfile.mkdtemp(prefix="mcore_", dir='./'))
+            uri = os.path.abspath(tempfile.mkdtemp(prefix=consts.prefix, dir='./'))
```
disconnect3d: Might be worth to consider one of:
* dropping `dir='./'` so that all mcores will be saved in `/tmp`
* adding a config parameter for `dir` so ppl can set it to `/tmp` or other e.g. persistent directory

manticore/utils/config.py
```diff
@@ -0,0 +1,208 @@
...
+from itertools import product
+
+
+_groups = {}
+
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigError(RuntimeError):
```
disconnect3d: Hmm... it's not a big deal, but I would be up for inheriting from `Exception`. With `RuntimeError` someone might think of 'oh it is not so broad as `Exception` so maybe I can catch all runtime errors (not being fully aware what are them) and end up wasting time if something breaks with different reason.

Below some random stuff I looked up while thinking about this.

---


Via https://docs.python.org/3/library/exceptions.html

Built-in exception classes inheritance tree:
```
BaseException
 +-- SystemExit
 +-- KeyboardInterrupt
 +-- GeneratorExit
 +-- Exception
(...)
      +-- RuntimeError
      |    +-- NotImplementedError
      |    +-- RecursionError
(...)
```

Some example that booms with a `RuntimeError`:
```python
while 1:
    try:
        def foo():
            foo()
        foo()
    except RuntimeError:
        print("lol")
```

manticore/utils/config.py
```diff
@@ -0,0 +1,208 @@
...
+        self.value = default
+        self.default = default
+        self.defined = defined
+
+    @property
+    def was_set(self) -> bool:
+        return self.value is not self.default
+
+
+class _group:
```
disconnect3d: I am not 100% sure but it seems we could inherit from a dict and remove some of the code

manticore/utils/config.py
```diff
@@ -0,0 +1,208 @@
...
+    possible_names = ['mcore.ini', 'manticore.ini']
+    names = [os.path.join('.', ''.join(x)) for x in product(['', '.'], possible_names)]
+
+    if path is not None:
+        names = [path]
+
+    for name in names:
+        if os.path.exists(name):
+            logger.info(f'Reading configuration from {name}')
+            with open(name, 'r') as ini_f:
```
disconnect3d: Funfact, there is a [TOCTOU](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use)/race condition here: we check if file exists and later open it.

An [EAPF style](https://docs.quantifiedcode.com/python-anti-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html) is less bug-prone in that matter (as long as the exception is caught early on, which is not always obvious)

manticore/__main__.py
```diff
@@ -121,10 +140,22 @@ def positive(value):
...
+
     parsed = parser.parse_args(sys.argv[1:])
     if parsed.procs <= 0:
         parsed.procs = 1
 
+    update_non_defaults(parsed, parser)
+
+    if parsed.config_print:
+        print(config.describe_options())
+        raise SystemExit
```
disconnect3d: `sys.exit(retcode)`?

manticore/utils/config.py
```diff
@@ -0,0 +1,208 @@
...
+    def get_description(self, name: str) -> str:
+        """
+        Return the description, or a help string of variable identified by |name|.
+        """
+        if name not in self._vars:
+            raise ConfigError("f{self.name}.{name} not defined.")
+
+        return self._vars[name].description
+
+    def was_set(self, name: str) -> bool:
```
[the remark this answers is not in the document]
yan: `_var.was_set` is a property, but this `was_set` takes a name parameter (in essence, a lookup func)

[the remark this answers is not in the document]
disconnect3d: Oh yeah I know. I just mean that it is a bit misleading when one might see `something.was_set(name)` and `something_else.was_set` - a different usage.

manticore/utils/config.py
```diff
@@ -0,0 +1,236 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ConfigError(Exception):
+    pass
+
+
+class _var:
+    def __init__(self, name: str='', default=None, description: str=None, defined: bool=True):
+        self.name = name
```
feliam: idea;importance:3%. Make name and other private. `._name`
And if go that way. getter for value return default if None.

yan: name is used externally to the class though (i.e. in module functions)

mossberg:
> A `manticore.ini` is also now produced in workspace directory upon a run's completion.

just tried it, looks like it's not implemented for ethereum yet

mossberg:
i happen to have a malformed `manticore.yml` in my pwd right now because i'm playing around :)

then i was confused why manticore seemed to be totally broken for all invocations..

```
[I] (py3) vagrant ubuntu-bionic /v/manticore (dev-config-system) ❯ manticore examples/evm/two_tx_ovf.sol
  File "/home/vagrant/.virtualenvs/py3/bin/manticore", line 11, in <module>
    load_entry_point('manticore', 'console_scripts', 'manticore')()
  File "/vagrant/manticore/manticore/__main__.py", line 247, in main
    args = parse_arguments()
  File "/vagrant/manticore/manticore/__main__.py", line 179, in parse_arguments
    process_config_values(parsed, parser)
  File "/vagrant/manticore/manticore/__main__.py", line 27, in process_config_values
    config.load_overrides(args.config)
  File "/vagrant/manticore/manticore/utils/config.py", line 185, in load_overrides
    parse_config(yml_f)
  File "/vagrant/manticore/manticore/utils/config.py", line 159, in parse_config
    for section_name, section in c.items():
AttributeError: 'NoneType' object has no attribute 'items'

```

then i realized it was because of my malformed file. let's bail out with a nicer message in case the yml file is malformed

mossberg:
In this situation, shouldn't the cli values of `--txlimit 2` and `--no-testcases` been respected, since the cli is higher priority? 

```
[I] (py3) vagrant ubuntu-bionic /v/manticore (dev-config-system) ❯ cat manticore.yml
cli:
    detect_all: true
    no_testcases: false
    txlimit: 1
[I] (py3) vagrant ubuntu-bionic /v/manticore (dev-config-system) ❯ manticore tests/binaries/detectors/etherleak_true_pos_argument.sol --txlimit 2 --no-testcases
2018-10-05 21:46:49,102: [9813] m.main:INFO: Beginning analysis
2018-10-05 21:46:49,104: [9813] m.ethereum:INFO: Starting symbolic create contract
2018-10-05 21:46:49,448: [9813] m.ethereum:INFO: Starting symbolic transaction: 0
2018-10-05 21:46:50,545: [9813] m.ethereum:INFO: 1 alive states, 4 terminated states
2018-10-05 21:46:50,563: [9813] m.ethereum:INFO: Generated testcase No. 0 - STOP
2018-10-05 21:46:51,733: [9920] m.ethereum:INFO: Generated testcase No. 1 - REVERT
2018-10-05 21:46:52,908: [9920] m.ethereum:INFO: Generated testcase No. 2 - RETURN
2018-10-05 21:46:53,824: [9920] m.ethereum:INFO: Generated testcase No. 3 - REVERT
2018-10-05 21:46:54,956: [9920] m.ethereum:INFO: Generated testcase No. 4 - REVERT
2018-10-05 21:46:56,611: [9813] m.ethereum:INFO: Results in /vagrant/manticore/mcore_878uarct

```

manticore/__main__.py
```diff
@@ -4,13 +4,54 @@
...
 
-sys.setrecursionlimit(10000)
+consts = config.get_group('main')
+consts.add('recursionlimit', default=10000,
+           description="Value to set for Python recursion limit")
+
+sys.setrecursionlimit(consts.recursionlimit)
+
+
+def process_config_values(args, parser):
```
mossberg: nit, i'm a little wary of adding stuff to this file :/ i just don't want it to get really big since it's supposed to just be a small thing and has been growing

feliam: Maybe this should go to config.py  (or helpers?)

mossberg:
> In this situation, shouldn't the cli values of `--txlimit 2` and `--no-testcases` been respected, since the cli is higher priority?

For me this is still not totally correct. the `--no-testcases` override is respected, but not the txlimit one.

yan:
> > In this situation, shouldn't the cli values of `--txlimit 2` and `--no-testcases` been respected, since the cli is higher priority?
> 
> For me this is still not totally correct. the `--no-testcases` override is respected, but not the txlimit one.


I just tested it, and it works for me. Try `print(parsed.txlimit)` after `process_config_values` is called?

PR #1150, opened 2018-09-20
Fix terminate state

### LINKED ISSUE #1076: Manticore.terminate() should abandon current state
### Manticore version

8142472c50617c130151ce701bb67a7f98972708

### Summary of the problem

Currently Manticore.terminate() does not abandon the current state. All it does it set the shutdown flag within the executor, which causes no more states to be loaded, and for the executor to stop _at the next execute loop cycle_.

Typically Manticore.terminate() is called from within a hook.

```
@m.hook(0x41414141)
def hook(state):
  # stuff
  m.terminate()  # expects for manticore to completely exit, and for the m.run() function to return at this point
```

the `@m.hook()` syntactic sugar is just a `will_execute_instruction` event handler. that event is fired in `abstractcpu.py:Cpu.execute` before the instruction is executed. so if the hook calls `m.terminate()` and returns, manticore will actually not stop then. that instruction will continue to execute, and manticore will potentially do (lots?) of other stuff. manticore will only stop until the next iteration of the execute loop begins.

in order to actually implement the semantics of Manticore.terminate() we should probably raise `TerminateState()` within m.terminate() to force the state to halt. since this forces a new iteration of the execute loop, this will effectively stop the engine.

https://github.com/trailofbits/manticore/blob/8142472c50617c130151ce701bb67a7f98972708/manticore/manticore.py#L613-L618

todo:

- raise `TerminateState` at the end of `Manticore.terminate()`. maybe something like `TerminateState('Mantiocre terminated')`

mossberg:
Fixes #1076 

- raises TerminateState in m.terminate()
- adds an except for Terminate state in Manticore.run(). this prevents TerminateStates (such as from m.terminate() and state.abandon() to 'break out' to the user) like if you do m.terminate() outside the executor loop (in the init hook, for example)

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1150)

### DIFF
--- a/manticore/manticore.py
+++ b/manticore/manticore.py
@@ -628,6 +628,7 @@ class Manticore(Eventful):
         a :func:`~hook`.
         '''
         self._executor.shutdown()
+        raise TerminateState('Terminated')
 
     def shutdown(self):
         '''
--- a/manticore/manticore.py
+++ b/manticore/manticore.py
@@ -606,20 +606,25 @@ class Manticore(Eventful):
         :param timeout: Analysis timeout, in seconds
         '''
         assert not self.running, "Manticore is already running."
-        self._start_run()
-
-        self._time_started = time.time()
-        if timeout > 0:
-            t = Timer(timeout, self.terminate)
-            t.start()
         try:
-            self._start_workers(procs, profiling=should_profile)
+            self._start_run()
 
-            self._join_workers()
-        finally:
+            self._time_started = time.time()
             if timeout > 0:
-                t.cancel()
-        self._finish_run(profiling=should_profile)
+                t = Timer(timeout, self.terminate)
+                t.start()
+            try:
+                self._start_workers(procs, profiling=should_profile)
+
+                self._join_workers()
+            finally:
+                if timeout > 0:
+                    t.cancel()
+            self._finish_run(profiling=should_profile)
+        except TerminateState:
+            #  Must have been raised outside of executor loop. likely by Manticore.terminate()
+            # or State.abandon() in the init hook.
+            pass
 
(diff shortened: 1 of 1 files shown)


manticore/manticore.py
```diff
@@ -628,6 +628,7 @@ def terminate(self):
         a :func:`~hook`.
         '''
         self._executor.shutdown()
+        raise TerminateState('Terminated')
```
feliam: This will cancel exploration of current state. 
And it will wait until next fork for all the other running states already in a worker.

manticore/manticore.py
```diff
@@ -606,20 +606,25 @@ def run(self, procs=1, timeout=0, should_profile=False):
...
+                t = Timer(timeout, self.terminate)
+                t.start()
+            try:
+                self._start_workers(procs, profiling=should_profile)
+
+                self._join_workers()
+            finally:
+                if timeout > 0:
+                    t.cancel()
+            self._finish_run(profiling=should_profile)
```
yan: Should `_finish_run` be invoked inside the `finally` block?

mossberg: that's not related to this PR but i guess?

i actually do not see the point of the try/finally here, since it doesn't catch anything.

mossberg:
this got stalled, and is relatively low value, so i'll close it. 

- it's less effective than i was anticipating because i didn't think about multi worker environments
- the use case it fixes (`this prevents TerminateStates (such as from m.terminate() and state.abandon() to 'break out' to the user) like if you do m.terminate() outside the executor loop (in the init hook, for example)`) is uncommon.

PR #1154, opened 2018-09-27
Continue cleanup of ethereum package

mossberg:
- create abi.py, move ABI class into it
- create account.py move EVMAccount and EVMContract into it
- create solidity.py move SolidityMetadata into it

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1154)


manticore/ethereum/solidity.py
```diff
@@ -0,0 +1,191 @@
+import pyevmasm as EVMAsm
+
+from .abi import ABI
+from ..exceptions import EthereumError
+
+
+class SolidityMetadata(object):
```
feliam: Hmm maybe for future. Rename this to just Metadata? ethereum.solidity.Metadata ?

mossberg: good idea! although i don't particularly mind SolidityMetadata either. having just `Metadata` floating around the codebase elsewhere could be a little vague as to what it refers to (then you need to track it back to where the import came from to see the solidity party)

PR #1160, opened 2018-09-29
Add new auto CONCRETE test generator for eth vmTests

feliam:
New auto concrete test generator for VMTests
Concrete VMTests tests added (Frontier)
Pyevmasm version bump and minimal `fork` support
Several correctness bugfixes. Correctness level: Concrete Frontier 100%

This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1160)


dguido:
XML to Python unit tests in concrete mode for vmTests are done on this branch. Caveat: the VMTests are just for certain forks and we support our own mixture of forks so it wont pass. This identified a few consensus bugs which are being fixed in this branch.

Next, we will do the same for GeneralStateTests, pinning on a specific fork we'll support. We'll also have to do the same thing for symbolic inputs (convert the XML to symbolic unit tests).

manticore/manticore.py
```diff
@@ -224,9 +224,12 @@ def unregister_plugin(self, plugin):
...
     def __del__(self):
-        plugins = list(self.plugins)
-        for plugin in plugins:
-            self.unregister_plugin(plugin)
+        try:
+            plugins = list(self.plugins)
+            for plugin in plugins:
+                self.unregister_plugin(plugin)
+        except:
+            pass
```
disconnect3d: What are the cases when it fails? Maybe we should `try ... except` only the `self.unregister_plugin` call?

disconnect3d: What are the cases when it fails? Maybe we should wrap only the `self.unregister_plugin` call with `try ... except`?

manticore/core/smtlib/visitors.py
```diff
@@ -548,19 +544,20 @@ def arithmetic_simplify(expression):
 
 
 def to_constant(expression):
-    value = arithmetic_simplify(expression)
+    value = simplify(expression)
     if isinstance(value, Constant):
         return value.value
     elif isinstance(value, Array):
-        if value.index_max:
+        if expression.index_max:
```
feliam: This should be `value`. And `simplify()` should maintain the index_max pre/pos simplification

manticore/manticore.py
feliam: This failed by `self` not even having a `.plugins` member. `__del__` is problematic.

manticore/core/smtlib/constraints.py
```diff
@@ -54,7 +54,7 @@ def add(self, constraint, check=False):
         if isinstance(constraint, bool):
             constraint = BoolConstant(constraint)
         assert isinstance(constraint, Bool)
-        constraint = simplify(constraint)
+        #constraint = simplify(constraint)
```
disconnect3d: Either remove or uncomment?

manticore/core/smtlib/expression.py
```diff
@@ -690,6 +691,7 @@ def underlying_variable(self):
         return array
 
     def read_BE(self, address, size):
+        from manticore.core.smtlib.visitors import translate_to_smtlib
```
disconnect3d: This import seems to be redundant

tests/auto/make_VMTests.py
```diff
@@ -0,0 +1,215 @@
...
+
+    testname = (os.path.split(filename)[1].replace('-','_')).split('.')[0]
+    bytecode = unhexlify(testcase['exec']['code'][2:])
+    disassemble = ''
+    try:
+        #add silent=True when evmasm supports it
+        disassemble = '\n                  '.join(EVMAsm.disassemble(bytecode).split('\n'))
+    except Exception as e:
+        pass
+    sha256sum = hashlib.sha256(open(filename, 'rb').read()).hexdigest()
```
disconnect3d: Should use `with open(filename 'rb') as f: sha256sum = hashlib.sha256(f.read()).hexdigest()`

Side note: this line may use a lot of memory if the file is very large (see https://stackoverflow.com/questions/22058048/hashing-a-file-in-python).

tests/test_binaries.py
```diff
@@ -303,31 +306,94 @@ def test_decree(self):
...
+            )
+
+    def test_eth_regressions_807(self):
+        issue = {'number': 807, 'contract': 'C', 'txlimit': 1}
+        self._simple_cli_run(
+                f'{issue["number"]}.sol', contract=issue['contract'], tx_limit=issue['txlimit'],
+                in_directory=issue.get('in_directory')
+            )
+
+    def test_eth_regressions_80(self):
```
disconnect3d:
```suggestion
    def test_eth_regressions_808(self):
```

tests/test_smtlibv2.py
```diff
@@ -262,34 +263,57 @@ def testBasicArrayConcatSlice(self):
...
+        #make array of 32->8 bits
+        array = cs.new_array(32, index_max=12)
+        array = array.write(0, hw)
+        array_slice = array[0:2]
+        self.assertTrue(self.solver.must_be_true(cs, array == hw))
+        self.assertTrue(self.solver.must_be_true(cs, array_slice[0] == array[0]))
+        self.assertTrue(self.solver.must_be_true(cs, array_slice[0:2][1] == array[1]))
+        array_slice[0] = ord('A')
+        self.assertTrue(self.solver.must_be_true(cs, array_slice[0] == ord('A')))
+        self.assertTrue(self.solver.must_be_true(cs, array_slice[0:2][1] == array[1]))
```
disconnect3d: Wondering, what happens if I do:
```
array_slice[0:10000]
```
when the slice has only 2 elements? Will it return me a slice of those two, as with Python lists?

We might want add a testcase for it.

feliam: Some tests added.

manticore/core/smtlib/expression.py
```diff
@@ -730,8 +731,10 @@ def __add__(self, other):
         new_arr = ArrayProxy(ArrayVariable(self.index_bits, self.index_max + len(other), self.value_bits, 'concatenation{}'.format(uuid.uuid1())))
         for index in range(self.index_max):
             new_arr[index] = simplify(self[index])
+        _concrete_cache = new_arr._concrete_cache
         for index in range(len(other)):
             new_arr[index + self.index_max] = simplify(other[index])
+        new_arr._concrete_cache.update(_concrete_cache)
```
disconnect3d: Are those fine in the end?

It seems as if it is sth like `d = A.dict; <some operations>; A.dict.update(d)` (`d` is not a copy of `A.dict`, just a reference/pointer to it).

manticore/platforms/evm.py
```diff
@@ -468,13 +468,13 @@ def __init__(self, constraints, address, data, caller, value, bytecode, world=No
         super().__init__(**kwargs)
         if data is not None and not issymbolic(data):
             data_size = len(data)
-            data_symbolic = constraints.new_array(index_bits=256, value_bits=8, index_max=data_size, name='DATA_{:x}'.format(address), avoid_collisions=True)
+            data_symbolic = constraints.new_array(index_bits=256, value_bits=8, index_max=data_size, name='DATA_{:x}'.format(address), avoid_collisions=True, default=0)
```
```suggestion
            data_symbolic = constraints.new_array(index_bits=256, value_bits=8, index_max=data_size, name=f'DATA_{address:x}', avoid_collisions=True, default=0)
```

manticore/platforms/evm.py
```diff
@@ -468,13 +468,13 @@ def __init__(self, constraints, address, data, caller, value, bytecode, world=No
...
             data_symbolic[0:data_size] = data
             data = data_symbolic
 
         if bytecode is not None and not issymbolic(bytecode):
             bytecode_size = len(bytecode)
-            bytecode_symbolic = constraints.new_array(index_bits=256, value_bits=8, index_max=bytecode_size, name='BYTECODE_{:x}'.format(address), avoid_collisions=True)
+            bytecode_symbolic = constraints.new_array(index_bits=256, value_bits=8, index_max=bytecode_size, name='BYTECODE_{:x}'.format(address), avoid_collisions=True, default=0)
```
```suggestion
            bytecode_symbolic = constraints.new_array(index_bits=256, value_bits=8, index_max=bytecode_size, name=f'BYTECODE_{address:x}', avoid_collisions=True, default=0)
```

manticore/platforms/evm.py
```diff
@@ -509,7 +509,8 @@ def extend_with_zeroes(b):
         #if len(bytecode) == 0:
         #    raise EVMException("Need code")
         self._constraints = constraints
-        self.memory = constraints.new_array(index_bits=256, value_bits=8, name='EMPTY_MEMORY_{:x}'.format(address), avoid_collisions=True)
+        # Uninitialized values in memory are 0 by spec
+        self.memory = constraints.new_array(index_bits=256, value_bits=8, name='EMPTY_MEMORY_{:x}'.format(address), avoid_collisions=True, default=0)
```
```suggestion
        self.memory = constraints.new_array(index_bits=256, value_bits=8, name=f'EMPTY_MEMORY_{address:x}', avoid_collisions=True, default=0)
```

manticore/core/smtlib/expression.py
```diff
@@ -730,8 +731,10 @@ def __add__(self, other):
         new_arr = ArrayProxy(ArrayVariable(self.index_bits, self.index_max + len(other), self.value_bits, 'concatenation{}'.format(uuid.uuid1())))
         for index in range(self.index_max):
             new_arr[index] = simplify(self[index])
+        _concrete_cache = new_arr._concrete_cache
```
feliam: ArrayProxy creates a fresh dict (assumed)

manticore/platforms/evm.py
```diff
@@ -32,6 +32,7 @@
 MASK160 = 2 ** 160 - 1
 TT255 = 2 ** 255
 TOOHIGHMEM = 0x1000
+DEFAULT_FORK = 'byzantium'
```
disconnect3d: Should probably be a config variable/parameter, but we can make it after merging.

manticore/core/smtlib/expression.py
feliam: The idea is that  you have two 5 bytes long arrays: xAxxx  xxBxx
1s has a concrete value in pos 1: A 
2nd has a concrete value in pos 2: B

The concatenation makes a fresh empty concrete cache dict.
It _should_ be populated with concrets from the 1st array and updated from something from the second array.

code is wrong. checking.

manticore/platforms/evm.py
```diff
@@ -218,12 +218,12 @@ def set_result(self, result, return_data=None):
...
         if result in {'RETURN', 'REVERT'}:
-            if not isinstance(return_data, (bytearray, Array)):
+            if not isinstance(return_data, (bytes, bytearray, Array)):
                 raise EVMException('Invalid transaction return_data type:', type(return_data).__name__)
         elif result in {'STOP', 'THROW', 'SELFDESTRUCT'}:
             if return_data is None:
                 return_data = bytearray()
-            if not isinstance(return_data, (bytearray, Array)) or len(return_data) != 0:
+            if not isinstance(return_data, (bytes, bytearray, Array)) or len(return_data) != 0:
                 raise EVMException('Invalid transaction return_data. To much data for STOP,THROW or SELFDESTRUCT')
```
```suggestion
                raise EVMException(f'Invalid transaction return_data. Too much data ({len(return_data)}) for STOP, THROW or SELFDESTRUCT')
```

manticore/platforms/evm.py
```diff
@@ -1726,16 +1746,19 @@ def __init__(self, constraints, storage=None, initial_block_number=None, initial
...
         self._transactions: List[Transaction] = list()
 
-        if initial_block_number is None:
-            #assume initial symbolic block
-            initial_block_number = constraints.new_bitvec(256, "BLOCKNUMBER", avoid_collisions=True)
-        self._initial_block_number = initial_block_number
-        if initial_timestamp is None:
+        if blocknumber is None:
+            #assume initial byzantium block
+            blocknumber = 4370000
```
disconnect3d: Don't we have this data in pyevmasm?

manticore/platforms/evm.py
```diff
@@ -2058,18 +2086,23 @@ def send_funds(self, sender, recipient, value):
 
     def get_code(self, address):
         if address not in self._world_state:
-            return bytearray()
+            return bytes()
```
disconnect3d: What's the reason of that change?

manticore/platforms/evm.py
```diff
@@ -2200,7 +2236,11 @@ def create_account(self, address=None, balance=0, code=None, storage=None, nonce
             raise EthereumError('The account already exists')
 
         if storage is None:
-            storage = self.constraints.new_array(index_bits=256, value_bits=256, name='STORAGE_{:x}'.format(address))
+            # Uninitialized values in a storage are 0 by spec
+            storage = self.constraints.new_array(index_bits=256, value_bits=256, name='STORAGE_{:x}'.format(address), avoid_collisions=True, default=0)
```
```suggestion
            storage = self.constraints.new_array(index_bits=256, value_bits=256, name=f'STORAGE_{address:x}', avoid_collisions=True, default=0)
```

manticore/platforms/evm.py
```diff
@@ -2200,7 +2236,11 @@ def create_account(self, address=None, balance=0, code=None, storage=None, nonce
...
+        else:
+            # TODO: Check storage type/ dict from int/bitvec to int/bitvec
+            pass
```
disconnect3d: Is this fine to `pass` here? If something is missing, maybe we shall raise an exception (?)

tests/EVM/VMTests/eth_vmBlockInfoTest.py
```diff
@@ -0,0 +1,313 @@
...
+        data = rlp.encode([Log(unhexlify('{:040x}'.format(l.address)), l.topics, to_constant(l.memlog)) for l in world.logs])
+        self.assertEqual(sha3.keccak_256(data).hexdigest(), '1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347')
+        
+        # test spent gas
+        self.assertEqual(world.current_vm.gas, 79995)
+
+    def test_difficulty(self):
+        '''
+            Textcase taken from https://github.com/ethereum/tests
+            File: /home/felipe/Projects/manticore/tests/auto/tests/VMTests/vmBlockInfoTest/difficulty.json
```
disconnect3d: We should probably remove path prefix from here (`/home/felipe/Projects/`)

manticore/platforms/evm.py
disconnect3d: Ok I guess its to be able to specify code with `b''`

manticore/platforms/evm.py
disconnect3d: https://github.com/trailofbits/pyevmasm/blob/master/pyevmasm/evmasm.py#L881-L913
we have the reverse, guess we need to add the reverse of it too :P

disconnect3d: So I'd say it is good to go now.

manticore/platforms/evm.py
feliam: So far this is only used to tune the unittests that needs older-than-byzantium assembler. Once we pass frontier in both concrete AND symbolic we are probably deleting those tests and upgrade to the next fork. So in the end we'll end up not needing this variable/constant/config thing.

manticore/platforms/evm.py
feliam: bytecode is generally not mutable also. bytearray sounded like a bad fit

manticore/platforms/evm.py
feliam: All fork setting is not coherent. We can change the assembler, the block number and the instruction handlers are always the same (tending to byzantium).

manticore/core/smtlib/visitors.py
```diff
@@ -582,19 +578,24 @@ def arithmetic_simplify(expression):
 
 
 def to_constant(expression):
-    value = arithmetic_simplify(expression)
+    '''
+        Iff the expression can be simplified to a Constant get the actual concrete value.
+        This discards/ignore any taint
+    '''
+    value = simplify(expression)
     if isinstance(value, Constant):
```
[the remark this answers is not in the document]
feliam: https://github.com/trailofbits/manticore/blob/dev-new-auto-evmtest/manticore/core/smtlib/visitors.py#L587

Issue #1169, opened 2018-10-02
Consensus errors


holiman:
I'm eyeing through the EVM implementation, and thought I'd add any consensus errors that I can spot to this ticket. The ticket may contain some false positives, apologies in advance if that is the case

1. Missing analysis of valid `JUMPDEST`s. The EVM does not allow jumping into a `data`-section, for example into the `32` bytes following a `PUSH32`. So if the code is
```
PUSH01 5b
PUSH1 01
JUMP
```
Then it appears that manticore evm would jump into the 'artificial' jumpdest `5b` at byte `1` and continue execution. 

2. Missing `JUMPDEST` check. Apparently [known already](https://github.com/trailofbits/manticore/blob/master/manticore/platforms/evm.py#L1366) 

3. `DELEGATECALL`/`CALLCODE` quirks. It looks like `DELEGATECALL` passes the current address as `sender` [here](https://github.com/trailofbits/manticore/blob/master/manticore/platforms/evm.py#L1497). That's how `CALLCODE` works, whereas `DELEGATECALL` should pass the original `sender` as `sender`. 

4. `SELFDESTRUCT` immediately deletes an address from the `world`: https://github.com/trailofbits/manticore/blob/master/manticore/platforms/evm.py#L1880 . 
5. `SELFDESTRUCT` adds deleted account to a [list](https://github.com/trailofbits/manticore/blob/master/manticore/platforms/evm.py#L1657), possibly adding the same account multiple times if `SELFDESTRUCT` is performed several times: https://github.com/trailofbits/manticore/blob/master/manticore/platforms/evm.py#L1881 

6. (I'm unsure about this): I can't find where the stack value normalization happens. For example, the [`SUB`](https://github.com/trailofbits/manticore/blob/98b9d360a92c8f4acf75c7e6327e295b3a8a7d0d/manticore/platforms/evm.py#L951) operation returns potentially a negative integer. What happens if that negative integer is used as `value` for a `CALL`? Will a negative `value` be sent? 

7. `63/64ths` rule does not seem implemented. Whenever a `CALL` type is made, the max allowed gas that is sent along to the child 63/64ths of the current `gas`.

dguido:
Thanks for the code review. We filed a few issues yesterday related to testing and plan to address them in this release cycle. In particular, we think that adding support for the Constantinople changes will take only a day or two and should be in soon.

https://github.com/trailofbits/manticore/issues/1163
https://github.com/trailofbits/manticore/issues/1164
[...]

PR #1172, opened 2018-10-02
Fix Consensus-errors.1 & 2

feliam:
Fix sub issues 1 & 2 from here:  https://github.com/trailofbits/manticore/issues/1169
```
Missing analysis of valid JUMPDESTs. The EVM does not allow jumping into a data-section, for example into the 32 bytes following a PUSH32. So if the code is
PUSH01 5b
PUSH1 01
JUMP
Then it appears that manticore evm would jump into the 'artificial' jumpdest 5b at byte 1 and continue execution.

Missing JUMPDEST check. Apparently known already
```

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1172)

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -356,16 +356,17 @@ class EVM(Eventful):
         self._valid_jumpdests = set()
 
 
-        def f(b):
+        #Compile the list of valid jumpdests via linear dissassembly
+        def extend_with_zeroes(b):
             try:
                 for x in b:
                     yield(to_constant(x))
                 for x in (0,)*32:
                     yield(x)
             except Exception as e:
-                print (e)
                 raise StopIteration
-        for i in EVMAsm.disassemble_all(f(bytecode)):
+
+        for i in EVMAsm.disassemble_all(extend_with_zeroes(bytecode)):
             if i.mnemonic == 'JUMPDEST':
                 self._valid_jumpdests.add(i.pc)
 
@@ -706,13 +707,20 @@ class EVM(Eventful):
         self._pc = last_pc
         self._checkpoint_data = None
 
-    def _set_check_jmpdest(self):
-        self._check_jumpdest = True
+    def _set_check_jmpdest(self, flag=True):
+        self._check_jumpdest = flag
 
     def _check_jmpdest(self):
-        if self._check_jumpdest:
+        if issymbolic(self._check_jumpdest):
+            should_check_jumpdest = solver.get_all_values(self.constraints, self._check_jumpdest)
+            if len(should_check_jumpdest) != 1:
+                raise EthereumError("Conditional not concretized at JMPDEST check")
+        else:
+            should_check_jumpdest = [self._check_jumpdest]
+
+        if should_check_jumpdest == [True]:
             self._check_jumpdest = False
-            if self.pc not in self._valid_jmpdests:
+            if self.pc not in self._valid_jumpdests:
                 raise InvalidOpcode()
 
     def _advance(self, result=None, exception=False):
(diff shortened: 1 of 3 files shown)


manticore/platforms/evm.py
```diff
@@ -469,6 +470,35 @@ def __init__(self, constraints, address, data, caller, value, bytecode, world=No
...
+            try:
+                for x in b:
+                    x = to_constant(x)
+                    if isinstance(x, int):
+                        yield(x)
+                    else:
+                        yield(0)
+                        
+                for x in (0,)*32:
+                    yield(x)
```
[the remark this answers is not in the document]
yan: `itertools.repeat(0, 32)`?

manticore/platforms/evm.py
```diff
@@ -801,6 +836,22 @@ def _rollback(self):
...
+
+    def _check_jmpdest(self):
+        if issymbolic(self._check_jumpdest):
+            should_check_jumpdest = solver.get_all_values(self.constraints, self._check_jumpdest)
+            if len(should_check_jumpdest) != 1:
+                raise EthereumError("Conditional not concretized at JMPDEST check")
+        else:
+            should_check_jumpdest = [self._check_jumpdest]
+
+        if should_check_jumpdest == [True]:
```
disconnect3d: Since we only support a symbolic case where it is concretized/has one solution (?) then maybe we should do retrieve the value in that branch:
```python
if len(should_check_jumpdest) != 1:
    raise EthereumError(...)

should_check_jumpdest = should_check_jumpdest[0]
```

and just assign `should_check_jumpdest = self._check_jumpdest` if it is not symbolic and then check it here as `if should_check_jumpdest is True:`

manticore/platforms/evm.py
```diff
@@ -469,6 +470,33 @@ def __init__(self, constraints, address, data, caller, value, bytecode, world=No
             bytecode_symbolic[0:bytecode_size] = bytecode
             bytecode = bytecode_symbolic
 
+        #TODO: Handle the case in which bytecode is  symbolic (This happens at
```
disconnect3d: If we have any ideas on how to deal with it, its good to write them down here

manticore/platforms/evm.py
```diff
@@ -469,6 +470,33 @@ def __init__(self, constraints, address, data, caller, value, bytecode, world=No
...
+            # appended at the end
+            try:
+                for x in b:
+                    x = to_constant(x)
+                    if isinstance(x, int):
+                        yield(x)
+                    else:
+                        yield(0)
+                for _ in range(32):
+                    yield(0)
```
disconnect3d: I am not that much into the topic, but we add 32 * 0 no matter of the bytecode length, shouldn't we do sth like `32 - len(bytecode)` or something similar? Just sanity check/question huh

tests/eth_general.py
```diff
@@ -860,3 +862,44 @@ def test_run_solc_with_remappings(self):
                 self.assertIn("lib/B.sol", source_list)
         finally:
             shutil.rmtree(d)
+
+    def test_jmpdest_check(self):
+        '''
+        '''
```
disconnect3d: Can we fill the docstring explaining the testcase or at least linking to something ( ? )

tests/eth_general.py
```diff
@@ -860,3 +862,44 @@ def test_run_solc_with_remappings(self):
...
+                new_vm.execute()
+        except evm.EndTx as e:
+            result = e.result
+            if e.result in ('RETURN', 'REVERT'):
+                returndata = e.data
+        # test spent gas
+        self.assertEqual(result, 'THROW')
+        self.assertEqual(new_vm.gas, 99992)
+        #check refund
+        #check logs
```
disconnect3d: Are we missing those? either TODO or remove?

manticore/platforms/evm.py
disconnect3d: So to sum up, do I understand correctly that in most cases we need to add only *some* zeros but in the worst case all of 32 zeros. And the fact we add 32 zeros always is okay i.e. it doesn't change any other behavior?

feliam: So this linear disassembly thing is only used to search for "JUMPDEST" instructions and whitelisting those as JUMP targets. Adding zeros (STOP opcode or zero padded PUSH operand) will not add new valid JUMPDEST targets.

PR #1173, opened 2018-10-02
Fixes sha3 algorithm

feliam:


manticore/platforms/evm.py
```diff
@@ -668,7 +668,12 @@ def _checkpoint(self):
             pc = self.pc
             instruction = self.instruction
             old_gas = self.gas
-            self._consume(instruction.fee)
+
+            # FIXME patch for pyevmasm
```
disconnect3d: Hmm is this change for this PR?

Issue #1158, opened 2018-09-28
Colored detector output


mossberg:
To make the findings stand out, we could add colored output.

![image](https://user-images.githubusercontent.com/2467355/46235937-8cadd980-c34a-11e8-9621-36c1dd6b4704.png)

redyoshi49q:
I have a few questions relating to this issue request:

-  Should color support be given a command line parameter?  Should it default to on or off?
-  Should support for 16 colors be assumed or tested for?  Is there a preferred method for testing for such support?
-  Is it safe to assume that there's no need to consider using 256 color or "true color" for this purpose, or is the want for a larger and more diverse color palette sufficient to use 256 color support if it exists (with 16 color and no color as fallbacks)?
-  Is there an output environment that should be assumed, or should the environment that Manticore is outputting to be tested for?  In my experience, color codes in a .bashrc can react differently in a virtual terminal, a terminal emulator, an Emacs shell, and an Emacs term.

mossberg:
> * Should color support be given a command line parameter?  Should it default to on or off?

i think it should probably default to on, with ability to disable



> * Should support for 16 colors be assumed or tested for?  Is there a preferred method for testing for such support?

i don't have a preferred method. i'd say we can probably assume support exists. we can probably assume a fairly regular ubuntu 18.04 ish setup, we explicitly recommend that in the readme.





> * Is it safe to assume that there's no need to consider using 256 color or "true color" for this purpose

yeah, probably safe for now.

in general, i don't think we need to implement the most incredibly robust solution right now. let's just do something simple that works "most of the time"?

dguido:
Closed in #1177

PR #1177, opened 2018-10-06
Colored Output

redyoshi49q:
This pull request implements #1158 .  A "--no-colors" command line flag was added to disable the insertion of ANSI color escape sequences, which are now outputted by default.  Currently, "DEBUG:", "WARNING:", "ERROR:", and "INFO:" should output in magenta, yellow, red, and blue, respectively, but these colors can be changed by editing the `coloring` dict object.

Since Manticore starts logging before command line options are parsed, there is a brief window during which ANSI color escape sequences might be outputted before the toggle to disable them takes effect.  In addition, if there are any starting points of Manticore's execution /other/ than `manticore/__main__.py`, then I may have failed to insert a check for the aforementioned command line toggle in those areas.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1177)

### DIFF
--- a/manticore/utils/log.py
+++ b/manticore/utils/log.py
@@ -28,18 +28,17 @@ class ContextFilter(logging.Filter):
     for k, v in coloring.items():
         color_map[k] = colors[v]
 
-    color_set = u'\x1b[%sm'
-    color_reset = u'\x1b[0m'
+    colored_levelname_format = u'\x1b[{}m{}:\x1b[0m'
+    plain_levelname_format = u'{}:'
 
     def colored_level_name(self, levelname):
         '''
         Colors the logging level in the logging record
         '''
-        if ( self.colors_disabled ):
-            return levelname + u':'
+        if self.colors_disabled:
+            return self.plain_levelname_format.format(levelname)
         else:
-            retval = (self.color_set % self.color_map[levelname]) + levelname + u':' + self.color_reset
-            return retval
+            return self.colored_levelname_format.format(self.color_map[levelname], levelname)
 
     def filter(self, record):
         if hasattr(self, 'stateid') and isinstance(self.stateid, int):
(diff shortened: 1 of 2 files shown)


mossberg:
thanks a lot @redyoshi49q! we'll get to this soon.

mossberg:
some pics:

image
image

manticore/utils/log.py
```diff
@@ -18,26 +17,53 @@ def summarized_name(self, name):
...
+        color_map[k] = colors[v]
+
+    color_set = u'\x1b[%sm'
+    color_reset = u'\x1b[0m'
+
+    def colored_level_name(self, levelname):
+        '''
+        Colors the logging level in the logging record
+        '''
+        if ( self.colors_disabled ):
```
mossberg: we can omit the parens here

manticore/utils/log.py
```diff
@@ -18,26 +17,53 @@ def summarized_name(self, name):
...
+    coloring = {u'DEBUG':u'magenta', u'WARNING':u'yellow',
+        u'ERROR':u'red', u'INFO':u'blue'}
+    colors =  dict(zip([u'black', u'red', u'green', u'yellow',
+        u'blue', u'magenta', u'cyan', u'white'], map(str, range(30, 30 + 8))))
+
+    color_map = {}
+    for k, v in coloring.items():
+        color_map[k] = colors[v]
+
+    color_set = u'\x1b[%sm'
```
mossberg: we can prefer the `'{}'.format()` style for this?

redyoshi49q:
@mossberg Your suggestions have been implemented (my C/C++ fluency eeked out a bit).

I made the changes using my phone (and thus haven't yet tested the lateat commit).  The change should be relatively safe, though, and I would expect Travis to catch stupid mistakes on my part.

PR #1181, opened 2018-10-08
Add check and warning for mismatched args

### LINKED ISSUE #1178: Check call signature counts in an manticore.ethereum.EVMContract call ?
### OS / Environment

macOS High Sierra
### Manticore version

Version: 0.2.1.1

### Python version

Python 3.3.6
### Summary of the problem

When a Solidity function is called in a `manticore.ethereum.EVMContract` object, the parameter count of the function is not checked. 

This contract from the  trufflecon2018 manticore example: 

```solidity
pragma solidity^0.4.24;
contract Overflow {
    uint public sellerBalance=0;

    function add(uint value) public returns (bool){
        sellerBalance += value; // complicated math, possible overflow
    }
}
```

Now here is an improper use of the call `add`, since there are two parameters

```python
m = ManticoreEVM()
with open('overflow.sol') as f: source_code = f.read()
user_account = m.create_account(balance=1000)
contract_account = m.solidity_create_contract(source_code,  owner=user_account)
symbolic_var1 = m.make_symbolic_value()
symbolic_var2 = m.make_symbolic_value()
contract_account.add(symbolic_var1, symbolic_var2)  # Uses one parameter, not two!
```

### Expected behavior

Some sort of error in calling `add()` which takes only one parameter
### Actual behavior

No error.

mossberg:
This adds a check and warning if a contract function is called with a number of arguments that does not match the signature.

The reason a warning is logged rather than an error thrown, is because passing an incorrect number of arguments to a contract invocation will result in an incorrect length tx data being used in the tx. If it's too short, the missing arguments will be implicitly zero (see yellow paper, oob reads in txdata are 0). If it's too long, the extra arguments won't be accessed by the contract. So imo neither case is technically "wrong".

Also refactors `EVMContract.__getattribute__` to be more readable using a namedtuple.

Fixes #1178

Running the script in the issue now produces

```
[I] (py3) vagrant ubuntu-bionic /v/m/1178 ❯ py x.py                         ❮ 10/08 05:39
2018-10-08 17:39:38,724: [4249] m.e.abi:WARNING: Number of provided arguments (2) does not match number of arguments in signature: add(uint256)
```

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1181)

### DIFF
--- a/manticore/ethereum/__init__.py
+++ b/manticore/ethereum/__init__.py
@@ -11,6 +11,7 @@ from ..core.smtlib import ConstraintSet, Operators, solver, BitVec, Array, Array
 from ..platforms import evm
 from ..core.state import State, TerminateState
 from ..utils.helpers import issymbolic, PickleSerializer
+from ..utils.log import init_logging
 import tempfile
 from subprocess import Popen, PIPE, check_output
 from multiprocessing import Process, Queue
@@ -28,7 +29,10 @@ from .account import EVMAccount, EVMContract
 from .abi import ABI
 from .solidity import SolidityMetadata
 
+
 logger = logging.getLogger(__name__)
+
+init_logging()  # FIXME(mark): emitting a warning in abi.py does not work unless this is called a second time here
 
 
 def flagged(flag):
--- a/manticore/ethereum/abi.py
+++ b/manticore/ethereum/abi.py
@@ -43,7 +43,7 @@ class ABI(object):
         num_args = len(args)
         num_sig_args = len(type_spec.split(','))
         if num_args != num_sig_args:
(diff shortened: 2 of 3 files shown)


manticore/ethereum/__init__.py
```diff
@@ -28,8 +29,11 @@
 from .abi import ABI
 from .solidity import SolidityMetadata
 
+
 logger = logging.getLogger(__name__)
 
+init_logging()  # FIXME(mark): emitting a warning in abi.py does not work unless this is called a second time here
```
disconnect3d: :( any chance to do sth with this fixme?

PR #1182, opened 2018-10-08
Add optional param to apply constraint to state

mossberg:
```
Would it be possible to have, from a state, a function that solves a variable and concretizes it's solution in the state constraints ? 
It's a frequent mistake, from a user script level to have multiple symb variables, call `solve_one` on each variable, without concretizing the solution of each variable prior solving another one
Like if you have `a+b` and you look for overflow, the solver may tell you that a=1 and b=1 if you ask for the two solutions independently
```

This adds it to solve_one solve_buffer, which are the main APIs that make sense.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1182)

### DIFF
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -370,22 +370,24 @@ class State(Eventful):
 
     ################################################################################################
     # The following should be moved to specific class StatePosix?
-    def solve_buffer(self, addr, nbytes):
+    def solve_buffer(self, addr, nbytes, constrain=False):
         '''
         Reads `nbytes` of symbolic data from a buffer in memory at `addr` and attempts to
         concretize it
 
         :param int address: Address of buffer to concretize
         :param int nbytes: Size of buffer to concretize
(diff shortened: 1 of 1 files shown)


manticore/core/state.py
```diff
@@ -367,22 +370,24 @@ def solve_min(self, expr):
...
     # The following should be moved to specific class StatePosix?
-    def solve_buffer(self, addr, nbytes):
+    def solve_buffer(self, addr, nbytes, constrain=False):
         '''
         Reads `nbytes` of symbolic data from a buffer in memory at `addr` and attempts to
         concretize it
 
         :param int address: Address of buffer to concretize
         :param int nbytes: Size of buffer to concretize
+        :param bool constrain: If True, constrain the buffer to the concretized value
```
disconnect3d: or `whether to constrain the buffer to a concrete value`

Issue #1130, opened 2018-09-08
--timeout ignored for EVM


computerality:
The `--timeout` option doesn't seem to work when analyzing a .sol file from commandline. Example `manticore --timeout 60 t.sol`

mossberg:
we just need to forward `args.timeout` in `__main__.py` into the `ManticoreEVM` constructor

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 50.0 DAI (50.0 USD @ $1.0/DAI)  attached to it__ as part of the Ethereum Community Fund via ECF Web 3.0 Infrastructure Fund fund__.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/trailofbits/manticore/1130/1447).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/trailofbits/manticore/1130/1447).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $43,935.34 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [evgeniuz](https://gitcoin.co/profile/evgeniuz) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/trailofbits/manticore/1130/1447?mutate_worker_action=approve&worker=evgeniuz) | [reject worker](https://gitcoin.co/issue/trailofbits/manticore/1130/1447?mutate_worker_action=reject&worker=evgeniuz))_.**

Added basic timeout support for ManticoreEVM, plan to add tests for this issue as well.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/trailofbits/manticore/1130/1447).

gitcoinbot:
@evgeniuz Hello from Gitcoin Core - are you still working on this issue? Please submit a WIP PR or comment back within the next 3 days or you will be removed from this ticket and it will be returned to an ‘Open’ status. Please let us know if you have questions!
* [x] warning (3 days)
[...]

PR #1186, opened 2018-10-11
added timeout support for ManticoreEVM

evgeniuz:
Fixes https://github.com/trailofbits/manticore/issues/1130

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1186)

### DIFF
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -13,6 +13,8 @@ logger = logging.getLogger('manticore.main')
 consts = config.get_group('main')
 consts.add('recursionlimit', default=10000,
            description="Value to set for Python recursion limit")
+consts.add('timeout', default=0,
+           description='Timeout, in seconds, for Manticore invocation')
 
 
 def parse_arguments():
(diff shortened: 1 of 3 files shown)


evgeniuz:
Plan to try to add tests for this issue.

mkosowsk:
Hi @evgeniuz how is progress coming along? Thanks!

evgeniuz:
Sorry, was a little busy this week, will submit updated PR tomorrow (I have suspicion that current version sets timeout per single transaction, not whole invocation).

evgeniuz:
Added timeout for solidity CLI runner (also, changed timeout handling in `run` method to avoid code duplication, now same decorator is used in both places).

evgeniuz:
Added tests.

evgeniuz:
Implemented according to spec (with test) and ready for review.

disconnect3d:
We have a `timeout` config variable for smt solver:
https://github.com/trailofbits/manticore/blob/master/manticore/core/smtlib/solver.py#L36-L37

and we should add one for the overall execution too. This should also have the default value of 0. @evgeniuz can you add it?

evgeniuz:
Added timeout to config.

Issue #1071, opened 2018-08-22
BranchLimited policy holds global lock for too long


mossberg:
In this function, the global lock is held for a bit too long. it should just be held to retrieve the visited/summaries variables, and then release, because the shared state is not written to.

https://github.com/trailofbits/manticore/blob/38ecc25f9bacb51a253966474a2668a4f7704540/manticore/core/executor.py#L138-L150

PR #1187, opened 2018-10-13
Issue #1071  --> Global lock held for a long time

Srinivas11789:
**Fix for issue #1071**

**- Problem:**
  * Global lock need to be held until the required variables are fetched
  * Lock should be released while processing further

**- Fix:**
  * Scope the lock loop only until the variables are set

**- Uncertain:**
  * A similar scenario in `choice` function at Uncovered policy, should it be fixed as well?
  * Do we have to explicitly unlock?
  * Should there be any error handle for failures in lock loop?

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1187)

### DIFF
--- a/manticore/core/executor.py
+++ b/manticore/core/executor.py
@@ -148,14 +148,14 @@ class BranchLimited(Policy):
         with self.locked_context() as policy_ctx:
             visited = policy_ctx.get('visited', dict())
             summaries = policy_ctx.get('summaries', dict())
-            lst = []
-            for id_, pc in summaries.items():
-                cnt = visited.get(pc, 0)
-                if id_ not in state_ids:
-                    continue
-                if cnt <= self._limit:
-                    lst.append((id_, visited.get(pc, 0)))
-            lst = sorted(lst, key=lambda x: x[1])
+        lst = []
+        for id_, pc in summaries.items():
+            cnt = visited.get(pc, 0)
+            if id_ not in state_ids:
+                continue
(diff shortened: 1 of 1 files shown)


manticore/core/executor.py
```diff
@@ -148,14 +148,14 @@ def choice(self, state_ids):
...
-                cnt = visited.get(pc, 0)
-                if id_ not in state_ids:
-                    continue
-                if cnt <= self._limit:
-                    lst.append((id_, visited.get(pc, 0)))
-            lst = sorted(lst, key=lambda x: x[1])
+        lst = []
+        for id_, pc in summaries.items():
+            cnt = visited.get(pc, 0)
+            if id_ not in state_ids:
```
disconnect3d: This check can be moved to the begining of the loop

manticore/core/executor.py
```diff
@@ -148,14 +148,14 @@ def choice(self, state_ids):
...
+                continue
+            if cnt <= self._limit:
+                lst.append((id_, visited.get(pc, 0)))
+        lst = sorted(lst, key=lambda x: x[1])
```
disconnect3d: Since we are returning `lst[0][0]` at the end we should probably remove the list and keep track of the "first item" (based on `item[1]`).

Side note: when you have a sorted list and you want to insert into it and keep it being sorted, https://docs.python.org/3/library/bisect.html can be used.

PR #1188, opened 2018-10-14
String representation for Manticore EVM

### LINKED ISSUE #981: Implement ManticoreEVM.__str__
Implement ManticoreEVM.__str__ so it prints out a summary of the state list status.
```
<ManticoreEVM | Alive states: 19 ; Terminated states: 2109>
```

MJ10:
Resolves #981 

* Implemented the __str__ method to print no. of alive and terminated states using `count_running_states` and `count_terminated_states` methods. 
* Also implemented __repr__ which is used as a "formal" string representation.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1188)

### DIFF
--- a/manticore/ethereum/__init__.py
+++ b/manticore/ethereum/__init__.py
@@ -1548,10 +1548,7 @@ class ManticoreEVM(Manticore):
         pass
 
     def __repr__(self):
-        return "<ManticoreEVM | Alive States: {}; Terminated States: {}>".format(
-            self.count_running_states(),
-            self.count_terminated_states()
-        )
+        return self.__str__()
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


MJ10:
@disconnect3d I'll make repr call str. Or should they both haves separate messages to be displayed?

disconnect3d:
@MJ10 if you have an idea or think that one of them could show different things (or more things), probably just go for it.

Otherwise its fine to call one in another.

MJ10:
I'm can't really think of anything else to add, which would be helpful, as I don't have much experience with Manticore. For now I'll make the change, but I welcome suggestions from anyone else with more experience working with Manticore.

PR #1192, opened 2018-10-16
Fixes issue #170: Cannot control input size from cli

### LINKED ISSUE #170: Can't control the stdin input size from cli
Add cli commandline argument to control max stdin input size

Srinivas11789:
**Summary:** 
Fixes #170 --> add new cli argument for input_size and change function calls to include this argument

**Changes:**
* **Default value:** set default value for argument to 256 as hardcoded before
* **__main__.py:** add new argument for input_size 
* **make_linux, linux:** function call changes

**Clarification:**
* Default value for input_size to `256`? (as hardcoded before)
* Variable name set to `input_size`?
* The default value for argument is set at both argument `definition at arg parse` and `function definitions - make_linux, linux` which might be redundant? (make_linux call has a number of default arguments and linux function is a classmethod)?

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1192)

### DIFF
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -56,7 +56,7 @@ def parse_arguments():
                         help='Show program version information')
     parser.add_argument('--config', type=str,
                         help='Manticore config file (.yml) to use. (default config file pattern is: ./[.]m[anti]core.yml)')
-    parser.add_argument('--stdin_size', type=int, default=256,
+    parser.add_argument('--stdin_size', type=int, default=STDIN_INPUT_DEFAULT_SIZE,
                         help='Control the maximum symbolic stdin size')
 
     bin_flags = parser.add_argument_group('Binary flags')
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -3,7 +3,7 @@ import sys
 import logging
 import argparse
 
-from . import Manticore
+from . import Manticore, STDIN_INPUT_DEFAULT_SIZE
 from .utils import log, config
 
 # XXX(yan): This would normally be __name__, but then logger output will be pre-
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -56,8 +56,8 @@ def parse_arguments():
                         help='Show program version information')
     parser.add_argument('--config', type=str,
                         help='Manticore config file (.yml) to use. (default config file pattern is: ./[.]m[anti]core.yml)')
-    parser.add_argument('--input_size', type=int, default=256,
-                        help='Control the maximum input size')
+    parser.add_argument('--stdin_size', type=int, default=256,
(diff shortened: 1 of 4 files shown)


disconnect3d:
The default value of 256 is fine.

Can you add some info about the argument to the docs -  https://github.com/trailofbits/manticore/blob/66afb1f011d70f722bc00809d4eef355df39e35e/docs/syminput.rst#symbolic-stdin ?

@yan @feliam @mossberg 

Maybe we should rename `input_size` to `stdin_size` to be more explicit? (it also doesn't influence evm part :F)

> The default value for argument is set at both argument `definition at arg parse` and `function definitions - make_linux, linux` which might be redundant? (make_linux call has a number of default arguments and linux function is a classmethod)?

The value should be specified in the code only once imho. It would be nice to fetch it from a parser but it is not exposed outside of the function and also it would have to be done in super ugly way:
```
parser._option_string_actions['--input_size'].default
```

so I guess we can put a `STDIN_INPUT_DEFAULT_SIZE = 256` somewhere?

Srinivas11789:
@disconnect3d Thanks for the comments.
* I did not know about the argument doc, will update.
* Agree with `STDIN_INPUT_DEFAULT_SIZE`

Srinivas11789:
**Update:**
* **_Info about argument in doc:_** Made minimal changes to reflect that stdin_size is configurable. Let me know if I need to reformat the `Symbolic stdin` section
* Changed variable name to `stdin_size`
* The `stdin_size` argument needs to be present in the `linux` function as well as `Manticore` class functions in Manticore file - Added a global constant `STDIN_INPUT_DEFAULT_SIZE` for manticore module to set default input size < read only for default value >. Let me know your thoughts on this.
Also,
    - I could alter this global constant with default value from parser as you suggested from main.py `parser._option_string_actions['--input_size'].default`, but the flow looks ugly.
    - Currently, only place the `stdin_size` defaults are set are the argparse and `STDIN_INPUT_DEFAULT_SIZE`

Please let me know your thoughts.

manticore/__main__.py
```diff
@@ -56,6 +56,8 @@ def positive(value):
                         help='Show program version information')
     parser.add_argument('--config', type=str,
                         help='Manticore config file (.yml) to use. (default config file pattern is: ./[.]m[anti]core.yml)')
+    parser.add_argument('--stdin_size', type=int, default=256,
```
disconnect3d:
```suggestion
    parser.add_argument('--stdin_size', type=int, default= STDIN_INPUT_DEFAULT_SIZE,
```

Srinivas11789:
@disconnect3d Thanks.
I like that change. I was wondering to import from the manticore module as well but was unsure of the effects. Fixed init.

Issue #1209, opened 2018-10-22
solidity_create_contract may create contract with wrong constructor argument values


stephan-tolksdorf:
`SolidityMetadata.__init__` does not properly handle the case where multiple ABI items have no name, which happens when the contract has both a constructor and an explicitly defined fallback function. In that case the fallback function metadata will override the constructor metadata. This causes `get_constructor_arguments()` to return `'()'` irrespective of the actual constructor parameters. If the contract constructor actually has parameters, this leads `ManticoreEVM.solidity_create_contract` to pass the wrong type signature to `ABI.serialize`. Since `ABI._serialize_tuple` does not check the length of its argument tuples, `ABI.serialize` will not report an error and return without serializing any arguments. This leads to the transaction data missing the constructor arguments data, which has the same effect as if all arguments were zero. (If no `args` are passed to `solidity_create_contract`, symbolic arguments will be created according to the wrong, empty tuple signature. In that case the tuples passed to `ABI.serialize` have the same 0 length, but the end result is still the same.)

I'll submit a PR that fixes the constructor and fallback function handling in `SolidityMetadata.__init__` and also adds a tuple length check to `ABI._serialize_tuple`.

`SolidityMetadata` currently also doesn't properly handle overloaded functions, but fixing that requires an API change to the `abi` property. I'll open a separate issue for that.

stephan-tolksdorf:
@mossberg  This bug was also fixed by PR #1212

mossberg:
thanks!

PR #1211, opened 2018-10-22
Fix constructor and fallback function handling in Solidity metadata import

stephan-tolksdorf:
Also adds a tuple length check to `ABI._serialize_tuple`.

Fixes issue #1209.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1211)

### DIFF
--- a/manticore/ethereum/abi.py
+++ b/manticore/ethereum/abi.py
@@ -137,6 +137,8 @@ class ABI(object):
     def _serialize_tuple(types, value, dyn_offset=None):
         result = bytearray()
         dyn_result = bytearray()
+        if len(types) != len(value):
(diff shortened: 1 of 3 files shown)


mossberg:
thanks for the pr @stephan-tolksdorf ! since the `self.abi` change made by the PR will just be overwritten by #1212 and the other change to `_serialize_tuple` also already exists in #1212 do you think it would make sense to close this PR and just merge #1212 ?

stephan-tolksdorf:
It's a bit unfortunate that GitHub doesn't properly support PRs that depend on other PRs.

I agree that it's easier to just merge the other PR and will close this PR. I'll have to come up with another PR for Hacktoberfest then 😉

Issue #1210, opened 2018-10-22
SolidityMetadata doesn't import the ABI data for overloaded functions properly


stephan-tolksdorf:
The `abi` property of `SolidityMetadata` is currently a 1-1 mapping from the function name to the Solidity ABI metadata, even though functions can be overloaded (i.e. there can be multiple functions with the same name and different parameter lists). If the contract contains overloaded functions, only one function from each overload set will be represented in the `abi` dict. In that case  `get_abi`, `get_func_argument_types` and `get_func_return_types` may return wrong results.

The `abi` property currently will also store items for event declarations, which can be overloaded too.

The `abi` property seems to be part of the public API, though at least within the Manticore project it doesn't seem to be used outside the class.

I'll submit a PR with a proposed fix.

disconnect3d:
Can you show a minimal working example when the functions return wrong results?

stephan-tolksdorf:
This is one example case where the functions return the wrong result:

        m = ManticoreEVM()
        source_code = '''
        contract C {
            function f() public {}
            function f(uint a) public payable {}
        }
        '''
        user_account = m.create_account(balance=1000, name='user_account')
        contract_account = m.solidity_create_contract(source_code, owner=user_account, name='contract_account')
        md = m.get_metadata(contract_account)
        sel1 = ABI.function_selector('f()')
        sel2 = ABI.function_selector('f(uint256)')
        assert(sel1 != sel2)
        abi1 = md.get_abi(ABI.function_selector('f()')) # Actually returns the abi for f(uint256).
        abi2 = md.get_abi(ABI.function_selector('f(uint256)'))
        assert(abi1['type'] == 'function')
        assert(abi2['type'] == 'function')
        assert(abi1 != abi2) # This assert fails.

Note that overloaded functions aren't the only problem that the PR fixes. Currently the implementation e.g. also mistakes event items for function items.

Note that the PR for this issue builds on my previous PR for the related issue #1209. The PR for that issue contains a test case where without the PR a contract with the wrong data is constructed.

mossberg:
addressed #1212

PR #1212, opened 2018-10-22
Fix handling of overloaded functions and some other issues in SolidityMetadata

stephan-tolksdorf:
This draft PR is my proposed fix for issue #1210. It builds on PR #1211.

This PR currently

1. Rewrites `SolidityMetadata.__init__` to fix the handling of overloaded functions.
2. Separates the abi items for functions, events, constructor and fallback function into separate properties (though the event items are not yet publicly exposed).
3. Removes the `abi` property.
4. Removes the `add_function` method.
5. Adds a decorator for deprecating functions.
6. Deprecates the `functions` and `hashes` properties due to their problematic handling of the fallback function and adds `function_signatures` and `function_selectors` as replacements.
7. Deprecates the `get_hash` method, which seems redundant in its current form.
8. Adds static methods for constructing function and tuple type signatures that can handle ABIv2 tuple arguments.
9. Adds the `constructor_abi`, `has_non_default_constructor` and `has_non_default_fallback_function` properties.
10. Adds some docstrings and type annotations.

I can of course split this PR up or remove changes, change names, etc. if you'd like me to.

If you're OK with the direction of this PR, I could add more tests for `SolidityMetadata`.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1212)

### DIFF
--- a/manticore/ethereum/__init__.py
+++ b/manticore/ethereum/__init__.py
@@ -13,6 +13,7 @@ from ..core.state import State, TerminateState
 from ..utils.helpers import issymbolic, PickleSerializer
 from ..utils import config
 from ..utils.log import init_logging
+from typing import Dict, Optional
 import tempfile
 from subprocess import Popen, PIPE, check_output
 from multiprocessing import Process, Queue
@@ -99,21 +100,20 @@ class FilterFunctions(Plugin):
             #Let's compile  the list of interesting hashes
             selected_functions = []
 
-            for func_hsh in md.hashes:
-                if func_hsh == b'\0\0\0\0':
-                    continue
+            for func_hsh in md.function_selectors:
                 abi = md.get_abi(func_hsh)
-                func_name = md.get_func_name(func_hsh)
+                if abi['type'] == 'fallback':
+                    continue
                 if self._mutability == 'constant' and not abi.get('constant', False):
                     continue
                 if self._mutability == 'mutable' and abi.get('constant', False):
                     continue
-                if not re.match(self._regexp, func_name):
+                if not re.match(self._regexp, abi['name']):
                     continue
                 selected_functions.append(func_hsh)
 
-            if self._fallback:
-                selected_functions.append(b'\0\0\0\0')
+            if self._fallback and md.has_non_default_fallback_function:
+                selected_functions.append(md.fallback_function_selector)
 
             if self._include:
                 # constrain the input so it can take only the interesting values
@@ -121,7 +121,7 @@ class FilterFunctions(Plugin):
                 state.constrain(constraint)
             else:
                 #Avoid all not selected hashes
-                for func_hsh in md.hashes:
+                for func_hsh in md.function_selectors:
                     if func_hsh in selected_functions:
                         constraint = tx.data[:4] != func_hsh
                         state.constrain(constraint)
@@ -464,7 +464,7 @@ class ManticoreEVM(Manticore):
 
         self.constraints = ConstraintSet()
         self.detectors = {}
-        self.metadata = {}
+        self.metadata: Dict[int, SolidityMetadata] = {}
 
         # The following should go to manticore.context so we can use multiprocessing
         self.context['ethereum'] = {}
@@ -1220,7 +1220,7 @@ class ManticoreEVM(Manticore):
             for i in range(offset, offset + size):
                 code_data.add((state.platform.current_vm.address, i))
 
-    def get_metadata(self, address):
+    def get_metadata(self, address) -> Optional[SolidityMetadata]:
         """ Gets the solidity metadata for address.
             This is available only if address is a contract created from solidity
         """
--- a/manticore/ethereum/abi.py
+++ b/manticore/ethereum/abi.py
@@ -137,6 +137,8 @@ class ABI(object):
     def _serialize_tuple(types, value, dyn_offset=None):
         result = bytearray()
         dyn_result = bytearray()
+        if len(types) != len(value):
+            raise ValueError(f"The number of values to serialize is {'less' if len(value) < len(types) else 'greater'} than the number of types")
         for ty_i, value_i in zip(types, value):
             result_i, dyn_result_i = ABI._serialize(ty_i, value_i, dyn_offset + len(dyn_result))
             result += result_i
--- a/manticore/ethereum/solidity.py
+++ b/manticore/ethereum/solidity.py
@@ -183,7 +183,7 @@ class SolidityMetadata(object):
         The content of the returned dict is described at https://solidity.readthedocs.io/en/latest/abi-spec.html#json_
         """
         if not isinstance(hsh, (bytes, bytearray)):
-            raise ValueError('The selector argument must be a concrete byte array')
+            raise TypeError('The selector argument must be a concrete byte array')
         sig = self._function_signatures_by_selector.get(hsh)
         if sig is not None:
             return dict(self._function_abi_items_by_signature[sig])
@@ -200,7 +200,7 @@ class SolidityMetadata(object):
         the empty tuple type signature ``'()'`` is returned.
         """
         if not isinstance(hsh, (bytes, bytearray)):
-            raise ValueError('The selector argument must be a concrete byte array')
+            raise TypeError('The selector argument must be a concrete byte array')
         sig = self._function_signatures_by_selector.get(hsh)
         return '()' if sig is None else sig[sig.find('('):]
 
@@ -212,7 +212,7 @@ class SolidityMetadata(object):
         the empty tuple type signature ``'()'`` is returned.
         """
         if not isinstance(hsh, (bytes, bytearray)):
-            raise ValueError('The selector argument must be a concrete byte array')
+            raise TypeError('The selector argument must be a concrete byte array')
         abi = self.get_abi(hsh)
         outputs = abi.get('outputs')
         return '()' if outputs is None else SolidityMetadata.tuple_signature_for_components(outputs)
@@ -222,7 +222,7 @@ class SolidityMetadata(object):
         or ``'{fallback}'`` if no such function exists.
         """
         if not isinstance(hsh, (bytes, bytearray)):
-            raise ValueError('The selector argument must be a concrete byte array')
+            raise TypeError('The selector argument must be a concrete byte array')
         sig = self._function_signatures_by_selector.get(hsh)
         return '{fallback}' if sig is None else sig[:sig.find('(')]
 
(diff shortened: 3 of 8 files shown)


disconnect3d:
@stephan-tolksdorf hey

> This draft PR

If this is a `work in progress`, please add `[WIP]` to the title (and remove it later on + send a message that it is ready for review; also if you have any questions, feel free to ask).

stephan-tolksdorf:
@disconnect3d I've updated the title. 

I'd like to add more tests to this PR, but before doing so, it would be great if I could get some feedback on the current version of the PR. In particular I'd like to know whether you're OK with the scope of the PR and how I went about solving the issue.

manticore/ethereum/__init__.py
```diff
@@ -99,29 +100,28 @@ def will_open_transaction_callback(self, state, tx):
             #Let's compile  the list of interesting hashes
             selected_functions = []
 
-            for func_hsh in md.hashes:
-                if func_hsh == b'\0\0\0\0':
-                    continue
+            for func_hsh in md.function_selectors:
                 abi = md.get_abi(func_hsh)
-                func_name = md.get_func_name(func_hsh)
+                if abi['type'] == 'fallback':
```
disconnect3d: :+1: this seems much better than `func_hsh == b'\0\0\0\0'`

manticore/ethereum/__init__.py
```diff
@@ -99,29 +100,28 @@ def will_open_transaction_callback(self, state, tx):
...
+                    continue
                 if self._mutability == 'constant' and not abi.get('constant', False):
                     continue
                 if self._mutability == 'mutable' and abi.get('constant', False):
                     continue
-                if not re.match(self._regexp, func_name):
+                if not re.match(self._regexp, abi['name']):
```
disconnect3d: Loud thinking: is using regex here a good idea? It might slow down things a lot (since this is a callback).

@feliam @yan @mossberg what do u think?

manticore/ethereum/abi.py
```diff
@@ -137,6 +137,8 @@ def _serialize_bytes(value):
     def _serialize_tuple(types, value, dyn_offset=None):
         result = bytearray()
         dyn_result = bytearray()
+        if len(types) != len(value):
```
disconnect3d: We should probably use `if __debug__` in such cases as described here https://stackoverflow.com/a/27455446

manticore/ethereum/solidity.py
```diff
@@ -1,10 +1,36 @@
...
+        return name + SolidityMetadata.tuple_signature_for_components(inputs)
+
+    @staticmethod
+    def tuple_signature_for_components(components: Sequence[Mapping[str, Any]]) -> str:
+        """Equivalent to ``function_signature_for_name_and_inputs('', components)``."""
+        ts = []
+        for c in components:
+            t: str = c['type']
+            if t.startswith('tuple'):
+                assert len(t) == 5 or t[5] == '['
```
disconnect3d: Shouldn't it be `len(t) >= 5`? as we use `t[5:]` later on

manticore/ethereum/abi.py
stephan-tolksdorf: I've implemented the check the same way it's done in `_serialize_array`.

Personally I don't think it's the right choice for a tool like Manticore to trade the safety of an always-on check against the minuscule performance improvement from disabling an assert like this in release mode.

manticore/ethereum/solidity.py
stephan-tolksdorf: No, since in this branch `len(t)` is always ≥ 5 because `t` starts with 'tuple' and the right side of the `or` expression should only be evaluated if the `t` has more than 5 characters.

feliam:
Hi @stephan-tolksdorf thank you for this. This is a clear improvement towards handling more complex metadata and contracts.  We are going to quick test this PR on our internal scripts list to check that nothing goes super wrong and get back to you. 
BUt other than that I think it is going in a good direction.

manticore/ethereum/solidity.py
disconnect3d: Oh it is `or` here. That makes sense :).

stephan-tolksdorf:
@feliam That's good to hear, thank you!

Note that this PR contains also the commit from the previous PR #1211. If and when we get to the merging, it would be good if both PRs are merged in the right order.

manticore/ethereum/abi.py
mossberg: imo i think @stephan-tolksdorf 's implementation is fine, and mirroring `_serialize_array` is correct. I actually think asserts and this explicit error checking are semantically different, and not necessarily interchangeable -- this case is more handling bad input, rather than expressing an invariant so I think the explicit check is more semantically correct. i don't think we should care too much about perf in this case. :)

manticore/ethereum/abi.py
```diff
@@ -137,6 +137,8 @@ def _serialize_bytes(value):
...
+            raise ValueError(f"The number of values to serialize is {'less' if len(value) < len(types) else 'greater'} than the number of types")
```
mossberg: pls use EthereumError for this, we're trying to shift over to that

mossberg:
i debugged the test failures; some of the tests were passing a symbolic argument when the constructor didn't take one. we didn't previously error, but now we do, so the tests failed. this patch fixes it:

```
diff --git a/tests/eth_benchmark.py b/tests/eth_benchmark.py
index a953419..5291621 100644
--- a/tests/eth_benchmark.py
+++ b/tests/eth_benchmark.py
@@ -41,7 +41,7 @@ class EthBenchmark(unittest.TestCase):
         self.mevm=None
         shutil.rmtree(self.worksp)

-    def _test(self, name, should_find):
+    def _test(self, name, should_find, use_ctor_sym_arg=False):
         """
         Tests DetectInvalid over the consensys benchmark suite
         """
@@ -52,8 +52,12 @@ class EthBenchmark(unittest.TestCase):

         filename = os.path.join(THIS_DIR, 'binaries', 'benchmark', f'{name}.sol')

+        if use_ctor_sym_arg:
+            ctor_arg = (mevm.make_symbolic_value(),)
+        else:
+            ctor_arg = ()

-        mevm.multi_tx_analysis(filename, contract_name='Benchmark', args=(mevm.make_symbolic_value(),))
+        mevm.multi_tx_analysis(filename, contract_name='Benchmark', args=ctor_arg)

         expected_findings = set(( (c, d) for b, c, d in should_find))
         actual_findings = set(( (c, d) for a, b, c, d in mevm.global_findings))
@@ -66,10 +70,10 @@ class EthBenchmark(unittest.TestCase):
         self._test('assert_constructor', {(23, 'INVALID instruction', True)})

     def test_assert_multitx_1(self):
-        self._test('assert_multitx_1', set())
+        self._test('assert_multitx_1', set(), True)

     def test_assert_multitx_2(self):
-        self._test('assert_multitx_2', {(150, 'INVALID instruction', False)})
+        self._test('assert_multitx_2', {(150, 'INVALID instruction', False)}, True)

     def test_integer_overflow_minimal(self):
         self._test('integer_overflow_minimal', {(163, 'Unsigned integer overflow at SUB instruction', False)})

```

mossberg:
i went ahead and applied the diff to the PR branch.

manticore/ethereum/abi.py
stephan-tolksdorf: @mossberg Good point regarding the difference of checking preconditions on inputs vs checking internal invariants!

manticore/ethereum/abi.py
stephan-tolksdorf: The rest of the code in that file currently uses `ValueError` and I thought that keeping the code locally consistent might be preferable. 

In general, I think it could be quite helpful to use separate exception hierarchies for (Python) programming errors and errors that arise during the normal EVM execution (runtime errors due to e.g. IO errors would be third category). Since `EthereumError` is a subclass of `ManticoreError`, using it for all exceptions in the Ethereum code base increases the risk that programming errors are inadvertently masked. 

For example, currently the Executor logs and suppresses all exceptions, even asserts, which can make it more difficult to diagnose errors and test fixes. Using separate exception hierarchies could make it easier to let exceptions caused by programming errors bubble up to the parent Manticore process/instance.

stephan-tolksdorf:
@mossberg Thank you!

Regarding deprecations: You could have a policy that all deprecated methods are removed in the next release, or whenever they become a maintenance burden. I like the deprecation notices because they both serve as documentation for the API changes and make it easier for users to upgrade their code bases, especially in cases like these where the property names are quite common and there are subtle differences in behaviour between the old and the new methods. If you still want them to be removed, I can do that of course.

mossberg:
> We are going to quick test this PR on our internal scripts to check that nothing goes super wrong and get back to you.

Internal scripts ran just fine with this 👍

stephan-tolksdorf:
In the commits I just pushed I've fixed the `get_func_argument_types` implementations, added type checks for the function selector arguments and added more tests.

The type checks use `ValueError`, like in `parse_tx`. If @mossberg confirms that he prefers `EthereumError` everywhere, I'll replace `ValueError` with `EthereumError`.

stephan-tolksdorf:
I'll take a look at the failing `eth_detectors.py` tests.

mossberg:
@stephan-tolksdorf thanks for the clear explanations :)

re: Errors: I retract my opinion on EthereumError vs builtin errors. I've thought about it more, and now think that since the errors are very "python-y" and _not domain specific_, it does make sense for us to raise a builtin error here vs an error from our custom hierarchy. Furthermore, these are not errors that should ever be caught/handled really...they indicate improper API usage that should be corrected for the right types. For these reasons I think builtin errors are the way to go here. I will say that for all the error checks that are type/isinstance related, `TypeError` is probably a better one to use than `ValueError`. thanks for making me thing more deeply on this!

re: Deprecation: I think you also make great points. It's not as if the maintenance burden of carrying deprecated functions is high, and it's probably worth it for the end user experience. let's leave them in.

> I'll take a look at the failing `eth_detectors.py` tests.

From my brief look, these failures might be the exact same issues that my previous patch above fixed , just in a different set of tests. (various instances in the tests where we provide ctor arguments to contracts that do not take any). I should have actually run the full eth tests locally 😭 it's 1am in my time zone so i can't look at it now, but they should be straightforward to fix. once tests are passing, lgtm to merge.

stephan-tolksdorf:
I agree that `TypeError` is more appropriate than `ValueError` for the type checks. Thanks for spotting that! I've replaced all `ValueError` uses in `solidity.py`. 

You were also right about the `eth_detectors.py` tests, which I've fixed the same way you fixed the `eth_benchmark.py` tests.

PR #1214, opened 2018-10-23
Convenient api for asserting invariant and generating testcase

mossberg:
A common use case for ethereum is checking that a certain condition is always true in a state, and generating a testcase if this is not so. Currently users need to implement this themselves using the various state methods (`state.can_be_true` etc) and `state.generate_testcase`, but since this is expected to be very common, we should implement it in the core.

Discussion 

- Is the API backwards? Should you specify the condition you want to never be able to happen?
- naming: is `assert_` a poor name? what would be better? verify, ensure, check, ... ?

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1214)

### DIFF
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -301,6 +301,23 @@ class State(Eventful):
     def is_feasible(self):
         return self.can_be_true(True)
 
+    # alternate names:
+    # state.check
+    # state.verify
+    # state.ensure
+    # ...
+    def assert_(self, expression, testcase_message, testcase_name='Assertion Failed'):
+        """
+        Check that an expression is always true, and if it is not always true, generate a testcase where the
+        expression is false.
+
+        :param condition: Expression (condition) to be asserted as always true
+        """
+        with self as temp_state:
+            temp_state.constrain(expression == False)
+            if temp_state.is_feasible():
+                temp_state.generate_testcase(name=testcase_name, message=testcase_message)
+
     def can_be_true(self, expr):
         expr = self.migrate_expression(expr)
         return self._solver.can_be_true(self._constraints, expr)


mossberg:
blocked on #1215 merge

manticore/core/state.py
```diff
@@ -301,6 +301,23 @@ def migrate_expression(self, expression):
     def is_feasible(self):
         return self.can_be_true(True)
 
+    # alternate names:
+    # state.check
+    # state.verify
+    # state.ensure
+    # ...
+    def assert_(self, expression, testcase_message, testcase_name='Assertion Failed'):
```
feliam: assert, verify, ensure kind of suggest to me something is done to the state, no?

This is not modifying the current state. We may even put this as a helper?
manticore.generate_testcase_if_fail(state, constraint, message)

(or if it breaks lots of scripts carrying the manticore back link in all states as we discussed)

mossberg:
Writing up some more details on this:

Context: The main way Manticore ethereum scripts generally work is: 1. Setup custom blockchain/contracts state in Manticore. 2. Submit symbolic transactions. 3. Iterate through all discovered blockchain states and verify that a certain invariant can never be violated / a certain invariant is always true in the state. This PR focuses on 3 in this list.

Here's some example code that illustrates what you effectively see at the end of basically all mcore eth scripts:

```
m = ManticoreEVM()
...
for state in m.running_states:
  balance = ...  # some symbolic variable that is extracted from the state

  # we want to verify that the balance can never be 0 for example
  with state as temp_state:
    temp_state.constrain(balance == 0)
    if temp_state.is_feasible():  # check if it's possible for the assertion to be violated
      m.generate_testcase('balance could be 0!!!! :( :( :(')  # generate testcase if it can be violated
```

the script needs to do 2 things:
1. check if it's possible for the assertion to be violated (very natural to do this using the state.can_be_true etc APIs)
2. generate a testcase if the assertion can be violated

This is an example of the client needing to do something the module can do, a violation of the "Don't Make the Client Do Anything the Module Could Do" API design principle [(ref)](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/32713.pdf).

this is a commonly used enough pattern such that it warrants a convenience API that does both of these things. thus, the vision is something like:


```
m = ManticoreEVM()
...
for state in m.running_states:
  balance = ...  # some symbolic variable that is extracted from the state

  # we want to verify that the balance can never be 0 for example
  state.assert_(balance != 0, 'balance could be 0!!!')  # checks that balance != 0 must be true for this state, and generates a testcase if this is not the case
```

where we have a single API for this commonly desired action.

the proposed method involves implementing the API on the `State` class. this requires the State to be able to internally trigger generation of testcases. This ultimately requires execution of `Manticore` level code, however State is at a deeper level of the hierarchy. 

Currently `State.generate_testcase` exists which effectively indirectly calls `ManticoreEVM._generate_testcase_callback` via the event system -- emitting the `will_generate_testcase` event. The PR currently implements this approach.

However, this is not desirable because the event system should actually not be enabled when the executor loop is not running (reasons for which are slightly tangential), which it is not at the point the client script is doing the `for state in m.running_states` thing.

So, I'm planning on rewriting this PR to allow State to generate a testcase in a different way. Rather than using the event system to execute the Manticore level code, we can add a back reference from the State to the Manticore/ManticoreEVM object. then State.generate_testcase can look like:

```
def generate_testcase(self, name, message):
  self._manticore._generate_testcase_callback(...)
```

However, this back reference does feel a little messy imo... It couples the State with the Manticore. One should be able to create a state standalone without having to have an 'owning' Manticore object?

mossberg:
invalidated by #1221 - this API as is does not make it very obvious that the workspace will be written to.

PR #1215, opened 2018-10-23
Copy event system state during state fork

mossberg:
We previously did not copy the event system state such as

- `_signals`: the various event subscriptions
- `_forwards`: the various objects that events needed to be forwarded to

when a State object was copied.

This causes bugs like `State.generate_testcase` not working because the events are not forwarded to the `Executor` class.

This also adds a `forward_events_from` call in one place we forgot it.

This does a full copy of the `_signals` and `_forwards` so further mutation of the child state does not affect the parent.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1215)

### DIFF
--- a/manticore/utils/event.py
+++ b/manticore/utils/event.py
@@ -157,6 +157,6 @@ class Eventful(object, metaclass=EventsGatherMetaclass):
             raise TypeError
         self._forwards[sink] = include_source
 
-    def copy_event_state(self, new_object: Eventful):
+    def copy_eventful_state(self, new_object: Eventful):
         new_object._forwards = copy.copy(self._forwards)
         new_object._signals = copy.copy(self._signals)
--- a/manticore/utils/event.py
+++ b/manticore/utils/event.py
@@ -157,6 +157,6 @@ class Eventful(object, metaclass=EventsGatherMetaclass):
             raise TypeError
         self._forwards[sink] = include_source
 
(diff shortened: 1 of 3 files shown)


mossberg:
now that #1214 is closed we could close this. however, it still stands that we might want this bug fix. i think there should definitely be buggy behavior if the forwards are not copied in a forked state -- i would expect that signals fired from the cpu/platform would not be able to reach Manticore/executor. however, i am 99% sure that has worked in the past/works currently. so i would need to debug further why exactly those work, given that the forwards are not copied in forked states.

mossberg:
@feliam any thoughts on this? irrespective of whether `State.generate_testcase` should even exist, currently due to this bug, the State cannot publish events and expect Manticore/Executor to be able to handle them.

manticore/utils/event.py
```diff
@@ -151,3 +156,7 @@ def forward_events_to(self, sink, include_source=False):
         if not isinstance(sink, Eventful):
             raise TypeError
         self._forwards[sink] = include_source
+
+    def copy_eventful_state(self, new_object: 'Eventful'):
```
disconnect3d:
```suggestion
    def copy_eventful_state(self, new_object: Eventful):
```

isn't it working this way?

mossberg: nope, Eventful is not declared yet

PR #1216, opened 2018-10-23
Rebased: add test for "arguments_linux_armv7" binary

### LINKED ISSUE #554: tests: add test for "arguments_linux_armv7" binary

disconnect3d:
This is rebased version of #961 and it should fix #554

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1216)


disconnect3d:
Hmm the assertion fails on travis while it works on my machine. Maybe it is bcoz of different libc versions? Not sure, but the expected set has more items than the actual one.

Gonna investigate it more.

scripts/travis_test.sh
```diff
@@ -5,6 +5,11 @@ RV=0
 set -o errexit
 set -o pipefail
 
+# vv remove this
+coverage run -m unittest discover tests/ -p test_binaries.py 2>&1 >/dev/null | tee travis_tests.log
+exit $?
+# ^^ REMOVE THIS
```
disconnect3d: todo/fixme: remove this ;p

PR #1218, opened 2018-10-24
Improve exceptions and errors

mossberg:
Use more specific, standard errors in some places.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1218)


disconnect3d:
@mossberg some tests fails here :s

Issue #731, opened 2018-02-02
preconstrain tx.data in evm symbolic transactions


mossberg:
currently, manticore's evm analysis redundantly explores the dispatch stub on every symbolic transaction. this results in those `REVERT` states in the second symbolic TX. after the first symbolic Tx, manticore knows all the paths that lead to code. we should preconstrain the first 4 bytes of tx.data in all following transactions to avoid this redundant exploration, which is both inefficient, and confusing UX.

```
2018-02-02 13:38:32,425: [36572] m.main:INFO: Beginning analysis
2018-02-02 13:38:32,515: [36572] m.ethereum:INFO: Starting symbolic transaction: 1
2018-02-02 13:38:32,831: [36572] m.ethereum:INFO: Generated testcase No. 0 - REVERT
2018-02-02 13:38:36,379: [36572] m.ethereum:INFO: Generated testcase No. 1 - REVERT
2018-02-02 13:38:39,795: [36572] m.ethereum:INFO: Finished symbolic transaction: 1 | Code Coverage: 55% | Terminated States: 3 | Alive States: 1
2018-02-02 13:38:39,796: [36572] m.ethereum:INFO: Starting symbolic transaction: 2
2018-02-02 13:38:40,192: [36572] m.ethereum:INFO: Generated testcase No. 2 - REVERT
2018-02-02 13:38:44,649: [36572] m.ethereum:INFO: Generated testcase No. 3 - REVERT
2018-02-02 13:38:49,734: [36572] m.ethereum:WARNING: Integer overflow at ADD instruction
2018-02-02 13:38:49,861: [36572] m.ethereum:INFO: Finished symbolic transaction: 2 | Code Coverage: 100% | Terminated States: 6 | Alive States: 2
2018-02-02 13:38:49,881: [36572] m.ethereum:INFO: Generated testcase No. 4 - STOP
2018-02-02 13:38:55,076: [36572] m.ethereum:INFO: Generated testcase No. 5 - STOP
```

mossberg:
we should revisit this now. i think FilterFunctions could be something similar to this?

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 200.0 DAI (200.0 USD @ $1.0/DAI)  attached to it__ as part of the Ethereum Community Fund via ECF Web 3.0 Infrastructure Fund fund__.__

[...]

Issue #1219, opened 2018-10-24
ethereum.ABI incorrectly (de)serializes certain types


stephan-tolksdorf:
The current implementation of `ethereum.ABI` does not properly differentiate between ['dynamic' and 'static' types](https://solidity.readthedocs.io/en/latest/abi-spec.html#formal-specification-of-the-encoding). In particular, it incorrectly treats fixed-size arrays of non-dynamic elements as dynamic types and it treats tuples with a dynamic element as non-dynamic types.

Fixing this completely requires adding new grammar rules to the YACC grammar in `abitypes.py` and making the corresponding changes to `abi.py`.

This a simple example where the invalid serialization leads to wrong results:
```
from manticore.ethereum import ManticoreEVM, ABI
from manticore.core.smtlib.visitors import to_constant

m = ManticoreEVM(procs=4) # initiate the blockchain
source_code = '''
pragma solidity ^0.4.24;
contract Trivial {
    function f(uint a) public returns (uint) {
        return a;
    }
    function f1(uint[1] a) public returns (uint) {
        return a[0];
    }
}
'''
creator_account = m.create_account(balance=1000)
contract = m.solidity_create_contract(source_code, owner=creator_account)

# Correctly prints 123.
contract.f(123)
print(ABI.deserialize('uint', to_constant(m.world.transactions[-1].return_data)))

# Incorrectly prints 32 (the offset of the data of the incorrectly dynamically encoded
# array from the start of the encoded arguments).
contract.f1([123])
print(ABI.deserialize('uint', to_constant(m.world.transactions[-1].return_data)))
```

mossberg:
thanks for filing this 👍

PR #1220, opened 2018-10-24
Preconstrain the data of symbolic human transactions

stephan-tolksdorf:
This PR implements a feature (enabled by default) to preconstrain the data of symbolic human transactions to avoid triggering uninteresting exceptions in the Solidity contract function dispatcher.

This PR fixes issue #731.

I've implemented the feature directly in `ManticoreEVM` and not e.g. as a plugin or in `EVMWorld` because there I also can constrain the transaction value depending on the function selector and thus avoid the 'not enough funds' fork in `EVMWorld._process_pending_transaction` for states that would lead to immediate REVERTs anyway. Doing it there also has the advantage that the constraint expression only is constructed a single time, not once for every running state.

With this PR applied, Manticore avoids all 6 states leading to REVERTs when it is run with the default options on `examples/evm/simple_int_overflow.sol` and covers all interesting code path with only 3 states.

The new behaviour can be disabled with the `--explore-contract-function-dispatcher` command line flag or by setting the public `preconstrains_symbolic_tx_data` property on `ManticoreEVM`.

With this PR applied, the `eth_general.test_emit_did_execute_end_instructions` test fails. This could be fixed by setting `preconstrains_symbolic_tx_data` to false. However, the [doccomment in that test](https://github.com/trailofbits/manticore/blob/master/tests/eth_general.py#L616) makes me think that it uses the wrong `sol` test file and was actually meant to be run on a test case similar to `examples/evm/simple_int_overflow.sol`.

Another test that fails is [`eth_detectors.test_delegatecall_not_ok`](https://github.com/trailofbits/manticore/blob/master/tests/eth_detectors.py#L212). I believe that the test is incorrect, because the [delegatecall in the fallback function](https://github.com/trailofbits/manticore/blob/master/tests/binaries/detectors/delegatecall_not_ok.sol#L24) could only ever call itself, since by definition the fallback function is only reached if the function selector in the transaction data doesn't match any selector of a normal function.

While playing with possible fixes to the `delegatecall_not_ok` test I noticed that when preconstraining the transaction data, Manticore is not able to find a non-throwing path for the following simple contract
```
contract C {
    function f(bytes b) {
       require(b.length == 64);
    }
}
```

I suspect the problem here is the sampling strategy for the `calldataload` instruction, but I haven't looked into this any further.

One way to improve Manticore's behaviour for this and similar test cases would be to also preconstrain the offsets of function arguments with the 'dynamic' ABI encoding, but I'll probably have to leave that for a separate PR. (Doing that is also partially blocked by issue #1219).

I'd greatly appreciate any feedback on the implementation approach, the naming and default values of the configuration options and the best way to deal with the test failures. (CC @mossberg @feliam @disconnect3d)

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1220)

### DIFF
--- a/tests/eth_general.py
+++ b/tests/eth_general.py
@@ -665,21 +665,18 @@ class EthTests(unittest.TestCase):
         class TestDetector(Detector):
             def did_evm_execute_instruction_callback(self, state, instruction, arguments, result):
                 if instruction.is_endtx:
-                    with self.locked_context('insns', dict) as d:
-                        d[instruction.semantics] = True
+                    with self.locked_context('endtx_instructions', set) as d:
+                        d.add(instruction.name)
 
         mevm = self.mevm
         p = TestDetector()
         mevm.register_detector(p)
 
-        filename = os.path.join(THIS_DIR, 'binaries/int_overflow.sol')
-        mevm.multi_tx_analysis(filename, tx_limit=2)
+        filename = os.path.join(THIS_DIR, 'binaries/simple_int_overflow.sol')
+        mevm.multi_tx_analysis(filename, tx_limit=2, tx_preconstrain=True)
 
-        self.assertIn('insns', p.context)
-        context = p.context['insns']
-        self.assertIn('STOP', context)
-        self.assertIn('RETURN', context)
-        self.assertIn('REVERT', context)
+        self.assertIn('endtx_instructions', p.context)
+        self.assertSetEqual(p.context['endtx_instructions'], {'INVALID', 'RETURN', 'STOP'})
 
     def test_call_with_concretized_args(self):
         """Test a CALL with symbolic arguments that will to be concretized.
@@ -918,6 +915,28 @@ class EthTests(unittest.TestCase):
 
         context = p.context.get('flags', {})
         self.assertTrue(context.get('found', False))
+
+    def test_preconstraints(self):
+        source_code = '''
+        contract C {
+            constructor() public {}
+            function f0() public {}
+            function f1(uint a) public payable {}
+        }
+        '''
+        m: ManticoreEVM = self.mevm
+
+        creator_account = m.create_account(balance=1000)
+        contract_account = m.solidity_create_contract(source_code, owner=creator_account, balance=0)
+
+        data = m.make_symbolic_buffer(320)
+        value = m.make_symbolic_value()
+        m.constrain(m.preconstraint_for_call_transaction(address=contract_account, data=data, value=value))
+        m.transaction(caller=creator_account, address=contract_account, data=data, value=value)
+
+        results = [state.platform.all_transactions[-1].result for state in m.all_states]
+        # The TXERROR indicates a state where the sent value is greater than the senders budget.
+        self.assertListEqual(sorted(results), ['STOP']*2 + ['TXERROR'])
 
 class EthHelpersTest(unittest.TestCase):
     def setUp(self):
--- a/manticore/__main__.py
+++ b/manticore/__main__.py
@@ -116,6 +116,9 @@ def parse_arguments():
     eth_flags.add_argument('--txaccount', type=str, default="attacker",
                            help='Account used as caller in the symbolic transactions, either "attacker" or "owner"')
 
+    eth_flags.add_argument('--txpreconstrain', action='store_true',
+                           help='Constrain human transactions to avoid exceptions in the contract function dispatcher')
+
     eth_flags.add_argument('--contract', type=str,
                            help='Contract name to analyze in case of multiple contracts')
 
--- a/manticore/ethereum/cli.py
+++ b/manticore/ethereum/cli.py
@@ -48,7 +48,9 @@ def ethereum_main(args, logger):
     logger.info("Beginning analysis")
 
     with m.shutdown_timeout(args.timeout):
-        m.multi_tx_analysis(args.argv[0], contract_name=args.contract, tx_limit=args.txlimit, tx_use_coverage=not args.txnocoverage, tx_send_ether=not args.txnoether, tx_account=args.txaccount)
+        m.multi_tx_analysis(args.argv[0], contract_name=args.contract, tx_limit=args.txlimit,
+                            tx_use_coverage=not args.txnocoverage, tx_send_ether=not args.txnoether,
+                            tx_account=args.txaccount, tx_preconstrain=args.txpreconstrain)
 
     # TODO unregister all plugins
 
--- a/manticore/ethereum/manticore.py
+++ b/manticore/ethereum/manticore.py
@@ -5,7 +5,7 @@ import string
 from multiprocessing import Queue, Process
 from queue import Empty as EmptyQueue
 from subprocess import check_output, Popen, PIPE
-from typing import Dict, Optional
+from typing import Dict, Optional, Union
 
 import io
 import os
@@ -16,7 +16,7 @@ import sha3
 import tempfile
 
 from ..core.manticore import ManticoreBase
-from ..core.smtlib import ConstraintSet, Array, ArrayProxy, BitVec, Operators
+from ..core.smtlib import ConstraintSet, Array, ArrayProxy, BitVec, Operators, BoolConstant, BoolOperation, Expression
 from ..core.state import TerminateState, AbandonState
 from .account import EVMContract, EVMAccount, ABI
 from .detectors import Detector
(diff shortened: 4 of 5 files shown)


manticore/ethereum/__init__.py
```diff
@@ -931,6 +963,9 @@ def _transaction(self, sort, caller, value=0, address=None, data=None, gaslimit=
             if '_pending_transaction' in state.context:
                 raise EthereumError("This is bad. It should not be a pending transaction")
 
+            if preconstraint is not None:
```
mossberg: this could run into issues if the first four bytes are already constrained to 1 solution, or are effectively concrete, for example if the API is used to execute a function with symbolic arguments

```
contract.func(m.make_symbolic_value())
```

will result in a tx data that is an ArrayProxy (and thus `issymbolic()`), even thought the first four bytes are really concrete. if we add contradicting constraints on the first four bytes (such as a function preconstrain), that will result in an infeasible state, and solver errors raised later.

we can semi-reliably detect if the func has is already constrained by checking if there can be more than 1 solution. something like

```
# ask for 2 solutions. if we only get one back, there can only be 1.
hash_not_constrained = len(state.solve_n(data[:4], 2)) > 1
if hash_not_constrained:
   # hash not constrained, it's safe to do the preconstrain stuff
```

manticore/ethereum/__init__.py
```diff
@@ -917,6 +927,28 @@ def _transaction(self, sort, caller, value=0, address=None, data=None, gaslimit=
             if len(data) == 0:
                 raise EthereumError("An initialization bytecode is needed for a CREATE")
 
+            if contract_metadata is not None and issymbolic(value) and self.preconstrains_symbolic_tx_data \
+                and not contract_metadata.constructor_abi['payable']:
+                    preconstraint = value == 0
+        else:
```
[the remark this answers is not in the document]
stephan-tolksdorf: Yeah, this case-by-case handling doesn't need to be exhaustive, so `elif sort == 'CALL'` without an assert seems better.

manticore/ethereum/__init__.py
```diff
@@ -447,14 +448,15 @@ def contract_accounts(self):
...
-    def __init__(self, procs=10, **kwargs):
+    def __init__(self, procs=10, preconstrain_symbolic_tx_data=True, **kwargs):
         """ A Manticore EVM manager
             :param int procs: number of workers to use in the exploration
         """
         self._accounts = dict()
         self._serializer = PickleSerializer()
 
         self._config_procs = procs
+        self.preconstrains_symbolic_tx_data = preconstrain_symbolic_tx_data
```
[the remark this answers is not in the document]
stephan-tolksdorf: I used to the -s form here to make it obvious that it is a boolean property and not a method that does something. Would you rather stick to the simple infinitive form?

manticore/ethereum/__init__.py
```diff
@@ -917,6 +927,28 @@ def _transaction(self, sort, caller, value=0, address=None, data=None, gaslimit=
             if len(data) == 0:
                 raise EthereumError("An initialization bytecode is needed for a CREATE")
 
+            if contract_metadata is not None and issymbolic(value) and self.preconstrains_symbolic_tx_data \
+                and not contract_metadata.constructor_abi['payable']:
+                    preconstraint = value == 0
+        else:
+            assert sort == 'CALL'
+            if not issymbolic(address) and issymbolic(data) and len(data) > 4 and self.preconstrains_symbolic_tx_data:
```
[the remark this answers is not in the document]
stephan-tolksdorf: I'll refactor this.

manticore/ethereum/__init__.py
stephan-tolksdorf: The constructed `preconstraint` is a logical disjunction over all possible function selectors (see the `preconstraint = Operators.OR(preconstraint, c)` line),  so I don't think that the implementation could add a contradicting constraint here, unless the concrete function has no entry in the `SolidityMetadata`, which would probably be a programming error.

Invoking the solver here might be bad for performance.

stephan-tolksdorf: I could try to detect simple cases where the selector is concrete without invoking the selector. 

An alternative might be to special-case the `transaction` call here https://github.com/trailofbits/manticore/blob/2a634b707c664f8d0150afa100e280fc55b5408d/manticore/ethereum/account.py#L133-L136

manticore/ethereum/__init__.py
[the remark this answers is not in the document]
mossberg: i see; i still think i prefer the simple infinitive. i think it's nice to have the formal constructor parameter match in name to the attribute it is written to. i also think the ambiguity is not too severe -- if there was confusion about what `preconstrain_symbolic_tx_data` is, it's not too hard to grep and find out

manticore/ethereum/__init__.py
mossberg: > The constructed `preconstraint` is a logical disjunction over all possible function selectors (see the `preconstraint = Operators.OR(preconstraint, c)` line), so I don't think that the implementation could add a contradicting constraint here, unless the concrete function has no entry in the `SolidityMetadata`, which would probably be a programming error

Ah, you're totally right, i retract my comment. when i wrote that i was incorrectly thinking that this preconstraint was constraining it to a specific function. so i guess even if the tx hash bytes were already constrained/concrete, adding this disjunction constraint would technically be redundant, but not do any harm.

stephan-tolksdorf:
@mossberg Thank you for the review! I've moved most of the implementation into a separate method and renamed the `preconstrains_...` attribute. The new implementation now also avoids adding the preconstraint in simple cases where a symbolic data buffer has a concrete function selector, e.g. when the transaction was created by the python wrapper in `EVMContract.__getattribute__`.

stephan-tolksdorf:
I've fixed the tests (following the rationale from my first comment above). 

Replacing `int_overflow.sol` with `simple_int_overflow.sol` in `test_emit_did_execute_end_instructions` also has the benefit of reducing the runtime of the `eth_general.py` tests by about 10s on my laptop.

I used the opportunity to make the `eth_general` import in `eth_detectors.py` a relative import. I hope that's OK.

stephan-tolksdorf:
@disconnect3d Thank you for the review! I'll look into moving the implementation into a plugin tomorrow.

I saw that @feliam created the ticket #1231 for the issues related to 'dynamic' ABI args that I briefly discussed at the end of my initial comment above. I suspect that preconstraining such arguments might be essential for allowing Manticore to efficiently and reliably explore the bodies of functions with such arguments. Hence, the code for generating these argument preconstraints arguably would belong into the core ManticoreEVM implementation and then probably should be colocated with the remaining code related to preconstraints. Did you discuss this aspect?

stephan-tolksdorf:
The current implementation only creates the preconstraint once and then adds this constraint to each running state at the time when the state is temporarily loaded for the constraint variable migration.

In order to make a plugin-based implementation similarly efficient, we'd need to emit a 'start_human_transaction' event before the variable migration loop and then emit another event during each iteration of the migration loop. The plugin would need to retain state between the events. 

If the plugin should do all the work in the 'start_human_transaction' event handler, it would either need to load and save each running state itself, which would duplicate the IO work in the migration loop, or `_transaction` would need to load all states into memory at the same time and pass them to the event subscribers as an array, which could balloon the memory usage of the main process.

Splitting the work between e.g. a `start_human_transaction` event on the main process and subsequent per-state events that are published in the worker processes would require copying the preconstraint to each worker process, e.g. via the shared context.

Do you prefer one of these plugin implementation approaches, or do you maybe have a better idea?

feliam:
Can not we just do a `m.constraint(_precondition_)` at main using metadata or info from previous tx depending on availability?

The fact that `m.constraint(_precondition_)` is eager and loads all the states in place and add the constrain to each on the spot is a different problem. 

```
Main():
    ...
    m.constrain(input[:4] == $funcid1 or input[:4] == $funcid2 or input[:4] == $fallback_chosen_id)
    #depending on payable  funcid==notpayable_function_id => value := 0.
    m.constrain(value ==0 or (input[:4] != $funcid1 and input[:4] != $funcid2));

**Comment by stephan-tolksdorf** [comment]
@feliam I've refactored the implementation such that preconstraints have to be explicitly constructed. Is this what you had in mind?

**Comment by feliam** [comment]
Yes this is similar to what I was thinking!
So this avoids sending ether to not-payable functions. lgtm

File: tests/eth_general.py [hunk 5]
```diff
@@ -919,6 +916,28 @@ def will_decode_instruction_callback(self, state, pc):
...
+        contract_account = m.solidity_create_contract(source_code, owner=creator_account, balance=0)
+
+        data = m.make_symbolic_buffer(320)
+        value = m.make_symbolic_value()
+        m.constrain(m.preconstraint_for_call_transaction(address=contract_account, data=data, value=value))
+        m.transaction(caller=creator_account, address=contract_account, data=data, value=value)
+
+        results = [state.platform.all_transactions[-1].result for state in m.all_states]
+        # The TXERROR indicates a state where the sent value is greater than the senders budget.
+        self.assertListEqual(sorted(results), ['STOP']*2 + ['TXERROR'])
```
disconnect3d: Is there a possibility to assert applied constraints?

stephan-tolksdorf: I could try to pattern match the constraint returned from `preconstraint_for_call_transaction`, but that would be a lot of effort and would make the test sensitive to implementation details of the constraint system. If the constraint was constructed wrongly, the state results would very likely include a `REVERT` or miss one of the items that the `assertListEqual` looks for.

disconnect3d: Ok, that makes sense. :+1:

disconnect3d:
Thanks! <3

PR #1221, opened 2018-10-24
Add convenience assertion & generate API

mossberg:
This invalidates #1214.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1221)

### DIFF
--- a/manticore/ethereum/__init__.py
+++ b/manticore/ethereum/__init__.py
@@ -1256,8 +1256,24 @@ class ManticoreEVM(Manticore):
     def workspace(self):
         return self._executor._workspace._store.uri
 
-    def generate_testcase(self, state, name, message=''):
-        self._generate_testcase_callback(state, name, message)
+    def generate_testcase(self, state, name, message='', condition=None):
+        """
+
+        :param state:
+        :param name:
+        :param message:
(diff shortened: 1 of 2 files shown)


manticore/ethereum/__init__.py
```diff
@@ -1256,8 +1256,22 @@ def unregister_detector(self, d):
...
+        :param only_if: if expr can be true, generate testcase. if is None, generate testcase unconditionally.
+        :return:
+        """
+        if only_if is None:
+            self._generate_testcase_callback(state, name, message)
+        else:
+            with state as temp_state:
+                temp_state.constrain(only_if)
+                if temp_state.is_feasible():
+                    self._generate_testcase_callback(temp_state, name, message)
```
feliam: Should it generate the testcase with the only_if constrained added or not?

mossberg: I think it should generate with the only_if constrained. For example say you want to check that the "balance cannot be zero".

```
m.generate_testcase(state, 'fail', 'balance CAN BE ZERO :O', only_if=balance == 0)
```

then i want to see the actual input that makes the balance == 0 in order to really see it with my own eyes, and also debug 

sound ok?

Issue #1103, opened 2018-08-31
Add special exception for abandoning a state


feliam:
This relies on the exception message. Refactor. Special exception or otherwise.
https://github.com/trailofbits/manticore/blob/cf72c12fd6a35ff9e62408fa23e033b94f804ceb/manticore/ethereum.py#L2201

mossberg:
We could make an `AbandonState(TerminateState)` exception.

PR #1222, opened 2018-10-24
Add AbandonState exception

ex0dus-0x:
Fixes https://github.com/trailofbits/manticore/issues/1103.

_(BTW, this is Alan, the new intern)_

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1222)

### DIFF
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -23,6 +23,11 @@ class TerminateState(StateException):
         super().__init__(message)
         self.testcase = testcase
 
+class AbandonState(TerminateState):
+    ''' Exception returned for abandoned states when
+        execution is finished
+    '''
+
 
 class Concretize(StateException):
     ''' Base class for all exceptions that trigger the concretization
@@ -196,7 +201,7 @@ class State(Eventful):
 
         Note: This must be called from the Executor loop, or a :func:`~manticore.Manticore.hook`.
         '''
-        raise TerminateState("Abandoned state")
+        raise AbandonState
 
     def new_symbolic_buffer(self, nbytes, **options):
         '''Create and return a symbolic buffer of length `nbytes`. The buffer is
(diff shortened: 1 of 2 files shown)


manticore/core/state.py
```diff
@@ -196,7 +201,7 @@ def abandon(self):
 
         Note: This must be called from the Executor loop, or a :func:`~manticore.Manticore.hook`.
         '''
-        raise TerminateState("Abandoned state")
+        raise AbandonState
```
stephan-tolksdorf: `AbandonState` inherits the `__init__` method from `TerminateState`, so this line misses the required string argument and hence will throw a different exception. One way to fix this issue would be to give `AbandonState` its own `__init__` method with a message argument that has the default value `'Abandoned state'`.

stephan-tolksdorf:
@disconnect3d @ex0dus-0x The commit actually causes [an error](https://travis-ci.org/trailofbits/manticore/jobs/445842911) in an example script, but since the `Executor` (logs and) suppresses all exceptions, this isn't recognized as an error by the test runner.

disconnect3d:
@stephan-tolksdorf Thanks for spotting this. I've reverted it in https://github.com/trailofbits/manticore/pull/1225

@ex0dus-0x can you fix it and open a new PR?

manticore/core/state.py
disconnect3d: :+1:

```python
In [1]: class A(Exception): pass

In [2]: class B(A):
   ...:     def __init__(self, msg):
   ...:         pass
   ...:

In [3]: class C(B): pass

In [4]:

In [4]: raise C
TypeError                                 Traceback (most recent call last)
<ipython-input-4-4a531f008761> in <module>()
----> 1 raise C

TypeError: __init__() missing 1 required positional argument: 'msg'
```

PR #1226, opened 2018-10-26
Add AbandonState Exception (Fixes #1222)

ex0dus-0x:
@stephan-tolksdorf @disconnect3d 

Added the changes. Is there anything I might need to do with the test cases in order for the executor to raise exceptions?

Thanks for the help!

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1226)

### DIFF
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -23,6 +23,14 @@ class TerminateState(StateException):
         super().__init__(message)
         self.testcase = testcase
 
+class AbandonState(TerminateState):
+    ''' Exception returned for abandoned states when
+        execution is finished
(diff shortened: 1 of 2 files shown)


stephan-tolksdorf:
CodeClimate wants two blank lines before the `AbandonState` declaration, but other than that this looks good to me.

Changing the exception handling in the Executor run loop would be out of scope for this PR/issue, I think.

Thanks for the PR!

disconnect3d:
> Changing the exception handling in the Executor run loop would be out of scope for this PR/issue, I think.

:+1:

Thanks!

Issue #676, opened 2017-12-28
AttributeError: 'EVM' object has no attribute 'header'


ggrieco-tob:
### OS / Environment

Distributor ID:	ArchLinux
Description:	ArchLinux
Release:	rolling-release

### Manticore version

Last revision (4c0f4bc6cf7509eb0f3399327fb494188b137671)

### Python version

Python 2.7.13

### Summary of the problem

Loading a simple contract triggers an AttributeError: 'EVM' object has no attribute 'header'

### Step to reproduce the behavior

This is a small solidity contract to reproduce this issue:

```
contract C {
    uint public n=0;
    function f() public{
        n = block.number;
    }
}
```

You can trigger this assertion using the manticore cli:

```
$ manticore error.sol                                                                
```

### Expected behavior

It shouldn't fail.

### Actual behavior

```
2017-12-28 11:08:32,891: [13648] m.main:INFO: Beginning analysis
2017-12-28 11:08:32,934: [13648] m.ethereum:INFO: Starting symbolic transaction: 1
2017-12-28 11:08:33,228: [13648] m.ethereum:INFO: Generated testcase No. 0 - REVERT
2017-12-28 11:08:34,781: [13648] m.ethereum:INFO: Generated testcase No. 1 - REVERT
2017-12-28 11:08:36,777: [13648] m.ethereum:INFO: Generated testcase No. 2 - REVERT
2017-12-28 11:08:38,781: [13648] m.ethereum:INFO: Generated testcase No. 3 - REVERT
2017-12-28 11:08:41,011: [13648] m.c.executor:ERROR: Exception: 'EVM' object has no attribute 'header'
Traceback (most recent call last):
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/core/executor.py", line 463, in run
    if not current_state.execute():
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/core/state.py", line 126, in execute
    result = self._platform.execute()
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/platforms/evm.py", line 2112, in execute
    self.current.execute()
  File "/home/g/.local/lib/python2.7/site-packages/manticore-0.1.6-py2.7.egg/manticore/platforms/evm.py", line 1302, in execute
    result = implementation(*arguments)
[...]

PR #1233, opened 2018-10-31
Add support for cross directory imports

disconnect3d:
Having a `main.sol`:
```
pragma solidity ^0.4.25;

import '../utils/DirImported.sol';
import 'utils/DirImported2.sol';

contract C {}
```

and such directory structure:
```
$ tree
.
├── main
│   └── main.sol
└── utils
    ├── DirImported.sol
    └── DirImported2.sol

2 directories, 3 files
```

and launching `manticore [options] main/main.sol` Manticore didn't respect current working directory and launched `solc [options] --allow-paths . main.sol` in `main` directory. As a result Manticore was failing due to `solc` being unable to import code from `utils` directory.

This PR fixes that issue and so Manticore will launch `solc [options] --allow-paths . main/main.sol` instead. A testcase has been added for that issue.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1233)

### DIFF
--- a/manticore/ethereum/__init__.py
+++ b/manticore/ethereum/__init__.py
@@ -336,27 +336,26 @@ class ManticoreEVM(Manticore):
             #logger.warning("Unsupported solc version %s", installed_version)
             pass
 
-        #shorten the path size so library placeholders wont fail.
-        #solc path search is a mess #fixme
-        #https://solidity.readthedocs.io/en/latest/layout-of-source-files.html
-        current_folder = os.getcwd()
-        abs_filename = os.path.abspath(source_file.name)
-        working_folder, filename = os.path.split(abs_filename)
+        # solc path search is a mess
+        # https://solidity.readthedocs.io/en/latest/layout-of-source-files.html
+
+        relative_filepath = source_file.name
+        working_dir = os.path.abspath(relative_filepath)[:-len(relative_filepath)]
 
         solc_invocation = [solc] + list(solc_remaps) + [
             '--combined-json', 'abi,srcmap,srcmap-runtime,bin,hashes,bin-runtime',
             '--allow-paths', '.',
-            filename
+            relative_filepath
         ]
 
-        p = Popen(solc_invocation, stdout=PIPE, stderr=PIPE, cwd=working_folder)
+        p = Popen(solc_invocation, stdout=PIPE, stderr=PIPE, cwd=working_dir)
         stdout, stderr = p.communicate()
 
         stdout, stderr = stdout.decode(), stderr.decode()
 
         # See #1123 - solc fails when run within snap
         # and https://forum.snapcraft.io/t/interfaces-allow-access-tmp-directory/5129
-        if stdout == '' and '""%s"" is not found' % filename in stderr:
+        if stdout == '' and '""%s"" is not found' % relative_filepath in stderr:
             raise EthereumError(
                 'Solidity compilation failed with error: {}\n'
                 'Did you install solc from snap Linux universal packages?\n'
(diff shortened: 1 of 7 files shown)


tests/test_binaries.py
```diff
@@ -1,19 +1,34 @@
-import io
 import unittest
 import sys
 import shutil
 import tempfile
 import os
-import hashlib
```
disconnect3d: Those are redundant imports.

tests/test_binaries.py
```diff
@@ -77,7 +96,7 @@ def _simple_cli_run(self, filename, contract=None, tx_limit=1):
         command.append(str(tx_limit))
         command.append(filename)
 
-        subprocess.check_call(command, stdout=subprocess.PIPE)
+        subprocess.check_call(command, stdout=subprocess.PIPE, cwd=working_dir)
```
feliam: This is not really checking anything. Just that the process wont take > 10 mins I guess. We should try to add more checks here. Eventually

disconnect3d: It checks whether Manticore failed its execution or not. Initially I thought that the listed issues had to crash all Manticore and it returns exit code != 0.

But I guess that https://github.com/trailofbits/manticore/issues/676 did return 0 exit code, so I am confused a bit.

disconnect3d:
One thing I am worried about is this code:
https://github.com/trailofbits/manticore/blob/master/manticore/ethereum/__init__.py#L280-L290

If the solidity contract path is very long, the `lib_placeholder` in:
```python
lib_name = lib_placeholder.split(':')[1].split('_')[0]
```

will end up not having a `:` in it as apparently `solc` truncates the path to a given number of characters... so `[1]` will end up in a `IndexError`.

This was thrown in one of the detector testcases that used `library {...}` - https://github.com/trailofbits/manticore/blob/master/tests/binaries/detectors/delegatecall_ok.sol

I mitigated it by adding a `working_dir` to the `solidity_create_contract` call. But in the end, if one doesn't pass a `working_dir` this may still fail.

Anyway, I think the `solidity_create_contract` should still have a possibility of specifying working directory, so imports will work fine in contracts used through the API.

manticore/ethereum/__init__.py
```diff
@@ -336,27 +336,33 @@ def _run_solc(source_file, solc_bin=None, solc_remaps=[]):
             #logger.warning("Unsupported solc version %s", installed_version)
             pass
 
-        #shorten the path size so library placeholders wont fail.
```
feliam: Can notwe use this path shortening thing in `solidity_create_contract` if the working_directory is not specified?

manticore/ethereum/__init__.py
```diff
@@ -336,27 +336,33 @@ def _run_solc(source_file, solc_bin=None, solc_remaps=[]):
...
-        p = Popen(solc_invocation, stdout=PIPE, stderr=PIPE, cwd=working_folder)
+        p = Popen(solc_invocation, stdout=PIPE, stderr=PIPE, **additional_kwargs)
         stdout, stderr = p.communicate()
 
         stdout, stderr = stdout.decode(), stderr.decode()
 
         # See #1123 - solc fails when run within snap
         # and https://forum.snapcraft.io/t/interfaces-allow-access-tmp-directory/5129
-        if stdout == '' and '""%s"" is not found' % filename in stderr:
+        if stdout == '' and '""%s"" is not found' % relative_filepath in stderr:
```
feliam: Maybe just f-string it?

manticore/ethereum/__init__.py
disconnect3d: We resolved this on a DM; for others/future ref:

if someone lanuches:
```manticore some_path```
it should launch `solc` with the same path and from the same working dir as the person is on -
unless the person wants it differently.

Previously we shortened the path but we can't do it as we can't know how much can we shorten it as there might be relative imports in contracts :s.

PR #1239, opened 2018-10-31
Use the address of symbolic conditional instructions

NeatMonster:
When two symbolic conditional instructions follow sequentially, the two paths of
the first instruction will be explored. But when the exploration reaches the
second instruction, because the at_symbolic_conditional flag has already been
set by the first one, only one of the two possible paths will be explored.

This commit fixes this issue by replacing the boolean flag by the address of the
instruction, which ensure the flag has been set for the current instruction.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1239)

### DIFF
--- a/tests/test_armv7cpu.py
+++ b/tests/test_armv7cpu.py
@@ -2,6 +2,7 @@ import unittest
 import struct
 from functools import wraps
 
+from manticore.core.cpu.abstractcpu import ConcretizeRegister
 from manticore.core.cpu.arm import Armv7Cpu as Cpu, Mask, Interruption
 from manticore.core.memory import Memory32, SMemory32
 from manticore.core.smtlib import *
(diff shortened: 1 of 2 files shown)


disconnect3d:
Can we add a test for it?

NeatMonster:
Here is an unit test for it. It ensures that two branches would be explored each time.

tests/test_armv7cpu.py
```diff
@@ -1695,3 +1696,35 @@ def test_arm_save_restore_cpu(self):
...
 
+    def test_symbolic_conditional(self):
+        asm = \
+        """
+          tst r0, r0
+          beq label
+          bne label
+        label:
+          nop
+        """
```
disconnect3d: Can we rewrite the code so the addresses are more explicit?

e.g.
```python
# in thumb mode each instruction has size of 2 bytes
# (the code starts at 0x1004)
asm = '\n'.join(
    '  tst r0, r0',  # 0x1004
    '  beq label'  # 0x1006
    '  bne label'  # 0x1008
    'label:'
    '  nop'  # 0x100a
)
```

NeatMonster: Done, see my latest commit. :smile:

NeatMonster:
I've made the changes as requested, and added some more comments.

PR #1242, opened 2018-10-31
Allow file offsets not aligned on page size

NeatMonster:
When mapping into memory uncommon ELF executables having their segments file
offsets not aligned on the system page size, the call to mmap_function() will
fail silently. It will actually return 0xffffffff, which isn't checked anywhere
and will only be detected when trying to read/write/free the memory area.

Because this ELF files triggering this issue are programmatically crafted from
binaries into a proprietary format, it won't be happening for all other
executables. But Manticore should support binaries as generic as possible.

This commit allows mapping into memory files having offsets not aligned on the
system page size by simply aligning the file offset when mmap'ing, and returning
a pointer shifted from the necessary correction. When munmap'ing the memory, the
address is aligned again to recovery the initial one given when mmap'ing.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1242)

### DIFF
--- a/manticore/utils/mappings.py
+++ b/manticore/utils/mappings.py
@@ -41,9 +41,9 @@ def mmap(fd, offset, size):
     prot = MMAP.PROT_READ | MMAP.PROT_WRITE
     flags = MMAP.MAP_PRIVATE
 
-    # When trying to map the contents of a file into memory, the offset must be a multiple of the
-    # page size (see `man mmap`). So we need to align it before passing it to mmap(). Doing so also increases
-    # the size of the memory area needed, so we need to account for that difference.
+    # When trying to map the contents of a file into memory, the offset must be a multiple of the page size (see
+    # `man mmap`). So we need to align it before passing it to mmap(). Doing so also increases the size of the memory
+    # area needed, so we need to account for that difference.
     aligned_offset = offset & ~0xfff
     size += offset - aligned_offset
 
@@ -52,20 +52,21 @@ def mmap(fd, offset, size):
     assert size > 0
 
     result = mmap_function(0, size, prot, flags, fd, aligned_offset)
+    assert result != ctypes.c_void_p(-1).value
 
-    # Now when returning the pointer to the user, we need to skip the corrected offset so that the
-    # user doesn't end up with a pointer to other region of the file than the one they requested.
+    # Now when returning the pointer to the user, we need to skip the corrected offset so that the user doesn't end up
+    # with a pointer to another region of the file than the one they requested.
     return ctypes.cast(result + offset - aligned_offset, ctypes.POINTER(ctypes.c_char))
 
 
 def munmap(address, size):
-    # When unmapping the memory area, we need to recover the pointer and the size that were given
-    # to mmap(). The pointer can be recovered by aligning it on the page size. The size needs to be
-    # increased by that difference to account for cases where the requested memory area was larger.
+    # When unmapping the memory area, we need to recover the pointer and the size that were given to mmap(). The
+    # pointer can be recovered by aligning it on the page size. The size needs to be increased by that difference to
+    # account for cases where the requested memory area was larger.
     address = ctypes.cast(address, ctypes.c_void_p).value
     aligned_address = address & ~0xfff
     size += address - aligned_address
-    aligned_address = ctypes.cast(aligned_address, ctypes.POINTER(ctypes.c_char))
+    assert size > 0
 
     result = munmap_function(aligned_address, size)
     assert result == 0
--- a/manticore/utils/mappings.py
+++ b/manticore/utils/mappings.py
@@ -41,17 +41,26 @@ def mmap(fd, offset, size):
     prot = MMAP.PROT_READ | MMAP.PROT_WRITE
     flags = MMAP.MAP_PRIVATE
 
+    # When trying to map the contents of a file into memory, the offset must be a multiple of the
+    # page size. So we need to align it before passing it to mmap(). But doing so also increases
+    # the size of the memory area needed, so we need to account for that difference.
+    aligned_offset = offset & ~0xfff
+    size += offset - aligned_offset
+
     if size & 0xfff != 0:
         size = (size & ~0xfff) + 0x1000
-
     assert size > 0
 
-    aligned_offset = offset & ~0xfff
     result = mmap_function(0, size, prot, flags, fd, aligned_offset)
+
+    # Now when returning the pointer to the user, we need to skip the corrected offset so that the
+    # user doesn't end up with a pointer to an other region of the file than the one he requested.
     return ctypes.cast(result + offset - aligned_offset, ctypes.POINTER(ctypes.c_char))
 
 
 def munmap(address, size):
+    # When unmapping the memory area, we need to recover the pointer that was returned by mmap().
+    # It's only a matter of aligning it back to the page size, which requires some ctypes casting.
     aligned_address = ctypes.cast(address, ctypes.c_void_p)
     aligned_address = aligned_address.value & ~0xfff
     aligned_address = ctypes.cast(aligned_address, ctypes.POINTER(ctypes.c_char))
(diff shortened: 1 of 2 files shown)


disconnect3d:
To be more specific/clear, the `man mmap` says (emphasis mine):

> The  contents  of a file mapping (as opposed to an anonymous mapping; see `MAP_ANONYMOUS` below), are initialized using length bytes starting at `offset` offset in the file (or other object) referred to by the file descriptor `fd`.  **`offset` must be a  multiple of the page size as returned by `sysconf(_SC_PAGE_SIZE)`**.

so we definitely need this.

manticore/utils/mappings.py
```diff
@@ -46,10 +46,15 @@ def mmap(fd, offset, size):
 
     assert size > 0
 
-    result = mmap_function(0, size, prot, flags, fd, offset)
-    return ctypes.cast(result, ctypes.POINTER(ctypes.c_char))
+    aligned_offset = offset & ~0xfff
+    result = mmap_function(0, size, prot, flags, fd, aligned_offset)
```
disconnect3d:
```suggestion
    result = mmap_function(0, size + offset - aligned_offset, prot, flags, fd, aligned_offset)
```

Assume that:
* offset = 0x1001
* size/length = 10

The `alligned_offset` is 0x1000. We mmap 10 bytes so the `result` will be an area of 10 bytes***.

Later, we return a pointer to `result+offset-aligned_offset` which effectively is `result+1`. Since `result` is of 10 bytes, we will end up pointing to an area of 9 bytes, while we wanted a 10 bytes area.

And that's the reason why we need to adjust `size` argument properly.

I have based my assumptions on the example taken from `man mmap` where they call `mmap` like this (`pa_offset` is the aligned size, `offset` is size):
```c
addr = mmap(NULL, length + offset - pa_offset, PROT_READ,
                       MAP_PRIVATE, fd, pa_offset);
```

and then they use the memory as (so it is similar to us returning the adjusted pointer):
```c
write(STDOUT_FILENO, addr + offset - pa_offset, length);
```
---
*** well, `mmap` will probably return a whole memory page (by default 4 kB on most or all kernels) but we probably can't assume that.

disconnect3d:
Hmm... it would also be nice to add those explanations to the code. Thanks!

manticore/utils/mappings.py
NeatMonster: > We mmap 10 bytes so the result will be an area of 10 bytes***.

I think your code already in place that aligns the `size` on the page size will have changed the `size` value to 0x1000, so you won't be mmap'ing 10 bytes but 0x1000.

NeatMonster:
> Hmm... it would also be nice to add those explanations to the code, either as a docstring or as a comment. Thanks!

Sure, I will add some comments around the changes.

> Also, any chance for adding a test for it?

If you can point out where an appropriate place for it would be, then yes!

manticore/utils/mappings.py
disconnect3d: Why 0x1000? size + offset - aligned_offset = 10 + 0x1001 - 0x1000 = 10 + 1 = 11, so yeah we mmap more bytes but we return pointer to the result+1.

NeatMonster: Please consider the following example:
```
def mmap(fd, offset, size):
    if size & 0xfff != 0:
        size = (size & ~0xfff) + 0x1000
    assert size > 0

    aligned_offset = offset & ~0xfff
    print("mmap_function(0, %#x, prot, flags, fd, %#x)" % (size, aligned_offset))
    result = 0x12340000 # page aligned address returned by mmap()
    return result + offset - aligned_offset

def munmap(address, size):
    aligned_address = address & ~0xfff
    print("munmap_function(%#x, %#x)" % (aligned_address, size))

offset = 0x1
size = 0x10

address = mmap(0, offset, size)
print("result is %#x" % address)
munmap(address, size)
```

and the corresponding output:
```
mmap_function(0, 0x1000, prot, flags, fd, 0x0)
result is 0x12340001
munmap_function(0x12340000, 0x10)
```

We're trying to map 0x10 bytes from a file starting at offset 0x1. You can see that we end up calling mmap() with a offset of 0 and a size of 0x1000. The address returned by mmap() is the address at which the beginning of the file is mapped, so we need to add 0x1 to the pointer returned to the user.

When the user tries to free the memory using this pointer, we align it back to recover the value returned by mmap(). We can use the original size because munmap() will unmap the whole page.

disconnect3d: Oh, I am sorry, I missed the operations above that deal with `size`. So yeah, this is correct :).

:+1:

NeatMonster: I did more thinking and it turns out that you were right!

Consider an offset of 0xfff and a size of 0x2. You would need to mmap (and then unmap) 0x2000 bytes starting at offset 0x0, but the current code would only mmap (and unmap) 0x1000.

I've pushed two other commits that fix this issue (with comments!).

NeatMonster:
> I believe that `tests/test_armv7cpu.py` should be a proper place. We have some decorators magic there that might help but they aren't necessary/don't have to be used.

Pardon my ignorance, but I can't seem to understand why the tests should be written there. While I've encountered this issue on converted ARM binaries, I think the tests should be of something more general, like ensuring that non page-aligned offset values will be accepted after the fix.

disconnect3d:
Oh yeah, I thought I am replying in one of the arm issues 🤣 . They should probably belong to `test_linux` :).

NeatMonster:
I've added a unit test to `tests/test_binaries.py`. I've decided to put it there because it is there that logging and various utilities are being tested. I can move it if you can find a more appropriate place.

I will add the unit tests and docstrings that you requested on my other PRs as soon as possible. Have you been able to take a look at https://github.com/trailofbits/manticore/pull/1239 though? Thanks for reminding me about the syntax highlighting! :smile:

manticore/utils/mappings.py
```diff
@@ -41,15 +41,31 @@ def mmap(fd, offset, size):
...
 
 def munmap(address, size):
-    result = munmap_function(address, size)
+    # When unmapping the memory area, we need to recover the pointer and the size that were given
+    # to mmap(). The pointer can be recovered by aligning it on the page size. The size needs to be
+    # increased by that difference to account for cases where the requested memory area was larger.
+    address = ctypes.cast(address, ctypes.c_void_p).value
+    aligned_address = address & ~0xfff
+    size += address - aligned_address
+    aligned_address = ctypes.cast(aligned_address, ctypes.POINTER(ctypes.c_char))
```
disconnect3d: Is the cast required here at all? I assume that since we call a glibc function the type doesn't really matter here?

tests/test_binaries.py
```diff
@@ -242,6 +243,21 @@ def test_brk_regression(self):
...
+        with open(filename, 'rb') as f:
+            for addr, size in [
+                (0x1, 0x1000),
+                (0x1, 0x1fffe),
+                (0x1, 0x1ffff),
+                (0xfff, 0x1),
+                (0xfff, 0x2),
+                (0xfff, 0x1000)
+            ]:
+                ret = mmap(f.fileno(), addr, size)
```
disconnect3d: We should probably assert that `ret` is different than -1 (or to be more specific: `0xffffffffffffffff` or `ctypes.c_uint64(-1).value`.

manticore/utils/mappings.py
NeatMonster: No it isn't. I've kept it to conform to the definition of `munmap_function.argtype` but I removed it in my latest commit.

tests/test_binaries.py
NeatMonster: `ret` would be a `ctypes.POINTER(ctypes.c_char)` here, so it would need to be casted back to a `ctypes.c_void_p` to be compared. I have added an assert into the code of mmap instead.

```python
In [1]: import ctypes

In [2]: ctypes.c_void_p(-1) == ctypes.c_void_p(-1)
Out[2]: False
```

NeatMonster:
I think you're right. The syscall return value should match the one that is expected by the program. But if my understanding is correct, it will require to modify others parts of the code, e.g. `memory.py`.

disconnect3d:
Yeah, let's leave it as it is, unless someone has a use case for that.

PR #1243, opened 2018-10-31
Added CBZ, TBB and TBH instructions semantics

NeatMonster:
CBZ was not implemented while CBNZ was. As far as Manticore is concerned, CBZ is
the same as CBNZ, but with two possible destination addresses exchanged.

TBB and TBH instructions are often used for switches. According to the ARM
documentation, the base register contains the address of a table of
bytes/half-words, and the index register contains an index into this table. The
selected value is then shifted left by one and zero-extended to 32-bits before
being added to the current PC. If the PC register is used as the base, the table
is then located immediately after the instruction.

See also ARM Architecture Reference Manual Thumb-2 Supplement.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1243)

### DIFF
--- a/tests/test_armv7cpu.py
+++ b/tests/test_armv7cpu.py
@@ -989,32 +989,32 @@ class Armv7CpuInstructions(unittest.TestCase):
     # CBZ/CBNZ
 
     @itest_setregs("R0=0")
-    @itest_custom("cbz r0, 0x2a", mode=CS_MODE_THUMB)
+    @itest_custom("cbz r0, #0x2a", mode=CS_MODE_THUMB)
     def test_cbz_taken(self):
         pre_pc = self.rf.read('PC')
         self.cpu.execute()
-        self.assertEqual(self.rf.read('PC'), pre_pc + 42)
+        self.assertEqual(self.rf.read('PC'), pre_pc + 0x2a)
 
     @itest_setregs("R0=1")
     @itest_custom("cbz r0, #0x2a", mode=CS_MODE_THUMB)
     def test_cbz_not_taken(self):
         pre_pc = self.rf.read('PC')
         self.cpu.execute()
-        self.assertEqual(self.rf.read('PC'), pre_pc + 2)
+        self.assertEqual(self.rf.read('PC'), pre_pc + 2)  # cbz is 2 bytes long
 
     @itest_setregs("R0=1")
-    @itest_custom("cbnz r0, 0x2a", mode=CS_MODE_THUMB)
+    @itest_custom("cbnz r0, #0x2a", mode=CS_MODE_THUMB)
     def test_cbnz_taken(self):
         pre_pc = self.rf.read('PC')
         self.cpu.execute()
-        self.assertEqual(self.rf.read('PC'), pre_pc + 42)
+        self.assertEqual(self.rf.read('PC'), pre_pc + 0x2a)
 
     @itest_setregs("R0=0")
     @itest_custom("cbnz r0, #0x2a", mode=CS_MODE_THUMB)
     def test_cbnz_not_taken(self):
         pre_pc = self.rf.read('PC')
         self.cpu.execute()
-        self.assertEqual(self.rf.read('PC'), pre_pc + 2)
+        self.assertEqual(self.rf.read('PC'), pre_pc + 2)  # cbnz is 2 bytes long
 
     # TBB/TBH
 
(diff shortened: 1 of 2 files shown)


disconnect3d:
Hey, can you functions description to docstrings and add tests for them?

NeatMonster:
I've added some docstrings/comments/unit tests for the CBZ/TBB/TBH instructions and for CBNZ which was already implemented.

tests/test_armv7cpu.py
```diff
@@ -986,6 +986,86 @@ def test_bl_neg(self):
...
+
+    @itest_setregs("R0=0")
+    @itest_custom("cbz r0, 0x2a", mode=CS_MODE_THUMB)
+    def test_cbz_taken(self):
+        pre_pc = self.rf.read('PC')
+        self.cpu.execute()
+        self.assertEqual(self.rf.read('PC'), pre_pc + 42)
+
+    @itest_setregs("R0=1")
+    @itest_custom("cbz r0, #0x2a", mode=CS_MODE_THUMB)
```
disconnect3d: Is there any reason why do we use sometimes `0x2a` and sometimes `#0x2a`?

tests/test_armv7cpu.py
```diff
@@ -986,6 +986,86 @@ def test_bl_neg(self):
...
         self.assertEqual(self.rf.read('LR'), pre_pc + 4)
 
+    # CBZ/CBNZ
+
+    @itest_setregs("R0=0")
+    @itest_custom("cbz r0, 0x2a", mode=CS_MODE_THUMB)
+    def test_cbz_taken(self):
+        pre_pc = self.rf.read('PC')
+        self.cpu.execute()
+        self.assertEqual(self.rf.read('PC'), pre_pc + 42)
```
disconnect3d:
```suggestion
        self.assertEqual(self.rf.read('PC'), pre_pc + 0x2a)  # branch taken
```

tests/test_armv7cpu.py
```diff
@@ -986,6 +986,86 @@ def test_bl_neg(self):
...
+        pre_pc = self.rf.read('PC')
+        self.cpu.execute()
+        self.assertEqual(self.rf.read('PC'), pre_pc + 2)
+
+    @itest_setregs("R0=1")
+    @itest_custom("cbnz r0, 0x2a", mode=CS_MODE_THUMB)
+    def test_cbnz_taken(self):
+        pre_pc = self.rf.read('PC')
+        self.cpu.execute()
+        self.assertEqual(self.rf.read('PC'), pre_pc + 42)
```
disconnect3d:
```suggestion
        self.assertEqual(self.rf.read('PC'), pre_pc + 0x2a)
```

tests/test_armv7cpu.py
```diff
@@ -986,6 +986,86 @@ def test_bl_neg(self):
...
+    @itest_custom("tbb [r0, r1]", mode=CS_MODE_THUMB)
+    def test_tbb(self):
+        # Write the table of offsets at 0xd000 (R0)
+        # Index is 1 (R1), offset will be 2x21 = 42
+        for i, offset in enumerate([11, 21, 31]):
+            self.mem.write(0xd000 + i, struct.pack('<B', offset))
+
+        pre_pc = self.rf.read('PC')
+        self.cpu.execute()
+        self.assertEqual(self.rf.read('PC'), pre_pc + 4 + 42)
```
disconnect3d: Why do we do `+4` here?

tests/test_armv7cpu.py
NeatMonster: Based on [this other StackOverflow answer](https://stackoverflow.com/a/51987780), I agree that it's better to use `#` to keep it consistent. I have made the necessary changes in my latest commit.

tests/test_armv7cpu.py
NeatMonster: Changed in my latest commit.

tests/test_armv7cpu.py
NeatMonster: Changed in my latest commit.

tests/test_armv7cpu.py
NeatMonster: Because `TBB` and `TBH` are 4 bytes long instructions. I have added a comment in my latest commit to hopefully make that clearer.

PR #1244, opened 2018-10-31
Added ADR, ADDW and SUBW instructions semantics

NeatMonster:
These 3 instructions can be used to load PC-relative addresses. The ADDW and
SUBW instructions can also be used with other source registers. When the PC
register is used as the source, the ARM documentation specifies that the offset
is added or subtracted from ALIGN(pc, 4).

See also the ARM Architecture Reference Manual Thumb-2 Supplement.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1244)

### DIFF
--- a/tests/test_armv7cpu.py
+++ b/tests/test_armv7cpu.py
@@ -976,33 +976,33 @@ class Armv7CpuInstructions(unittest.TestCase):
     def test_adr(self):
         pre_pc = self.rf.read('PC')
         self.cpu.execute()
-        self.assertEqual(self.rf.read('R0'), pre_pc + 4 + 16)
+        self.assertEqual(self.rf.read('R0'), (pre_pc + 4) + 16)  # adr is 4 bytes long
 
     # ADDW
 
     @itest_setregs("R1=0x1234")
-    @itest_thumb("addw r0, r1, 0x2a")
+    @itest_thumb("addw r0, r1, #0x2a")
     def test_addw(self):
-        self.assertEqual(self.rf.read('R0'), 0x1234 + 42)
+        self.assertEqual(self.rf.read('R0'), 0x1234 + 0x2a)
 
-    @itest_custom("addw r0, pc, 0x2a", mode=CS_MODE_THUMB)
+    @itest_custom("addw r0, pc, #0x2a", mode=CS_MODE_THUMB)
     def test_addw_pc_relative(self):
         pre_pc = self.rf.read('PC')
         self.cpu.execute()
-        self.assertEqual(self.rf.read('R0'), pre_pc + 4 + 42)
+        self.assertEqual(self.rf.read('R0'), (pre_pc + 4) + 0x2a)  # addw is 4 bytes long
 
     # SUBW
 
     @itest_setregs("R1=0x1234")
-    @itest_thumb("subw r0, r1, 0x2a")
+    @itest_thumb("subw r0, r1, #0x2a")
     def test_subw(self):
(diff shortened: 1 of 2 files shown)


disconnect3d:
Hey, can you functions description to docstrings and add tests for them?

NeatMonster:
I've added some docstrings/unit tests for the ADR/ADDW/SUBW instructions.

tests/test_armv7cpu.py
```diff
@@ -970,6 +970,40 @@ def test_str(self):
...
+        pre_pc = self.rf.read('PC')
+        self.cpu.execute()
+        self.assertEqual(self.rf.read('R0'), pre_pc + 4 + 16)
+
+    # ADDW
+
+    @itest_setregs("R1=0x1234")
+    @itest_thumb("addw r0, r1, 0x2a")
+    def test_addw(self):
+        self.assertEqual(self.rf.read('R0'), 0x1234 + 42)
```
disconnect3d:
```suggestion
        self.assertEqual(self.rf.read('R0'), 0x1234 + 0x2a)
```

tests/test_armv7cpu.py
```diff
@@ -970,6 +970,40 @@ def test_str(self):
...
+
+    @itest_custom("addw r0, pc, 0x2a", mode=CS_MODE_THUMB)
+    def test_addw_pc_relative(self):
+        pre_pc = self.rf.read('PC')
+        self.cpu.execute()
+        self.assertEqual(self.rf.read('R0'), pre_pc + 4 + 42)
```
disconnect3d:
```suggestion
        self.assertEqual(self.rf.read('R0'), pre_pc + 4 + 0x2a)
```

tests/test_armv7cpu.py
```diff
@@ -970,6 +970,40 @@ def test_str(self):
...
+
+    # SUBW
+
+    @itest_setregs("R1=0x1234")
+    @itest_thumb("subw r0, r1, 0x2a")
+    def test_subw(self):
+        self.assertEqual(self.rf.read('R0'), 0x1234 - 42)
```
disconnect3d:
```suggestion
        self.assertEqual(self.rf.read('R0'), 0x1234 - 0x2a)
```

tests/test_armv7cpu.py
```diff
@@ -970,6 +970,40 @@ def test_str(self):
...
+
+    @itest_custom("subw r0, pc, 0x2a", mode=CS_MODE_THUMB)
+    def test_subw_pc_relative(self):
+        pre_pc = self.rf.read('PC')
+        self.cpu.execute()
+        self.assertEqual(self.rf.read('R0'), pre_pc + 4 - 42)
```
disconnect3d:
```suggestion
        self.assertEqual(self.rf.read('R0'), pre_pc + 4 - 0x2a)
```

disconnect3d:
Loud thinking: maybe the instruction tests should assert ALL register states and so one of the decorators (or a new one) should do sth like:

```python
test_function.expected_registers = {copy initial register values}
test_function()
# the test_function does sth like test_function.expected_registers['r0'] = expected_value
self.assertEqual(currentRegs(), test_function.expected_registers)
```

so we are sure a given instruction also doesn't change any other state...

Another approach could be "marking a state change" during test runs and expect functions to "handle this state changes" e.g. by removing them - so we could assert that "all state changes were handled by the test".

We don't have to do it in this PR, it's just loud thinking about something we should probably introduce at some point.

Issue #1237, opened 2018-10-31
Feasible state found if value (in .transaction) is set to 0 but not if using a symbolic value


ralph-pichler:
### OS / Environment

MacOS 10.14

### Manticore version

Version: 0.2.2

### Python version

Python 3.7.0

### Dependencies

asn1crypto==0.24.0
attrdict==2.0.0
capstone==3.0.5.post1
certifi==2018.10.15
cffi==1.11.5
chardet==3.0.4
coincurve==9.0.0
configparser==3.5.0
coverage==4.5.1
cytoolz==0.9.0.1
dictionaries==0.0.1
eth-abi==1.2.2
eth-account==0.3.0
eth-hash==0.2.0
eth-keyfile==0.5.1
eth-keys==0.2.0b3
eth-rlp==0.1.2
eth-tester==0.1.0b33
eth-typing==1.3.0
eth-utils==1.2.2
ethereum==2.3.2
future==0.16.0
hexbytes==0.1.0
idna==2.7
Jinja2==2.10
manticore==0.2.2
MarkupSafe==1.0
mock==2.0.0
mythril==0.18.12
parsimonious==0.8.1
pbkdf2==1.3
pbr==5.0.0
persistent==4.4.2
ply==3.11
plyvel==1.0.5
py-ecc==1.4.3
py-flags==1.1.2
py-solc==3.2.0
pycairo==1.17.1
pycparser==2.19
pycryptodome==3.6.6
pyelftools==0.25
pyethash==0.1.27
pyevmasm==0.1.0
PyGObject==3.30.1
pysha3==1.0.2
PyYAML==3.13
repoze.lru==0.7
requests==2.20.0
rlp==1.0.3
scrypt==0.8.6
semantic-version==2.6.0
six==1.11.0
toolz==0.9.0
transaction==2.3.0
unicorn==1.0.1
urllib3==1.24
z3-solver==4.8.0.0
zope.interface==4.5.0

### Summary of the problem

A feasible state is found if value in .transaction is 0 but not if value is a symbolic value

### Step to reproduce the behavior

The contract is from the DEVCON4 workshop:
```solidity
pragma solidity ^0.4.24;
contract UnprotectedWallet{
    address public owner;

    modifier onlyowner {
        require(msg.sender==owner);
        _;
    }

    constructor() public {
        owner = msg.sender;
    }
    function changeOwner(address _newOwner) public {
       owner = _newOwner;
    }

    function deposit() payable public { 
    }

[...]

PR #1246, opened 2018-10-31
Fix exception handling in EVM.transact._pre_func

stephan-tolksdorf:
Fixes issue #1237.

I will fix the `DetectExternalCallAndLeak` tests later and, if still necessary, add a test for the previously buggy behaviour.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1246)

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -526,6 +526,7 @@ class EVM(Eventful):
         self._allocated = 0
         self._on_transaction = False  # for @transact
         self._checkpoint_data = None
+        self._published_pre_instruction_events = False
 
         # Used calldata size
         min_size = 0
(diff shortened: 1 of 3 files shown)


stephan-tolksdorf:
The `will_evm_execute_instruction` event is published multiple times for the same `CALL` if the call has arguments that need to be concretized. The way the `DetectExternalCallAndLeak` detector is implemented makes me think that the event should only be published before any concretization.  @feliam @mossberg: Is that correct?

feliam:
(The bugs in which concrete and symbolic execution diverge are the worst. Thank you for fixing this.)

Yes it should be published once per state.

And I think it may be better to publish it before concretization so a an event handler/plugin could interpose and control hot it gets concretized?

stephan-tolksdorf:
@feliam I've changed the EVM event publishing logic to only emit the pre-instruction once (before concretization) and fixed the tests.

Issue #1255, opened 2018-11-05
Events are lost after a state fork during the execution of an EVM message call


stephan-tolksdorf:
`EVMWorld.__setstate__` only installs an event sink for the top-most VM in the callstack. If the callstack has more than one item, i.e. if the state is forked during a message call, events will no longer be published after the last message call returns (until the next state fork occurs).

In the following example this bug leads to an Ethereum leak not being detected:

```python
from manticore.ethereum import ManticoreEVM, DetectExternalCallAndLeak

m = ManticoreEVM()
source_code = '''
pragma solidity ^0.4.24;

contract Lib {
   function isSeven(uint a) public pure returns (bool) {
       if (a == 7) {
           return true;
       } else {
           return false;
       }
   }
}

contract Wallet {
   Lib private lib;

   constructor() public payable {
       lib = new Lib();
   }

   function luckyNumber(uint a) public {
       if (lib.isSeven(a)) {
           msg.sender.transfer(address(this).balance);
       }
   }
}
'''

m.register_detector(DetectExternalCallAndLeak())

owner = m.create_account(name='owner', balance=1000)

wallet = m.solidity_create_contract(source_code, name='wallet', contract_name='Wallet', owner=owner, balance=1000)
attacker = m.create_account(name='attacker', balance=0)

wallet.luckyNumber(m.make_symbolic_value(), caller=attacker)

print(f'findings: {m.global_findings}') # Prints an empty set.
print(f'workspace: {m.workspace}')
m.finalize() # The generated traces have large gaps.
```

I'll prepare a pull request.

dguido:
I guess not!

PR #1256, opened 2018-11-05
Fix event forwarding after state fork in EVM message call

stephan-tolksdorf:
Fixes issue #1255.

Also adds type hints to some private attributes of the EVMWorld class.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1256)

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -5,10 +5,11 @@ import io
 import copy
 import inspect
 from functools import wraps
+from typing import List, Set, Tuple, Union
 from ..exceptions import EthereumError
 from ..utils.helpers import issymbolic, get_taints, taint_with, istainted
 from ..platforms.platform import *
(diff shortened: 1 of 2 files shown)


manticore/platforms/evm.py
```diff
@@ -1750,16 +1750,14 @@ def __setstate__(self, state):
         self._transactions = state['transactions']
         self._initial_block_number = state['initial_block_number']
         self._initial_timestamp = state['_initial_timestamp']
-        self._do_events()
+
+        for _, _, _, _, vm in self._callstack:
```
feliam: I wonder if you have considered just adding the lacking self._do_events() to close_transaction?

stephan-tolksdorf: Yeah, but that would override/duplicate the registration if the previous vm was already registered, i.e. if the the top-most item was appended in `_open_transaction`. I thought doing it this way would be a bit cleaner and easier to follow.

Issue #1247, opened 2018-11-01
Make mcore installable for specific targets


feliam:
* Work out setup.py so it is possible to install EVM(or other arch) only manticore.
* Fix and refactor imports so manticore does not need to have say Unicorn when in EVM mode 
* `pip install manticore .[intel]` will install manticore for x86/amd64 stuff
* `pip install manticore .[eth]` if they want ethereum stuff
* `pip install manticore .[decree]` if they want decree stuff
* `pip install manticore .[intel, eth]` should install intel and ethereum stuff (not sure about the syntax here)
* `pip install manticore .` should install all of it  (if possible
* similar behavior should be doable via `python setup.py ..`

...

mossberg:
proposed solution for this is to invert the defaults and have `pip install manticore` only install evm deps. in the setup.py `install_requires` will have the basic things for evm to work. then if a user desires binary analysis deps, they can do `manticore[native]`, which will use `extras_require` to install capstone, unicorn, pyelftool, etc.

disconnect3d:
In the end we decided to make it like this (merged in #1257 ):

* `pip install manticore` - installs core and ethereum dependencies
* `pip install manticore[native]` - installs core, ethereum and native dependencies

as rather most of our users use Manticore for ethereum related stuff and ethereum dependencies can be installed without compiling additional modules.

PR #1257, opened 2018-11-05
Make Manticore installable for separate targets

disconnect3d:
https://github.com/trailofbits/manticore/issues/1247

As a result:
* `pip install manticore`
  * installs only core dependencies: `manticore-0.2.2 pyyaml-3.13 z3-solver-4.8.0.0.post1`
  * in such state, trying to run `manticore <binary_or_solidity_file>` will end up with [1], same goes for importing anything from `manticore.native` or `manticore.ethereum`

* `pip install manticore[evm]` - installs both core and evm dependencies; if users try to use the native stuff, they end up with [1]

* `pip install manticore[native]` - installs both core and native dependencies; also if users try to use evm, they end up with [1]
* `pip install manticore[evm,native]` - installs core, evm and native dependencies

Note that we also have other extra dependencies that can be installed: `[dev]`, `[dev-noks]` (no keystone), `[redis]`.

[1] When any of the dependency is missing, such text is displayed and the user can choose one of the options:
```
$ manticore ~/test.sol

[*] It seems that your Manticore installation lacks targets dependencies.
[*]
[*] Choose your option:
[*] 1) get more detailed information
[*] 2) install Ethereum vm target dependencies
[*] 3) install native targets (x86, x86-64, armv7, decree) dependencies
[*] 4) install both native and evm targets dependencies
[*] 5) exit

Your choice (1/2/3/4/5):
```
If the user has evm or native dependencies the option will say `(installed)` at the end.

**NOTE: This PR lacks README changes but I think we can make them in next PR.**

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1257)


setup.py
```diff
@@ -21,26 +29,31 @@ def rtd_dependent_deps():
...
+            'capstone>=3.0.5',
+            'pyelftools',
+            'unicorn',
+            'ply',
+        ],
+        'evm': [
+            'pysha3',
+            # In 0.1.1, pyevmasm changed its gas cost calculations, so Manticore will need to update its
+            # unit tests to match before we can upgrade pyevmasm
+            'pyevmasm==0.1.0',
```
feliam: evm will need ply for the ABI parser I think

disconnect3d: So it's a common dependency, thanks. I haven't run tests for those things yet.

I will have to move some files/classes/exceptions here and there to make things work properly.

I am thinking of creating `manticore/native` directory and move there e.g. `manticore/core/cpu`, `manticore/binary`. Not sure about:
* `manticore/core/memory.py` -  but it doesn't use any native dependencies, so I think I will keep it for now
* `manticore/core/plugin.py` - this will have to be split as it has some native plugins inside
* `manticore/core/state.py` - it uses `from .cpu.abstractcpu import ConcretizeRegister` which has capstone dependency, so I will have to split it somehow. It uses it in `class State(Eventful)` in `def execute` :S.

feliam: I'd say:
* manticore/core/memory.py -> native
*manticore/core/plugin.py -> split +1
* manticore/core/state.py -> refactor. ConcretizeRegister is not fo abstract State. Maybe we need AbstractState, StateLinux and StateEVM ?

examples/evm/asm_to_smtlib.py
```diff
@@ -90,7 +90,7 @@ def set_storage_data(self, address, offset, value):
     def log(self, address, topics, memlog):
         pass
 
-    def send_funds(address, recipient, value):
+    def send_funds(self, address, recipient, value):
```
disconnect3d: This function missed the `self`, so I added it...

manticore/__init__.py
```diff
@@ -1,3 +1,6 @@
-from .manticore import Manticore, STDIN_INPUT_DEFAULT_SIZE  # noqa
-from .models import variadic  # noqa
-from .utils.helpers import issymbolic  # noqa
+from manticore.utils import log
+
+# Default symbolic input size
+STDIN_INPUT_DEFAULT_SIZE = 256
```
disconnect3d: This is a bit weird place to have it in, but the point is it needs to be importable from `__main__.py` and from the `manticore/native/manticore.py`.

manticore/__main__.py
```diff
@@ -1,21 +1,45 @@
...
 
-from . import Manticore, STDIN_INPUT_DEFAULT_SIZE
-from .utils import log, config
-
-# XXX(yan): This would normally be __name__, but then logger output will be pre-
-# pended by 'm.__main__: ', which is not very pleasing. hard-coding to 'main'
-logger = logging.getLogger('manticore.main')
+import logging
+from . import STDIN_INPUT_DEFAULT_SIZE
+# from manticore import STDIN_INPUT_DEFAULT_SIZE
```
disconnect3d: Remove this line

manticore/core/manticore.py
```diff
@@ -154,19 +72,11 @@ def __init__(self, path_or_state, argv=None, workspace_url=None, policy='random'
         # Link Executor events to default callbacks in manticore object
         self.forward_events_from(self._executor)
 
-        if isinstance(path_or_state, str):
```
disconnect3d: The `path_or_state` logic has been moved to the children class: `native/manticore.py:Manticore` since `make_initial_state` creates a target for a native state.

manticore/core/manticore.py
```diff
@@ -227,66 +137,14 @@ def unregister_plugin(self, plugin):
...
+        # If an exception is thrown in a child/inherited class's __init__ before calling super().__init__
+        # the self.plugins is not assigned
+        # because of that, we need to use a `getattr` here, otherwise we might get two exceptions
+        # (and the one from here is irrelevant and confusing)
+        plugins = list(getattr(self, 'plugins', []))
         for plugin in plugins:
             self.unregister_plugin(plugin)
 
-    @classmethod
-    def linux(cls, path, argv=None, envp=None, entry_symbol=None, symbolic_files=None, concrete_start='', stdin_size=STDIN_INPUT_DEFAULT_SIZE, **kwargs):
```
disconnect3d: Moved to `native/manticore.py:Manticore`

manticore/core/manticore.py
```diff
@@ -227,66 +137,14 @@ def unregister_plugin(self, plugin):
...
-        :return: Manticore instance, initialized with a Decree State
-        :rtype: Manticore
-        """
-        try:
-            return cls(make_decree(path, concrete_start), **kwargs)
-        except KeyError:  # FIXME(mark) magic parsing for DECREE should raise better error
-            raise Exception(f'Invalid binary: {path}')
-
-    @classmethod
-    def evm(cls, **kwargs):
```
disconnect3d: **I removed this** instead of moving to `ethereum/manticore.py:Manticore`. I feel like this is not used.
@feliam @mossberg @yan please reply if it is not okay/if we use this classmethod anywhere.

manticore/core/manticore.py
```diff
@@ -648,33 +506,6 @@ def is_shutdown(self):
         ''' Returns True if shutdown was requested '''
         return self._executor.is_shutdown()
 
-    #############################################################################
-    #############################################################################
-    #############################################################################
-    # Move all the following elsewhere Not all manticores have this
-    def _get_symbol_address(self, symbol):
```
disconnect3d: Although it is not used anywhere, I have moved it to `native/manticore.py:Manticore`.

manticore/core/plugin.py
```diff
@@ -245,36 +242,6 @@ def did_finish_run_callback(self):
         logger.info('Coverage: %d different instructions executed', len(executor_visited))
 
 
-class ConcreteTraceFollower(Plugin):
```
disconnect3d: Moved to `native/plugins.py`

manticore/core/state.py
```diff
@@ -138,40 +134,6 @@ def __exit__(self, ty, value, traceback):
         self._child = None
         self.platform.constraints = self.constraints
 
-    def execute(self):
```
disconnect3d: Moved to `native/state.py:State`. The `ethereum/state.py:State` has its own implementation that just does `return self._platform.execute()` for now.

manticore/core/state.py
```diff
@@ -417,20 +379,6 @@ def solve_buffer(self, addr, nbytes, constrain=False):
                 cs_to_use.add(c == result[-1])
         return result
 
-    def invoke_model(self, model):
```
disconnect3d: Moved to `native/state.State`, same goes for the `manticore/models.py` file.

manticore/ethereum/__init__.py
```diff
@@ -1,1657 +1,16 @@
-import random
-import binascii
-import string
-import re
-import os
-import pyevmasm as EVMAsm
+from manticore.utils.install_helper import ensure_evm_deps
```
disconnect3d: Most or all of this file was moved to `ethereum/manticore.py`.
The `FilterFunction` was moved to `ethereum/plugins.py`

manticore/native/__init__.py
```diff
@@ -0,0 +1,6 @@
+from manticore.utils.install_helper import ensure_native_deps
+
+ensure_native_deps()
+
+from .manticore import Manticore
+from ..utils.helpers import issymbolic
```
disconnect3d: @mossberg @feliam @yan Do we need any more imports here?

manticore/native/cli.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def main(args, logger_):
+    env = {key: val for key, val in [env[0].split('=') for env in args.env]}
+
+    m = Manticore(args.argv[0], argv=args.argv[1:], env=env, entry_symbol=args.entrysymbol,
+                  workspace_url=args.workspace, policy=args.policy,
+                  concrete_start=args.data, stdin_size=args.stdin_size)
+
+    m.verbosity(args.v)
```
disconnect3d: disclaimer: this was previously set as `Manticore.verbosity(args.v)` I believe we should make it an instance method instead of static method (this isn't covered by this PR; shall it?).

manticore/utils/install_helper.py
```diff
@@ -0,0 +1,109 @@
+import subprocess
+import sys
+import pkg_resources
+
+
+REQUIREMENTS_TO_IMPORTS = {
```
disconnect3d: This thing might be discussable. Effectively, the evm and native requirements are duplicated here and in `setup.py`. This is because we need to try import all of them to ensure that either evm or native dependencies are installed (and this dict maps pypi package name to its import).

So... in theory we could provide the evm/native package requirements JUST HERE and import it in setup.py, but I am not sure if we want this.

On the other hand I already import this stuff in setup.py to ensure that those lists stay synchronised. 

@mossberg @feliam @yan Please provide feedback; is that fine?

manticore/utils/install_helper.py
```diff
@@ -0,0 +1,109 @@
...
+        try:
+            __import__(import_name)
+        except ImportError:
+            return False
+
+    return True
+
+
+_has_native = _has_deps('native')
+_has_evm = _has_deps('evm')
```
disconnect3d: This just checks the dependencies once for manticore run.

manticore/utils/install_helper.py
```diff
@@ -0,0 +1,109 @@
...
+            '2': 'evm',
+            '3': 'native',
+            '4': 'evm,native'
+        }
+        deps = choice2depname[choice]
+
+        # We need to use (proper) `manticore==<version>` here
+        version = pkg_resources.get_distribution('manticore').version
+
+        cmd = [sys.executable, '-m', 'pip', 'install', f'manticore[{deps}]=={version}']
```
disconnect3d: Note that this won't work before we make the next Manticore release as it launches `pip install manticore[deps]==version`.

If you want to test it out locally, change `manticore[{deps}]` with `.` and run manticore from the repository directory (it might work with changing `.` to path to manticore repo but I'm not sure about that).

setup.py
```diff
@@ -21,35 +81,12 @@ def rtd_dependent_deps():
...
-        'pyelftools',
-        'unicorn',
-        'ply',
-        'pysha3',
-        # In 0.1.1, pyevmasm changed its gas cost calculations, so Manticore will need to update its
-        # unit tests to match before we can upgrade pyevmasm
-        'pyevmasm==0.1.0',
-        'pyyaml',
-    ] + rtd_dependent_deps(),
+                         'pyyaml',
```
disconnect3d: Fix indent

tests/mockmem.py
```diff
@@ -1,39 +1,52 @@
 from manticore.core.smtlib import Operators
 
-class Memory:  #todo Mock
+
+class Memory:  # todo Mock
```
disconnect3d: I felt the urge to fire reformat on this file, doh.

tests/test_binaries.py
```diff
@@ -146,17 +146,18 @@ def test_logger_verbosity(self):
...
+        self.assertEqual(len(output_lines), 6)
+
+        self.assertIn(b'Loading program', output_lines[0])
+        self.assertIn(b'Verbosity set to 1.', output_lines[1])
+        self.assertIn(b'Generated testcase No. 0 - Program finished with exit status: 0', output_lines[2])
+        self.assertIn(b'Generated testcase No. 1 - Program finished with exit status: 0', output_lines[3])
+        self.assertIn(b'Results in ', output_lines[4])
+        self.assertIn(b'Total time: ', output_lines[5])
 
-        for line in testcase_info:
```
disconnect3d: Uh, this for here was wrong, the `testcase_info` was empty because there were not that much output to account for `[2:-5]` slice. As a result, the assertion below never fired.

But the test was helpful anyway.

tests/test_memory.py
```diff
@@ -1772,27 +1776,36 @@ def test_symbolic_force_access(self):
...
-        import manticore.utils.mappings
+        from manticore.native import mappings
         import ctypes
 
+        old_cdll = ctypes.cdll
+        old_mapping_sys = mappings.sys
+
         ctypes.cdll = mock.MagicMock()
-        manticore.utils.mappings.sys = mock.MagicMock()
+        mappings.sys = mock.MagicMock()
```
disconnect3d: Note that those monkey patches here are super dangerous as we don't revert them.

I added a simple reverting of them, but if an assertion fails, this won't be reverted (shall be done with some context manager).

manticore/core/manticore.py
yan: This is likely okay, since we're not after native/evm symmetry anyway.

manticore/__init__.py
yan: should this be a value adjustable from the `config` system?

manticore/__main__.py
```diff
@@ -1,21 +1,44 @@
...
+    sys.setrecursionlimit(consts.recursionlimit)
+
+    log.set_verbosity(args.v)
+
+    if args.argv[0].endswith('.sol'):
+        from manticore.ethereum.cli import main as evm_main
+        evm_main(args, logger)
+    else:
+        from manticore.native.cli import main as native_main
+        native_main(args, logger)
```
yan: IMHO, i'd rather import the above `main` functions without aliasing them; it might make grepping for them harder. I think `from a.b.c import main; main()` is pretty straightforward.

manticore/__init__.py
mossberg: ideally, yes

scripts/travis_test.sh
```diff
@@ -143,7 +143,7 @@ if [ "$should_run_tests" = true ]; then
     echo "Measuring code coverage..."
     measure_cov "manticore/core/smtlib/*" 80
     measure_cov "manticore/core/cpu/x86.py" 50
-    measure_cov "manticore/core/memory.py" 85
```
yan: Leaving comment here, but it's about travis in general. We should only install `[native]` for linux/cpu tests and only install `[evm]` for evm tests to make sure the delineation is clear and complete.

manticore/core/manticore.py
```diff
@@ -227,66 +137,14 @@ def unregister_plugin(self, plugin):
         plugin.manticore = None
 
     def __del__(self):
-        plugins = list(self.plugins)
+        # If an exception is thrown in a child/inherited class's __init__ before calling super().__init__
+        # the self.plugins is not assigned
+        # because of that, we need to use a `getattr` here, otherwise we might get two exceptions
+        # (and the one from here is irrelevant and confusing)
+        plugins = list(getattr(self, 'plugins', []))
```
yan: Why is `list()` required here?

manticore/core/workspace.py
```diff
@@ -125,7 +125,7 @@ def save_state(self, state, key):
         """
         Save a state to storage.
 
-        :param manticore.core.State state:
+        :param manticore.core.BaseState state:
```
yan: `BaseState` or `StateBase`?	(Ditto for below)

manticore/native/manticore.py
```diff
@@ -0,0 +1,192 @@
...
+from elftools.elf.elffile import ELFFile
+from elftools.elf.sections import SymbolTableSection
+
+from manticore.core.manticore import ManticoreBase
+from manticore import STDIN_INPUT_DEFAULT_SIZE
+
+from manticore.core.smtlib import ConstraintSet
+from manticore.utils import log
+from manticore.utils.helpers import issymbolic
+from .state import State
```
yan: Mixing absolute and relative imports. Prefer relative?

manticore/platforms/evm.py
```diff
@@ -94,10 +94,11 @@ def dump(self, stream, state, mevm, conc_tx=None):
 
         :param stream: Output stream to write to. Typically a file.
         :param manticore.core.state.State state: state that the tx exists in
-        :param manticore.ethereum.ManticoreEVM mevm: manticore instance
+        :param ethereum.manticore.ManticoreEVM mevm: manticore instance
```
mossberg: should this be `manticore.ethereum.manticore.ManticoreEVM`?

manticore/native/state.py
```diff
@@ -0,0 +1,44 @@
+from manticore.core.state import StateBase, Concretize, TerminateState
+from manticore.native.memory import ConcretizeMemory, MemoryException
+
+
+class State(StateBase):
```
mossberg: i wonder if we should not use the same name `State` for both ethereum and native. `NativeState` and `EthereumState`?

manticore/platforms/evm.py
disconnect3d: Either `manticore.ethereum.ManticoreEVM` or as you wrote, will use the first one.

manticore/native/state.py
disconnect3d: @yan what do you think here?

I am up for using `NativeState` and `EthereumState`, being more explicit is probably better here.

manticore/native/manticore.py
```diff
@@ -0,0 +1,192 @@
...
+        :param kwargs: Forwarded to the Manticore constructor
+        :return: Manticore instance, initialized with a Decree State
+        :rtype: Manticore
+        """
+        try:
+            return cls(_make_decree(path, concrete_start), **kwargs)
+        except KeyError:  # FIXME(mark) magic parsing for DECREE should raise better error
+            raise Exception(f'Invalid binary: {path}')
+
+    def invoke_model(self, model):
```
mossberg: i think this should be on the native State, rather than on native Manticore

manticore/native/state.py
yan: I don't have strong opinions on this. I would actually have been fine with `native.State` and `evm.State`, since module/namespace can describe the kind, not the name itself, but `NativeState` and `EthereumState` are both fine with me.

manticore/__main__.py
disconnect3d: The reason for that is that the `__main__.py` also contains `main` function, heh.

But I guess we can just name them as `native_main` and` evm_main`?

manticore/core/manticore.py
disconnect3d: It's not, fixed.

manticore/native/manticore.py
disconnect3d: Nice catch! I have no idea how I moved it to Manticore :|.

disconnect3d:
> manticore.native.Manticore() no longer takes a path
It still does accept `path_or_state` for native Manticore :)

disconnect3d:
Speaking of helpers we can:
* expose it through `manticore` e.g. `from manticore import issymbolic` (as long as they don't import native stuff)
* expose it through both `manticore.ethereum` and `manticore.native`
* expose it only through `manticore.helpers`

Hard to say what's the best option but I would be up for leaving it the same as it was, so it doesn't bother people that much?

@yan @mossberg

disconnect3d:
___
*[examples/script/introduce_symbolic_bytes.py, line 4 at r8](https://reviewable.io/reviews/trailofbits/manticore/1257#-LSVssDi6L9IjP0jzqA8:-LSntKmD5EHLNEi4W1IT:begb22g) ([raw file](https://github.com/trailofbits/manticore/blob/90f1371b861cfead1b68aca21fb0c5d4322622df/examples/script/introduce_symbolic_bytes.py#L4)):*

`issymbolic` should come from `manticore.utils.helpers`, not there.
</blockquote></details>

See https://github.com/trailofbits/manticore/pull/1257#issuecomment-443191505

disconnect3d:
___
*[manticore/core/cpu/abstractcpu.py, line 1 at r8](https://reviewable.io/reviews/trailofbits/manticore/1257#-LSVrGu21-4mqKdc2b5T:-LSnt_JG0nAExYa7SDnc:bqequbx) ([raw file](https://github.com/trailofbits/manticore/blob/90f1371b861cfead1b68aca21fb0c5d4322622df/manticore/core/cpu/abstractcpu.py#L1)):*

at some point this `core/cpu` would get moved to `native/cpu` or something right?
</blockquote></details>

Done, moved to native/cpu

disconnect3d:
___
*[manticore/core/cpu/abstractcpu.py, line 1 at r8](https://reviewable.io/reviews/trailofbits/manticore/1257#-LSVul9z21nVvafrC_jg:-LSntb2SB8gJKycFtAVy:b9dkjj3) ([raw file](https://github.com/trailofbits/manticore/blob/90f1371b861cfead1b68aca21fb0c5d4322622df/manticore/core/cpu/abstractcpu.py#L1)):*

Should `abstractcpu.py` be inside `native` or inside `core`?
</blockquote></details>

I moved it in native/cpu

disconnect3d:
___
*[manticore/core/cpu/x86.py, line 15 at r8](https://reviewable.io/reviews/trailofbits/manticore/1257#-LSVrXYuAZYnqnVmxy2l:-LSnth-eAMa8LfwucMai:b-896fix) ([raw file](https://github.com/trailofbits/manticore/blob/90f1371b861cfead1b68aca21fb0c5d4322622df/manticore/core/cpu/x86.py#L15)):*

relative pls
</blockquote></details>

Done.

disconnect3d:
> > Speaking of helpers we can:
> > 
> > * expose it through `manticore` e.g. `from manticore import issymbolic` (as long as they don't import native stuff)
> > * expose it through both `manticore.ethereum` and `manticore.native`
> > * expose it only through `manticore.helpers`
> > 
> > Hard to say what's the best option but I would be up for leaving it the same as it was, so it doesn't bother people that much?
> 
> I say let's go with the first one

That's what I did in https://github.com/trailofbits/manticore/pull/1257/commits/2b2d71f1f3dfa0563b8ff38b55951213ff86b392 :)

PR #1267, opened 2018-11-08
Change generate_testcase name to have a default val

disconnect3d:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1267)

### DIFF
--- a/manticore/ethereum/__init__.py
+++ b/manticore/ethereum/__init__.py
@@ -1304,7 +1304,7 @@ class ManticoreEVM(Manticore):
     def workspace(self):
         return self._executor._workspace._store.uri
 
-    def generate_testcase(self, state, name, message='', only_if=None):
+    def generate_testcase(self, state, message='', only_if=None, name='custom'):
         """
         Generate a testcase to the workspace for the given program state. The details of what
         a testcase is depends on the type of Platform the state is, but involves serializing the state,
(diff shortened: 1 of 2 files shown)


disconnect3d:
@mossberg Hmm shall we also change `def _generate_testcase_callback(self, state, name, message=''):` argument order? I think it is fine since it is more like an internal/private function?

manticore/ethereum/__init__.py
```diff
@@ -1320,14 +1320,13 @@ def generate_testcase(self, state, name, message='', only_if=None):
         For example, invariant: "balance" must not be 0. We can check if this can be violated and generate a
         testcase::
 
-            m.generate_testcase(state, 'balance', 'balance CAN be 0', only_if=balance == 0)
+            m.generate_testcase(state, 'balance CAN be 0', only_if=balance == 0, name='balance')
```
mossberg: let's actually just remove the `name='balance'` thing here. the 'balance' parameter was only there previously because some kind of `name` was required. now that's it's not, it's simpler to omit it.

manticore/ethereum/__init__.py
```diff
@@ -1304,7 +1304,7 @@ def unregister_detector(self, d):
     def workspace(self):
         return self._executor._workspace._store.uri
 
-    def generate_testcase(self, state, name, message='', only_if=None):
+    def generate_testcase(self, state, message='', only_if=None, name='custom'):
```
mossberg: another idea, maybe 'user' would actually be a better default name here

manticore/ethereum/__init__.py
disconnect3d:
```suggestion
            m.generate_testcase(state, 'balance CAN be 0', only_if=balance == 0)
```

manticore/ethereum/__init__.py
disconnect3d:
```suggestion
    def generate_testcase(self, state, message='', only_if=None, name='user'):
```

PR #1276, opened 2018-11-20
Add hook injector script

ex0dus-0x:
This script injects hooks into memory addresses based on a user-specified pattern for a symbol. 

This harnesses `pyelftools` to check input as a regex against symbol table entries. Memory addresses are retrieved from the `st_value` attribute, and then hooked through `mcore.add_hook` with a hook callback.

Is not yet fully tested, but just wanted to make this known for feedback, etc.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1276)

### DIFF
--- a/scripts/hook_injector.py
+++ b/scripts/hook_injector.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python3
+'''
+hook_injector.py
+
+    This example harnesses the power of
+    Manticore as a debugger during program analysis,
+    providing introspection of hooks injected through
+    user-specified pattern inputs.
+
+'''
+import re
+import argparse
+
+from manticore import Manticore
+from elftools.elf.elffile import ELFFile
+from elftools.elf.sections import SymbolTableSection
+
+
+class Introspector(object):
+    """
+    Introspector object is used to correctly parse
+    an ELF binary with a symbol table
+    """
+
+    def __init__(self, target_f, pattern):
+
+        # collected mem addresses from symbol pattern
+        self.functions = []
+
+        # instantiate manticore with binary target
+        self.mcore = Manticore(target_f)
+
+        with open(target_f, 'rb') as binary:
+
+            # parse binary
+            elf = ELFFile(binary)
+            if elf is None:
+                raise Exception("symbols are unsupported")
+
+            # retrieve symbols from all sections in binary
+            for section in elf.iter_sections():
+
+                # check if symbol table
+                if not isinstance(section, SymbolTableSection):
+                    continue
+
+                # retrieve ALL symbols
+                syms = section.iter_symbols()
+                if syms is None:
+                    raise Exception("no symbols found")
+
+                # create dict of symbols and respective memory addr, which is represented by
+                # 'st_value' attribute in symbol table entry
+                symbols = dict((sym.name, sym.entry['st_value']) for sym in syms)
+
+                # regex matching against symbols
+                for name, addr in symbols.items():
+                    if re.match(pattern, name):
+                        self.functions.append(addr)
+
+
+    def attach_and_run(self, callback):
+        """
+        Attach Manticore hooks to all memory addresses
+        and run symbolic executor
+        """
+        
+        for addr in self.functions:
+            self.mcore.add_hook(addr, callback)
+
+        self.mcore.verbosity = 2
+        self.mcore.workers = 1
+        self.mcore.run()
+
+
+
+def main():
+
+    # initialize parser and parse arguments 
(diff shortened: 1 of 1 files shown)


scripts/hook_injector.py
```diff
@@ -0,0 +1,100 @@
...
+from elftools.elf.sections import SymbolTableSection
+
+
+class Introspector(object):
+    """
+    Introspector object is used to correctly parse
+    an ELF binary with a symbol table
+    """
+
+    def __init__(self, target_f, pattern):
```
disconnect3d: A class with one or no methods apart from `__init__` can be just a function. Side note: there's also a great talk about not writing classes in Python ( https://www.youtube.com/watch?v=o9pEzgHorH0 ).

scripts/hook_injector.py
```diff
@@ -0,0 +1,100 @@
...
+
+        with open(target_f, 'rb') as binary:
+
+            # parse binary
+            elf = ELFFile(binary)
+            if elf is None:
+                raise Exception("symbols are unsupported")
+
+            # retrieve symbols from all sections in binary
+            for section in elf.iter_sections():
```
disconnect3d: Loud thinking: maybe we shall move this (yielding/generating sections or symbols or returning a list of them?) into `Manticore` and use such method also in here https://github.com/trailofbits/manticore/blob/master/manticore/manticore.py#L655-L676 ?

scripts/hook_injector.py
```diff
@@ -0,0 +1,100 @@
...
+    This example harnesses the power of
+    Manticore as a debugger during program analysis,
+    providing introspection of hooks injected through
+    user-specified pattern inputs.
+
+'''
+import re
+import argparse
+
+from manticore import Manticore
```
disconnect3d: This will need to be changed to `from manticore.native import Manticore` after https://github.com/trailofbits/manticore/pull/1257 is merged (shall be soonish/today)

scripts/hook_injector.py
disconnect3d: The function should probably just return what is in `self.functions` currently.

scripts/hook_injector.py
```diff
@@ -0,0 +1,100 @@
...
+class Introspector(object):
+    """
+    Introspector object is used to correctly parse
+    an ELF binary with a symbol table
+    """
+
+    def __init__(self, target_f, pattern):
+
+        # collected mem addresses from symbol pattern
+        self.functions = []
```
disconnect3d: The name is misleading: this stores addresses of symbols that are matching a given pattern, maybe `matched_addresses` or `matching_addresses`?

scripts/hook_injector.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python3
+'''
+hook_injector.py
+
+    This example harnesses the power of
+    Manticore as a debugger during program analysis,
+    providing introspection of hooks injected through
+    user-specified pattern inputs.
```
disconnect3d: Any chance to have a minimal working example here? E.g. a little C code and its output example? Thanks to that the script usability would be immediately visible for future readers :D.

scripts/hook_injector.py
```diff
@@ -0,0 +1,100 @@
...
+        self.mcore.run()
+
+
+
+def main():
+
+    # initialize parser and parse arguments 
+    parser = argparse.ArgumentParser(description='Inject hooks at user-specified patterns')
+    parser.add_argument('-i', '--input', type=str,
+                        help="Pattern input for hook injection")
```
disconnect3d:
```suggestion
                        help="Regex pattern input for hook injection")
```

scripts/hook_injector.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+    def attach_and_run(self, callback):
+        """
+        Attach Manticore hooks to all memory addresses
+        and run symbolic executor
+        """
+        
+        for addr in self.functions:
+            self.mcore.add_hook(addr, callback)
```
disconnect3d: Imho this code should be moved to `main`.

scripts/hook_injector.py
```diff
@@ -0,0 +1,100 @@
...
+
+        self.mcore.verbosity = 2
+        self.mcore.workers = 1
+        self.mcore.run()
+
+
+
+def main():
+
+    # initialize parser and parse arguments
```
disconnect3d: The comment is not needed

mossberg:
i agree with @yan, i think manticore-examples would be a better home for this script :) can we move it to a pr over there?

PR #1279, opened 2018-11-23
Make gas calculation faithfulness configurable

feliam:
* split extra gas calculation to its own method
* add config constant for choosing  gas faithfulness type  (faithful => slow)

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1279)


manticore/platforms/evm.py
```diff
@@ -21,10 +23,25 @@
...
+# 4: OOG soon. If it may NOT be enough gas we ignore the normal case. A constraint is added to assert the gas is NOT enough and the other state is ignored.
+# 99: Ignore gas. Do not account for it. Do not OOG.
+consts = config.get_group('evm')
+consts.add('oog', default=0, description='Default behavior for symbolic gas.'
+                                         '0: Fully faithfull. Test at every instruction. Forks.'
+                                         '1: Mostly faithful. Test at BB limit. Forks.'
+                                         '2: Incomplete. Concretize gas to MIN/MAX values. Forks.'
+                                         '3: Try to not fail due to OOG. If it can be enough gas use it. Ignore the path to OOG. Wont fork'
+                                         '4: Try OOG only. Fail soon. Ignore the path with enough gas.'
+                                         '99: Ignore gas. Instructions wont consume gas')
```
```suggestion
                                         "99: Ignore gas. Instructions won't consume gas")
```

manticore/platforms/evm.py
```diff
@@ -21,10 +23,25 @@
...
+# This configuration variable allows the user to control and perhaps relax the gas calculation
+# 0: gas is faithfully accounted and checked at instruction level. State may get forked in OOG/NoOOG
+# 1: gas is faithfully accounted and checked at basic blocks limits. State may get forked in OOG/NoOOG
+# 2: Concretize gas. If the fee to be consumed gets to be symbolic. Choose some potential values and fork on those.
+# 3: Try not to OOG. If it may be enough gas we ignore the OOG case. A constraint is added to assert the gas is enough and the OOG state is ignored.
+# 4: OOG soon. If it may NOT be enough gas we ignore the normal case. A constraint is added to assert the gas is NOT enough and the other state is ignored.
+# 99: Ignore gas. Do not account for it. Do not OOG.
+consts = config.get_group('evm')
+consts.add('oog', default=0, description='Default behavior for symbolic gas.'
+                                         '0: Fully faithfull. Test at every instruction. Forks.'
```
```suggestion
                                         '0: Fully faithful. Test at every instruction. Forks.'
```

manticore/platforms/evm.py
```diff
@@ -244,18 +261,38 @@ class EVMException(Exception):
...
 
-class ConcretizeStack(EVMException):
+class ConcretizeArgument(EVMException):
     '''
-    Raised when a symbolic memory cell needs to be concretized.
+    Raised when a symbolic argument needs to be concretized.
     '''
 
-    def __init__(self, pos, expression=None, policy='MINMAX'):
+    def __init__(self, pos, expression=None, policy='SAMPLED'):
```
disconnect3d: What are the consequences of changing the policy from MINMAX to SAMPLED?

manticore/platforms/evm.py
```diff
@@ -244,18 +261,38 @@ class EVMException(Exception):
...
         self.policy = policy
 
 
+class ConcretizeFee(EVMException):
+    '''
+    Raised when a symbolic gas fee needs to be concretized.
+    '''
+
+    def __init__(self, policy='MINMAX'):
+        self.message = "Concretizing evm instruction gas fee {}"
```
disconnect3d: Do we want the `{}` here? If so, let's add a comment that it is intended

manticore/platforms/evm.py
```diff
@@ -722,53 +760,78 @@ def _pop(self):
...
+                #explore both options / fork
+                #FIXME this will reenter here and generate redundant queries
+                enough_gas_solutions = get_possible_solutions()
+                if len(enough_gas_solutions) == 2:
+                    #if gas can be both enough an insuficient, fork
+                    raise Concretize("Concretize gas fee",
+                                     expression=Operators.UGT(self._gas, fee),
+                                     setstate=None,
+                                     policy='ALL')
+                elif enough_gas_solutions[0] == False:
```
```suggestion
                elif enough_gas_solutions[0] is False:
```

manticore/platforms/evm.py
```diff
@@ -722,53 +760,78 @@ def _pop(self):
...
+                    #if gas if only insuficient OOG!
+                    logger.debug("Not enough gas for instruction")
+                    raise NotEnoughGas()
+                else:
+                    assert enough_gas_solutions[0] == True
```
```suggestion
                    assert enough_gas_solutions[0] is True
```

manticore/platforms/evm.py
```diff
@@ -21,10 +23,25 @@
...
+# When gas is concrete the gas checks and calculation are pretty straigth forward
+# Though Gas can became symbolic in normal bytecode execution for example at instructions
+# MSTORE, MSTORE8, EXP, ... and every instruction with internal operation restricted by gas
+# This configuration variable allows the user to control and perhaps relax the gas calculation
+# 0: gas is faithfully accounted and checked at instruction level. State may get forked in OOG/NoOOG
+# 1: gas is faithfully accounted and checked at basic blocks limits. State may get forked in OOG/NoOOG
+# 2: Concretize gas. If the fee to be consumed gets to be symbolic. Choose some potential values and fork on those.
+# 3: Try not to OOG. If it may be enough gas we ignore the OOG case. A constraint is added to assert the gas is enough and the OOG state is ignored.
+# 4: OOG soon. If it may NOT be enough gas we ignore the normal case. A constraint is added to assert the gas is NOT enough and the other state is ignored.
+# 99: Ignore gas. Do not account for it. Do not OOG.
```
disconnect3d: Can we have a mode in which we adjust the gas so it exactly accounts all the given tx (or conditions/constraints)?

disconnect3d: Like what if one do `require(gasNow() == 0x1234)` ?

manticore/platforms/evm.py
```diff
@@ -816,6 +879,13 @@ def _push_results(self, instruction, result):
             assert instruction.pushes == 0
             assert result is None
 
+    def _calculate_extra_gas(self, *arguments):
+        current = self.instruction
+        implementation = getattr(self, f"{current.semantics}_gas", None)
```
disconnect3d: What are the possible values here? Maybe an `if, elif, else` instead?

manticore/platforms/evm.py
```diff
@@ -834,23 +905,26 @@ def _checkpoint(self):
...
+        ''' Revert the stack, gas, pc and memory allocation so it looks like before executing the instruction '''
+        last_pc, last_gas, last_instruction, last_arguments, fee, allocated = self._checkpoint_data
         self._push_arguments(last_arguments)
         self._gas = last_gas
         self._pc = last_pc
+        self._allocated = allocated
         self._checkpoint_data = None
 
     def _set_check_jmpdest(self, flag=True):
+        # TODO: Document. This enables the check that the next instruction must be a JUMPDEST
```
disconnect3d: Can we document it in this PR?

manticore/platforms/evm.py
```diff
@@ -915,19 +989,45 @@ def setstate(state, value):
...
+            def setstate(state, value):
+                current_vm = state.platform.current_vm
+                _pc, _old_gas, _instruction, _arguments, _fee, _allocated = current_vm._checkpoint_data
+                new_arguments = []
+                for old_arg in _arguments:
+                    if len(new_arguments) == pos:
+                        new_arguments.append(value)
+                    else:
+                        new_arguments.append(old_arg)
+                    current_vm._checkpoint_data = (_pc, _old_gas, _instruction, new_arguments, _fee, _allocated)
```
disconnect3d: shouldn't this be out of the `for` loop?
```suggestion
                current_vm._checkpoint_data = (_pc, _old_gas, _instruction, new_arguments, _fee, _allocated)
```

manticore/platforms/evm.py
```diff
@@ -1200,10 +1308,12 @@ def ADDRESS(self):
         '''Get address of currently executing account'''
         return self.address
 
+    def BALANCE_gas(self, account):
+        BALANCE_SUPPLEMENTAL_GAS = 380
+        return BALANCE_SUPPLEMENTAL_GAS
```
disconnect3d: Maybe just return 380 and add a comment?

manticore/platforms/evm.py
```diff
@@ -1251,6 +1361,12 @@ def CALLDATASIZE(self):
         '''Get size of input data in current environment'''
         return self._calldata_size
 
+    def CALLDATACOPY_gas(self, mem_offset, data_offset, size):
+        GCOPY = 3             # cost to copy one 32 byte word
```
disconnect3d: Just wondering why does it have a prefix 'G'? Is it a 'GAS COPY COST'?

manticore/platforms/evm.py
```diff
@@ -1385,12 +1509,18 @@ def POP(self, a):
...
+        return self._get_memfee(address + 32)
+
     def MLOAD(self, address):
         '''Load word from memory'''
         self._allocate(address, 32)
         value = self._load(address, 32)
         return value
 
+    def MSTORE_gas(self, address, value):
+        return self._get_memfee(address + 32)
```
disconnect3d: It's probably good to add a comment why is it +32 or +1 here and there (at different funcs)

manticore/platforms/evm.py
```diff
@@ -1621,6 +1781,10 @@ def STATICCALL(self, gas, address, in_offset, in_size, out_offset, out_size):
 
         return self.world.last_transaction.return_value
 
+    def REVERT_gas(self, offset, size):
+        memfee = self._get_memfee(offset, size)
```
disconnect3d: just `return self._get_memfee(offset, size)` - same for other calls

manticore/platforms/evm.py
```diff
@@ -244,18 +261,38 @@ class EVMException(Exception):
     pass
 
 
-class ConcretizeStack(EVMException):
+class ConcretizeArgument(EVMException):
     '''
-    Raised when a symbolic memory cell needs to be concretized.
+    Raised when a symbolic argument needs to be concretized.
     '''
```
disconnect3d: Can we use `"""` instead of `'''`?

manticore/platforms/evm.py
```diff
@@ -722,53 +760,78 @@ def _pop(self):
         return self.stack.pop()
 
     def _consume(self, fee):
+
+        #Check type and bitvec size
```
```suggestion
        # Check type and bitvec size
```

manticore/platforms/evm.py
```diff
@@ -21,10 +23,25 @@
...
-#fixme make it global using this https://docs.python.org/3/library/configparser.html
-#and save it at the workspace so results are reproducible
-config = namedtuple("config", "out_of_gas")
-config.out_of_gas = None  # 0: default not enough gas, 1 default to always enough gas, 2: for on both
+#Gas behaivor configuration
+# When gas is concrete the gas checks and calculation are pretty straigth forward
+# Though Gas can became symbolic in normal bytecode execution for example at instructions
+# MSTORE, MSTORE8, EXP, ... and every instruction with internal operation restricted by gas
+# This configuration variable allows the user to control and perhaps relax the gas calculation
+# 0: gas is faithfully accounted and checked at instruction level. State may get forked in OOG/NoOOG
```
yan: Can we assign these options symbolic names, like `pedantic` (0), `complete` (1), `eager` (2), `optimistic` (3), `defer` (4) and `ignore` (99)? Afraid that if we store the command line with the saved flag, it's not very clear what the value was initially.

manticore/platforms/evm.py
feliam: MINMAX selects the maximun and the minimun value.
and SAMPLED takes MIN/MAX _and_ some other values in the middle.

manticore/platforms/evm.py
feliam: That comes from inspiration gained from other projects.
Name suggestions welcome.

disconnect3d:
One additional feature request here: can we add a display of gas to https://github.com/trailofbits/manticore/blob/master/manticore/platforms/evm.py#L1651-L1728 so e.g. the `VerboseTrace` plugin will show it?

We should probably also display the gas cost of given executed instructions. But this can be done in another PR.

manticore/platforms/evm.py
feliam: In that case `gasNow()` will return (for option 1) a symbolic value describing it. 
And the state will be forked in `=0x1234` and  `!=0x1234`.
If you want to ignore one of those states is more a task for general state selection/policy.

manticore/platforms/evm.py
disconnect3d: Ok right, I get it now. This probably slows down the execution a lot (we do some more calls and calls are  kinda expensive in Python) but let's not focus on that now.

feliam: Hmm yes. Better ways for distpatching gas calculation are wellcome. Lets discuss it in a diff PR/Issue.

Though, we can disable it all by not calling `_calculate_extra_gas` at all when at that specific mode.

It wont be.. well.. real EVM but ppl do not care much about it.

manticore/platforms/evm.py
feliam: Fixed/improved  in _get_memfee (and _use_data)

manticore/platforms/evm.py
```diff
@@ -21,10 +23,25 @@
 
 logger = logging.getLogger(__name__)
 
-#fixme make it global using this https://docs.python.org/3/library/configparser.html
-#and save it at the workspace so results are reproducible
-config = namedtuple("config", "out_of_gas")
-config.out_of_gas = None  # 0: default not enough gas, 1 default to always enough gas, 2: for on both
+#Gas behaivor configuration
```
```suggestion
# Gas behaviour configuration
```

manticore/platforms/evm.py
```diff
@@ -608,6 +644,16 @@ def __setstate__(self, state):
         super().__setstate__(state)
 
     def _get_memfee(self, address, size=1):
+        """
+            This calculates the ammount of extra gas needed for accessing to
```
```suggestion
            This calculates the amount of extra gas needed for accessing to
```
Also, can we remove the additional indent made here?

manticore/platforms/evm.py
```diff
@@ -608,6 +644,16 @@ def __setstate__(self, state):
...
+            previously unused memory.
+
+            `address` is the base memory offset and,
```
```suggestion
        :param address: base memory offset
```

manticore/platforms/evm.py
```diff
@@ -608,6 +644,16 @@ def __setstate__(self, state):
...
+            `size` is the size of the memory access.
```
```suggestion
        :param size: size of the memory access
```

PR #1289, opened 2018-12-06
Fix command.sh not saved in ethereum

disconnect3d:
For native binaries Manticore saves `command.sh` and `manticore.yml` in
a `_did_finish_run_callback`: https://github.com/trailofbits/manticore/blob/c29f7cf8672014cf02cfc1271ec1d0a3cb3983af/manticore/core/manticore.py#L524-L533

For ethereum smart contracts this callback is
called two times (not sure why), so we supress it instead: https://github.com/trailofbits/manticore/blob/c29f7cf8672014cf02cfc1271ec1d0a3cb3983af/manticore/ethereum/manticore.py#L1523-L1526

This PR moves this logic to a `ManticoreBase._save_run_data` method, so
it can be used in both native and ethereum Manticore engines.

As a result:
* native engine works as it worked
* ethereum engine will now save `command.sh` file
* ethereum engine will now print the elapsed time

Additionally, a bug with `command.sh` output has been fixed: when one
passed argument that had spaces, e.g.:
```
manticore "path to contract/with_spaces.sol"
```

this resulted in a `command.sh` having such content:

```
manticore path to contract/with_spaces.sol
```

which is wrong as it shall be `manticore "path to contract/with_spaces.sol"` or `manticore 'path to contract/with_spaces.sol'`.

This has been fixed by processing all arguments with `shlex.quote`, so
they are quoted properly.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1289)

### DIFF
--- a/manticore/core/manticore.py
+++ b/manticore/core/manticore.py
@@ -9,6 +9,7 @@ from multiprocessing import Process
 from threading import Timer
 
 import functools
+import shlex
 import types
 
 from ..core.executor import Executor
@@ -522,8 +523,11 @@ class ManticoreBase(Eventful):
         self._coverage_file = path
 
     def _did_finish_run_callback(self):
+        self._save_run_data()
+
+    def _save_run_data(self):
         with self._output.save_stream('command.sh') as f:
-            f.write(' '.join(sys.argv))
+            f.write(' '.join(map(shlex.quote, sys.argv)))
 
         with self._output.save_stream('manticore.yml') as f:
             config.save(f)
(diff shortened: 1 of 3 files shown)


feliam:
`run()` is called once for each tx in Ethereum. There each tx is like running one native program after the other but maintaining the FS/storage. We may need more/different event names.
Do we still need command.sh now that we have manticore.yml?

disconnect3d:
> We may need more/different event names.

Yeah..

The `command.sh` is nice as we can just copy and fire it but I am not that super convinced that we need it.

@yan or @mossberg any opinions here?

PR #1300, opened 2018-12-12
Don't clone Manticore in Docker image

disconnect3d:
So it is possible to build image from given Manticore version (and not only from master).

Changes:
- use `LABEL`s instead of `MAINTAINER`
- use `DEBIAN_FRONTEND=noninteractive` for all `apt-get install` (not necessarily required, but probably a good practice)
- use local Manticore files instead of cloning it from Github (so we can build a container from given commit)
- change from `manticore` user to `root`, see below for more info

**NOTE: This changes the user in the container from `manticore` to `root`.** Still, I believe it is not that bad. My reasoning:
* The container is not used to host anything - it is mostly used locally
* It is now possible to install additional software if needed; otherwise, to do that the user would have to run the container with `--user root` flag

Also note that I didn't remove `git` from installed software. I believe our users might still want to use git e.g. to clone some repository (although they might rather use bind mounts instead, if their repositories are private).

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1300)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential software-p
     apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y solc ethereum
 
-# Install Manticore
 ADD . /manticore
 RUN cd manticore && pip3 install .[native]
-z
+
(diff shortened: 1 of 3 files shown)


mossberg:
this is cool :D

PR #1302, opened 2018-12-13
Add symbol resolution feature for binary analysis

ex0dus-0x:
This PR adds symbol resolution through `core.Manticore.resolve()` for ELF binaries through DWARF debug information and symbol tables.

```
@m.hook(m.resolve('MyLib_function', line=4))
def hook(state):
    m.terminate()

@m.hook(m.resolve('MyLib_otherfunc'))
def hook(state):
    m.abort()
```

If the user does not supply a `line` argument, the first instance of the symbol will be resolved and its memory address will be hooked on.

I have yet to implement unit tests (if necessary), but this initial implementation seems to be working. Feedback for improvements and optimizations welcome!

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1302)

### DIFF
--- a/manticore/native/manticore.py
+++ b/manticore/native/manticore.py
@@ -137,33 +137,6 @@ class Manticore(ManticoreBase):
         for cb in self._hooks.get(None, []):
             cb(state)
 
-    #############################################################################
-    #############################################################################
-    #############################################################################
-    # Move all the following elsewhere Not all manticores have this
-    def _get_symbol_address(self, symbol):
-        '''
-        Return the address of |symbol| within the binary
-        '''
-
-        # XXX(yan) This is a bit obtuse; once PE support is updated this should
-        # be refactored out
-        if self._binary_type == 'ELF':
-            self._binary_obj = ELFFile(open(self._binary, 'rb'))
-
-        if self._binary_obj is None:
-            return NotImplementedError("Symbols aren't supported")
-
-        for section in self._binary_obj.iter_sections():
-            if not isinstance(section, SymbolTableSection):
-                continue
-
-            symbols = section.get_symbol_by_name(symbol)
-            if not symbols:
-                continue
-
-            return symbols[0].entry['st_value']
-
 
 def _make_initial_state(binary_path, **kwargs):
     with open(binary_path, 'rb') as f:
--- a/manticore/native/manticore.py
+++ b/manticore/native/manticore.py
@@ -26,8 +26,6 @@ class Manticore(ManticoreBase):
         if isinstance(path_or_state, str):
             if not os.path.isfile(path_or_state):
                 raise OSError(f'{path_or_state} is not an existing regular file')
-
-            self._binpath = path_or_state
             initial_state = _make_initial_state(path_or_state, argv=argv, **kwargs)
         else:
             initial_state = path_or_state
@@ -79,6 +77,15 @@ class Manticore(ManticoreBase):
             return cls(_make_decree(path, concrete_start), **kwargs)
         except KeyError:  # FIXME(mark) magic parsing for DECREE should raise better error
             raise Exception(f'Invalid binary: {path}')
+
+    @property
+    def binary_path(self):
+        """
+        Assumes that self._initial_state.platform.program always
+        refers to current program. Might not be true in case program
+        calls execve().
+        """
+        return self._initial_state.platform.program
 
     ###############################
     # Hook Callback Methods
@@ -143,37 +150,10 @@ class Manticore(ManticoreBase):
             cb(state)
 
     ###############################
-    # Symbol Resolution Methods
+    # Symbol Resolution
     ###############################
 
-    def _get_addr_by_line(dwarfinfo, symbols, line):
-        """
-        Identifies a memory address through an input
-        line number through DWARF debug information.
-        """
-        for symbol in symbols:
-            # iterate over compiler units to find line program
-            for CU in dwarfinfo.iter_CUs():
-                lineprog = dwarfinfo.line_program_for_CU(CU)
-                prevstate = None
-
-                for entry in lineprog.get_entries():
-                    # check for new state assignment
-                    if entry.state is None or entry.state.end_sequence:
-                        continue
-
-                    # check between address range between prev and current state containing address
-                    if prevstate and prevstate.address <= symbol.entry['st_value'] < entry.state.address:
-                        state_line = prevstate.line
-
-                        # line number might be offset, so check to see if line number is within a range
-                        if int(state_line) - 2 <= line <= int(state_line) + 2:
-                            return symbol.entry['st_value']
-
-        # if address not found, raise exception
-        raise NotImplementedError(f"The {self._binpath} ELF file does not support line programs")
-
-    def resolve(self, symbol, line=None):
+    def resolve(self, symbol):
         """
         A helper method used to resolve a symbol name into a memory address when
         injecting hooks for analysis.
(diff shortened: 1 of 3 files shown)


manticore/core/manticore.py
```diff
@@ -278,6 +283,69 @@ def _join_workers(self):
...
 
+    #
+    # Symbol resolution
+    #
+
+    def _addr_by_symline(dwarfinfo, symbols, line):
+        '''
+        Return address for a symbol and line it is on through DWARF
+        debug info
+        '''
```
disconnect3d: Please use triple `"` characters for docstrings (yeah I know we have it mixed, we have to switch them all).

manticore/core/manticore.py
```diff
@@ -10,6 +10,8 @@
 
 import functools
 import types
+from elftools.elf.elffile import ELFFile
+from elftools.elf.sections import SymbolTableSection
```
disconnect3d: Those imports along with the functionality needs to be moved to `native/manticore.py` or probably even better - to `binary/`.

Since Manticore can be used for analysing both native binaries and ethereum smart contracts we want to keep `ManticoreBase` not dependant on any of them.

manticore/core/manticore.py
```diff
@@ -278,6 +283,69 @@ def _join_workers(self):
...
+        :type line: int or None
+        '''
+        with open(self._path, 'rb') as f:
+
+            elffile = ELFFile(f)
+            if elffile is None:
+                return NotImplementedError("Symbols aren't supported")
+ 
+            if not elffile.has_dwarf_info():
+                raise NotImplementedError("DWARF debug info not supported")
```
disconnect3d: The error message is a bit misleading, maybe sth like this?:
```suggestion
                raise NotImplementedError(f"The {self._path} ELF file is missing DWARF debug info")
```

manticore/core/manticore.py
```diff
@@ -278,6 +283,69 @@ def _join_workers(self):
...
+        :type symbol: string
+
+        :param line: if more functions present, optional line number can be included
+        :type line: int or None
+        '''
+        with open(self._path, 'rb') as f:
+
+            elffile = ELFFile(f)
+            if elffile is None:
+                return NotImplementedError("Symbols aren't supported")
```
disconnect3d: Can this ever happen?

I believe `ELFFile` is a class and since `__init__` can't really return `None` this will probably raise an exception instead of leaving us with `None`. As in:

* `ELFFile.__init__` calls https://github.com/eliben/pyelftools/blob/master/elftools/elf/elffile.py#L65-L82
* `ELFFile._identify_file` makes some assertions https://github.com/eliben/pyelftools/blob/master/elftools/elf/elffile.py#L409-L417

manticore/core/manticore.py
```diff
@@ -278,6 +283,69 @@ def _join_workers(self):
...
+            for section in elffile.iter_sections():
+                if not isinstance(section, SymbolTableSection):
+                    continue
+
+                symbols = section.get_symbol_by_name(symbol)
+                if not symbols:
+                    continue
+                
+                if line is None:
+                    return symbols[0].entry['st_value']
```
disconnect3d: Can there be a situation when there are many symbols with same name i.e. `len(symbols) > 1` is `True`?

What shall we do in such situation?

manticore/core/manticore.py
```diff
@@ -278,6 +283,69 @@ def _join_workers(self):
             while len(self._workers) > 0:
                 self._workers.pop().join()
 
+    #
+    # Symbol resolution
+    #
+
+    def _addr_by_symline(dwarfinfo, symbols, line):
```
disconnect3d: How does one fetch this information (dwarf line for a given symbol)?

I am afraid it might not be that used by the users :S.

manticore/core/manticore.py
disconnect3d: Maybe we should raise an Exception and display info for all symbols to the user? So they can use `line=X` later on (and have the info about `line` from here).

disconnect3d:
Side note: if a PR is a *work in progress* put a `[WIP]` in the title.

Sadly, GitHub doesn't support it by default, but for example GitLab prevents from merging PRs that have `WIP` in their title.

This can be added to given repo on GitHub using an app but I am not that convinced that we want to do it.

manticore/core/manticore.py
ex0dus-0x: If there are many symbols that exist with the same name, the user can select the one that they would want to resolve through its line number. With this being an optional parameter, if `line` is not specified, `m.resolve` returns the address for the first instance of the symbol it discovers.

I think an exception could be used, but instead to be thrown if no symbols are not found _at all_.

manticore/core/manticore.py
ex0dus-0x: Thanks for the info! `ELFFile._identify_file` calls `elf_assert`, which raises `ELFError`, so the error-checking on part isn't necessary.

manticore/core/manticore.py
ex0dus-0x: `pyelftools` exposes an API to DWARF that can accurately identify the function's actual line number and the file that it comes from, which I believe this is made available through the Line Number Table. [This example](https://github.com/eliben/pyelftools/blob/master/examples/dwarf_decode_address.py) best shows this.

ex0dus-0x:
Hey @disconnect3d, thanks for all the feedback!

I've slowly added the changes requested, with some more optimizations. However, I still feel that this `resolve` functionality should be present in `native.Manticore` alongside `hook` and `add_hook`, as it is a method meant to be directly called through the API. `addr_by_symline` could definitely be in `binary/binary.py` in some way, since a lot of the ELF parsing functionality is already present there.

I'm going to make a PR that moves over the binary stuff from `ManticoreBase` over to `native.Manticore` and then from there finalize refactoring this PR accordingly.

ex0dus-0x:
At this point, the branch has been updated with the changes from https://github.com/trailofbits/manticore/pull/1309, and all symbol resolution methods migrated accordingly as well.

Also added as a an addition to the test cases in `tests/test_manticore.py` for the symbol resolution feature.

manticore/native/manticore.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, path_or_state, argv=None, workspace_url=None, policy='random'
             if not os.path.isfile(path_or_state):
                 raise OSError(f'{path_or_state} is not an existing regular file')
 
+            self._binpath = path_or_state
```
disconnect3d: Hmm... for `Linux` we already have the binary path in `self.state.platform.program`, just for clarification:
* `_make_initial_path` creates a given `Platform` object - `Linux`
* for `Linux`: https://github.com/trailofbits/manticore/blob/master/manticore/platforms/linux.py#L404
* the `State` itself has a `_platform` attribute and `platform` property - https://github.com/trailofbits/manticore/blob/master/manticore/core/state.py#L82

But this is not true for `Decree` - we should probably add this property there. Or maybe the `class Platform` should assign `self._program_path` or sth.

ex0dus-0x: I can first set a `path` property in `class Platform`:

```python
class Platform(Eventful):
    def __init__(self, path, **kwargs):
        self.path = path
        super().__init__(**kwargs)
```

and then assign `self._binpath`to `initial_state.platform.path`. Since DECREE binaries can't be resolved through this feature, calling `resolve()` will throw a `NotImplementedError`.

disconnect3d: I'm not sure about having this assigned twice. Maybe we should have a private property instead e.g. in `ManticoreBase` that would resolve to `self._initial_state.platform.path`?

On the other hand - if a Linux binary calls `execve` with another path, that property will still resolve to the old/first program path.

@yan @mossberg any comments on this?

mossberg: > Maybe we should have a private property instead e.g. in ManticoreBase that would resolve to self._initial_state.platform.path?

i like this solution :) the execve issue is esoteric enough i think that we can maybe just put a comment around our sys_execve model, and leave it.

mossberg:
so the `m.resolve()` API (intended for the `@m.hook(m.resolve())` pattern) is one API choice. another choice might be to extend the existing `@m.hook()` API in some way, for example, allowing a string to be passed in, where the API will assume that the string is a name of a function. have we given this any consideration?

off the top of my head
- m.resolve allows users to resolve things unrelated to hooking. (how common would this be?)
- `@m.hook(m.resolve())` is a little bit of typing... if these two are nearly always going to be used together, i think it might be good to consolidate somehow
- `@m.hook('strcmp', line=234)`??

mossberg:
however, we should do that in separate PRs anyway. in any case, the decision is where `m.resolve()` should exist at all, and i think the `@m.hook(m.resolve('strcmp')+5)` use case is actually pretty compelling, so it sounds good to me to add it, and then later consider modifying `m.hook`

ex0dus-0x:
With @mossberg , added some other changes that would fixes up a few things. First, commit adds a `self.program` attribute for DECREE.  A property in `native.Manticore` is created that returns `self._initial_state.platform.program` (see comments about `execve` program name case) to remove use of `self._binpath`. Exception-handling was made better in `resolve` with a unit test to prove so as well.

PR #1303, opened 2018-12-13
Fix bug, check for dunder

mossberg:
allow sol funcs with 1 underscore, which is consistent with the above check in add_function

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1303)

### DIFF
--- a/manticore/ethereum/account.py
+++ b/manticore/ethereum/account.py
@@ -102,7 +102,7 @@ class EVMContract(EVMAccount):
                 contract_account.add(1000)
 
         """
-        if not name.startswith('_'):
+        if not name.startswith('__'):
             self.__init_hashes()
             if self._hashes is not None and name in self._hashes.keys():
                 def f(*args, signature: Optional[str]=None, caller=None, value=0, gas=0xffffffffffff, **kwargs):


disconnect3d:
For some reason travis hung :(
Can we also add a test for it?

manticore/ethereum/account.py
```diff
@@ -102,7 +102,7 @@ def __getattribute__(self, name):
                 contract_account.add(1000)
 
         """
-        if not name.startswith('_'):
+        if not name.startswith('__'):
```
disconnect3d: This check is wrong - the `__private` attributes are a syntax sugar for `_CLS_private` where `CLS` is the name of the class where a method that uses `__private` is defined.

```
In [9]: class A:
   ...:     def foo_invoker(self):
   ...:         print("A.foo_invoker")
   ...:         self.__foo()
   ...:
   ...:     def __foo(self):
   ...:         print("A.foo")
   ...:
   ...:     def __getattribute__(self, attr):
   ...:         print(f"Getting {attr}")
   ...:         return object.__getattribute__(self, attr)
   ...:

In [10]: A().foo_invoker()
Getting foo_invoker
A.foo_invoker
Getting _A__foo
A.foo
```

disconnect3d: On the other hand, this would still need to check for all protected attributes (the one starting with `_`) that the EVMAccount/EVMContract defines.

I've proposed a better fix for that issue in #1306.

PR #1304, opened 2018-12-13
Fix logging regressions

ehennenfent:
The recent refactor required us to move around some of the import statements in order to avoid circular imports. This caused certain modules to only be imported (and thus, create loggers) _after_ the final call to `log.init_logging`. As as result, they'd stay on the default verbosity instead of the verbosity specified via `manticore.verbosity`. This patch creates a custom Logger class that will set the correct verbosity level for each logger at the time that it's initialized, so we no longer need a separate initialization routine. 

Should be ready to go, just waiting on Travis.

### DIFF
--- a/manticore/__init__.py
+++ b/manticore/__init__.py
@@ -1,8 +1,6 @@
 from .utils import config, log
 from .utils.helpers import issymbolic, istainted
 
-log.init_logging()
-
 consts = config.get_group('main')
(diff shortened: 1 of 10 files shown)


ehennenfent:
The standalone `manticore` client sets the verbosity to 1, so the default level is still INFO for `manticore.core.manticore`.

PR #1305, opened 2018-12-14
Refactor generate_testcase events and fix eth verbose trace plugin

disconnect3d:
This PR fixes `VerboseTrace` plugin for ethereum and adds a testcase for it.

It also refactors some small parts of code and refactors/changes things a bit.

### Important changes
1. `VerboseTrace` ethereum plugin saves output to `<testcase>.verbose_trace` for each generated testcase.

2. The `Generated testcase No. X - MSG` message is now generated by `ManticoreBase` and so is standarized between ethereum and native.

3. The `will_generate_testcase_callback` interface/arguments has been changed from: `state, test_id, msg` to `state, testcase, msg`. This is because:
- as a `test_id` we passed a name or prefix instead of id (I'd assume e.g. a number or something unique). Second of all, we now pass a `Testcase` object
- it also enables plugins to write output for a given generated testcase (see how `VerboseTrace` does that).

4. The `Executor` class does not publish `will_generate_testcase` event anymore and instead forwards testcase generation to `ManticoreBase` through a `internal_generate_testcase` event. This event SHALL NEVER be used externally. When `ManticoreBase` receives it, it creates a `Testcase` object and publishes a `will_generate_testcase` event sending this testcase instance to the clients.

5. Removed `State.generate_testcase`. Note that a testcase can still be generated in two ways:
* ~by using `Executor.generate_testcase` - this is used internally by `Executor.run` to generate testcases for a terminated state and solver errors~ **EDIT: Removed**
* by using `ManticoreEVM.generate_testcase(self, state, message='', only_if=None, name='user')` which is used by scripts to check smart contracts. **As a result, there is no way to generate a testcase for native stuff. Is that okay? Shall we move this to `ManticoreBase`?**

### Refactoring changes

1. The `self.last_reg_state = {}` attribute of `Plugin` class is never used and has been removed.

2. Removed inheritance from `object` - an old thing from Python 2, not needed in Python 3.
2. Added some comments to `Solver._init` to make it more explicit what is it and what is it for.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1305)


manticore/core/manticore.py
```diff
@@ -39,7 +39,7 @@ class ManticoreBase(Eventful):
     :ivar dict context: Global context for arbitrary data storage
     '''
 
-    _published_events = {'start_run', 'finish_run'}
+    _published_events = {'start_run', 'finish_run', 'generate_testcase'}
```
disconnect3d: Btw. adding `generate_testcase` here prevents from getting a warning when we publish a `will_generate_testcase` event.

I am not sure if it is good that the prefix is ommitted - as this probably means we can also freely publish `did_generate_testcase` which is not there anywhere...

manticore/core/workspace.py
```diff
@@ -31,7 +31,26 @@ def manager():
...
 
-class Store(object):
+class Testcase:
+    def __init__(self, workspace, prefix):
+        self._num = workspace._increment_id()
+        self._prefix = prefix
+        self._ws = workspace
+
+    @property
+    def prefix(self):
```
disconnect3d: I added a `prefix` property here which is used/passed to output's `. save_testcase` for generating native testcases: https://github.com/trailofbits/manticore/pull/1305/files#diff-92b1b59d64fb129749dcf0ede70b5131R39

This `save_testcase` should probably be abstracted out.

manticore/core/workspace.py
```diff
@@ -482,12 +487,12 @@ def save_testcase(self, state, prefix, message=''):
         :param str message: The message to add to output
         :return: A state id representing the saved state
         """
-
         self._named_key_prefix = prefix
-        self._increment_id()
```
disconnect3d: NOTE: This has been removed because we already increment the id when we create a `Testcase` instance.

This `Testcase` instance is not used here *yet*. It shall be, in the end.

manticore/core/workspace.py
```diff
@@ -507,15 +512,6 @@ def save_summary(self, state, message):
             summary.write(f"Command line:\n  '{' '.join(sys.argv)}'\n")
             summary.write(f'Status:\n  {message}\n\n')
 
-            # FIXME(mark) This is a temporary hack for EVM. We need to sufficiently
-            # abstract the below code to work on many platforms, not just Linux. Then
-            # we can remove this hack.
-            if getattr(state.platform, 'procs', None) is None:
-                import pprint
-                summary.write("EVM World:\n")
-                summary.write(pprint.pformat(state.platform._global_storage))
```
disconnect3d: This is some old code that doesn't work in the end. There is no `_global_storage` or `global_storage` anywhere in the codebase (and so it crashed with a missing `AttributeError`).

manticore/core/executor.py
```diff
@@ -485,7 +472,7 @@ def run(self):
 
                         logger.debug("Generic terminate state")
                         if e.testcase:
-                            self.generate_testcase(current_state, str(e))
+                            self._publish('internal_generate_testcase', current_state, 'test', str(e))
```
yan: Leave off `'test'` to use the default and just use `message=str(e)`?

manticore/core/manticore.py
```diff
@@ -80,14 +80,18 @@ def __init__(self, initial_state, workspace_url=None, policy='random', **kwargs)
...
         self._assertions = {}
         self._coverage_file = None
         self.trace = None
 
         # FIXME move the following to a plugin
-        self.subscribe('will_generate_testcase', self._generate_testcase_callback)
         self.subscribe('did_finish_run', self._did_finish_run_callback)
+        self.subscribe('internal_generate_testcase', self._publish_generate_testcase)
+
+    def _publish_generate_testcase(self, state, prefix='test', message=''):
```
disconnect3d: Hmm... we probably don't want child class to be able to override this one. So I am going to change it into `__` instead of `_`.

This _should be obvious_ but honestly, it is easy to forget how some things work.

TLDR: creating an `__attribute` in reality creates a `_CLS_attribute` where `CLS` is the class name and using `self.__attribute` resolves to `_CLS2_attribute` where `CLS2` is the class in which the method that uses the attribute is defined.

An example below.
```
In [10]: class A:
    ...:     def __init__(self):
    ...:         self.x = self._foo
    ...:         self.y = self.__bar
    ...:     def _foo(self):
    ...:         print("A.foo")
    ...:     def __bar(self):
    ...:         print("A.bar")
    ...:
    ...:

In [11]: class B(A):
    ...:     def _foo(self):
    ...:         print("B.foo")
    ...:     def  __bar(self):
    ...:         print("B.bar")
    ...:

In [12]: B().x()
B.foo

In [13]: B().y()
A.bar
```

PR #1310, opened 2018-12-18
__main__.py: fetch manticore version from installed module

disconnect3d:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1310)


disconnect3d:
NOTE: The only disadvantage here is it will actually display *the installed Manticore version* and not the one that is in setup.py but it should not be a problem at all.

PR #1311, opened 2018-12-18
Improve solidity_create_contract UX

disconnect3d:
Previously, when user made a mistake when creating a contract from solidity code , we just returned `None` as the contract's constructor REVERTed.

This PR introduces checks for `solidity_create_contract` so exceptions are thrown in such situations:
* if contract's constructor is not payable and `balance` is passed
* if contract's constructor is payable but the owner account has insufficient balance

It also adds tests for those things.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1311)


manticore/ethereum/manticore.py
```diff
@@ -590,6 +590,15 @@ def solidity_create_contract(self, source_code, owner, name=None, contract_name=
                     else:
                         constructor_data = b''
 
+                    if balance != 0:
```
disconnect3d: @feliam Question: do we ever want to traverse such paths when creating contracts from solidity?

Shall we introduce a `force` keyword argument for such cases? If so, I would be worried we might never actually use it (hmm, but the tests should be good enough to be able to check e.g. things changed between solc versions?).

feliam: We want to traverse *all* paths.

feliam:
We should simply raise a descriptive? exception when a contract creation reverts. For any reason. And if we know why, report it. It is better not to base on what the solidity code/compiler says. WYSINWYX 

```
def create_contract(.):
```
 has the same issue.

PR #1312, opened 2018-12-19
Update Capstone to 4.0.1

disconnect3d:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1312)


disconnect3d:
Re: failing arm thumb `THB` instruction tests: it seems some address is calculated in a different way due to different value of `operand.shift.type`.

I have no idea if the value is correct, so I asked on capstone: https://github.com/aquynh/capstone/issues/1317

disconnect3d:
Re: I've sent a PR to Capstone: https://github.com/aquynh/capstone/pull/1330

We might need to wait for a minor release of Capstone before merging this PR.

PR #1313, opened 2018-12-20
Speed up ArmV7Operand.type assertions

disconnect3d:
So accessing `operand.type` will now just return the value instead of building a mapping and getting a value for given key.

We will also fail fast if there will ever be an unrecognized type for a given operand.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1313)

### DIFF
--- a/manticore/native/cpu/arm.py
+++ b/manticore/native/cpu/arm.py
@@ -53,25 +53,30 @@ def instruction(body):
     return abstract_instruction(instruction_implementation)
 
 
+_TYPE_MAP = {
+    cs.arm.ARM_OP_REG: 'register',
+    cs.arm.ARM_OP_MEM: 'memory',
+    cs.arm.ARM_OP_IMM: 'immediate',
+    cs.arm.ARM_OP_PIMM: 'coprocessor',
+    cs.arm.ARM_OP_CIMM: 'immediate'
+}
+
+
 class Armv7Operand(Operand):
     def __init__(self, cpu, op, **kwargs):
         super().__init__(cpu, op, **kwargs)
+        self.__type = _TYPE_MAP[self.op.type]
 
     @property
     def type(self):
-        type_map = {
-            cs.arm.ARM_OP_REG: 'register',
-            cs.arm.ARM_OP_MEM: 'memory',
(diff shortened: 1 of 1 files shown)


manticore/native/cpu/arm.py
```diff
@@ -53,25 +53,30 @@ def instruction_implementation(cpu, *args, **kwargs):
...
-            cs.arm.ARM_OP_REG: 'register',
-            cs.arm.ARM_OP_MEM: 'memory',
-            cs.arm.ARM_OP_IMM: 'immediate',
-            cs.arm.ARM_OP_PIMM: 'coprocessor',
-            cs.arm.ARM_OP_CIMM: 'immediate'
-        }
-
-        return type_map[self.op.type]
+        """
+        Corresponds to capstone's `operand.type` (cs.arm.RAM_OP_*).
```
disconnect3d:
```suggestion
        Corresponds to capstone's `operand.type` (cs.arm.ARM_OP_*).
```

manticore/native/cpu/arm.py
```diff
@@ -53,25 +53,30 @@ def instruction_implementation(cpu, *args, **kwargs):
...
+    cs.arm.ARM_OP_IMM: 'immediate',
+    cs.arm.ARM_OP_PIMM: 'coprocessor',
+    cs.arm.ARM_OP_CIMM: 'immediate'
+}
+
+
 class Armv7Operand(Operand):
     def __init__(self, cpu, op, **kwargs):
         super().__init__(cpu, op, **kwargs)
+        self.__type = _TYPE_MAP[self.op.type]
```
mossberg: jw, why two underscores? maybe just 1 is fine

disconnect3d: Mostly to be more bullet proof against future changes of `Operand` - e.g. if it ever get a `_type` attribute, having a `_type` here would overwrite it. Using `__type` prevents it.

But I guess it doesn't have a `_type` and we can do it for now.

mossberg: ah i see. either way is fine with me then!

PR #1314, opened 2018-12-20
Refactor events and test_events

disconnect3d:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1314)

### DIFF
--- a/manticore/utils/event.py
+++ b/manticore/utils/event.py
@@ -8,10 +8,10 @@ logger = logging.getLogger(__name__)
 
 
 class EventsGatherMetaclass(type):
-    '''
+    """
     Metaclass that is used for Eventful to gather events that classes declare to
     publish.
-    '''
+    """
     def __new__(cls, name, parents, d):
         eventful_sub = super(EventsGatherMetaclass, cls).__new__(cls, name, parents, d)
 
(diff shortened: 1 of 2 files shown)


manticore/utils/event.py
```diff
@@ -139,22 +139,19 @@ def _publish_impl(self, _name, *args, **kwargs):
                 sink._publish_impl(_name, *args, **kwargs)
 
     def subscribe(self, name, method):
-        if not inspect.ismethod(method):
-            raise TypeError
+        assert inspect.ismethod(method), f'{method.__class__.__name__} is not a method'
```
mossberg: i'm not sure if asserts are more appropriate for this situation? these are not invariants, it's more argument checking, so i think raising TypeErrors might be better (although the error messages could be improved)

disconnect3d: Do our users call `subscribe`? It seems to me that subscribing to an event is an internal thing and the events are exposed to users only through plugins/detectors.

If it is internal - then having an assertion here should be good enough ( ? ).

mossberg: you're right, generally users should opt for the Plugin/Detector interface rather than using the raw subscribe one. however, i'm not sure whether this is an external or internal interface matters too much; imo, it's more about whether the condition being checked is an invariant or for input error checking. in this case, it seems more for error checking, and `TypeError` actually seems like a more semantically correct and descriptive exception to raise, since this is actually a type check

disconnect3d: Honestly I think that in the end ("in a production build") such checks are not needed (the code itself should never call this with bad type) but it is good to have it as a defence against ourselves when making changes. This way, having it as assertion is better, at least in theory (I wonder if there is a way to ship code without assertions).

mossberg: yeah, I see your point. i think for this particular case it doesn't matter too much, i don't think `subscribe` is in any hot code paths. at the end of the day i think i do slightly prefer the TypeError since it is more specific, and from what i can see, having it shouldn't affect performance in a noticeable way. but it's only a slight preference :)

PR #1319, opened 2019-01-02
added warning and error to sys_arch_prctl

sharon-lin:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1319)

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -1412,11 +1412,11 @@ class Linux(Platform):
         ARCH_SET_FS = 0x1002
         ARCH_GET_FS = 0x1003
         ARCH_GET_GS = 0x1004
-        if code not in (ARCH_SET_GS, ARCH_SET_FS, ARCH_GET_FS, ARCH_GET_GS):
+        if code not in {ARCH_SET_GS, ARCH_SET_FS, ARCH_GET_FS, ARCH_GET_GS}:
             logger.debug("code not in expected options ARCH_GET/SET_FS/GS")
             return -errno.EINVAL 
         if code != ARCH_SET_FS:
-            raise NotImplementedError("code is not set to 0x1002")
+            raise NotImplementedError("Manticore supports only arch_prctl with code=ARCH_SET_FS (0x1002) for now")
         self.current.FS = 0x63
         self.current.set_descriptor(self.current.FS, addr, 0x4000, 'rw')
         return 0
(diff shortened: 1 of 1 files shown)


manticore/platforms/linux.py
```diff
@@ -1412,7 +1412,11 @@ def sys_arch_prctl(self, code, addr):
         ARCH_SET_FS = 0x1002
         ARCH_GET_FS = 0x1003
         ARCH_GET_GS = 0x1004
-        assert code == ARCH_SET_FS
+        if code not in (ARCH_SET_GS, ARCH_SET_FS, ARCH_GET_FS, ARCH_GET_GS):
+            logger.debug("code not in expected options ARCH_GET/SET_FS/GS")
+            return -errno.EINVAL 
+        if code != ARCH_SET_FS:
+            raise NotImplementedError("code is not set to 0x1002")
```
disconnect3d:
```suggestion
            raise NotImplementedError("Manticore supports only arch_prctl with code=ARCH_SET_FS (0x1002) for now")
```

manticore/platforms/linux.py
```diff
@@ -1412,7 +1412,11 @@ def sys_arch_prctl(self, code, addr):
         ARCH_SET_FS = 0x1002
         ARCH_GET_FS = 0x1003
         ARCH_GET_GS = 0x1004
-        assert code == ARCH_SET_FS
+        if code not in (ARCH_SET_GS, ARCH_SET_FS, ARCH_GET_FS, ARCH_GET_GS):
+            logger.debug("code not in expected options ARCH_GET/SET_FS/GS")
+            return -errno.EINVAL
```
disconnect3d:
```suggestion
            return -errno.EINVAL
```
(remove trailing whitespace)

manticore/platforms/linux.py
```diff
@@ -1412,7 +1412,11 @@ def sys_arch_prctl(self, code, addr):
         ARCH_SET_FS = 0x1002
         ARCH_GET_FS = 0x1003
         ARCH_GET_GS = 0x1004
-        assert code == ARCH_SET_FS
+        if code not in (ARCH_SET_GS, ARCH_SET_FS, ARCH_GET_FS, ARCH_GET_GS):
```
disconnect3d:
```suggestion
        if code not in {ARCH_SET_GS, ARCH_SET_FS, ARCH_GET_FS, ARCH_GET_GS}:
```

It doesn't matter here but usually it is better to use set literals instead of a tuple/list. Sets have average of O(1) while lists have O(n) for `item in collection` ([source](https://wiki.python.org/moin/TimeComplexity)).

It can also be tested in ipython interactive shell:
image

PR #1323, opened 2019-01-04
Make is_human a property

disconnect3d:
This PR changes the `Transaction.is_human` method to a property and adds a docstring that explains what we understand as a "human transaction" along with an example.

The motivation for this change is that we had two places in the codebase where we used it **as if it would be a property while it wasn't**. This code is probably bugged - probably as I hadn't confirmed it with an example but certainly making `if tx.is_human:` would always be true:

https://github.com/trailofbits/manticore/blob/580b358678c471b252c381d1733b98a6a2ad9bde/manticore/ethereum/plugins.py#L51-L54

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1323)

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -200,14 +200,14 @@ class Transaction:
     @property
     def is_human(self):
         """
-        Returns whether this is a human or internal transaction.
+        Returns whether this is a transaction made by human (in a script).
 
         As an example for:
             contract A { function a(B b) { b.b(); } }
             contract B { function b() {} }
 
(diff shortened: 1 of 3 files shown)


manticore/platforms/evm.py
```diff
@@ -197,7 +197,18 @@ def sort(self, sort):
     def result(self):
         return self._result
 
+    @property
     def is_human(self):
+        """
+        Whether it is a human or internal transaction.
```
```suggestion
        Returns whether this is a human or internal transaction.
```

Issue #1295, opened 2018-12-10
Gas related bug


bingen:
### OS / Environment
I was using docker version.
Ubuntu 16.04.5 LTS on host
Ubuntu 18.04.1 LTS on container

### Manticore version
(I was using docker version.)
Manticore 0.2.2

### Python version
I was using docker version.
Python 3.6.6

### Dependencies
asn1crypto==0.24.0
capstone==3.0.5.post1
cryptography==2.1.4
future==0.17.1
idna==2.6
keyring==10.6.0
keyrings.alt==3.0
manticore==0.2.2
ply==3.11
pycrypto==2.6.1
pyelftools==0.25
pyevmasm==0.1.0
pygobject==3.26.1
pysha3==1.0.2
python-apt==1.6.3
pyxdg==0.25
PyYAML==3.13
SecretStorage==2.3.1
six==1.11.0
unattended-upgrades==0.1
unicorn==1.0.1
z3-solver==4.8.0.0.post1

### Summary of the problem
Exception while running manticore

### Step to reproduce the behavior
```
docker run --rm -ti -v `pwd`:/src trailofbits/manticore bash
ulimit -s unlimited
manticore --detect-all --contract Staking /src/flattened_contracts/Staking.sol
```
You can find contract here:
https://gist.github.com/bingen/abd7c11829a72c5bd71c18b33225d1d9
(just dismiss the last one, the Echidna one)

### Actual behavior
This is the exception output:
```
2018-12-10 01:13:05,979: [271] m.c.s.solver:WARNING: Found an unknown core, probably a solver timeout
2018-12-10 01:16:52,201: [271] m.c.executor:ERROR: Exception: 
Traceback (most recent call last):
  File "/home/manticore/.local/lib/python3.6/site-packages/manticore/core/executor.py", line 467, in run
    if not current_state.execute():
  File "/home/manticore/.local/lib/python3.6/site-packages/manticore/core/state.py", line 141, in execute
    result = self._platform.execute()
  File "/home/manticore/.local/lib/python3.6/site-packages/manticore/platforms/evm.py", line 2160, in execute
    self.current_vm.execute()
  File "/home/manticore/.local/lib/python3.6/site-packages/manticore/platforms/evm.py", line 901, in execute
    last_pc, last_gas, instruction, arguments = self._checkpoint()
  File "/home/manticore/.local/lib/python3.6/site-packages/manticore/platforms/evm.py", line 824, in _checkpoint
    self._consume(instruction.fee)
[...]

PR #1328, opened 2019-01-08
Hotfix regressions

disconnect3d:
Fixes issues introduced in recent PR:
* `ConcretizeStack` has been renamed to `ConcretizeArgument`
* removes redundant calculation/variable in `CALLDATACOPY_gas` (which could crash the executor e.g. on example from #1295)
* Fixes/removes undeclared `_value` variable

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -1037,7 +1037,7 @@ class EVM(Eventful):
             def setstate(state, value):
                 current_vm = state.platform.current_vm
                 _pc, _old_gas, _instruction, _arguments, _fee, _allocated = current_vm._checkpoint_data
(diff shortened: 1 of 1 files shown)


manticore/platforms/evm.py
```diff
@@ -1037,7 +1037,7 @@ def setstate(state, value):
             def setstate(state, value):
                 current_vm = state.platform.current_vm
                 _pc, _old_gas, _instruction, _arguments, _fee, _allocated = current_vm._checkpoint_data
-                current_vm._checkpoint_data = (_pc, _old_gas, _instruction, _arguments, _value, _allocated)
+                current_vm._checkpoint_data = (_pc, _old_gas, _instruction, _arguments, _fee, _allocated)
```
feliam: `_value` to `value` not to `fee`.
This need to change the fee part of `current_vm._checkpoint_data` to `value`

PR #1337, opened 2019-01-10
Fix missing consts in CLI args

disconnect3d:
Brings back config variables into CLI arguments.

Also removed duplicated arguments.

This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1337)


manticore/core/manticore.py
```diff
@@ -39,16 +43,19 @@ class ManticoreBase(Eventful):
 
     _published_events = {'start_run', 'finish_run', 'generate_testcase'}
 
-    def __init__(self, initial_state, workspace_url=None, policy='random', **kwargs):
+    def __init__(self, initial_state, workspace_url=None, policy='random', timeout=None, **kwargs):
```
feliam: Do we need the timeout here if it is already a global config value ?

disconnect3d: It's the question where should the user specify it if they want to do so. Do you think they should use `config.core.timeout` instead?

disconnect3d: Let's leave it as an argument here as someone might want to use it when playing with scripts.

feliam: Ok. Note though that she may still need to do some config.py setting even from a script.

manticore/core/manticore.py
```diff
@@ -414,7 +421,7 @@ def _finish_run(self, profiling=False):
 
         self._publish('did_finish_run')
 
-    def run(self, procs=1, timeout=0, should_profile=False):
+    def run(self, procs=1, should_profile=False, timeout=None):
```
disconnect3d:
```suggestion
    def run(self, procs=1, timeout=None, should_profile=False):
```

PR #1339, opened 2019-01-12
Fix writing findings address and pc as hex values

werew:
The values of `address` and `pc` were incorrectly written as decimals.


feliam:
There are tests that check for the old wrong behavior: https://travis-ci.org/trailofbits/manticore/jobs/478819295#L556 :(

disconnect3d:
@feliam yeah, i'm going to fix em.

PR #1350, opened 2019-01-15
Symbolic memory model bugfixes

ehennenfent:
Handles edge cases that sometimes occur with symbolic memory models. Nothing major, just merging so we can delete the branch.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1350)


manticore/core/smtlib/expression.py
```diff
@@ -308,6 +308,10 @@ def __ror__(self, other):
     def __invert__(self):
         return BitVecXor(self, self.cast(self.mask))
 
+    def __len__(self):
+        """Returns the bitvector size, in bytes(!)"""
+        return self.size // 8
```
disconnect3d: @feliam is it fine to return len of a `BitVec` in bytes?

feliam: idk. It is a *bit* vector. 
pyZ3 does not use __len__ but rather a bv.size()

We should carefully check that it is not used for bits anywhere.

disconnect3d: @ehennenfent any context why do we need to add this in the first place?

ehennenfent: Pretty sure the original reason was to allow [this assert](https://github.com/trailofbits/manticore/blob/master/manticore/native/cpu/abstractcpu.py#L608) to pass when using a memory model that returns a BitVec.

feliam: We need a way to read the size of a bitvec. Maybe using __len_ is not the best way. 
We already have `.size`

Are we sure we'll never support a 4 bit bitvec?

It may be better if we remove it (__len__).

manticore/native/cpu/x86.py
```diff
@@ -583,7 +583,7 @@ def _update_cache(self, name, value):
             self._cache.pop(affected, None)
 
     def read(self, name):
-        name = self._alias(name)
+        name = str(self._alias(name))
```
ehennenfent: For context, this prevents an issue on line 591, where Python doesn't like evaluating `if 'FLAGS' in name:` if name is not a string. I don't remember how I triggered it, but since the name should already be a string in most reasonable cases, I don't think this will break any functionality.

manticore/core/smtlib/expression.py
ehennenfent: We've since found a way to make the symbolic memory model work without this workaround, so if you think it's better to delete than risk setting ourselves up for failure later, that's fine with me.

disconnect3d: I removed it.

PR #1351, opened 2019-01-15
[WIP] Opportunistic state merging for Manticore

vaibhavbsharma:
This PR is a work-in-progress as of now. The central idea is to merge states that happen to be at the same program location. This requires us to implement a `is_merge_possible` predicate that compares two states and a `merge` method that creates a merged `State` from two mergeable `State` objects. 

I am still working on finishing the implementation of `is_merge_possible` by adding checks for the array of byte values in various `Map` objects that describe the state's memory.  The next step is to implement the `merge` method by merging just the CPU and copy everything else from one of the two states. 

So far, the changes affect the main loop in `Executor`. But, I am waiting for 3 APIs to be implemented that will allow me to move my code out of `Executor` and into a separate state-merging plugin. The design of the APIs is as follows:
1. `load_state` loads a state given a `state_id` without deleting it.
2. `delete_state` deletes a state given a `state_id`
3. `replace_state` replaces the state with a value `state_id` in its workspace with another state given in its argument
All three APIs return an error if the state with `state_id` does not exist in `Executor`'s workspace. @feliam is currently working on implementing these APIs. 

With these APIs implemented, I plan to create a state-merging plugin that adds a callback for `will_load_state`. The callback would merge the state that is about to be executed with other states if any happen to be at the same program location. To figure out which states are at the same program location without loading all the states into memory, the plugin would add a handler for `did_enqueue_state` and maintain a global dictionary that maps `state_id` values to that state's program counter.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1351)


examples/script/basic_statemerging.py
```diff
@@ -0,0 +1,29 @@
...
+    m = Manticore(path, policy='random')
+
+    def will_load_state_callback(_, state_id):
+        print("about to load state_id = " + str(state_id))
+
+    def did_load_state_callback(_, state, state_id):
+        print("loaded state_id = " + str(state_id) + " at cpu = " + hex(state.cpu.PC))
+
+    m.subscribe('will_load_state', will_load_state_callback)
+    m.subscribe('did_load_state', did_load_state_callback)
```
disconnect3d: Loud thinking: do we want users to directly subscribe for events? @feliam

manticore/core/executor.py
```diff
@@ -340,6 +341,24 @@ def get(self):
...
+
+    @sync
+    def _delete_state(self, state_id):
+        if state_id in self._states:
+            del self._states[self._states.index(state_id)]
+        return self._workspace.rm_state(state_id)
+
+    @sync
+    def _replace_state(self, state_id, state):
+        # self._workspace.rm_state(state_id)
```
disconnect3d: Delete comment?

ehennenfent:
Closing this as it's now being tracked in #1482

PR #1356, opened 2019-01-17
Unicorn Preload

ehennenfent:
Modifies the Unicorn emulator module to allow it to be used for "preloading" large binaries. It uses Unicorn to execute x64 instructions in bulk while Manticore handles IO and syscalls. State changes are aggressively written from Manticore back to Unicorn, and lazily written from Unicorn to Manticore before a syscall.

The following script demonstrates this. The user can register a plugin at startup that tells Manticore to use Unicorn to quickly execute the initialization instructions leading up to the start of `main`. The binary in question is [`multiple-styles`](https://gist.github.com/ehennenfent/a5ad9746615d1490c618a88b98769c10) from an old example.
```python
from manticore.native import Manticore
from manticore.core.plugin import Plugin

address_of_main = 0x4009ae

class concretePlugin(Plugin):

    def will_start_run_callback(self, state, *_args):
        state.cpu.emulate_until(address_of_main)

m = Manticore("multiple-styles", concrete_start='coldlikeminisoda')
m.register_plugin(concretePlugin())
m.run()
```
Once `main` is reached, Manticore takes over and continues execution as normal. In the case of this example, the full solution is `coldlikeminisodas`, so Manticore generates two different test cases for the remaining byte. The performance improvement for this example is only marginal because the bulk of the time is taken up by the solver, and relatively little initialization is required before executing `main`. However, more complex binaries can see very significant speedups.

There are still some limitations, but work on those will come in future PR's. They include:
* Unicorn can't handle symbolic data, so we skip over writing it back. We should instead try to concretize it to a single value, and only continue executing in Unicorn if we can do that successfully.
* Full memory maps are often copied lazily into Unicorn, and unmaps and permissions changes are not well modeled, which can lead to Unicorn's memory maps becoming desynchronized from Manticore.
* Copying large amounts of data between Manticore and Unicorn can be dismally slow. A number of optimizations could improve this, including mapping files directly into Unicorn's memory rather than copying Manticore's memory maps.

This PR also boosts unimplemented linux system calls to warnings, which I've found to be a helpful UX change when debugging.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1356)


manticore/core/smtlib/expression.py
```diff
@@ -1026,6 +1026,9 @@ def array(self):
     def index(self):
         return self.operands[1]
 
+    def __repr__(self):
+        return "smtlib.expression.ArraySelect Object with index %s:\n%s" % (self.index, self.array)
```
disconnect3d: Maybe:
```suggestion
        return f"ArraySelect obj with index={self.index}:\n{self.array}"
```

manticore/native/cpu/abstractcpu.py
```diff
@@ -563,6 +575,18 @@ def __setattr__(self, name, value):
...
+        '''
+        Tells the CPU to set up a concrete unicorn emulator and use it to execute instructions
+        until target is reached.
+
+        :param int target: Where Unicorn should hand control back to Manticore. Set to 0 for all instructions.
+        '''
+        self._concrete = True
+        self._break_unicorn_at = target
+        if hasattr(self, 'emu'):
+            self.emu._stop_at = target
```
disconnect3d: Can we do this just in `concrete_emulate`?

manticore/native/cpu/abstractcpu.py
```diff
@@ -589,6 +613,31 @@ def write_int(self, where, expression, size=None, force=False):
 
         self._publish('did_write_memory', where, expression, size)
 
+    def _raw_read(self, where, size=1):
+        '''
```
disconnect3d:
```suggestion
        """
```

manticore/native/cpu/abstractcpu.py
```diff
@@ -589,6 +613,31 @@ def write_int(self, where, expression, size=None, force=False):
...
+        Selects bytes from memory. Attempts to do so faster than via read_bytes.
+
+        :param int where: address to read from
+        :param size: number of bytes to read
+        :return: the bytes in memory
+        :rtype: bytes
+        '''
```
disconnect3d:
```suggestion
        """
```

manticore/native/cpu/abstractcpu.py
```diff
@@ -589,6 +613,31 @@ def write_int(self, where, expression, size=None, force=False):
 
         self._publish('did_write_memory', where, expression, size)
 
+    def _raw_read(self, where, size=1):
+        '''
+        Selects bytes from memory. Attempts to do so faster than via read_bytes.
+
+        :param int where: address to read from
```
disconnect3d: TIL that it works this way too... but I'd move it to type annotation.
```suggestion
        :param where: address to read from
```

manticore/native/cpu/abstractcpu.py
```diff
@@ -589,6 +613,31 @@ def write_int(self, where, expression, size=None, force=False):
 
         self._publish('did_write_memory', where, expression, size)
 
+    def _raw_read(self, where, size=1):
```
disconnect3d:
```suggestion
    def _raw_read(self, where: int, size=1) -> bytes:
```

manticore/native/cpu/abstractcpu.py
```diff
@@ -589,6 +613,31 @@ def write_int(self, where, expression, size=None, force=False):
...
+        '''
+        Selects bytes from memory. Attempts to do so faster than via read_bytes.
+
+        :param int where: address to read from
+        :param size: number of bytes to read
+        :return: the bytes in memory
+        :rtype: bytes
```
disconnect3d:
```suggestion
```

manticore/native/cpu/abstractcpu.py
```diff
@@ -840,17 +889,25 @@ def execute(self):
...
-            if implementation is not None:
-                implementation(*insn.operands)
-
-            else:
-                text_bytes = ' '.join('%02x' % x for x in insn.bytes)
-                logger.warning("Unimplemented instruction: 0x%016x:\t%s\t%s\t%s",
-                               insn.address, text_bytes, insn.mnemonic, insn.op_str)
+            if self._concrete and 'SYSCALL' in name:
+                self.emu.sync_unicorn_to_manticore()
+            if self._concrete and 'SYSCALL' not in name:
```
disconnect3d: It might be good to actually change this part a bit and have just one `if self._concrete` and a `if 'SYSCALL' in name: ... else: ...` inside.

```suggestion
            elif self._concrete and 'SYSCALL' not in name:
```

manticore/native/cpu/abstractcpu.py
```diff
@@ -868,16 +925,48 @@ def _publish_instruction_as_executed(self, insn):
         self._publish('did_execute_instruction', self._last_pc, self.PC, insn)
 
     def emulate(self, insn):
+        '''
+        Pick the right emulate function (maintains API compatiblity)
+
+        :param insn: single instruction to emulate/start emulation from
+        '''
```
disconnect3d:
```suggestion
        """
```

manticore/native/cpu/abstractcpu.py
```diff
@@ -868,16 +925,48 @@ def _publish_instruction_as_executed(self, insn):
         self._publish('did_execute_instruction', self._last_pc, self.PC, insn)
 
     def emulate(self, insn):
+        '''
```
disconnect3d:
```suggestion
        """
```

manticore/native/cpu/abstractcpu.py
```diff
@@ -868,16 +925,48 @@ def _publish_instruction_as_executed(self, insn):
...
+            self.concrete_emulate(insn)
+        else:
+            self.backup_emulate(insn)
+
+    def concrete_emulate(self, insn):
+        '''
+        Start executing in Unicorn from this point until we hit a syscall or reach break_unicorn_at
+
+        :param capstone.CsInsn insn: The instruction object to emulate
+        '''
```
disconnect3d:
```suggestion
        """
```

manticore/native/cpu/abstractcpu.py
```diff
@@ -868,16 +925,48 @@ def _publish_instruction_as_executed(self, insn):
...
+        :param insn: single instruction to emulate/start emulation from
+        '''
+
+        if self._concrete:
+            self.concrete_emulate(insn)
+        else:
+            self.backup_emulate(insn)
+
+    def concrete_emulate(self, insn):
+        '''
```
disconnect3d:
```suggestion
        """
```

manticore/native/cpu/x86.py
```diff
@@ -714,8 +714,10 @@ def set_descriptor(self, selector, base, limit, perms):
         assert selector > 0 and selector < 0xffff
         assert base >= 0 and base < (1 << self.address_bit_size)
         assert limit >= 0 and limit < 0xffff or limit & 0xfff == 0
-        # perms ? not used yet Also is not really perms but rather a bunch of attributes
+        #perms ? not used yet Also is not really perms but rather a bunch of attributes
```
disconnect3d:
```suggestion
        # perms ? not used yet Also is not really perms but rather a bunch of attributes
```

manticore/platforms/linux.py
```diff
@@ -1719,6 +1720,8 @@ def sys_mmap(self, address, size, prot, flags, fd, offset):
 
         perms = perms_from_protflags(prot)
 
+        cpu._publish('will_map_memory', address, size, perms, "" if fd == 0 or flags & 0x20 else self.files[fd].name, offset)
```
disconnect3d: What does  `flags & 0x20` check for? (might be good to add a comment)

Any chance to change `"" if fd == 0 or flags & 0x20 else self.files[fd].name` to just `self.files[fd].name` and make it so we set a name on stdin/out/err manually?

Hmm, stdout/stderr/in shall have names actually:
```
In [4]: sys.stdout
Out[4]: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>

In [5]: sys.stdout.name
Out[5]: '<stdout>'

File: manticore/platforms/linux.py [hunk 15]
```diff
@@ -1730,6 +1733,8 @@ def sys_mmap(self, address, size, prot, flags, fd, offset):
             # FIXME Check if file should be symbolic input and do as with fd0
             result = cpu.memory.mmapFile(address, size, perms, self.files[fd].name, offset)
 
+        cpu._publish('did_map_memory', address, size, perms, "" if fd == 0 or flags & 0x20 else self.files[fd].name, offset, result)
```
disconnect3d: If we don't change `"" if fd == 0 or flags & 0x20 else self.files[fd].name` we should at least calculate it once :P

manticore/platforms/linux.py
```diff
@@ -1753,7 +1758,9 @@ def sys_mprotect(self, start, size, prot):
         :return: C{0} on success.
         '''
         perms = perms_from_protflags(prot)
+        self.current._publish('will_protect_memory', start, size, perms)
```
disconnect3d: This is more verbose, but maybe we should change it to `will_mprotect` and `did_mprotect`?

Or even `will_sys_mprotect`.

I guess that will be more explicit that it is related to mprotect syscall.

manticore/platforms/linux.py
```diff
@@ -1910,7 +1921,7 @@ def sys_acct(self, path):
         return -1
 
     def sys_exit(self, error_code):
-        'Wrapper for sys_exit_group'
+        '''Wrapper for sys_exit_group'''
```
disconnect3d:
```suggestion
        """Wrapper for sys_exit_group"""
```
It might be good to add a sentence here why is exit a wrapper for exit_group - is it that we lack some functionality here? Does it work the same in kernel and it is a technical debt from the past?

manticore/platforms/linux.py
```diff
@@ -1921,24 +1932,38 @@ def sys_exit_group(self, error_code):
...
+        logger.warning("Unimplemented system call: sys_sysinfo")
+        return -1
+
+    def sys_madvise(self, infop):
+        logger.warning("Unimplemented system call: sys_madvise")
         return -1
 
+    def sys_futex(self, uaddr, op, val, timeout, uaddr2, val3):
+        logger.warning("sys_futex isn't really implemented")
+        logger.debug(f"Futex at: {hex(uaddr)} -- {op}:{val} ({self.current.read_int(uaddr)})")
```
disconnect3d: This should not use f-string literals - as the output log string is produced by us (and so having a logging level that doesn't include debug will still waste time formatting the string).

manticore/utils/emulate.py
```diff
@@ -1,92 +1,189 @@
...
+        permissions |= UC_PROT_READ
+    if 'w' in m_perms:
+        permissions |= UC_PROT_WRITE
+    if 'x' in m_perms:
+        permissions |= UC_PROT_EXEC
+    return permissions
+
+
+# https://stackoverflow.com/a/1094933
+def sizeof_fmt(num, suffix='B'):
```
disconnect3d: The link is self-explanatory but it might be good to add a docstring and change name - e.g. `humanize_size` (not sure).

manticore/utils/emulate.py
```diff
@@ -1,92 +1,189 @@
...
+        # Keep track of all memory mappings. We start with just the text section
+        self.mem_map = {}
+        for m in cpu.memory.maps:
+            permissions = UC_PROT_NONE
+            if 'r' in m.perms:
+                permissions |= UC_PROT_READ
+            if 'w' in m.perms:
+                permissions |= UC_PROT_WRITE
+            if 'x' in m.perms:
+                permissions |= UC_PROT_EXEC
```
disconnect3d: This could use `convert_permissions`

manticore/utils/fallback_emulator.py
```diff
@@ -0,0 +1,287 @@
...
+        :param uc: The Unicorn instance.
+        :param address: The address which is contained by the mapping.
+        :rtype Map
+        '''
+
+        m = self._cpu.memory.map_containing(address)
+
+        permissions = UC_PROT_NONE
+        if 'r' in m.perms:
+            permissions |= UC_PROT_READ
```
disconnect3d: could use `convert_permissions`

manticore/utils/emulate.py
```diff
@@ -1,92 +1,189 @@
...
+# https://stackoverflow.com/a/1094933
+def sizeof_fmt(num, suffix='B'):
+    for unit in ' KMGTPEZ':
+        if abs(num) < 1024.0:
+            return "%3.1f%s%s" % (num, unit if unit != ' ' else '', suffix)
+        num /= 1024.0
+    return "%.1f%s%s" % (num, 'Y', suffix)
+
+
+class ConcreteUnicornEmulator(object):
```
disconnect3d: Does it make sense to inherit from base emulator?

Also, py3 doesn't require to inherit from object.
```suggestion
class ConcreteUnicornEmulator:
```

manticore/utils/emulate.py
```diff
@@ -1,92 +1,189 @@
...
-        }
-
-        # Keep track of all the memory Unicorn needs while executing this
-        # instruction
-        self._should_be_written = {}
+        cpu.subscribe('did_write_memory', self.write_back_memory)
+        cpu.subscribe('did_write_register', self.write_back_register)
+        cpu.subscribe('did_set_descriptor', self.update_segment)
+        cpu.subscribe('did_map_memory', self.map_memory_callback)
+        # cpu.subscribe('did_protect_memory', self.protect_memory)
```
disconnect3d: Remove comment?

tests/native/test_unicorn_concrete.py
```diff
@@ -0,0 +1,57 @@
...
+        self.m = Manticore(os.path.join(dirname, 'binaries', 'arguments_linux_amd64'),
+                           argv=['argv', 'mc', 'argface'])
+        self.concrete_instance = Manticore(os.path.join(dirname, 'binaries', 'arguments_linux_amd64'),
+                                           argv=['argv', 'mc', 'argface'])
+
+        self.concrete_instance.register_plugin(ConcretePlugin())
+        self.concrete_instance.register_plugin(RegisterCapturePlugin())
+        self.m.register_plugin(RegisterCapturePlugin())
+
+    def test_register_comparison(self):
```
disconnect3d: Aren't those tests taking a bit long time to execute? If so, it might be good to make both assertions in one test.

manticore/platforms/linux.py
```diff
@@ -114,7 +115,11 @@ def stat(self):
 
     def ioctl(self, request, argp):
         # argp ignored...
```
disconnect3d: Since we don't ignore `argp` anymore, the comment is redundant
```suggestion
```

manticore/platforms/linux.py
```diff
@@ -387,6 +392,7 @@ class Linux(Platform):
 
     # from /usr/include/asm-generic/resource.h
     RLIMIT_NOFILE = 7  # /* max number of open files */
+    RLIMIT_STACK = 3
```
disconnect3d: Loud thinking: we could probably use https://docs.python.org/3/library/resource.html#resource-limits ?

manticore/platforms/linux.py
```diff
@@ -1545,23 +1554,23 @@ def sys_ARM_NR_set_tls(self, val):
...
+        logger.warning(f"SIGACTION, Ignoring changing signal handler for signal {signum}")
         return 0
 
     def sys_rt_sigprocmask(self, cpu, how, newset, oldset):
         '''Wrapper for sys_sigprocmask'''
         return self.sys_sigprocmask(cpu, how, newset, oldset)
 
     def sys_sigprocmask(self, cpu, how, newset, oldset):
-        logger.debug(f"SIGACTION, Ignoring changing signal mask set cmd:{how}", )
+        logger.warning(f"SIGACTION, Ignoring changing signal mask set cmd:{how}", )
```
disconnect3d: Since this is `logging`, we should rather use it's internal formatting (i.e. we don't want to format if we discard the log later on). This being said if we change it to `warning` it's rather okay to format it right away?

manticore/utils/fallback_emulator.py
```diff
@@ -0,0 +1,287 @@
...
+from unicorn import *
+from unicorn.x86_const import *
+from unicorn.arm_const import *
+
+from capstone import *
+
+logger = logging.getLogger(__name__)
+
+
+class UnicornEmulator(object):
```
disconnect3d:
```suggestion
class UnicornEmulator:
```

manticore/native/cpu/abstractcpu.py
ehennenfent: I've actually made both of the optimizations you describe here independently without realizing until I tested them that they subtly break the logic. You could refactor it to have a nested `if` at the top, but then you need to change the final `else` to `if not self.concrete or 'SYSCALL' in name` so I don't think it makes any substantial improvements for readability.

manticore/native/memory.py
```diff
@@ -540,6 +547,9 @@ def page_mask(self):
     def maps(self):
         return self._maps
 
+    def setCPU(self, cpu):
+        self.cpu = cpu
```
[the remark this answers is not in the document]
ehennenfent: Good point. A getter/setter would make more sense.

manticore/native/cpu/abstractcpu.py
```diff
@@ -620,8 +681,24 @@ def write_bytes(self, where, data, force=False):
...
+            not issymbolic(data) and \
+            self._concrete  # fast write breaks symbolic strcmp/len model, so only use in concrete mode
+
+        if can_write_raw:
+            logger.debug("Using fast write")
+            offset = mp._get_offset(where)
+            if type(data) is str:
+                data = bytes(data.encode('utf-8'))
+            mp._data[offset:offset + len(data)] = data
+            self._publish('did_write_memory', where, data, 8 * len(data))
```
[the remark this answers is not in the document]
ehennenfent: I'd have to go find it elsewhere in the code but I'm pretty sure `write_memory` uses bit length by convention

manticore/native/cpu/abstractcpu.py
```diff
@@ -620,8 +681,24 @@ def write_bytes(self, where, data, force=False):
         :type data: str or list
         :param force: whether to ignore memory permissions
         '''
-        for i in range(len(data)):
-            self.write_int(where + i, Operators.ORD(data[i]), 8, force)
+
+        mp = self.memory.map_containing(where)
+        can_write_raw = type(mp) is AnonMap and \
+            type(data) in {str, bytes} and \
+            (mp.end - mp.start + 1) >= len(data) >= 1024 and \
```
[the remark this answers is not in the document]
ehennenfent: Completely arbitrary. The issues mentioned in the comment on line 690 made me worried that this introduces some subtle side effect that I haven't noticed yet, so I limited it to writes where there's enough data to be written that using the fast method actually provides a noticeable speed improvement. Mind you, I don't have numbers to back that up, I just kinda decided "1kb sounds about right". 
You raise a good point though. I'll add a comment detailing the motivation behind the limit.

manticore/native/cpu/abstractcpu.py
```diff
@@ -620,8 +681,24 @@ def write_bytes(self, where, data, force=False):
         :type data: str or list
         :param force: whether to ignore memory permissions
         '''
-        for i in range(len(data)):
-            self.write_int(where + i, Operators.ORD(data[i]), 8, force)
+
+        mp = self.memory.map_containing(where)
+        can_write_raw = type(mp) is AnonMap and \
+            type(data) in {str, bytes} and \
```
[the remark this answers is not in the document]
ehennenfent: Oh hey, that's nifty. Trying it out in a REPL, it looks like it _should_ be a tuple, so I'll fix it locally and push.

manticore/native/cpu/abstractcpu.py
ehennenfent: This is the function that the user interfaces with in order to invoke Unicorn mode, so I think it makes sense to have it as a separate function. I'll post an example script in a comment on the PR.

manticore/native/cpu/abstractcpu.py
```diff
@@ -589,6 +617,39 @@ def write_int(self, where, expression, size=None, force=False):
...
+        """
+        Selects bytes from memory. Attempts to do so faster than via read_bytes.
+
+        :param where: address to read from
+        :param size: number of bytes to read
+        :return: the bytes in memory
+        """
+        map = self.memory.map_containing(where)
+        start = map._get_offset(where)
+        if type(map) is FileMap:
```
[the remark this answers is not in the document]
ehennenfent: It does make sense to use `type` in this case because we access protected members of the FileMap that aren't guaranteed to behave the same way in a subclass. Strictly speaking, the right way to do this is probably to add unprotected bulk read and write functions to filemaps and anonmaps, but for now I'll just save the type in a variable so we won't need to calculate it twice.

ehennenfent:
Sample invocation using the [multiple-styles](https://gist.github.com/ehennenfent/a5ad9746615d1490c618a88b98769c10) example
```
from manticore.native import Manticore
from manticore.core.plugin import Plugin

class concretePlugin(Plugin):

    def will_start_run_callback(self, state, *_args):
        state.cpu.emulate_until(0x4009ae)

m = Manticore("multiple-styles", concrete_start='coldlikeminisoda')
m.register_plugin(concretePlugin())
m.verbosity(2)
m.run()
```

PR #1359, opened 2019-01-21
Symbolic VMTests

disconnect3d:
Changes the EVM tests auto-generator to be able to generate symbolic tests.

VMTests were generated and added to the commit - so we have both concrete and symbolic tests for frontier.

The symbolic tests works by creating all possible symbolic values as `BitVecVariable`s and by preconstraining their value to the initial value. At the end for all the `post` data (which is specified in VMTests json files) solver is asked for solutions and it is checked whether the value is correct (as expected - as specified in the json file) and if there is only one solution.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1359)


feliam:
Moved the effort to https://github.com/trailofbits/manticore/pull/1431.

PR #1363, opened 2019-01-22
WIP: Support lsr.w on ARMv7 THUMB

nkaretnikov:
`lsr.w R0, R0, R2` is processed in `_SR` in the `if rest and
rest[0].type == 'register'` branch despite being THUMB.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1363)


nkaretnikov:
WIP because I really dislike how the code looks now. But I can't figure out how to make `nosetests tests/native/test_armv*` pass otherwise.

nkaretnikov:
(I guess it might be easier to understand the issue if you just run the new test on master and look at the backtrace.)

nkaretnikov:
BTW, I'm okay with this sitting here for a while if it's not a blocker for others. My motivation for working on this was to get a small testcase working (but I'm not there yet) for ARMv7 to see how the current code works and whether there are any issues.

manticore/native/cpu/arm.py
```diff
@@ -455,12 +464,7 @@ def _shift(cpu, value, _type, amount, carry):
             amount = 1
```
disconnect3d:
```
# XXX: Capstone should set the value of an RRX shift to 1, which is
# asserted in the manual, but it sets it to 0, so we have to check
if _type in (cs.arm.ARM_SFT_RRX, cs.arm.ARM_SFT_RRX_REG) and amount != 1:
```
While we are here, could you check if this is still the  case in latest Capstone (4.0.1), please?

manticore/native/cpu/arm.py
```diff
@@ -455,12 +464,7 @@ def _shift(cpu, value, _type, amount, carry):
...
 
         elif _type in range(cs.arm.ARM_SFT_ASR_REG, cs.arm.ARM_SFT_RRX_REG + 1):
-            if cpu.mode == cs.CS_MODE_THUMB:
-                src = amount.read()
-            else:
-                src_reg = cpu.instruction.reg_name(amount).upper()
-                src = cpu.regfile.read(src_reg)
-            amount = Operators.EXTRACT(src, 0, 8)
+            amount = Operators.EXTRACT(amount, 0, 8)
```
disconnect3d: So the `amount` would be a register that stores the shift value now?

disconnect3d: Note that below we use  e.g. `ASR_C` from `bitwise.py` and this expcets amount to be a number; is it always fine?

nkaretnikov: I won't have much time to look into this in the near future due to other priorities, but to address your point about `bitwise`: there are tests for ASR and ROR in the testsuite, which pass with this change. That's why the change at `manticore/native/cpu/arm.py:105` was necessary. But it's hard to say for sure whether this would work for all inputs.

PR #1366, opened 2019-01-23
Aarch64

nkaretnikov:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1366)


nkaretnikov:
My Aarch64 work will be happening here. Currently this just includes code by @disconnect3d (with minor changes and updated for current master) and is completely untested. Things might be changing a lot while this is still in development, but feel free to review and comment if you feel like it.

nkaretnikov:
CC @yan (just in case)

nkaretnikov:
2464 unique instruction encodings get executed when running a "Hello, world!" program on my system

When running a "Hello, world!" program on my system, 2464 unique instruction encodings get executed.
Filtering by unique mnemonics, I get 85:
```
add
addp
adr
adrp
and
ands
asr
b
b.eq
b.ge
b.gt
b.hi
b.hs
b.le
b.lo
b.ls
b.lt
b.ne
bfxil
bic
bics
bl
blr
br
cbnz
cbz
ccmp
cinc
clz
cmeq
cmn
cmp
csel
cset
csetm
csinc
dc
dmb
dup
eor
ld1
ldaxr
ldp
ldr
ldrb
ldrh
ldrsw
ldur
ldxr
lsl
lsr
madd
mov
movk
movn
movz
mrs
msr
msub
mul
neg
nop
orr
rbit
ret
rev
sbfiz
stp
str
strb
strh
stur
sub
subs
svc
sxtw
tbnz
tbz
tst
ubfiz
ubfx
udiv
umulh
uxtb
uxth
```

nkaretnikov:
Pushed a "Hello, world!" demo, based on the existing `count_instructions` example. It counts and prints executed instructions and outputs some information about `puts` and `exit`:

```
# PYTHONPATH="." ./examples/script/aarch64/count_instructions.py | egrep "puts|exit"
/manticore/manticore/native/cpu/aarch64.py:285: UserWarning: Aarch64 support is experimental
  warnings.warn('Aarch64 support is experimental')
puts argument: hello
puts result: 6
exit status: 42
```

nkaretnikov:
Pushed a script for the "basic" demo.

CLI output:
```
# manticore ./examples/script/aarch64/basic
2019-04-18 10:44:26,296: [62731] m.c.manticore:INFO: Verbosity set to 1.
2019-04-18 10:44:26,366: [62731] m.n.manticore:INFO: Loading program ./examples/script/aarch64/basic
2019-04-18 10:44:33,949: [62731] m.c.manticore:INFO: Generated testcase No. 0 - Program finished with exit status: 0
2019-04-18 10:44:35,876: [62731] m.c.manticore:INFO: Generated testcase No. 1 - Program finished with exit status: 0
2019-04-18 10:44:35,898: [62731] m.c.manticore:INFO: Results in /manticore/mcore_qo2s6qa7
2019-04-18 10:44:35,898: [62731] m.c.manticore:INFO: Total time: 8.974636793136597

# hexdump -C mcore_qo2s6qa7/test_00000000.stdin
00000000  00 00 00 00                                       |....|
00000004

# hexdump -C mcore_qo2s6qa7/test_00000001.stdin
00000000  00 80 01 80                                       |....|
00000004

# printf "\x00\x00\x00\x00" | ./examples/script/aarch64/basic
Message: It is less than or equal to 0x41

# printf "\x00\x80\x01\x80" | ./examples/script/aarch64/basic
Message: It is greater than 0x41
```

nkaretnikov:
# AArch64 status

This can be moved to the wiki once the code is reviewed.

## Issues

All issues in the code are marked with `XXX`.  To summarize (in no particular
order):

* Support system registers other than: `CPACR_EL1`, `DCZID_EL0`, `TPIDR_EL0`.

* Check the current exception level when using privileged instructions.

* Implement the `DC ZVA` instruction.  I avoided it by setting `DCZID_EL0` to
  `0b10000`, which marks it as unsupported.  This instruction is used in glibc
  to efficiently zero-out memory.  If it's unsupported, a different code path is
  used.  This works fine, but may result in differences between traces when
  comparing to other emulators/debuggers.  The said instruction is available in
  QEMU and on my Raspberry Pi 3 B+.

* Add more fields to the register table.  For example, mark some registers as
  read-only.  Currently, I check for that in `read` and `write` methods, but
  it'd be nice to localize that.  Also, check reserved bits.

* When using Unicorn, allow to work with system registers via the Unicorn API.
  With Unicorn 1.0.1, the API can only be used for `NZCV`.  Anything else can
  be handled only via the `MSR (register)` and `MRS` instructions.  In HEAD,
  more registers are available, so revise this when upgrading.  See
  `arm64_reg_write` and `arm64_reg_read` in Unicorn.

* Add more instructions.  I've added only those needed to run the "Hello,
  world!" and "basic" binaries.  But some instruction variants might be even
  unused by the said binaries because my initial goal was to add all variants
  sharing the same mnemonic.  Then it became too time-consuming when I started
  working on SIMD.  So I shifted to adding only those that are required.  There
  are comments indicating which variants are implemented and which ones are
  missing next to every instruction.

* Remove workarounds for Capstone and Unicorn issues once those are fixed and
  dependencies are upgraded.  Some have already been fixed upstream.  You can
  find them by searching for "github" in the code.

* Some instructions may be trapped when executed.  This is not implemented,
  but there should be a comment in the instruction body of every such
  instruction.

* Maybe share some code between Armv7 and Aarch64 (or even x86).  This includes
  both the testsuite (for example, `itest*` decorators) and some helper
  functions (such as `_add_with_carry`).  Maybe test under Unicorn and with
  symbolic values (more on that below).

* Maybe look into using `tests/auto_generators`.

* Maybe refactor the code working with `NZCV` bits.  The code was initially
  written to support just concrete values, symbolic support was added later.
  So there might be opportunities for improvement (see `_ccmp_imm_reg`).

* Later it might be necessary to implement the `DMB` instruction.  Currently,
  it's essentially the same as `NOP` (with a few different checks).

* Fully implement exclusive load (`LDAXR`, `LDXR`) and store (`STXR`, `STLXR`)
  instructions, including alignment checks.  Currently, they just work like
  regular loads and stores.  Unicorn implements them but doesn't conform to the
  spec.  Armv7 doesn't implement `LDREX` and `STREX` either.  Exclusive stores
  are untested (but the rest of the instruction set should be covered).

* The status of floating-point support is unclear as it's not utilized by the
  mentioned binaries.

* `sys_readlink` and `sys_readlinkat` need to be extended to check for
  other error conditions and return accordingly.  Additionally, `sys_readlinkat`
  needs to support `dir_fd` as it currently behaves just like `sys_readlink`.

* Support machine names other than `aarch64` and different endiannesses.

* Added the `timeout` argument to `emu_start`.  Otherwise, Unicorn would hang on
  some instructions (opened an issue).  This isn't acceptable because all
  instructions implemented in Python are tested against Unicorn as well.
  However, it's not perfect either.  From time to time, one of the tests fails
  because some register is just not set before the timeout.  That is usually
  one of the `LDR` tests and the part that fails is writeback.  Maybe it'd be
  better to rollback to the previous behavior, but then the Unicorn part of the
  testsuite would need to be disabled.  I think hanging is better than rarely
  producing incorrect values as the former is easier to spot.  And the `timeout`
  option along with Unicorn tests could be enabled just for development, using
  an environment variable or something like that.

* In `test_aarch64cpu.py`, there's a lot of repetition.  There should be a way
  to refactor this without sacrificing readability and without adding even more
  code.  There are some attempts in that direction (search for "Bundle").  Maybe
  it's worth trying to migrate to nose2 as the current version has some
  limitations:
  https://stackoverflow.com/questions/12376598/conflicts-between-unittest-and-nose-frameworks

* Vector and system registers are not tested in `test_aarch64rf.py`.
  Reserved bits and read-only registers are not checked either.


## Code

For the sake of correctness, I've been trying to match the ARM manual as close
as possible.  For example, if an instruction is an alias, the execution goes
through the alias.  I also check the instruction encoding (found some issues
this way) and instruction arguments, when practical.  There might be some
opportunities for optimization when the code is well-tested.  And all that is
not foolproof as the code relies on Capstone.

In case you're not familiar with the ARM manual: when reviewing the
instructions, you should look at "C6.2 Alphabetical list of A64 base
instructions" and "C7.2 Alphabetical list of A64 Advanced SIMD and
floating-point instructions".  Everything else should be referenced from the
code.


## Tests

Tests follow the same structure as those of Armv7.  The only difference is
`test_aarch64cpu.py`.  My main goal with instruction tests was to organize them
such that the same tests could be run against the native Python code (with
concrete and symbolic arguments) and Unicorn.  There is a base class that
defines all tests, and three classes that use it.

There are some minor differences in behavior between Manticore and Unicorn, but
those tests should be pretty easy to find just by looking for ones with the
`if-then-else` statement.

Some immediate tests are skipped for the symbolic class because there is no
convenient way to provide symbolic inputs in such cases.  Those use the
`skip_sym` decorator.  Also, some symbolic tests have a single symbolic argument
that is PC, which gets concretized during execution (`ConcretizeRegister` is
raised by `execute` in `abstractcpu.py`), so the instructions don't really see
symbolic arguments in those cases, but that's expected.  The symbolic class is
modeled after symbolic tests in `test_x86.py` (another opportunity for
refactoring).

Read here on how to run the testsuite (do not forget to adjust the paths):
https://github.com/trailofbits/manticore/wiki/Hacking-on-Manticore

Some useful options to pass to `nosetests` are `-s`, `-v`, and `-m`.
`-v` is useful to see the `SKIP` messages, but may clutter the output.


## Examples

There are two example scripts in `examples/script/aarch64`: one for "Hello,
world!" and one for "basic".  The latter can be tested with symbolic and concrete
inputs (the comment at the start of each script should explain how to run it).

The binaries and sources are in the same directory.  I bundle everything just
for convenience and to make it reproducible.  Can be moved somewhere else.

The binaries work with the `manticore` CLI as well.

I compared the output produced by Manticore (files in `mcore*` except
instruction traces) after running "basic" in CLI mode (symbolic input) between
x86_64 and AArch64.  The results seem to match.

I also compared the instruction trace produced by my version of `count_instructions.py`
with the `stepi` (single-stepping) trace produced by gdb.  Both on AArch64.

The trace produced by gdb is twice the size of Manticore's and differs in some
places.  The differences are due to the following:
* gdb and Manticore take different conditional branches
* gdb loops many times while Manticore goes through (essentially the same as
  above)
* gdb doesn't print instructions between exclusive load and store instructions
  (due to exclusive access).

I haven't looked into why different paths are taken.  Could be caused by the
environment (how gdb sets things up) or differences in available instructions
(see `DC ZVA` above).  But at least there's nothing outright wrong (like jumps
leading to nowhere).  There are significant sections that match too.  The
observed/captured output (the printed string and exit code) is the same.

nkaretnikov:
@ehennenfent I'm removing the WIP prefix to signal that this is ready for review. The only things I intend to fix (as part of this PR) are codeclimate issues and anything that'll pop-up during review.

ehennenfent:
Going to close and reopen to see if that fixes Travis

nkaretnikov:
Fixed codeclimate issues in all files affected by this branch, not just my changes, including ignored paths.
If it's too much, I can fix only those introduced by me, but the above was easier to track locally.

There are some things left, but:
- they are usually ignored ("Consider refactoring"), if I understood @ehennenfent correctly
- fixes would lead to even uglier (IMO) code ("Visually indented line with same indent as next logical line")
- or those issues are just hard to fix ("Consider simplifying this complex logical expression").

Maybe it's worth ignoring some issues in the config: https://pep8.readthedocs.io/en/latest/intro.html#configuration

(Note that the number of issues is different below from what GitHub reports, but that's because I provide paths myself.)

```
% codeclimate analyze $(git diff --name-only master)
Starting analysis
Running structure: Done!
Running duplication: Done!
Running pep8: Done!

== manticore/native/cpu/aarch64.py (8 issues) ==
271-5318: `Aarch64Cpu` has 170 functions (exceeds 35 allowed). Consider refactoring. [structure]
302-331: Consider simplifying this complex logical expression. [structure]
305: Visually indented line with same indent as next logical line [pep8]
838: Visually indented line with same indent as next logical line [pep8]
1220-1234: Consider simplifying this complex logical expression. [structure]
3032-3060: Consider simplifying this complex logical expression. [structure]
3035: Visually indented line with same indent as next logical line [pep8]
4680-4694: Consider simplifying this complex logical expression. [structure]

== manticore/platforms/linux.py (1 issue) ==
873-1071: Function `load` has a Cognitive Complexity of 74 (exceeds 40 allowed). Consider refactoring. [structure]

== tests/native/test_aarch64cpu.py (2 issues) ==
1-14120: File `test_aarch64cpu.py` has 11416 lines of code (exceeds 9000 allowed). Consider refactoring. [structure]
172-14027: `Aarch64Instructions` has 1,412 functions (exceeds 35 allowed). Consider refactoring. [structure]

== tests/native/test_armv7unicorn.py (1 issue) ==
96-1328: `Armv7UnicornInstructions` has 161 functions (exceeds 35 allowed). Consider refactoring. [structure]

Analysis complete! Found 12 issues.
```

nkaretnikov:
Just to demonstrate accepted fixes (that I know of), I'll let you judge:

```
# 3: Visually indented line with same indent as next logical line [pep8]
if (True or
    False):
    pass

# Accepted.
if (
    True or
    False
):
    pass

# Accepted.
if (True or
   False):
    pass

# Accepted.
if (True or
        False):
    pass
```

nkaretnikov:
Squashed and rebased on master:
https://github.com/nkaretnikov/manticore/commits/aarch64-squash-rebase

Verified that there are no major differences by `diff`ing my changes on this branch to my changes on the new one.

Will be splitting the new branch into smaller PRs next.

nkaretnikov:
@ehennenfent Split [nkaretnikov/aarch64-squash-rebase](https://github.com/nkaretnikov/manticore/commits/aarch64-squash-rebase) (this PR squashed and rebased on master) into multiple PRs:
- https://github.com/trailofbits/manticore/pull/1418
- https://github.com/trailofbits/manticore/pull/1419
- https://github.com/trailofbits/manticore/pull/1420
- https://github.com/trailofbits/manticore/pull/1421
- https://github.com/trailofbits/manticore/pull/1422
- https://github.com/trailofbits/manticore/pull/1423

The idea here is to merge them in this particular order as each PR builds on the previous one. If git can't deal with commits on a feature branch that are already present on master, one will have to cherry-pick or rebase first.

After applying all of them, there should be no changes against [nkaretnikov/aarch64-squash-rebase](https://github.com/nkaretnikov/manticore/commits/aarch64-squash-rebase).

The advantage of this split method is that it doesn't introduce or remove any code in-between.
The disadvantage is that you'll need to merge some commits to master without tests. But you can always test locally first, and the tests will end up in master anyway when the last commit is merged. So this shouldn't be a problem.

nkaretnikov:
I'm closing this one to avoid confusion.

PR #1368, opened 2019-02-07
Remove redundant hash methods

disconnect3d:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1368)


disconnect3d:
Closing as it breaks some code. We probably need to refactor some other piece so `ArrayProxy`s are not hashed, but it is okay as it is now.

PR #1371, opened 2019-02-08
Fix ethereum Manticore API issue, where creating a new account with the code field failed

### LINKED ISSUE #1370: [Ethereum] Manticore API and EVM implementation disagree on type of code field
### OS / Environment
macOS 10.14.2

### Manticore version
Version: 0.2.4

### Python version
Python 3.7.1

### Summary of the problem
When creating a new account using the Manticore API as shown below, the API and the actual EVM implementation disagree on the type of the code field.

I have identified three problems/behaviours so far:

1. When creating an account using supplying simply a str argument, the manticore api tries to convert it into a bytearray (manticore/ethereum/manticore.py line 770):
```python
        if isinstance(code, str):
            code = bytearray(code)
        if code is not None and not isinstance(code, (bytearray, Array)):
            raise EthereumError("code bad type")
```
However this call fails since we do not supply an encoding:

`TypeError: string argument without an encoding`

2. When supplying a bytearray as asked, the api tries to create the account in the underlying evm implementation, which however expects a bytes type (manticore/platforms/evm.py line 2388):
```python
        if code is None:
            code = bytes()
        else:
            if not isinstance(code, (bytes, Array)):
                raise EthereumError('Wrong code type')
```

3. When supplying a bytes type, we fail the check for a bytearray in the manticore API.

Any of the uncommented lines produces an error:
```python
m = ManticoreEVM()

# test_account = m.create_account(code="0x00AAFF")
# test_account = m.create_account(code=bytearray("0x00AAFF", "utf-8"))
# test_account = m.create_account(code=bytes("0x00AAFF", "utf-8"))

```

### Expected behavior
Account created in the environment

### Actual behavior
See above

Joool:
Fixes #1370

I assumed bytes to be the correct type since the evm implementation uses it. Also the code field is immutable so it seemed like a good fit.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1371)

### DIFF
--- a/manticore/ethereum/manticore.py
+++ b/manticore/ethereum/manticore.py
@@ -744,7 +744,7 @@ class ManticoreEVM(ManticoreBase):
             :type balance: int or BitVecVariable
             :param address: the address for the new account (optional)
             :type address: int
-            :param code: the runtime code for the new account (None means normal account) (optional)
+            :param code: the runtime code for the new account (None means normal account), str or bytes (optional)
             :param name: a global account name eg. for use as reference in the reports (optional)
             :return: an EVMAccount
         """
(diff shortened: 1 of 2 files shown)


disconnect3d:
Hey, thanks for this PR!

@Joool could you add testcases that you described in #1370 to our tests, please?

Joool:
done :)

PR #1378, opened 2019-02-26
examples/linux: Refactor, fix crackme.py

woodruffw:
This is just the result of running `black` on `crackme.py`, plus a change to ~~how the program is emitted to circumvent some locale dumbness in the Docker image~~ the Dockerfile to set the correct locale.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1378)


woodruffw:
Found the regression: #1300 broke the earlier fix in #1137. I'll re-apply the original fix.

PR #1384, opened 2019-03-06
System Call Audit

ehennenfent:
As of 03/06/19, Manticore supports roughly 1/4th of the Linux system calls. For most binaries this is fine, but sometimes we run into binaries that require unimplemented system calls. The dev-syscall-audit branch will track efforts to evaluate which of the unimplemented system calls are most important to implement, and identify discrepancies between our implementations and the behavior expected by Linux programs. 

I've started off by adding auto-generated stubs for most of the currently unimplemented calls. These stubs change the default behavior, so that instead of throwing a `SyscallNotDefined` exception when manticore reaches an unimplemented syscall, it will simply ignore it and continue execution. 

This PR adds the [wrapt](https://wrapt.readthedocs.io/en/latest/) library as a dependency. It's used for making sure that decorators properly copy wrapped function attributes. It's used to fix the problem of wrapped system calls having incorrect argument signatures

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1384)


manticore/platforms/platform.py
```diff
@@ -1,10 +1,20 @@
...
+logger = logging.getLogger(__name__)
+
 
 class OSException(Exception):
     pass
 
 
+@wrapt.decorator
+def unimplemented(wrapped, _instance, args, kwargs):
+    logger.warning("Unimplemented system call: %s", wrapped.__name__)
```
ekilmer: Is there any way to get the PC value here? So that it would print something like:
```
m.p.platform:WARNING: 0x402048: Unimplemented system call: sys_setsockopt
```

Another thing that might be useful is to display the args to the system call?
```diff
-     logger.warning("Unimplemented system call: %s", wrapped.__name__)
+     logger.warning("Unimplemented system call: %s -- args: %s", wrapped.__name__, args)
```
```
m.p.platform:WARNING: Unimplemented system call: sys_setsockopt -- args: (3, 1, 15, 140737488354584, 4)
```

ehennenfent: 84afdf0 should be a little bit better. If I find the time, I'll work on modularizing the system call printing. I have some uncommitted changes I keep pushing between branches and it would be nice to be able to relegate those to a script.

ekilmer: The change looks good! Thank you 👍

feliam:
___
*[manticore/core/worker.py, line 161 at r5](https://reviewable.io/reviews/trailofbits/manticore/1384#-LerRPwG7rJYYjsXw2vB:-Lf0kuc94AMHVO5GaHh2:bxfmvo2) ([raw file](https://github.com/trailofbits/manticore/blob/8e38d9beb0c0166cf06ec38b49a1ecf87d8e7a23/manticore/core/worker.py#L161)):*
<details><summary><i>Previously, ehennenfent (Eric Hennenfent) wrote…</i></summary><blockquote>

It looks like most other callbacks ignore the `exc` parameter, so we should be good on that front. The reason I re-added this is that we have some internal tooling that needs to be able to catch exceptions in Manticore. Basically everything gets caught right here and squashed. With this callback, we can still get access to those, even if it's not with a `try...except`
</blockquote></details>

The state here is not terminated but it is killed. 
I think that we now have a list for exceptional states. "killed_states" 
Maybe we need to have different events for killing and terminating a correct state.
Maybe this tools need to subscribe to "will_kill_state" ?

PR #1385, opened 2019-03-07
ManticoreBase refactor

feliam:
* Executor.py deleted
* ManticoreBase now responsible for base functionality doh
* Worker is the base class for a worker
* 3 types of workers:
      * Single: runs in main. Perfect for profiling underlaying archs
      * Threads
      * Multiprocessing Processs
* fs:Workspace lock does not depend on an external process multiprocessing manager.
* state lists are well defined and at ManticoreBase level (All subclasses get the same kind of lists)
* Added Enums as valid type for config.py

Lots of documentation added and refined.
```

                   Manticore symbolically explores program states.


        Manticore phases
        ****************

        Manticore has multiprocessing capabilities. Several worker processes
        could be registered to do concurrent exploration of the READY states.
        Manticore can be itself at different phases: STANDBY, RUNNING.

                      +---------+               +---------+
                ----->| STANDBY +<------------->+ RUNNING |
                      +---------+               +----+----+

        = Phase STANDBY =
        Manticore starts at STANDBY with a single initial state. Here the user
        can inspect, modify and generate testcases for the different states. The
        workers are paused and not doing any work. Actions: run()


        = Phase RUNNING =
        At RUNNING the workers consume states from the READY state list and
        potentially fork new states or terminate states. A RUNNING manticore can
        be stopped back to STANDBY. Actions: stop()


        States and state lists
        **********************

        A state contains all the information of the running program at a given
        moment. State snapshots are saved to the workspace often. Internally
        Manticore associates a fresh id with each saved state. The memory copy
        of the state is then changed by the emulation of the specific arch.
        Stored snapshots are periodically updated using: _save() and _load().

                      _save     +-------------+  _load
            State  +----------> |  WORKSPACE  +----------> State
                                +-------------+

        During exploration Manticore spawns a number of temporary states that are
        maintained in different lists:

                Initial
                State
                  |   +-+---{fork}-----+
                  |   | |              |
                  V   V V              |
                +---------+        +---+----+      +------------+
                |  READY  +------->|  BUSY  +----->| TERMINATED |
                +---------+        +---+----+      +------------+
                     |
                     |                             +--------+
                     +---------------------------->| KILLED |
                                                   +--------+

        At any given time a state must be at the READY, BUSY, TERMINATED or
        KILLED list.

        State list: READY
        =================
        The READY list holds all the runnable states. Internally a state is
        added to the READY list via method `_put_state(state)`. Workers take
        states from the READY list via the `_get_state(wait=True|False)` method.
        A worker mainloop will consume states from the READY list and mark them
        as BUSYwhile working on them. States in the READY list can go to BUSY or
        KILLED


        State list: BUSY
        =================
        When a state is selected for exploration from the READY list it is
        marked as busy and put in the BUSY list. States being explored will be
        constantly modified  and only saved back to storage when moved out of
        the BUSY list. Hence, when at BUSY the stored copy of the state will be
        potentially outdated. States in the BUSY list can go to TERMINATED,
        KILLED or they can be {forked} back to READY. The forking process
        could involve generating new child states and removing the parent
        from all the lists.


        State list: TERMINATED
        ======================
        TERMINATED contains states that have reached a final condition and raised
        TerminateState. Workers mainloop simpliy move the states that requested
        termination to the TERMINATED list. This is a final list.

        ```An inherited Manticore class like ManticoreEVM could internally revive
        the states in TERMINATED that pass some condition and move them back to
        READY so the user can apply a following transaction.```

        State list: KILLED
        ==================
        KILLED contains all the READY and BUSY states found at a cancel event.
        Manticore supports interactive analysis and has a prominetnt event system
        A useror ui can stop or cancel the exploration at any time. The unfinnished
        states cought at this situation are simply moved to its own list for
        further user action. This is a final list.

```

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1385)


ehennenfent:
Went through and added a few comments. No significant objections once the tests are passing. Can you add any significant API changes[1] to Changelog.MD under the "Unreleased" section? 

1: Offhand, I see that a lot of evm methods are deprecated now and that we have functions like `generate_testcases` that must be called after analysis

feliam:
___
*[examples/evm/simple_value_check.sol, line 1 at r1](https://reviewable.io/reviews/trailofbits/manticore/1385#-LbP7Cpg0rdXWlx1bRqj:-Lbx7Rl509QXesNsrxlz:b-ygn0ox) ([raw file](https://github.com/trailofbits/manticore/blob/913c2583c6843d339ee4dc17495e863f2e1fa414/examples/evm/simple_value_check.sol#L1)):*
<details><summary><i>Previously, ehennenfent (Eric Hennenfent) wrote…</i></summary><blockquote>

Was there a reason behind removing the version pragma in this file and leaving it in the others?
</blockquote></details>

No particular reason. I will upgrade all examples to the newest and greatest soliditi version

PR #1386, opened 2019-03-11
Better doc for solc fails

feliam:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1386)

### DIFF
--- a/manticore/ethereum/manticore.py
+++ b/manticore/ethereum/manticore.py
@@ -311,7 +311,7 @@ class ManticoreEVM(ManticoreBase):
 
         contracts = output.get('contracts', [])
         if len(contracts) != 1 and contract_name is None:
-            raise EthereumError(f'Solidity file must contain exactly one contract or you must use a contract parameter to specify one. {", ".join(contracts)}')
+            raise EthereumError(f'Solidity file must contain exactly one contract or you must use a `--contract` parameter to specify one. Contracts found: {", ".join(contracts)}')
 
         name, contract = None, None
         if contract_name is None:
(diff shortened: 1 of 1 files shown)


manticore/ethereum/manticore.py
```diff
@@ -311,14 +311,14 @@ def _compile(source_code, contract_name, libraries=None, solc_bin=None, solc_rem
 
         contracts = output.get('contracts', [])
         if len(contracts) != 1 and contract_name is None:
-            raise EthereumError('Solidity file must contain exactly one contract or you must use a contract parameter to specify one.')
+            raise EthereumError(f'Solidity file must contain exactly one contract or you must use a contract parameter to specify one. {", ".join(contracts)}')
```
disconnect3d:
```suggestion
            raise EthereumError(f'Solidity file must contain exactly one contract or you must use a `--contract` parameter to specify one. Contracts found: {", ".join(contracts)}')
```
maybe like this?

manticore/ethereum/manticore.py
```diff
@@ -328,7 +328,7 @@ def _compile(source_code, contract_name, libraries=None, solc_bin=None, solc_rem
         name = name.split(':')[1]
 
         if contract['bin'] == '':
-            raise EthereumError('Solidity failed to compile your contract.')
+            raise EthereumError('Solidity failed to generate bytecode for your contract. Check if all the abstract functions are implemented')
```
disconnect3d: > Check if all the abstract functions are implemented
is this always the case? if not, maybe this:
```suggestion
            raise EthereumError('Solidity failed to generate bytecode for your contract. It might be that you forgot to implement all of the abstract functions?')
```
Also, can't we print out warnings/errors from solc here? (maybe the `warnings` variable have them?)

PR #1406, opened 2019-04-18
Add crytic-compile support

montyly:
This PR adds the support for [`crytic-compile`](https://github.com/crytic/crytic-compile) and allows manticore to work natively on many platforms.

The support of `crytic-compile` comes with a set of [standardized options](https://github.com/crytic/crytic-compile/blob/master/crytic_compile/cryticparser/cryticparser.py), that are available on the manticore command line, or through `solidity_create_contract`. These options are the same than for all the other tools using `crytic-compile` (including Slither and evm_cfg_builder).

Examples of usage:

### Etherscan:
```manticore 0xD850942eF8811f2A866692A623011bDE52a462C1 --contract VEN```

### Embark
```
embark demo
cd embark_demo
manticore . --embark-overwrite-config
```

### Truffle
```
mkdir metacoin
cd metacoin
truffle unbox metacoin
manticore . --contract MetaCoin
```

Notes:
- This PR fixes https://github.com/trailofbits/manticore/issues/1398

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1406)

### CODE REVIEW

**Comment by montyly** [comment]
The PR is ready to be reviewed.

What is the best strategy to add integration tests? Should I init a truffle and embark directories, let manticore run for 1 min on them, and check that some txs were generated?

**Comment by ehennenfent** [comment]
Hey! Full integration tests probably aren't necessary, but we should add some unit tests before merging. @feliam suggested we initialize a couple of contracts via crytic-compile and then check that they're valid. I'm not sure to what degree that would be redundant to crytic-compile's own tests. Any thoughts?

**Comment by montyly** [comment]
Yes that makes sense.

Crytic-compile unit-tests check for the compilation on all the platform. What is missing is the tests of the python API itself, which Manticore is the biggest user for now. 

I can write:
- One short contract within a truffle repo   
- One short contract using a library through delegatecall, within a truffle repo  

And then check that Manticore explores all the paths, and deploy correctly the library, in both scenarios.

**Comment by montyly** [comment]
I have updated the branch to `master`. 

The best way to test truffle is to do it through their metacoin demo, which contains a library. For example:
```
mkdir test
cd test
truffle unbox metacoin
manticore . --contract MetaCoin --workspace output
```
And then checking the number of testcase generated in `output`. It should be 28.

I am not sure how to integrate this with travis? The readme of https://github.com/trailofbits/manticore/tree/master/tests indicates that ` DO NOT create directories in tests/`

Any tips on how to do it?

**Comment by ehennenfent** [comment]
@montyly It looks like this test only takes a few minutes. We don't want to make the regular ethereum tests take any longer, but you could rename the `ethereum_vm` tests job to `ethereum_2` or something, and add the metacoin tests to that. You'll need to add the truffle/manticore commands to the relevant section of the test runner script, shown below. 

https://github.com/trailofbits/manticore/blob/e4539006b13619427d1d3bd983fea5b0873a81b0/scripts/travis_test.sh#L109-L114

**Comment by montyly** [comment]
The unit test was added. The PR is ready to be reviewed

File: scripts/travis_test.sh [hunk 1]
```diff
@@ -108,9 +129,11 @@ run_examples() {
 case $1 in
     ethereum_vm)
         make_vmtests
+        install_truffle
+        RV=$?
         echo "Running only the tests from 'tests/$1' directory"
         run_tests_from_dir $1
```
ehennenfent: @montyly Are you missing a call to `run_truffle_tests`?

montyly: Indeed, my bad

montyly:
@ehennenfent travis is fixed and the test succeeded

PR #1415, opened 2019-04-25
Performance optimization : use set instead of list

catenacyber:
Found by profiling ais3_crackme from manticore-examples

Profiling result ie
`cat cprof.txt | awk '{print $2 $0}' | sort -n | tail`
gives
```
1.929   825101    1.929    0.000    4.210    0.000 helpers.py:13(issymbolic)
2.862      277    2.862    0.010    2.866    0.010 {method 'readline' of '_io.TextIOWrapper' objects}
3.491  1173293    3.491    0.000    3.742    0.000 x86.py:469(__contains__)
8.196   161737    8.196    0.000    8.196    0.000 {built-in method posix.read}
```

So x86.py:469(__contains__) is the second place where we spend the most time

The patch replaces the list by a set to get better performance for the `in` operation : O(1) instead of O(n)

Thanks to https://stackoverflow.com/questions/20234935/python-in-operator-speed/20234990#20234990

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1415)

### DIFF
--- a/manticore/native/cpu/x86.py
+++ b/manticore/native/cpu/x86.py
@@ -454,9 +454,9 @@ class AMD64RegFile(RegisterFile):
         for name in ('AF', 'CF', 'DF', 'IF', 'OF', 'PF', 'SF', 'ZF'):
             self.write(name, False)
 
-        self._all_registers = tuple(self._table) + \
-            ('FP0', 'FP1', 'FP2', 'FP3', 'FP4', 'FP5', 'FP6', 'FP7', 'EFLAGS', 'RFLAGS') + \
-            tuple(self._aliases)
+        self._all_registers = set(self._table.keys()) | \
+            set(['FP0', 'FP1', 'FP2', 'FP3', 'FP4', 'FP5', 'FP6', 'FP7', 'EFLAGS', 'RFLAGS']) | \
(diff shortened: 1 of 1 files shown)


mossberg:
Cool! How much does this improve performance?

catenacyber:
And with this patch, we drop from 3 seconds to 0.4 seconds for the 1173293 calls to this function in ais3_crackme

catenacyber:
(total time for ais3_crackme is about one minute)

disconnect3d:
> (total time for ais3_crackme is about one minute)

Total time improvement, right? That's awesome.

Can we do that for other architectures too - in future PRs?

E.g. https://github.com/trailofbits/manticore/blob/master/manticore/native/cpu/arm.py#L298-L304

disconnect3d:
Btw if you use PyCharm it can display profiling results in a table letting you sort via given column and also display a colored code-graph so it is easy to see what calls what.

Also iirc the profiling results can be sorted within some cprofile function instead of in bash :).

catenacyber:
> Can we do that for other architectures too - in future PRs?

I do not see a `__contains__` function in arm.py 
@disconnect3d do you have a test case to profile ?

PR #1418, opened 2019-04-29
aarch64: add everything except instructions

nkaretnikov:
Based on disconnect3d/aarch64/skeleton.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1418)


nkaretnikov:

nkaretnikov:
I'm so glad you suggested to rename the example script! I did that and realized it's now broken. I don't get the same behavior as on a pre-rebase branch. I suspect something is changed in the syscall layer, investigating...

PR #1419, opened 2019-04-29
aarch64: add instructions

nkaretnikov:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1419)


nkaretnikov:

ehennenfent:
We're going to retract the request to remove the asserts for now. Since this is a big new branch, we'll accept the performance hit of having some extra sanity checks. Once this has been out in the real world for a while, we'll convert them to type hints ourselves.

nkaretnikov:
Okay, sure. I've caught a few issues this way, but this doesn't seem to justify performance implications. Also, the current style is really verbose, which might be a disadvantage for those reading the code.

Have you considered running with `-O` in production? This will disable all asserts, so asserts in critical places would need to be converted to exceptions first.

As for the hints, could you show me how to do it if you know offhand? I can't figure this out: one would need to match a constant against the type attribute.

PR #1420, opened 2019-04-29
aarch64: add instruction tests: A-B

nkaretnikov:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1420)


nkaretnikov:

PR #1433, opened 2019-05-14
AArch64: fix ldrb size

nkaretnikov:
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1433)

### DIFF
--- a/manticore/native/cpu/aarch64.py
+++ b/manticore/native/cpu/aarch64.py
@@ -1020,7 +1020,9 @@ class Aarch64Cpu(Cpu):
         if ldr:
             result = cpu.read_int(base + imm, size)
             if sextend:
-                result = Operators.SEXTEND(result, size, cpu.address_bit_size)
+                result = Operators.SEXTEND(result, size, reg_op.size)
+            else:
+                result = Operators.ZEXTEND(result, reg_op.size)
             reg_op.write(result)
         else:
             reg = reg_op.read()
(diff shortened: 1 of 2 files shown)


nkaretnikov:
Force-pushed after changing the test constant to a more suitable value.

nkaretnikov:
On AArch64, the symbolic tests work by setting a register to a symbolic value. For load instructions, this register is the stack pointer. Turns out, it's not enough as the memory can also be symbolic. This PR adds a test for one pair of store/load instructions and fixes a related bug (incorrect load due to a size mismatch between the loaded value and target register).

nkaretnikov:
BTW, you can trigger the bug by reverting the top commit and running the new test:
```
# git revert @
# nosetests -s tests/native/test_aarch64cpu.py:Aarch64SymInstructions -m 'test_strb_ldrb_imm_base32'
[...]
  File "/manticore/manticore/native/cpu/aarch64.py", line 3249, in LDRB
    cpu._LDRB_immediate(reg_op, mem_op, mimm_op)
  File "/manticore/manticore/native/cpu/aarch64.py", line 3222, in _LDRB_immediate
    cpu._ldr_str_immediate(reg_op, mem_op, mimm_op, ldr=True, size=8)
  File "/manticore/manticore/native/cpu/aarch64.py", line 1024, in _ldr_str_immediate
    reg_op.write(result)
  File "/manticore/manticore/native/cpu/aarch64.py", line 5438, in write
    self.cpu.regfile.write(self.reg, value)
  File "/manticore/manticore/native/cpu/aarch64.py", line 194, in write
    assert value.size == size
AssertionError
```

PR #1438, opened 2019-05-16
Blacken Manticore

ehennenfent:
Before our next release, we'll run the `black` autoformatter on Manticore and add a CI check to make sure it's been run on incoming pull requests.

I'd _prefer_ to have a separate CI check for `black` so that we could make it an optional requirement, but seeing as CodeClimate doesn't currently support it, Travis will have to do for now. 

Apologies for the commit spam if it turns out I've messed up the Travis files

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1438)


CONTRIBUTING.md
```diff
@@ -30,6 +30,10 @@ more documentation, look [here](https://guides.github.com/activities/forking/).
 
 Some pull request guidelines:
 
+- We use the [`black`](https://black.readthedocs.io/en/stable/index.html) auto-formatter
+  to enforce style conventions in Manticore. To ensure your code is properly 
+  formatted, run `black -t py36 -l 100 .` in the manticore directory before
```
ekilmer: Make sure to update this when you've decided on the final line length in `.travis.yml`. Right now travis does `120` but here it says `100`

scripts/travis_install.sh
```diff
@@ -26,12 +26,22 @@ function install_cc_env {
     pip install awscli
 }
 
-# install CodeClimate env in all conditions
-install_cc_env
+# Install black for initial formatting stage
+if [ "$1" == "format" ]; then
+    pip install black
```
ekilmer: Since the python packages get cached, do we want to stick with the first version we install or should this be `pip install -U black` to check for upgrades each time? 

Note: I'm not entirely sure how this works, but I _think_ it will stay on the first version that gets cached, or is that not how this works?

CONTRIBUTING.md
ehennenfent: You caught that much faster than I was expecting. My thought was to suggest 100, but only enforce 120. Thinking about it more, lingering 120-character lines in the repo would risk getting reformatted by later, unrelated pull requests. Perhaps I should rewrite the above command to only look at modified files, or we should simply pick one number and stick to it.

ekilmer: I think one number would make things much simpler for both the CI check and contributing, but I can see how a few 120-character lines could read better in special circumstances.

Are there any flags that black will read to possibly allow certain lines to not be reformatted? I'm not too familiar with the options

ehennenfent: In other news, Travis has just reminded me that `black` will extend lines that have been unnecessarily shortened to fit the specified max line length, so we can only choose one number. I'll change it to 100 for now but I'm willing to hear arguments otherwise. Personally, I do find 100 to be a bit cramped, but the `black` docs raise the points that line lengths over ~100 characters start to cause problems for:
* Side-by-side code review
* Visually impaired people
* Presenting code on slides

We'll see how it goes

scripts/travis_install.sh
ehennenfent: Hmm. On one hand, I'm wary of `black` getting updated and our CI suddenly no longer passing. On the other hand, it's currently beta software, so we probably don't want to fix the version number forever. 

I'll add a -U for now. If it ever breaks our CI, we'll then immediately fix the version number at whatever the last good version was. That way we'll at least know to fix the problems instead of picking one version and then forgetting to ever update it.

.codeclimate.yml
```diff
@@ -48,6 +48,10 @@ plugins:
         enabled: false
       E701:
         enabled: false
+      E203:
+        enabled: false
+      W503:
```
disconnect3d: Loud thinking: afaik we can comment in yaml using `#` as in Python, so it might be good to add a readable text what we disable.

Let's do it in another PR.

.travis.yml
```diff
@@ -36,6 +37,10 @@ cache:
 
 jobs:
   include:
+    - stage: format
+      env: TEST_TYPE=format
+      script:
+        - git diff --name-only $TRAVIS_COMMIT_RANGE | scripts/pyfile_exists.py | xargs black -t py36 -l 100 --check
```
disconnect3d: woaaa (via https://docs.travis-ci.com/user/environment-variables/), that's nice!
```
TRAVIS_COMMIT_RANGE: The range of commits that were included in the push or pull request. (Note that this is empty for builds triggered by the initial commit of a new branch.)
```

CONTRIBUTING.md
disconnect3d: I'd be up for 120 but 100 is also ok as I agree with the points. OTOH you don't use 120 lengthy lines that often, maybe for some documentation(?) or when you have too many scopes (but in this case u probably should move sth to a function/method if it's not a hot code)

examples/evm/minimal.py
```diff
@@ -23,25 +23,26 @@
...
-'''
+"""
 
-user_account = m.create_account(balance=1000, name='user_account')
+user_account = m.create_account(balance=1000, name="user_account")
 print("[+] Creating a user account", user_account.name_)
 
-contract_account = m.solidity_create_contract(source_code, owner=user_account, name='contract_account')
+contract_account = m.solidity_create_contract(
+    source_code, owner=user_account, name="contract_account"
```
disconnect3d: This had a length of 103! heh

.travis.yml
ehennenfent: It's not explicit about this in the description, but for PR builds, it'll give you all the files that differ between `master` and the pull request. For branch builds, it'll just give you the files that have changed since the last build.

PR #1444, opened 2019-05-22
x86: movhps support

### LINKED ISSUE #1432: Implement MOVHPS on x86
We've had a couple of issues submitted (#1425, #1411) where the included log file has a line like `m.n.c.abstractcpu:WARNING: Unimplemented instruction: [addr]:	[bytes]	movhps [args]`. Since we've encountered it in the wild, we should add support for the [MOVHPS](https://www.felixcloutier.com/x86/movhps) instruction to Manticore. 

We're prepared to offer a bounty for this issue.

woodruffw:
Adds support for `movhps`.

The implementation of `movhps` isidentical to `movhpd` since it moves two single-precision floats at the same time. As a result, we _could_ deduplicate the two into a single `movhp` or whatever method if desired.

Closes #1432.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1444)

### DIFF
--- a/manticore/native/cpu/x86.py
+++ b/manticore/native/cpu/x86.py
@@ -5508,6 +5508,27 @@ class X86Cpu(Cpu):
             dest.write(Operators.CONCAT(128, src.read(), value))
 
     @instruction
+    def MOVHPS(cpu, dest, src):
+        """
+        Moves high packed single-precision floating-point value.
+
+        Moves two packaed single-precision floating-point values from the source operand
+        (second operand) to the destination operand (first operand). The source and destination
+        operands can be an XMM register or a 64-bit memory location. The instruction allows
+        single-precision floating-point values to be moved to and from the high quadword of
+        an XMM register and memory. It cannot be used for register to register or memory to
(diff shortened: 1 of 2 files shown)


disconnect3d:
@woodruffw btw can we also add a testcase for it?

manticore/native/cpu/x86.py
```diff
@@ -5507,6 +5507,27 @@ def MOVHPD(cpu, dest, src):
             value = Operators.EXTRACT(dest.read(), 0, 64)  # low part
             dest.write(Operators.CONCAT(128, src.read(), value))
 
+    @instruction
+    def MOVHPS(cpu, dest, src):
+        """
+        Moves high packed single-precision floating-point value.
+
+        Moves two packaed single-precision floating-point values from the source operand
```
ehennenfent:
```suggestion
        Moves two packed single-precision floating-point values from the source operand
```

woodruffw:
> btw can we also add a testcase for it?

Yep, I'll add one now.

PR #1456, opened 2019-06-04
Moving issymbolic to smtlib

catenacyber:
Performance optimization avoiding runtime imports

Found by profiling ais3_crackme from manticore_examples with yappi (for multithreading profiling)
`issymbolic` was the leading function with 5.75 seconds in itself for about 2 minutes total execution

This change reduces it to 2.5 seconds

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1456)

### DIFF
--- a/manticore/__init__.py
+++ b/manticore/__init__.py
@@ -6,7 +6,8 @@ if sys.version_info < (3, 6):
 
 from .utils import config, log
 from .utils.log import set_verbosity
-from .utils.helpers import issymbolic, istainted
+from .utils.helpers import istainted
+from .core.smtlib import issymbolic
 from .ethereum.manticore import ManticoreEVM
 
 __all__ = [issymbolic.__name__, istainted.__name__, ManticoreEVM.__name__, set_verbosity.__name__]
--- a/manticore/__init__.py
+++ b/manticore/__init__.py
@@ -6,8 +6,7 @@ if sys.version_info < (3, 6):
 
 from .utils import config, log
 from .utils.log import set_verbosity
-from .utils.helpers import istainted
-from .core.smtlib import issymbolic
+from .core.smtlib import issymbolic, istainted
 from .ethereum.manticore import ManticoreEVM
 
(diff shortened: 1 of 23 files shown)


catenacyber:
Total time without profiling goes from 47 seconds to 40 seconds for ais3_crackme

catenacyber:
CI failure does not seem related. Is it ?

ehennenfent:
Thanks for the PR! Why do you think moving the function has improved performance? I've seen replacing `import *` improve performance by eliminating `LOAD_GLOBAL`s, but I'm not sure why moving the function from one module to another makes a difference. 

I've kicked off a re-test of master just to make sure we haven't encountered any EVM regressions. If [this build](https://travis-ci.com/trailofbits/manticore/builds/114290376) fails we'll know that moving the function has caused a problem.

catenacyber:
> Why do you think moving the function has improved performance?

Sorry, I have not been clear.
Moving the function does not improve performance.
But this is a needed side effect of removing the runtime `import` (yes executed at runtime 18000 thousands times looking for the right file)
ie removing this line https://github.com/trailofbits/manticore/blob/master/manticore/utils/helpers.py#L22 improves performance.
But we cannot just remove this line.
(There may another solution as to move the function to this other file)

> If this build fails

It succeeded :-)

ehennenfent:
I see what you mean now about why removing that inline `import` improves performance. I'm afraid I misspoke about the Travis build though. Since it passed, that means the issue with the tests must have been introduced by this change. It's not clear to me what's going on as all the errors are assertion errors, not, say, ImportErrors like one might expect from having moved a function around. It _is_ probably telling that all the failing tests are ethereum tests. The only notable difference I can see is that in some of the EVM files, `issymbolic` is imported _before_ some of the other imports that it used to be after. @feliam any ideas?

catenacyber:
I think I found my bug
`python3 -m manticore  --txlimit 1 contracts/1102.sol ` was failing

PR #1465, opened 2019-06-12
Perf optimization for writing one byte in memory

catenacyber:
Tested on ais3_crackme example

Function `write` called 19164 times goes down from 2.076 seconds cumulated to 0.914 seconds by using this new `write1` function (on 42 seconds total time)

This code may be less readable though...

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1465)

### DIFF
--- a/manticore/native/memory.py
+++ b/manticore/native/memory.py
@@ -1179,16 +1179,30 @@ class SMemory(Memory):
                     self._symbols.setdefault(base + offset, []).append((condition, value[offset]))
         else:
 
+            # regroups into areas to minimize calls to super().write
+            areas = []
+            startArea = 0
+            sizeArea = 0
             for offset in range(size):
                 if issymbolic(value[offset]):
+                    if sizeArea > 0:
+                        areas.append((startArea, sizeArea))
+                        sizeArea = 0
+                        startArea = offset + 1
                     if not self.access_ok(address + offset, "w", force):
                         raise InvalidMemoryAccess(address + offset, "w")
                     self._symbols[address + offset] = [(True, value[offset])]
                 else:
-                    # overwrite all previous items
-                    if address + offset in self._symbols:
-                        del self._symbols[address + offset]
-                    super().write1(address + offset, value[offset], force)
+                    sizeArea+=1
+            if sizeArea > 0:
+                areas.append((startArea, sizeArea))
+            for (startArea, sizeArea) in areas:
+                if sizeArea == 1:
+                    super().write1(address+startArea, value[startArea], force)
+                else:
+                    super().write(address+startArea, value[startArea:startArea+sizeArea], force)
+
+
 
     def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=False):
         """
--- a/manticore/native/memory.py
+++ b/manticore/native/memory.py
@@ -970,7 +970,7 @@ class Memory(object, metaclass=ABCMeta):
         if self._recording_stack:
             self._recording_stack[-1].append((addr, [val]))
 
-        m[addr] = val
+        m[addr : addr + 1] = [val]
 
     def write(self, addr, buf, force=False):
         size = len(buf)
--- a/manticore/native/memory.py
+++ b/manticore/native/memory.py
@@ -301,7 +301,7 @@ class AnonMap(Map):
                 self._data = [
                     Operators.ORD(b) for b in self._data
                 ]  # completely convert everything to a list if we need to store some symbolic data, just copying fpse
-            else if not isinstance(value[0], int):
+            elif not isinstance(value[0], int):
                 value = [Operators.ORD(n) for n in value]
             self._data[index] = value
         else:
--- a/manticore/native/memory.py
+++ b/manticore/native/memory.py
@@ -1193,16 +1193,16 @@ class SMemory(Memory):
                         raise InvalidMemoryAccess(address + offset, "w")
                     self._symbols[address + offset] = [(True, value[offset])]
                 else:
-                    sizeArea+=1
+                    sizeArea += 1
             if sizeArea > 0:
                 areas.append((startArea, sizeArea))
             for (startArea, sizeArea) in areas:
                 if sizeArea == 1:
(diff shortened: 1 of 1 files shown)


ehennenfent:
It looks like most of the Travis failures are `TypeError: 'int' object is not subscriptable` on line 299 of memory.py. Can you add a type check for that, or does it negate the performance improvement?

catenacyber:
Thank you @ehennenfent 

I just changed the order of type checks.
It looks like the line `self._data[index] = Operators.ORD(value)` in `AnonMap.__setitem__` was not reachable because we were first trying to access `value[0]` 
Now we only access `value[0]` if we are dealing with a slice

catenacyber:
> Checking the memory permissions on each byte is going to be slow.

But we have only one byte as that is the point of this PR.
Or do you mean that this loop (one level up the stack) should have been optimized instead ?
https://github.com/trailofbits/manticore/blob/81f8f97bf6fe3fc61c020e9081be6157c9d455a4/manticore/native/memory.py#L1180

ehennenfent:
> do you mean that this loop (one level up the stack) should have been optimized instead?

Not necessarily "instead", but yes, we should modify the loop such that it's not necessary to check the memory permissions for every single byte. We can save that for a different PR.

catenacyber:
> Not necessarily "instead", but yes, we should modify the loop such that it's not necessary to check the memory permissions for every single byte. We can save that for a different PR.

I guess I will do the upper loop modification first and then see if this PR is always relevant

catenacyber:
With the upper loop modification, I get 0.6 seconds cumulated for ais3_crackme with command `python3 -m cProfile -o cprofall.txt win.py`
```
19167    0.100    0.000    0.576    0.000 memory.py:1160(write)
```

With only the upper loop change (ie not using `write1`), I got 0.7 seconds and without any changes 2.1 seconds

manticore/native/memory.py
```diff
@@ -1165,16 +1179,30 @@ def write(self, address, value, force=False):
...
+                        areas.append((startArea, sizeArea))
+                        sizeArea = 0
+                        startArea = offset + 1
                     if not self.access_ok(address + offset, "w", force):
                         raise InvalidMemoryAccess(address + offset, "w")
                     self._symbols[address + offset] = [(True, value[offset])]
                 else:
-                    # overwrite all previous items
-                    if address + offset in self._symbols:
-                        del self._symbols[address + offset]
```
disconnect3d: I think that the CI fails with a timeout because we removed this. It might be that it ends up in an infinite loop.

Also, the `test_mix_of_concrete_and_symbolic` test fails.

In this test we:
1. Create memory with 0, 2, 4, 6 offsets/indexes/memory cells as concrete values and 1, 3, 5, 7 as symbolic.
2. Assert if those concretes/symbolic addresses are really concrete/symbolic.
3. Swap indexes of concrete/symbolic memory cells and reinitialise memory bytes to concrete/symbolic values.
4. We reassert those. **Here, the test fails.**

This is because we don't delete the old symbolic mappings, so in `mem._symbols` we have old 1, 3, 5, 7 symbolic cells and 0, 2, 4, 6 new symbolic cells. **The 1, 3, 5, 7 should not be there.**

disconnect3d: I am going to try fixing that by readding the `del ...` and see if it works.

dguido:
oops, I think you may have accidentally automatically closed this because of the text in #1506

disconnect3d:
I spend some time on this code/issue, analyzed it and found out that yes, indeed we need this or similar improvement, but the code present here is not complete and a bit overcomplex imho.

I am closing this issue and I am going to propose another solution along with other improvements.

By the way, the benchmark used here - `manticore-examples/test_ais3_crackme` - is not a good benchmark to show the issue.

A better one is here:
```python
import time
from manticore.core.smtlib import ConstraintSet
from manticore.native.memory import SMemory32

def test_write_concrete():
    t1 = time.time()
    mem = SMemory32(ConstraintSet())
    buf = mem.mmap(None, 0x1000, "rw")

    for i in range(1000):
        mem[buf:buf+0x1000] = "a" * 0x1000

    t2 = time.time()
    print(f"test_write_concrete: {t2-t1} seconds")

test_write_concrete()
```

Here are my results:
* current `master` or `cdda4ee`: `test_write_concrete: 54.29297089576721 seconds`
* this PR: `test_write_concrete: 2.8424699306488037 seconds`
* my solution, that does not introduce `write1` and only chunks the writes: `test_write_concrete: 3.17004132270813 seconds`

Just to be clear, the current PR has the following issues:
* it does not clear symbolic chunk writes when writing concrete values
* it does not support writes with partial symbolic and concrete values which works on master

manticore/native/memory.py
catenacyber: Indeed

catenacyber:
Bravo for the rework, glad I could help.

> By the way, the benchmark used here - manticore-examples/test_ais3_crackme - is not a good benchmark to show the issue as it does not involve a lot of memory writes.

Interesting.
I am looking for good benchmarks in a general way.
test_ais3_crackme is good as it is a real use case
We will not find worst case scenario in it but we can improve the performance in the general case.

> my solution, that does not introduce write1 

I agree that write1 is not so good to introduce
But Eric told me to keep it here https://github.com/trailofbits/manticore/pull/1465#issuecomment-511000613

> it does not support writes with partial symbolic and concrete values which works on master 

How so ?

disconnect3d:
> Bravo for the rework, glad I could help.

Thanks :).

> test_ais3_crackme is good as it is a real use case
Yeah, as a real use case it is rather good but it might be hard (and slow) to measure the right impact/improvement with it in regards to the given functionality.

> I agree that write1 is not so good to introduce
> But Eric told me to keep it here #1465 (comment)

I guess we can play with it in the future, but for now its speed-up is probably negligible.

> > it does not support writes with partial symbolic and concrete values which works on master

> How so ?

Assignments like `mem[0:2] = (cs.new_bitvec(8), b'a')` stopped working. We haven't tests for it before, but I've added such in #1509.

PR #1466, opened 2019-06-12
Enforce black formatting

Tiecoon:
create a pyproject.toml so users don't have to worry about command line flags for black

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1466)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
     - stage: format
       env: TEST_TYPE=format
       script:
-        - git diff --name-only $TRAVIS_COMMIT_RANGE | python3 scripts/pyfile_exists.py | xargs black -t py36 -l 100 --check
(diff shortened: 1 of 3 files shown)


ekilmer:
Would be nice to propagate this to `.travis.yml` also. Should appear on [line 43](https://github.com/trailofbits/manticore/blob/a0966dca9b2243a1235afed9a431f00e69ad358c/.travis.yml#L43)

ehennenfent:
@Tiecoon Just to make sure having a `pyproject.toml` in the root doesn't break the install process, can you run `pip install .` in a fresh virtual environment and make sure it works okay?

Tiecoon:
it works fine here, unicorn still complains if i just pip install .[native], but it does that on master for me as well currently.

Tiecoon:
i can figure out how to specify a build system in the pyproject.toml.

PR #1469, opened 2019-06-18
Dev apidocs refactor

Tiecoon:
Looking for feedback and if there are any common members or classes that people feel should be added or emphasized more if they use them frequently in manticore scripts.

build with `$ make html` in maticore/docs.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1469)


docs/base.rst
```diff
@@ -0,0 +1,7 @@
+ManticoreBase
+=============
+
+.. autoclass:: manticore.core.manticore.ManticoreBase
+   :members:
+   :special-members: __init__
+   :exclude-members: all_states, ready_states, count_ready_states, count_busy_states, killed_states, count_killed_states, terminated_states, count_terminated_states
```
feliam: maybe: finalize, locked_context,

Tiecoon: You want those excluded from the base?

Currently I exclude the ones above because I thought they were more appropriate for the states section. All other documented members for ManticoreBase should be present.

https://manticore.readthedocs.io/en/latest/base.html#manticore.core.manticore.ManticoreBase.finalize

https://manticore.readthedocs.io/en/latest/base.html#manticore.core.manticore.ManticoreBase.locked_context

feliam: oh exclude/

docs/native.rst
```diff
@@ -76,3 +76,64 @@ To implement a user-defined model, implement your model as a Python function, an
...
+
+    Manticore.linux('./binary', ['++++++', '++++++'], dict(VAR1='+++++', VAR2='++++++'))
+
+Symbolic stdin
+^^^^^^^^^^^^^^
+
+Manticore by default is configured with 256 bytes of symbolic stdin data which is configurable
+with the ``stdin_size`` kwarg of :meth:`manticore.native.Manticore.linux` , after an optional
+concrete data prefix, which can be provided with the ``concrete_start`` kwarg of
+:meth:`manticore.native.Manticore.linux`.
```
disconnect3d: hmm, seems like some mismatch with ReST format?
image

Tiecoon: Where are you seeing this?

https://manticore.readthedocs.io/en/latest/native.html#symbolic-input

image

disconnect3d: Oh it renders there properly. I checked it via github, my bad then. LGTM :)

manticore/native/cpu/x86.py
```diff
@@ -5701,7 +5701,8 @@ def MOVHPD(cpu, dest, src):
         :param src: source operand.
         """
         if src.size == 128:
-            assert dest.size == 64
+            print(dest.size)
```
ekilmer: Why is this `print` here? Just something that snuck through while debugging? :)

Tiecoon: Ah yeah, a capstone bug lead me down that route. That function normally never gets called the opcode is manually changed to movs due to a mnemonic conflict it seems. Where one movhpd says it can't write to memory and the other can. If I remember correctly. I'll fix that after this meeting.

PR #1472, opened 2019-06-20
Fix sycall getdents

ekilmer:
Based on description from http://man7.org/linux/man-pages/man2/getdents.2.html and tested on example within that document.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1472)

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -2685,7 +2685,7 @@ class Linux(Platform):
             stat = item.stat()
             print(f"FILE MODE: {item.name} :: {stat.st_mode:o}")
 
-            # https://elixir.bootlin.com/linux/latest/source/include/linux/fs_types.h#L27
+            # https://elixir.bootlin.com/linux/v5.1.15/source/include/linux/fs_types.h#L27
             d_type = (stat.st_mode >> 12) & 15
 
             packed = struct.pack(
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -2676,7 +2676,7 @@ class Linux(Platform):
 
         item = next(dent_iter, None)
         while item is not None:
-            fmt = f"LLH{len(item.name) + 1}sxc"
+            fmt = f"LLH{len(item.name) + 1}sB"
             size = struct.calcsize(fmt)
             if len(buf) + size > count:
                 # Don't overflow buffer
@@ -2685,8 +2685,11 @@ class Linux(Platform):
             stat = item.stat()
             print(f"FILE MODE: {item.name} :: {stat.st_mode:o}")
 
+            # https://elixir.bootlin.com/linux/latest/source/include/linux/fs_types.h#L27
+            d_type = (stat.st_mode >> 12) & 15
+
             packed = struct.pack(
-                fmt, item.inode(), size, size, bytes(item.name, "utf-8") + b"\x00", b"\x00"
+                fmt, item.inode(), size, size, bytes(item.name, "utf-8") + b"\x00", d_type
             )
             buf += packed
             item = next(dent_iter, None)
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -2,6 +2,7 @@ import binascii
 import ctypes
 import errno
 import fcntl
+import itertools
 import logging
 import socket
 import struct
@@ -2648,6 +2649,9 @@ class Linux(Platform):
             return -e.errno
         return 0
 
+    # A cache for keeping state when reading directories { fd: dent_iter }
+    sys_getdents_c = {}
+
     def sys_getdents(self, fd, dirent, count) -> int:
         """
         Fill memory with directory entry structs
@@ -2663,19 +2667,44 @@ class Linux(Platform):
             logger.info("Can't get directory entries for a file")
             return -1
 
-        with os.scandir(file.path) as dent_iter:
-            for index, item in zip(range(count), dent_iter):
-                fmt = f"LLH{len(item.name) + 1}sxc"
-                size = struct.calcsize(fmt)
+        if fd not in self.sys_getdents_c:
+            # First call on this file descriptor
+            self.sys_getdents_c[fd] = os.scandir(file.path)
 
-                stat = item.stat()
-                print("FILE MODE:", item.name, " :: ", stat.st_mode)
+        dent_iter = self.sys_getdents_c[fd]
 
-                buf += struct.pack(
-                    fmt, item.inode(), size, size, bytes(item.name, "utf-8") + b"\x00", b"\x00"
-                )
+        item = next(dent_iter, None)
+        while item is not None:
+            fmt = f"LLH{len(item.name) + 1}sxc"
+            size = struct.calcsize(fmt)
+            if len(buf) + size > count:
+                # Don't overflow buffer
+                break
 
-        self.current.write_bytes(dirent, buf)
+            stat = item.stat()
+            print(f"FILE MODE: {item.name} :: {stat.st_mode:o}")
+
+            packed = struct.pack(
+                fmt, item.inode(), size, size, bytes(item.name, "utf-8") + b"\x00", b"\x00"
+            )
+            buf += packed
+            item = next(dent_iter, None)
+
+        if item:
+            # Prepend the last valid item that didn't fit to the list for next time
+            self.sys_getdents_c[fd] = itertools.chain([item], dent_iter)
+        else:
+            # If everything fit, then save just the dent_iter
+            self.sys_getdents_c[fd] = dent_iter
+
+        if len(buf) > 0:
+            # Write out to buffer if we have something to write
+            self.current.write_bytes(dirent, buf)
+        else:  # len(buf) == 0
+            # When the buffer is 0, that means we've read all directory entries
+            # Delete the state and don't write a zero buffer back to dirent
+            del self.sys_getdents_c[fd]
+
         return len(buf)
 
     def sys_nanosleep(self, rqtp, rmtp) -> int:
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -450,6 +450,8 @@ class Linux(Platform):
         self.program = program
         self.clocks = 0
         self.files = []
+        # A cache for keeping state when reading directories { fd: dent_iter }
+        self._getdents_c = {}
         self._closed_files = []
         self.syscall_trace = []
         # Many programs to support SLinux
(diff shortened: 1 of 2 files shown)


manticore/platforms/linux.py
```diff
@@ -2648,6 +2649,9 @@ def sys_unlink(self, pathname) -> int:
             return -e.errno
         return 0
 
+    # A cache for keeping state when reading directories { fd: dent_iter }
+    sys_getdents_c = {}
```
ekilmer: I would love suggestions on the best way to do this. I want this to be a static variable for the `getdents` system call because it needs to store state. I saw that decorators could accomplish this, but do we do anything like this in Manticore that I could look at for inspiration?

ehennenfent: Are you sure you _really_ want there to be just one copy of this dict for _all_ the states? I'm not sure exactly what you're trying to accomplish here, but if you're just trying to store data in the `Linux` object, you can set `self._getdents_c = {}` in the `__init__` function as normal, then add corresponding entries to the `__getstate__` and `__setstate__` methods so that it's properly handled during state transitions.

ekilmer: I do not want one copy of the dict for _all_ states. I'd like to do as you described, and I tried it a couple times, although I guess I missed adding the field to both `__getstate__` and `__setstate__`. It seems to work now. Thank you!

ekilmer:
@ehennenfent I think there are still some issues with writing/using the `dirent` structs associated with this syscall. I'm getting incorrect behavior when running the following code with two files here:
```
root@ec7482a48d26:/manticore/test# ls -al /data
total 16
drwxr-xr-x 1 root root 4096 Jun 13 19:31 .
drwxr-xr-x 1 root  root  4096 Jun 24 18:09 ..
-rw-r--r-- 1 root root  145 Jun  4 15:29 messages.csv
-rw-r--r-- 1 root root  215 Jun  4 15:29 users.csv
```

and this code

```c
#define _GNU_SOURCE

#include <dirent.h>
#include <stdbool.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

bool ends_with_csv(const char* filename);

void main() {
  DIR* data_directory = opendir("/data");

  if (NULL == data_directory) exit(-1);

  struct dirent* entry;

  while ((entry = readdir(data_directory))) {
    if (DT_REG != entry->d_type) {
      printf("FAIL DT_REG - %s\n", entry->d_name);
      continue;
    }
    if (!ends_with_csv((char*)(&entry->d_name))) {
      printf("FAIL ENDS CSV - %s\n", entry->d_name);
      continue;
    }
    printf("Winner!!! - %s\n", entry->d_name);
  }
}

bool ends_with_csv(const char* filename) {
  size_t name_len = strlen(filename);
  if (0 != filename[name_len]) return false;

  if ('v' != filename[name_len - 1]) return false;
  if ('s' != filename[name_len - 2]) return false;
  if ('c' != filename[name_len - 3]) return false;
  if ('.' != filename[name_len - 4]) return false;


  return true;
}
```

Compiling with `gcc test.c` and running `./a.out` gives me:
```
Winner!!! - users.csv
FAIL DT_REG - ..
FAIL DT_REG - .
Winner!!! - messages.csv
```

However, Manticore output from `manticore -v a.out` gives the following relevant log lines
```
...
2019-06-24 18:49:54,962: [1120] m.p.linux:DEBUG: Opening file /lib/x86_64-linux-gnu/libc.so.6 for real fd 9
2019-06-24 18:49:54,964: [1120] m.p.platform:DEBUG: sys_openat(0xffffff9c, "/lib/x86_64-linux-gnu/libc.so.6", 0x80000, 0) = 3
2019-06-24 18:49:54,997: [1120] m.p.platform:DEBUG: sys_read(3, "ELF", 0x340) = 0x340
2019-06-24 18:49:55,183: [1120] m.p.platform:DEBUG: sys_newfstat(3, ") = 0
2019-06-24 18:49:55,585: [1120] m.p.platform:DEBUG: sys_mmap(0, 0x3f0ae0, 5, 0x802, 3, 0) = 0x7fffff9c4000
2019-06-24 18:49:55,611: [1120] m.p.platform:DEBUG: sys_mprotect(0x7fffffbab000, 0x200000, 0) = 0
2019-06-24 18:49:55,655: [1120] m.p.platform:DEBUG: sys_mmap(0x7fffffdab000, 0x6000, 3, 0x812, 3, 0x1e7000) = 0x7fffffdab000
2019-06-24 18:49:55,825: [1120] m.p.platform:DEBUG: sys_mmap(0x7fffffdb1000, 0x3ae0, 3, 50, 0xffffffff, 0) = 0x7fffffdb1000
2019-06-24 18:49:56,020: [1120] m.p.linux:DEBUG: sys_close(3)
2019-06-24 18:49:56,020: [1120] m.p.platform:DEBUG: sys_close(3) = 0
2019-06-24 18:49:57,588: [1120] m.p.platform:DEBUG: sys_mmap(0, 0x2000, 3, 34, 0xffffffff, 0) = 0x7ffffffd4000
2019-06-24 18:49:58,266: [1120] m.p.platform:DEBUG: sys_arch_prctl(0x1002, 0x7ffffffd4f00) = 0
2019-06-24 18:50:14,006: [1120] m.p.platform:DEBUG: sys_mprotect(0x7fffffdab000, 0x4000, 1) = 0
2019-06-24 18:50:15,619: [1120] m.p.platform:DEBUG: sys_mprotect(0x555555754000, 0x1000, 1) = 0
2019-06-24 18:50:17,065: [1120] m.p.platform:DEBUG: sys_mprotect(0x7ffffffdc000, 0x1000, 1) = 0
2019-06-24 18:50:17,094: [1120] m.p.platform:DEBUG: sys_munmap(0x7ffffffd6000, 0x57f6) = 0
2019-06-24 18:50:18,064: [1120] m.p.linux:DEBUG: Opening file /data for real fd 3
2019-06-24 18:50:18,064: [1120] m.p.platform:DEBUG: sys_openat(0xffffff9c, "/data", 0x90800, 0) = 3
2019-06-24 18:50:18,083: [1120] m.p.platform:DEBUG: sys_newfstat(3, "n") = 0
2019-06-24 18:50:18,149: [1120] m.n.c.abstractcpu:WARNING: Unimplemented instruction: 0x00007fffffdcc631:       0f ae 44 24 40  fxsave  [rsp + 0x40]
2019-06-24 18:50:18,459: [1120] m.n.c.abstractcpu:WARNING: Unimplemented instruction: 0x00007fffffdcc646:       0f ae 4c 24 40  fxrstor [rsp + 0x40]
2019-06-24 18:50:33,355: [1120] m.n.c.abstractcpu:WARNING: Unimplemented instruction: 0x00007fffffdcc631:       0f ae 44 24 40  fxsave  [rsp + 0x40]
2019-06-24 18:50:33,596: [1120] m.n.c.abstractcpu:WARNING: Unimplemented instruction: 0x00007fffffdcc646:       0f ae 4c 24 40  fxrstor [rsp + 0x40]
2019-06-24 18:50:33,898: [1120] m.p.platform:DEBUG: sys_brk(0) = 0x555555756000
2019-06-24 18:50:33,916: [1120] m.p.platform:DEBUG: sys_brk(0x555555777000) = 0x555555777000
FILE MODE: users.csv :: 100644
FILE MODE: messages.csv :: 100644
2019-06-24 18:50:34,216: [1120] m.p.platform:DEBUG: sys_getdents(3, "^>6", 0x8000) = 63
2019-06-24 18:50:34,506: [1120] m.p.platform:DEBUG: sys_newfstat(1, 0x7ffffffff620) = 0
2019-06-24 18:50:34,542: [1120] m.p.platform:DEBUG: sys_ioctl(1, 0x5401, 0x7ffffffff580) = -22
2019-06-24 18:50:35,208: [1120] m.p.platform:DEBUG: sys_getdents(3, "^>6", 0x8000) = 0
2019-06-24 18:50:35,734: [1120] m.p.linux:DEBUG: WRITE(1, 0x55555575e2a0, 51) -> <'FAIL DT_REG - users.csv'                       >
2019-06-24 18:50:35,734: [1120] m.p.linux:DEBUG: WRITE(1, 0x55555575e2a0, 51) -> <'FAIL DT_REG - messages.csv'                    >
2019-06-24 18:50:35,734: [1120] m.p.linux:DEBUG: WRITE(1, 0x55555575e2a0, 51) -> <''                                              >
2019-06-24 18:50:35,736: [1120] m.p.platform:DEBUG: sys_write(1, "FAIL DT_REG - users.csv\nFAIL DT_", 51) = 51
```

I'm not quite sure what is happening, but I think our `d_type` is in the wrong location. According to the documentation, [glibc `readdir`](http://man7.org/linux/man-pages/man3/readdir.3.html) looks for `d_type` here
```c
struct dirent {
    ino_t          d_ino;       /* Inode number */
    off_t          d_off;       /* Not an offset; see below */
    unsigned short d_reclen;    /* Length of this record */
    unsigned char  d_type;      /* Type of file; not supported
                                              by all filesystem types */
    char           d_name[256]; /* Null-terminated filename */
};
``` 
but the [`getdents` syscall](http://man7.org/linux/man-pages/man2/getdents.2.html) says it should be at the end.
```c
struct linux_dirent {
    unsigned long  d_ino;     /* Inode number */
    unsigned long  d_off;     /* Offset to next linux_dirent */
    unsigned short d_reclen;  /* Length of this linux_dirent */
    char           d_name[];  /* Filename (null-terminated) */
               /* length is actually (d_reclen - 2 -
               offsetof(struct linux_dirent, d_name)) */
               /*
    char           pad;       // Zero padding byte
    char           d_type;    // File type (only since Linux
                       // 2.6.4); offset is (d_reclen - 1)
               */
}
```
The syscall documentation also says that all programs should handle a `0` (`DT_UNKNOWN`) file type, but I think since most Linux users are using glibc(maybe?) we should implement the glibc version of `dirent`?

I have yet to test changing the location of `d_type`, so I'll report back if I can get it to work.

But for now, I don't think this should be merged.

ehennenfent:
As a general rule, you should follow the conventions established by the Linux kernel when making syscall-related changes to Manticore. Often, the documented behavior for a given system call is in fact the behavior of the _glibc wrapper_ for that system call. The _real_ system call may in fact do something different.

ekilmer:
> As a general rule, you should follow the conventions established by the Linux kernel when making syscall-related changes to Manticore. Often, the documented behavior for a given system call is in fact the behavior of the _glibc wrapper_ for that system call. The _real_ system call may in fact do something different.

Understood 👍 . It seems I was wrong about the order needing to be switched, so I've just ended up filling in the value for the `d_type` at the end of the struct, in place of the padding byte (15d1350), as referenced in the `getdents` documentation

> The d_type field is implemented since Linux 2.6.4.  It occupies a space that was previously a zero-filled padding byte in the linux_dirent structure.

This should be good to go now!

Issue #884, opened 2018-04-30
Add the pathgrind examples to test manticore


ggrieco-tob:
There are more than [20 different small C programs](https://github.com/codelion/pathgrind/tree/master/testcase) to test the input generation in the pathgrind tool (but sadly this project is abandoned!). One particularly interesting example is the re-implementation of the [ms08_067 XP/Vista vulnerability](https://github.com/codelion/pathgrind/blob/master/testcase/ms08_067.c). Our tool should be able to discover such vulnerability exploring all the paths.

PR #1491, opened 2019-07-20
pathgrind test cases added to /examples/pathgrind (issue #884)

yeti-detective:
I added the test cases from the pathgrind repository but I don't know if I am a powerful enough user to make use of them. Let me know if any code is required.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1491)


ggrieco-tob:
At least, we need to:
1. Remove `.exe` files.
2. Test all the them, to make sure they finish in reasonable time and the results are expected.
3. Integrate them in the test scripts.

yeti-detective:
I can analyze with `manticore <file>` all of the test*.c binaries if I compile them to my native architecture x86-64.
I've also tried it with binaries compiled to Intel 80386 (`clang -m32 -o test*_32 test*.c`), which I don't think is correct. It looks like manticore or python is freezing after some unimplemented instruction warnings, the final one seems always to be: `2019-08-11 22:38:41,747: [31212] m.n.c.abstractcpu:WARNING: Unimplemented instruction: 0x00000000f7e50d9a:	f4	hlt`
Analysis on binaries compiled from ms08_067.c seem to running into a path explosion condition & I can't figure out how to constrain that.
Any advice on the constraints for ms08_067.c or the 32 bit binaries is appreciated.

yeti-detective:
Also, I have .pkl files for the 32 bit analyses & they do load with pickle.load() & appear to be Manticore State objects.

ehennenfent:
Hey @yeti-detective, thank you for the PR. I notice there don't seem to be any any `.py` files committed to this PR. Typically, we use the examples section to showcase Python scripts that solve problems using Manticore. As the pathgrind test cases can be downloaded from their own repository, there's no need to include a copy of them here unless you're also including a script that showcases a particular Manticore-based analysis over them. 

Also, we try to keep binary files of this size out of the main repository to avoid slowing down checkout times. This PR would probably be better suited for the
[manticore-examples repo](https://github.com/trailofbits/manticore-examples).

yeti-detective:
@ehennenfent thank you for the feedback. I agree the binaries should not be included as they can be compiled from the source files. 

What I think would best satisfy the original posted [issue](https://github.com/trailofbits/manticore/issues/884) is a script that could reveal the ms08_67 vulnerability. However, the main method of the ms08_67 binary doesn't take any inputs, and I haven't been able to figure out how to properly constrain that analysis.

disconnect3d:
> However, the main method of the ms08_67 binary doesn't take any inputs, and I haven't been able to figure out how to properly constrain that analysis.

Are we speaking about  https://github.com/codelion/pathgrind/blob/master/testcase/ms08_067.c#L172? If so, it seems it takes the input from stdin, so it should be possible to constrain it?

disconnect3d:
Also it might be more feasible to close this PR and move the discussion to https://github.com/trailofbits/manticore/issues/884.

ehennenfent:
Archiving this in favor of keeping discussion on #884

PR #1495, opened 2019-07-25
WASM Support

ehennenfent:
[WIP]

At present, this adds a skeleton for WASM support. It's capable of parsing, instantiating, and executing the following WASM module:
```
int main(){
  return 42;
}
```
Most of the instructions still need to be implemented, and it doesn't support any symbolic data yet. I also need to work on the control flow, as WASM expects entire expressions (sequences of instructions) to be executed as units and needs some adaptation to work with Manticore's one-instruction-at-a-time model.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1495)


manticore/platforms/wasm.py
```diff
@@ -0,0 +1,90 @@
...
+            # TODO - create function stubs that have the correct signatures
+            self.store.funcs.append(
+                HostFunc(self.module.types[i.desc], import_dict.get(i.name, stub))
+            )
+            imports.append(FuncAddr(len(self.store.funcs) - 1))
+        self.instance.instantiate(self.store, self.module, imports)
+        self.instantiated = True
+
+    def invoke(self, name="main", argv=[]):
+        self.instance.invoke_by_name(name, self.stack, self.store, list(argv))
```
disconnect3d: Huh, `list(agv)` can be nice way of stopping the potential 'default mutable value' poblem :)

manticore/platforms/wasm.py
```diff
@@ -0,0 +1,90 @@
...
+        rets = 0
+        for export in self.instance.exports:
+            if export.name == funcname and isinstance(export.value, FuncAddr):
+                rets = len(self.store.funcs[export.value].type.result_types)
+
+        while self.instance.exec_instruction(self.store, self.stack):
+            pass
+        if rets == 0:
+            return []
+        if rets == 1:
```
disconnect3d:
```suggestion
        elif rets == 1:
```

manticore/platforms/wasm.py
```diff
@@ -0,0 +1,90 @@
...
+            return [self.stack.pop()]
+        else:
+            return [self.stack.pop() for _i in range(rets)]
```
disconnect3d:
```suggestion
            return [self.stack.pop() for _ in range(rets)]
```

could also be:
```python
result, self.stack = self.stack[rets:], self.stack[:rets]
return result
```

manticore/platforms/wasm.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def execute(self):
+        """
+        """
```
disconnect3d: Missing docstring

manticore/platforms/wasm.py
```diff
@@ -0,0 +1,90 @@
...
+        # Handle interrupts et al in a try/except here
+        # self.current.execute()
+        if not self.instantiated:
+            raise RuntimeError("Trying to execute before instantiation!")
+        if not self.instance.exec_instruction(self.store, self.stack):
+            ret = None
+            if not self.stack.empty():
+                ret = self.stack.pop()
+                print("WASM Execution returned", ret)
+            raise TerminateState(f"Execution returned {ret}")
```
disconnect3d: Random thought: do we need  to produce the string here? It is probably not used that often

It would probably be better to pass `ret` into `TerminateState` and construct the string only when it is ended (via `__str__`?).

If we wanted different msgs bettween engines this could be a `WasmTerminateState` with a different `__str__`.

manticore/wasm/cli.py
```diff
@@ -0,0 +1,19 @@
+from .manticore import ManticoreWASM
+
+
+def wasm_main(args, _logger):
+    env = {key: val for key, val in [env[0].split("=") for env in args.env]}
```
disconnect3d: Are those environment variables or WASM env like imported functions? I guess this should be the latter - if so - why do we parse `key=value` string here? I mean, it could be passed as a dict instead?

disconnect3d: if this is one or the another: it is probably good to add a comment/docstring about it here or/and in `ManticoreWASM`.

disconnect3d: It seems this are envvars - hmm after rethinkinig it: maybe some wasm engines supports them (e.g. via `getenv`) so it might be reasonable to keep them; though it would probably be nice to explicitly add them to `ManticoreWASM.__init__` and document why/what is the reasoning beyond accepting them (imho it is not be that obvious straight away?).

manticore/wasm/manticore.py
```diff
@@ -0,0 +1,80 @@
...
+from .state import State
+from ..core.manticore import ManticoreBase
+from ..core.smtlib import ConstraintSet
+
+logger = logging.getLogger(__name__)
+
+
+class ManticoreWASM(ManticoreBase):
+    """
+    """
```
disconnect3d: Docstring

manticore/wasm/module_structure.py
```diff
@@ -0,0 +1,280 @@
...
+class Memory:
+    type: MemoryType
+
+    def allocate(self, store: Store, mem_type: MemoryType) -> MemAddr:
+        a = MemAddr(len(store.mems))
+        store.mems.append(
+            MemInst(
+                [
+                    0x0 for _i in range(mem_type.min * (64 * 1024))
+                ],  # TODO - these should be symbolic, right?
```
disconnect3d: TODO: yes;  them, envvars (if wasm engines use them) and stdin

disconnect3d: Also: just `[0x0] * mem_type.min * (64 * 1024)`

Btw why `64*1024`?

manticore/wasm/module_structure.py
```diff
@@ -0,0 +1,280 @@
...
+    desc: ImportDesc
+
+
+@dataclass
+class Export:
+    name: Name
+    desc: ExportDesc
+
+
+class Module:
```
disconnect3d: Random thought: maybe this should just inherit from a dict...

manticore/wasm/state.py
```diff
@@ -0,0 +1,6 @@
+from ..core.state import StateBase
+
+
+class State(StateBase):
+    def execute(self):
+        return self._platform.execute()
```
disconnect3d: Loud thinking: might be worth to move it to `StateBase` and just inherit from it here? It is the same in ethereum's `State` and native has a more extended impl

manticore/wasm/types.py
```diff
@@ -0,0 +1,226 @@
...
+from ..core.smtlib import issymbolic, BitVec
+import wasm
+
+# TODO - These need to be symbolic fixed-size representations
+U32: type = type("U32", (int,), {})
+U64: type = type("U64", (int,), {})
+Byte: type = type("Byte", (int,), {})
+
+
+class I32(int):
```
disconnect3d: Is there any good reason for those to be classes instead of functions?

Also, we use `classmethod` here without using the `cls` argument so if we keep it as class, it could be `staticmethod`s.

manticore/wasm/types.py
```diff
@@ -0,0 +1,252 @@
...
+    @classmethod
+    def cast(cls, other):
+        if issymbolic(other):
+            return other
+        return F32(other)
+
+
+class F64(float):
+    def __new__(cls, val):
+        if int==type(val):
```
disconnect3d:
```suggestion
        if type(val) is int:
```

or `isinstance`.

disconnect3d: Same above

ehennenfent:
I'm going to promote this to "Ready to Review", as it's now passing both the symbolic and concrete tests. It's not completely done, but there's a lot to look at, and almost all of what's currently there is camera-ready.

A (questionably formatted) set of docs can be found on [RTD](https://manticore.readthedocs.io/en/dev-wasm/wasm.html#), but looking at the source will probably be easier. 

The major things that still need to get taken care of:

 - [ ] Find a workaround for floating point support. Felipe suggested compiling an integer floating point library to WASM, which seems promising.
 - [ ] Dump similarly-meaningful testcases to those produced by ManticoreNative or ManticoreEVM
 - [ ] Add integration tests for working with imported functions.
 - [ ] Publish events throughout WASM execution
 - [ ] Backport Py3.7-dependent code to Py3.6
 - [ ] Investigate optimization opportunities, especially surrounding the atomic stack

manticore/wasm/types.py
ehennenfent: Yes, the reason here being that the types of the values are used frequently for validation

manticore/wasm/module_structure.py
ehennenfent: 64*1024 is the standard WASM memory page size

manticore/wasm/cli.py
ehennenfent: This was a mistake, at the time. I use `env` to pass in imported functions, but this was still using the code for passing in environment variables on the command line. At present, there's no way to specify imported functions via the CLI, since I'm not sure how that would work.

manticore/platforms/wasm.py
ehennenfent: The exception string gets saved into a file when you call `finalize`, so I think it's worth keeping

manticore/wasm/types.py
```diff
@@ -0,0 +1,472 @@
...
+        :param other: Value to convert to F64
+        :return: If other is symbolic, other. Otherwise, F64(other)
+        """
+        if issymbolic(other):
+            return other
+        return cls(other)
+
+
+ValType = type  #: https://www.w3.org/TR/wasm-core-1/#syntax-valtype
+Value_t: typing.Tuple[typing.Type] = (I32, I64, F32, F64, BitVec)  # type: ignore
```
ekilmer: More description of type ignore here also

manticore/platforms/wasm.py
```diff
@@ -0,0 +1,363 @@
...
+                    self.import_module(
+                        i.module, exec_start, stub_missing
+                    )  # TODO - prevent circular imports
+            # If it's registered, but hasn't been imported yet, import it
+            elif not self.instantiated[self.module_names[i.module]]:
+                self.import_module(i.module, exec_start, stub_missing)
+            imported_version = self._get_export_addr(i.name, i.module)
+            if imported_version is None:
+                # check for manually provided version
+                imported_version = self.get_export(i.name, i.module)  # type: ignore
```
ekilmer: More description of type ignore here also

manticore/platforms/wasm.py
```diff
@@ -0,0 +1,363 @@
...
+            else:
+                if isinstance(i.desc, TypeIdx):  # Imported function
+                    func_type = module.types[i.desc]
+                    if i.module == "env" and imported_version:
+                        imported_version = HostFunc(func_type, imported_version)
+                    self.store.funcs.append(
+                        imported_version
+                        if imported_version
+                        else HostFunc(
+                            func_type, partial(stub, len(func_type.result_types))  # type: ignore
```
ekilmer: More description of type ignore here also

manticore/platforms/wasm.py
```diff
@@ -0,0 +1,363 @@
...
+        """
+        Prepares the underlying ModuleInstance for execution. Calls import_module under the hood, so this is probably
+        the only import-y function you ever need to call externally.
+
+        TODO: stubbed imports should be symbolic
+
+        :param env_import_dict: Dict mapping strings to functions. Functions should accept the current ConstraintSet as the first argument.
+        :param exec_start: Whether or not to automatically execute the `start` function, if it is set.
+        :return: None
+        """
```
ekilmer: Not sure if there is a way to automatically check for unsynchronized docs with the params in the docs vs in the actual code, but the docs are missing `supplemental_env` and `stub_missing`

manticore/utils/log.py
```diff
@@ -100,11 +100,12 @@ def get_levels():
             ("manticore.platforms.*", logging.DEBUG),
             ("manticore.ethereum", logging.DEBUG),
             ("manticore.core.plugin", logging.DEBUG),
+            ("manticore.wasm.*", logging.INFO),
```
ekilmer: Should there be something about wasm in the first level, labeled `#1` above?

readthedocs.yml
```diff
@@ -2,4 +2,4 @@ build:
     image: latest
 
 python:
-    version: 3.6
+    version: 3.7
```
ekilmer: Should this actually be Python 3.7? I don't know much about how this file is used

manticore/wasm/structure.py
```diff
@@ -0,0 +1,1703 @@
...
+                    )
+            elif sec_id == SEC_IMPORT:  # https://www.w3.org/TR/wasm-core-1/#import-section%E2%91%A0
+                for i in section_data.payload.entries:
+                    ty_map = i.get_decoder_meta()["types"]
+                    mod_name = strip_quotes(ty_map["module_str"].to_string(i.module_str))
+                    field_name = strip_quotes(ty_map["field_str"].to_string(i.field_str))
+                    if i.kind == 0:
+                        m.imports.append(Import(mod_name, field_name, TypeIdx(i.type.type)))
+
+                    elif i.kind == 1:
```
ekilmer: Having a literal `1` here feels weird, but the specification does it too, so this is probably fine.

manticore/wasm/structure.py
```diff
@@ -0,0 +1,1703 @@
...
+        self.funcs = state["funcs"]
+        self.tables = state["tables"]
+        self.mems = state["mems"]
+        self.globals = state["globals"]
+
+
+def _eval_maybe_symbolic(constraints, expression) -> bool:
+    if issymbolic(expression):
+        return solver.must_be_true(constraints, expression)
+    return True if expression else False
```
ekilmer: Could this be written as
```
return expression is not None
```
I'm also not quite sure what we're checking for in `expression`

ehennenfent: I think this is a typing thing. `expression` isn't explicitly constrained to `typing.Union[Bitvec, bool]` because mypy can't figure out that `issymbolic` will only be true for `BitVec`s, so to confirm that this will always return a bool, I do `return True if expression else False` instead of just `return expression`

manticore/utils/log.py
ehennenfent: By default, that'll be `logging.warning`, which should be fine. I use logging.info for printing WASM instructions and logging.debug for printing stack operations, so there's not a lot of middle ground for purely informational messages. I guess I could maybe make a logging.verbose level?

manticore/platforms/wasm.py
ehennenfent: Pycharm will do this if you go to Code -> Inspect Code. Fixed most of these, except for some cases where I didn't bother documenting `store` and `stack`

readthedocs.yml
ehennenfent: This is the version of the Python interpreter RTD uses when building the sphinx docs. It's probably fine either way. Since Py3.7 should be backwards compatible, I'm in favor of leaving it because we always seem to forget to update the RTD files whenever we make other changes.

manticore/wasm/structure.py
```diff
@@ -0,0 +1,1703 @@
...
+        assert len(ty.param_types) == len(
+            argv
+        ), f"Function {funcaddr} expects {len(ty.param_types)} arguments"
+        # for t, v in zip(ty.param_types, argv):
+        #     assert type(v) == type(t)
+
+        # Create dummy frame, which is required by the spec but doesn't serve any obvious purpose. It also never
+        # gets popped.
+        dummy_frame = Frame([], ModuleInstance())
+        stack.push(dummy_frame)  # type: ignore
```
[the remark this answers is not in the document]
ehennenfent: Due to blackening I typically wasn't able to put it on the same line, but I've added annotating comments on the previous line from each `type: ignore`.

It may become problematic to have mypy, which partially relies on comments, fighting with `black`, which treats comments as unimportant parts of the AST it can move to wherever.

manticore/utils/log.py
ekilmer: `logging.warning` is fine. I was just wondering because there seemed to be a pattern with enabling `logging.info` in level 1 for the different languages
```
        # 1
        [...
         ("manticore.ethereum.*", logging.INFO),
         ("manticore.native.*", logging.INFO),
         ...
         ]
```

manticore/wasm/structure.py
ekilmer: Got it. That makes sense if `expression` will always be a `bool` at this point and this is just to make mypy happy 👍

manticore/wasm/structure.py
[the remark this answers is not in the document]
ekilmer: :( Okay, yeah a comment above is just as good! Thank you!!

PR #1497, opened 2019-07-29
Replaced mkstmp with TemporaryFile.

MrSynAckSter:
This is a draft pull request to trigger CI tests.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1497)

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -2977,7 +2977,7 @@ class SLinux(Linux):
         if symbolic_path:
             import tempfile
 
-            fd, path = tempfile.mkstemp()
+            fd, path = tempfile.TemporaryFile()
             with open(path, "wb+") as f:
                 f.write("+" * 64)
             self.symbolic_files.append(path)


ehennenfent:
Hmm, I think it's a legitimate concern that the file will get garbage collected. I could be wrong about this, but I'm guessing that since we only keep a reference to the path (not the file pointer) Python will think it's safe to garbage collect the file. 

Fortunately, it shouldn't be too difficult to save the file pointers in a dict stored on the object, and remove the appropriate one whenever `_close` is called.

disconnect3d:
> This is a draft pull request to trigger CI tests.

So do we care about it/are we really thinking about this change? :P

---

Just to be clear, via https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp:
> Unlike TemporaryFile(), the user of mkstemp() is responsible for deleting the temporary file when done with it.

And the question is: do we want to keep the file or do we want it to be removed after its not needed?

> [...] I'm guessing that since we only keep a reference to the path (not the file pointer) Python will think it's safe to garbage collect the file. Perhaps @disconnect3d can weigh in?

The code as of now is not correct. It also seems that path is not tested via our tests as they  passed; I hope manticore-examples repo would fail on that :F.

Anyway, the `tempfile.TemporaryFile` returns a file object and not a `(fd, path)` as `tempfile.mkstemp` do.

disconnect3d:
Reopen if needed.

PR #1498, opened 2019-07-30
Add SHL/SHR/SAR

### LINKED ISSUE #1166: EVM - implement EIP 145: Bitwise shifting instructions
We need to add support for bitwise shifting instructions added in Constantinople fork.

More detailed information can be found here: https://eips.ethereum.org/EIPS/eip-145

Things that needs to be done:
* [ ] Add tests from EIP [1]
* [ ] Think of more cases the EIP doesn't specify and add them to the test suite.
* [ ] Add concrete execution implementation, make the symbolic parts fail (don't test this failures) [2].
* [ ] Add symbolic execution implementation [3]

[1] The tests should be minimal, e.g. doing only the necessary operations like creating a super small contract from bytecode that uses given instruction. The stuff should also be documented well: the instruction implementation should have a link to EIP in its docstring, when creating contract from bytecode we should have some comments describing what instructions are there. So in the end if someone else looks at the code later on they will immediately know what is going on.
[2] The symbolic part failure should be done by checking `issymbolic(argument)`, ofc only for the parts that can be symbolic.
[3] Please keep in mind that there are certain things that appears to be concrete all the time but **there are cases in which they might not be**. One of the examples can be `EVM.pc`. When we apply a `IntegerOverflowDetector`, the value of pc might get tainted (tainting means marking values so if other values depend on them, they are also marked and so we can detect how certain variables depend on each other) and so changed to `Constant`; we had a bug related to this that was fixed in https://github.com/trailofbits/manticore/pull/1347.

**The concrete and symbolic parts should be done in separate PRs.**

ehennenfent:
Closes #1166 
Adds instruction semantics and tests from [EIP-145](https://eips.ethereum.org/EIPS/eip-145).

Does not update for the Constantinople gas calculation semantics, or, for that matter, most of the other Constantinople changes. Further work will be handled in later PR's; the motivation behind this merge is to get the shift instructions into master sooner rather than later. 

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1498)

### DIFF
--- a/tests/auto_generators/make_VMTests.py
+++ b/tests/auto_generators/make_VMTests.py
@@ -375,7 +375,9 @@ class EVMTest_{os.path.splitext(os.path.basename(filename_or_folder))[0]}(unitte
                 continue
 
             filename = os.path.join(folder, filename)
-            testcase = dict(json.loads(open(filename).read()))
+            fp = open(filename)
+            testcase = dict(json.loads(fp.read()))
+            fp.close()
             for name, testcase in testcase.items():
                 output += (
                     gen_test(testcase, filename, symbolic=symbolic, skip="Performance" in filename)
@@ -387,7 +389,9 @@ class EVMTest_{os.path.splitext(os.path.basename(filename_or_folder))[0]}(unitte
     else:
         folder = None
         filename = os.path.abspath(filename_or_folder)
-        testcase = dict(json.loads(open(filename).read()))
+        fp = open(filename)
+        testcase = dict(json.loads(fp.read()))
+        fp.close()
         for name, testcase in testcase.items():
             output += (
                 gen_test(testcase, filename, symbolic=symbolic, skip="Performance" in filename)
--- a/tests/ethereum/EVM/test_EVMSHIFT.py
+++ b/tests/ethereum/EVM/test_EVMSHIFT.py
@@ -1,0 +1,885 @@
+import struct
+import unittest
+import json
+from manticore.platforms import evm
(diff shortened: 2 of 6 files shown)


tests/ethereum/EVM/test_EVMSHIFT.py
```diff
@@ -0,0 +1,885 @@
...
+        world = evm.EVMWorld(constraints)
+
+        address = 0x222222222222222222222222222222222222200
+        caller = 0x111111111111111111111111111111111111100
+        value = 10000
+        bytecode = b"\x1b"
+        data = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+        gas = 1000000
+
+        new_vm = evm.EVM(constraints, address, data, caller, value, bytecode, gas=gas, world=world)
```
feliam: Eventually... we'll need a way to specify which fork are we testing. This opcode means a different thing in previous forks. Yes, ETH changed the meaning of the words that are used to write contracts. Smart.

tests/auto_generators/make_VMTests.py
```diff
@@ -387,7 +389,9 @@ def _solve(self, constraints, val):
     else:
         folder = None
         filename = os.path.abspath(filename_or_folder)
-        testcase = dict(json.loads(open(filename).read()))
+        fp = open(filename)
```
feliam:
```with open(filename) as fp: 
    :man_shrugging: 
```

PR #1506, opened 2019-08-20
Improve smemory tests

disconnect3d:
Improves SMemory tests and adds more docs to it.

This is going to help debug, fix and either merge or close #1465.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1506)


disconnect3d:
The test passes for me locally, so I think we can merge.

PR #1507, opened 2019-08-20
Added storage change detection plugin

feliam:
This plugin will prevent the exploration of states that did not changed the state in the previous TX.

Transactions that did not change the storage are considered a final state and are not further explored.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1507)


manticore/ethereum/plugins.py
```diff
@@ -142,3 +142,73 @@ class VerboseTraceStdout(Plugin):
...
+
+
+class KeepOnlyIfStorageChanges(Plugin):
+    """ TODO: reword
+        This state will ignore states that are the result of executing a
+        transaction that did not write to the storage.
+
+        When this plugin is enabled transactions that wont write to the storage
+        are considered not to change the evm world state and hence ignored as a
+        starting point for the following human transaction.
```
disconnect3d: Maybe:
```
This plugin discards all transactions that results in states where 
the underlying EVM storage did not change or in other words,
there were no writes to it.

This allows to speed-up EVM engine exploration as we don't
explore states that have the same storage (contract data).

However, keep in mind that if the (contract) code relies on
account balance and the balance is not a symbolic value
it might be that a certain state will not be covered by the
execution when this plugin is used.
```

manticore/ethereum/plugins.py
```diff
@@ -142,3 +142,73 @@ class VerboseTraceStdout(Plugin):
...
+    """
+
+    def did_open_transaction_callback(self, state, tx, *args):
+        """ We need a stack. Each tx (internal or not) starts with a "False" flag
+            denoting that it did not write anything to the storage
+        """
+        state.context["written"].append(False)
```
disconnect3d: Hmmm... how does it differ from `self.locked_context(string) as ctx`? That would be a global state and here we use a context local/tied to the state?

manticore/ethereum/plugins.py
```diff
@@ -142,3 +142,73 @@ class VerboseTraceStdout(Plugin):
...
+                flag = flag or ether_sent
+            state.context["written"][-1] = state.context["written"][-1] or flag
+
+    def did_evm_write_storage_callback(self, state, *args):
+        """ Turn on the corresponding flag is the storage has been modified.
+        Note: subject to change if the current transaction is reverted"""
+        state.context["written"][-1] = True
+
+    def will_run_callback(self, *args):
+        """Initialize the flag stack at each human tx/run()"""
```
disconnect3d: Are `ready_states` == `human tx`?

manticore/ethereum/plugins.py
```diff
@@ -142,3 +142,73 @@ class VerboseTraceStdout(Plugin):
...
+        for st in self.manticore.ready_states:
+            st.context["written"] = [False]
+
+    def did_run_callback(self):
+        """When  human tx/run just ended remove the states that have not changed
+         the storage"""
+        with self.manticore.locked_context("ethereum.saved_states", list) as saved_states:
+            # Normallyy the ready states are consumed and forked, eth save the
```
disconnect3d:
```suggestion
            # Normally the ready states are consumed and forked, eth save the
```

manticore/ethereum/plugins.py
```diff
@@ -142,3 +142,73 @@ class VerboseTraceStdout(Plugin):
...
+            # be ignored.
+            for state_id in list(saved_states):
+                st = self.manticore._load(state_id)
+                if not st.context["written"][-1]:
+                    self.manticore._terminated_states.append(st.id)
+                    saved_states.remove(st.id)
+
+    def generate_testcase(self, state, testcase, message):
+        with testcase.open_stream("summary") as stream:
+            if not state.context.get("written", (False,))[-1]:
```
disconnect3d: Shall this ever happen that `written` won't be there? When/why?

tests/ethereum/test_general.py
```diff
@@ -1001,6 +1001,7 @@ def did_evm_execute_instruction_callback(self, state, instruction, arguments, re
                         with self.locked_context(name) as d:
                             d.append(instruction.pc)
                 except Exception as e:
+                    print(e)
```
disconnect3d: Probably remove?

PR #1509, opened 2019-08-21
Improve performance of bulk read/writes to memory

disconnect3d:
This is a better version of https://github.com/trailofbits/manticore/pull/1465.

This PR improves performance of `SMemory` concrete writes so we write concrete bytes in chunks instead of one by one.

Unlike #1465, this PR does not introduce `write1` which was a specialized `write` for single concrete bytes write.

I benchmarked master vs this PR vs #1465 via this code:
```python
import time
def test_write_concrete():
    from manticore.core.smtlib import ConstraintSet
    from manticore.native.memory import SMemory32

    string = 'a' * 0x1000

    mem = SMemory32(ConstraintSet())
    buf = mem.mmap(None, 0x1000, "rw")

    t1 = time.time()
    for i in range(1000):
        mem[buf:buf+0x1000] = string
    t2 = time.time()

    print(f"test_write_concrete: {t2-t1} seconds")

    assert mem[buf:buf+0x1000] == [b'a'] * 0x1000

test_write_concrete()
```

Results TLDR - master 54s, this PR 3.1s, #1465 2.8s.

image

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1509)

### DIFF
--- a/manticore/native/memory.py
+++ b/manticore/native/memory.py
@@ -1180,18 +1180,41 @@ class SMemory(Memory):
                     self._symbols.setdefault(base + offset, []).append((condition, value[offset]))
         else:
 
+            # Symbolic writes are stored in self._symbols and concrete writes are offloaded to super().write(...)
+            #
+            # Symbolic values are written one by one and concrete are written in chunks
+            # To do this, the logic below tracks start index of the part where concrete values are
+            # and commits/triggers writes of those concrete values when a symbolic value is hit or after the loop
+            #
+            concrete_start = None
+
             for offset in range(size):
                 ea = address + offset
 
                 if issymbolic(value[offset]):
+                    # Commit (offload) concrete write
+                    if concrete_start is not None:
+                        super().write(address+concrete_start, value[concrete_start:offset], force)
+                        concrete_start = None
+
+                    # Commit symbolic write if access is okay
                     if not self.access_ok(ea, "w", force):
                         raise InvalidMemoryAccess(ea, "w")
                     self._symbols[ea] = [(True, value[offset])]
+
                 else:
-                    # overwrite all previous items
+                    # For concrete value, remove symbolic chunks that were stored there
+                    # (Note: there might be many because of conditions later leading to different program paths)
                     if ea in self._symbols:
                         del self._symbols[ea]
-                    super().write(ea, [value[offset]], force)
+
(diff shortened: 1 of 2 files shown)


ehennenfent:
Have you tried this with BitVectors longer than 8 bits? I _think_ it should extract the individual symbolic bytes and write them in the correct places, but we should add an explicit test for it if possible.

disconnect3d:
> Have you tried this with BitVectors longer than 8 bits? I think it should extract the individual symbolic bytes and write them in the correct places, but we should add an explicit test for it if possible.

Yes and it does not work. I also think it is irrelevant to the changes made in this PR, though I haven't tested it deeply (e.g. doing the same tests on master branch).

I looked over this via a simple test:
```python
    def test_set_symbolic_bigbyte(self):
        cs = ConstraintSet()
        mem = SMemory32(cs)
        addr = mem.mmap(None, 0x100, "rwx")

        mem[addr] = cs.new_bitvec(32)
        x = mem[addr]
        pass
```

It turns out that we do not enforce a given symbolic memory byte to be of maximum or only 8-bit size :/.

This code also fails on `x = mem[addr]` but this is because:
- the `mem[addr] = cs.new_bitvec(32)` saves the BitVec32 to `self._symbols[addr]`
- when we do `mem[addr]` it executes this code: https://github.com/trailofbits/manticore/blob/0.3.1/manticore/native/memory.py#L1143-L1153
- the `result = list(map(Operators.ORD, super().read(address, size, force)))` there reads the memory and converts all items using `Operators.ORD`
- one of those items is the BitVec32 and `ORD` converts it to `BitVecExtract(s, 0, 7)` truncating it to 7 (!!) bits (*it shall be 8 here*): https://github.com/trailofbits/manticore/blob/0.3.1/manticore/core/smtlib/operators.py#L22
- then, in the memory read code we do: `return list(map(Operators.CHR, result))`
- the `CHR` (https://github.com/trailofbits/manticore/blob/0.3.1/manticore/core/smtlib/operators.py#L34) then tries to do `BitVecExtract(s, 0, 8)` on `s` here is the `BitVecExtract` with `.size=7`. The extract can't be used for extending bit vectors and so an assertion fails here.

Side note: there is no `BitVec32` class, it's just a `BitVec<Whatever>` with a `.size=32`.

PR #1511, opened 2019-08-22
Fixes ConstraintSet.new_bitvec size check

disconnect3d:
Previously the `new_bitvec` checked for `size == 1 or size % 8 == 0`.

This check allowed to create a bitvec with negative size, as in `cs.new_bitvec(-8)`.

It was also too constrained, as we might have cases where we want to constraint ourselves to e.g. 5 bits out of 8 bits of a byte). Sure, this can be done via setting a constrain over an 8-bit (1 Byte) bitvector, but it might be not the most efficient solution(?).

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1511)

### DIFF
--- a/manticore/core/smtlib/constraints.py
+++ b/manticore/core/smtlib/constraints.py
@@ -357,7 +357,7 @@ class ConstraintSet:
             :return: a fresh BitVecVariable
         """
         if size <= 0:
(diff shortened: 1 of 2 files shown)


disconnect3d:
Leaving it here to see if any tests fails. I am also going to add a testcase for this specific situation.

PR #1512, opened 2019-08-22
Fix operators.ORD for BitVecs of size > 8

disconnect3d:
TLDR: `ORD(s)` should return something that has 8 bits instead of 7, which it incorrectly did for BitVecs with size>8.

Note that the `ORD` still does not support `BitVec`s with size<8, but I am not sure if we need that support for now. To be clear, such situation ends up as an assertion failure that is being done in `BitVecExtract`, so when assertions are gone (for optimized launches) we might want to extend `ORD` to support it.

---
This change is [Reviewable](https://reviewable.io/reviews/trailofbits/manticore/1512)

### DIFF
--- a/manticore/core/smtlib/operators.py
+++ b/manticore/core/smtlib/operators.py
@@ -19,7 +19,7 @@ def ORD(s):
         if s.size == 8:
             return s
         else:
-            return BitVecExtract(s, 0, 7)
+            return BitVecExtract(s, 0, 8)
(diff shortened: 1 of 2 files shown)


disconnect3d:
I will probably add a regression test for this case.

disconnect3d:
Added a testcase for this. The test, when run on master branch, fails like this:

image

And passes on this PR.

PR #1514, opened 2019-08-26
Add partial implementation of recvfrom syscall

ekilmer:
This partial implementation does not handle src_addr and addrlen
arguments, which means this recvfrom acts like a recv syscall

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -2117,21 +2117,21 @@ class Linux(Platform):
 
     def sys_recv(self, sockfd, buf, count, flags, trace_str="_recv"):
         data: bytes = bytes()
-        if count != 0:
-            if buf not in self.current.memory or (buf + count) not in self.current.memory:
-                logger.info("buf or (buf + count) points to invalid address. Returning EFAULT")
-                return -errno.EFAULT
+        if buf not in self.current.memory or (buf + count) not in self.current.memory:
+            logger.info("buf or (buf + count) points to invalid address. Returning EFAULT")
+            return -errno.EFAULT
 
-            try:
-                sock = self._get_fd(sockfd)
-            except FdError:
-                return -errno.EINVAL
+        try:
+            sock = self._get_fd(sockfd)
+        except FdError:
+            return -errno.EBADF
 
-            if not isinstance(sock, Socket):
-                return -errno.ENOTSOCK
-            data = sock.read(count)
-            self.syscall_trace.append((trace_str, sockfd, data))
-            self.current.write_bytes(buf, data)
+        if not isinstance(sock, Socket):
+            return -errno.ENOTSOCK
+
+        data = sock.read(count)
+        self.syscall_trace.append((trace_str, sockfd, data))
+        self.current.write_bytes(buf, data)
 
         return len(data)
 
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -2115,18 +2115,23 @@ class Linux(Platform):
         fd = self._open(sock)
         return fd
 
-    def sys_recv(self, sockfd, buf, count, flags):
-        try:
-            sock = self.files[sockfd]
-        except IndexError:
-            return -errno.EINVAL
+    def sys_recv(self, sockfd, buf, count, flags, trace_str="_recv"):
+        data: bytes = bytes()
+        if count != 0:
+            if buf not in self.current.memory:
+                logger.info("buf points to invalid address. Returning EFAULT")
+                return -errno.EFAULT
 
-        if not isinstance(sock, Socket):
-            return -errno.ENOTSOCK
+            try:
+                sock = self._get_fd(sockfd)
+            except FdError:
+                return -errno.EINVAL
 
-        data = sock.read(count)
-        self.current.write_bytes(buf, data)
-        self.syscall_trace.append(("_recv", sockfd, data))
+            if not isinstance(sock, Socket):
+                return -errno.ENOTSOCK
+            data = sock.read(count)
+            self.syscall_trace.append((trace_str, sockfd, data))
+            self.current.write_bytes(buf, data)
 
         return len(data)
 
@@ -2138,19 +2143,7 @@ class Linux(Platform):
             logger.warning("sys_recvfrom: Unimplemented non-NULL addrlen")
 
         # TODO Unimplemented src_addr and addrlen, so act like sys_recv
-        try:
-            sock = self.files[sockfd]
-        except IndexError:
-            return -errno.EINVAL
-
-        if not isinstance(sock, Socket):
-            return -errno.ENOTSOCK
-
-        data = sock.read(count)
-        self.current.write_bytes(buf, data)
-        self.syscall_trace.append(("_recvfrom", sockfd, data))
-
-        return len(data)
+        return self.sys_recv(sockfd, buf, count, flags, trace_str="_recvfrom")
 
     def sys_send(self, sockfd, buf, count, flags):
         try:
@@ -2962,7 +2955,7 @@ class SLinux(Linux):
 
         return super().sys_write(fd, buf, count)
 
-    def sys_recv(self, sockfd, buf, count, flags):
+    def sys_recv(self, sockfd, buf, count, flags, trace_str='_recv'):
         if issymbolic(sockfd):
             logger.debug("Ask to read from a symbolic file descriptor!!")
             raise ConcretizeArgument(self, 0)
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -2118,8 +2118,8 @@ class Linux(Platform):
     def sys_recv(self, sockfd, buf, count, flags, trace_str="_recv"):
         data: bytes = bytes()
         if count != 0:
-            if buf not in self.current.memory:
-                logger.info("buf points to invalid address. Returning EFAULT")
+            if buf not in self.current.memory or (buf + count) not in self.current.memory:
+                logger.info("buf or (buf + count) points to invalid address. Returning EFAULT")
                 return -errno.EFAULT
 
             try:
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -2130,6 +2130,28 @@ class Linux(Platform):
 
         return len(data)
 
+    def sys_recvfrom(self, sockfd, buf, count, flags, src_addr, addrlen):
+        if src_addr != 0:
+            logger.warning("sys_recvfrom: Unimplemented non-NULL src_addr")
+
+        if addrlen != 0:
+            logger.warning("sys_recvfrom: Unimplemented non-NULL addrlen")
+
+        # TODO Unimplemented src_addr and addrlen, so act like sys_recv
+        try:
+            sock = self.files[sockfd]
+        except IndexError:
+            return -errno.EINVAL
+
+        if not isinstance(sock, Socket):
+            return -errno.ENOTSOCK
+
+        data = sock.read(count)
+        self.current.write_bytes(buf, data)
+        self.syscall_trace.append(("_recvfrom", sockfd, data))
+
+        return len(data)
+
     def sys_send(self, sockfd, buf, count, flags):
         try:
             sock = self.files[sockfd]
@@ -2959,6 +2981,33 @@ class SLinux(Linux):
 
         return super().sys_recv(sockfd, buf, count, flags)
 
+    def sys_recvfrom(self, sockfd, buf, count, flags, src_addr, addrlen):
+        if issymbolic(sockfd):
+            logger.debug("Ask to read from a symbolic file descriptor!!")
+            raise ConcretizeArgument(self, 0)
+
+        if issymbolic(buf):
+            logger.debug("Ask to read to a symbolic buffer")
+            raise ConcretizeArgument(self, 1)
+
+        if issymbolic(count):
+            logger.debug("Ask to read a symbolic number of bytes ")
+            raise ConcretizeArgument(self, 2)
+
+        if issymbolic(flags):
+            logger.debug("Submitted a symbolic flags")
+            raise ConcretizeArgument(self, 3)
+
+        if issymbolic(src_addr):
+            logger.debug("Submitted a symbolic source address")
+            raise ConcretizeArgument(self, 4)
+
+        if issymbolic(addrlen):
+            logger.debug("Submitted a symbolic address length")
+            raise ConcretizeArgument(self, 5)
+
+        return super().sys_recvfrom(sockfd, buf, count, flags, src_addr, addrlen)
+
     def sys_accept(self, sockfd, addr, addrlen):
         # TODO(yan): Transmit some symbolic bytes as soon as we start.
         # Remove this hack once no longer needed.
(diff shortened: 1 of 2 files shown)


manticore/platforms/linux.py
```diff
@@ -2126,6 +2126,28 @@ def sys_recv(self, sockfd, buf, count, flags):
...
+        # TODO Unimplemented src_addr and addrlen, so act like sys_recv
+        try:
+            sock = self.files[sockfd]
+        except IndexError:
+            return -errno.EINVAL
+
+        if not isinstance(sock, Socket):
+            return -errno.ENOTSOCK
+
+        data = sock.read(count)
```
disconnect3d: What if `read` fails?

disconnect3d: i.e.
```
close(sockfd);
sys_recvfrom(sockfd, ...)
```

disconnect3d: Also: what happens if `buf == nullptr`?

manticore/platforms/linux.py
```diff
@@ -2955,6 +2977,33 @@ def sys_recv(self, sockfd, buf, count, flags):
...
+
+        if issymbolic(flags):
+            logger.debug("Submitted a symbolic flags")
+            raise ConcretizeArgument(self, 3)
+
+        if issymbolic(src_addr):
+            logger.debug("Submitted a symbolic source address")
+            raise ConcretizeArgument(self, 4)
+
+        if issymbolic(addrlen):
```
disconnect3d: Loud thinking: I wish Python had inline functions (`issymbolic` should be inlined ;D).

tests/native/test_syscalls.py
```diff
@@ -141,6 +142,56 @@ def test_chmod(self):
...
+        sock_fd = self.linux.sys_socket(socket.AF_INET, socket.SOCK_STREAM, 0)
+        self.assertEqual(sock_fd, 3)
+        self.linux.sys_bind(sock_fd, None, None)
+        self.linux.sys_listen(sock_fd, None)
+        conn_fd = self.linux.sys_accept(sock_fd, None, 0)
+        self.assertEqual(conn_fd, 4)
+        self.linux.sys_close(conn_fd)
+
+        conn_fd = -1
+        # Fails with "Name socket4 already in use"
```
ehennenfent: Since we've closed FD 4 already, shouldn't this be okay? The tests pass so is this just an outdated comment?

manticore/platforms/linux.py
```diff
@@ -2115,21 +2115,36 @@ def sys_accept4(self, sockfd, addr, addrlen, flags):
...
 
-    def sys_recv(self, sockfd, buf, count, flags):
-        try:
-            sock = self.files[sockfd]
-        except IndexError:
-            return -errno.EINVAL
+    def sys_recv(self, sockfd, buf, count, flags, trace_str="_recv"):
+        data: bytes = bytes()
+        if count != 0:
+            if buf not in self.current.memory:
```
ehennenfent: You could also consider checking if `buf + count in self.current.memory`

tests/native/test_syscalls.py
ekilmer: We did close FD 4, but `sys_accept` should be able to either open a new FD number or re-use FD 4 (native run re-uses 4). Currently, it tries to reopen FD 4, but we haven't purged FD 4 from the available new FDs for open files list when we close it (or something, I forget exactly), so there is an error when trying open a new FD after closing.

Here is a C program that might be a bit more concrete:

```c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <netinet/in.h>
#include <signal.h>
#include <string.h>

#define PORT 8081
#define BUFFER_SIZE 80
#define TRUE 0
#define FALSE 1

int server_fd, new_client;
struct sockaddr_in address;

int setupServer()
{
    int opt = 1; // reuse address

    // create socket file descriptor, attach to 8081
    server_fd = socket(AF_INET, SOCK_STREAM, 0);
    if (setsockopt(server_fd, SOL_SOCKET, SO_REUSEADDR | SO_REUSEPORT, &opt, sizeof(opt))) {
      printf("Cannot setsockopt\n");
      return FALSE;
    }
    address.sin_family = AF_INET;
    address.sin_addr.s_addr = INADDR_ANY;
    address.sin_port = htons(PORT);
    if (bind(server_fd, (struct sockaddr *)&address, sizeof(address))) {
      printf("Cannot bind\n");
      return FALSE;
    }

    printf("Listening on port %i...\n", PORT);
    return TRUE;
}

int main() {
  int setup = setupServer();
  if (setup != 0) {
    return 1;
  }

  int addrlen = sizeof(address);

  listen(server_fd, 10);
  printf("Listening on fd %d...\n", server_fd);
  new_client = accept(server_fd, (struct sockaddr *)&address, (socklen_t *)&addrlen);
  printf("Accepting on fd %d...\n", new_client);
  close(new_client);
  new_client = accept(server_fd, (struct sockaddr *)&address, (socklen_t *)&addrlen);
  printf("Accepting on fd %d...\n", new_client);
  close(new_client);
  close(server_fd);
}
```

Some manticore output (from this branch)

```
...
2019-09-03 21:40:56,816: [47] m.p.platform:DEBUG: sys_arch_prctl(0x1002, 0x7ffffffd5f00) = 0
2019-09-03 21:41:20,429: [47] m.p.platform:DEBUG: sys_mprotect(0x7fffffdab000, 0x4000, 1) = 0
2019-09-03 21:41:24,487: [47] m.p.platform:DEBUG: sys_mprotect(0x555555754000, 0x1000, 1) = 0
2019-09-03 21:41:26,867: [47] m.p.platform:DEBUG: sys_mprotect(0x7ffffffdc000, 0x1000, 1) = 0
2019-09-03 21:41:26,900: [47] m.p.platform:DEBUG: sys_munmap(0x7ffffffd7000, 0x406e) = 0
2019-09-03 21:41:28,207: [47] m.p.platform:DEBUG: sys_socket(2, 1, 0) = 3
2019-09-03 21:41:28,222: [47] m.p.platform:WARNING: Unimplemented system call at 0x7fffffae6e6a: sys_setsockopt(0x3, 0x1, 0xf, 0x7ffffffffdb4, 0x4)
2019-09-03 21:41:28,222: [47] m.p.platform:DEBUG: sys_setsockopt(3, 1, 15, "", 4) = 0
2019-09-03 21:41:28,242: [47] m.p.platform:DEBUG: sys_bind(3, "", 16) = 0
2019-09-03 21:41:28,471: [47] m.p.platform:DEBUG: sys_newfstat(1, 0x7ffffffff600) = 0
2019-09-03 21:41:28,508: [47] m.p.platform:DEBUG: sys_ioctl(1, 0x5401, 0x7ffffffff560) = -22
2019-09-03 21:41:28,598: [47] m.n.c.abstractcpu:WARNING: Unimplemented instruction: 0x00007fffffdcc631: 0f ae 44 24 40  fxsave [rsp + 0x40]
2019-09-03 21:41:28,972: [47] m.n.c.abstractcpu:WARNING: Unimplemented instruction: 0x00007fffffdcc646: 0f ae 4c 24 40  fxrstor[rsp + 0x40]
2019-09-03 21:41:49,827: [47] m.n.c.abstractcpu:WARNING: Unimplemented instruction: 0x00007fffffdcc631: 0f ae 44 24 40  fxsave [rsp + 0x40]
2019-09-03 21:41:50,031: [47] m.n.c.abstractcpu:WARNING: Unimplemented instruction: 0x00007fffffdcc646: 0f ae 4c 24 40  fxrstor[rsp + 0x40]
2019-09-03 21:41:50,309: [47] m.p.platform:DEBUG: sys_brk(0) = 0x555555756000
2019-09-03 21:41:50,330: [47] m.p.platform:DEBUG: sys_brk(0x555555777000) = 0x555555777000
2019-09-03 21:41:51,237: [47] m.p.platform:DEBUG: sys_listen(3, 10) = 0
2019-09-03 21:41:51,531: [47] m.p.platform:DEBUG: sys_accept(3, "", "") = 4
2019-09-03 21:41:51,832: [47] m.p.linux:DEBUG: sys_close(4)
2019-09-03 21:41:51,832: [47] m.p.platform:DEBUG: sys_close(4) = 0
2019-09-03 21:41:51,846: [47] m.c.worker:ERROR: Exception in state 1: ValueError('Name socket4 already used',)
  File "/usr/local/lib/python3.6/dist-packages/manticore/platforms/linux.py", line 2420, in execute
    self.current.execute()
  File "/usr/local/lib/python3.6/dist-packages/manticore/native/cpu/abstractcpu.py", line 1004, in execute
    raise e
  File "/usr/local/lib/python3.6/dist-packages/manticore/native/cpu/abstractcpu.py", line 990, in execute
    implementation(*insn.operands)
  File "/usr/local/lib/python3.6/dist-packages/manticore/native/cpu/abstractcpu.py", line 1133, in new_method
    return old_method(cpu, *args, **kw_args)
  File "/usr/local/lib/python3.6/dist-packages/manticore/native/cpu/x86.py", line 5663, in SYSCALL
    raise Syscall()
manticore.native.cpu.abstractcpu.Syscall: CPU Syscall

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.6/dist-packages/manticore/core/worker.py", line 121, in run
    current_state.execute()
  File "/usr/local/lib/python3.6/dist-packages/manticore/native/state.py", line 29, in execute
  File "/usr/local/lib/python3.6/dist-packages/manticore/platforms/linux.py", line 2427, in execute
    self.syscall()
  File "/usr/local/lib/python3.6/dist-packages/manticore/platforms/linux.py", line 2254, in syscall
  File "/usr/local/lib/python3.6/dist-packages/manticore/native/cpu/abstractcpu.py", line 425, in invoke
    ret = super().invoke(model, prefix_args)
  File "/usr/local/lib/python3.6/dist-packages/manticore/native/cpu/abstractcpu.py", line 367, in invoke
    result = model(*arguments)
  File "/usr/local/lib/python3.6/dist-packages/manticore/platforms/linux.py", line 3013, in sys_accept
    symb = self.constraints.new_array(name=f"socket{fd}", index_max=nbytes)
  File "/usr/local/lib/python3.6/dist-packages/manticore/core/smtlib/constraints.py", line 397, in new_array
    raise ValueError(f"Name {name} already used")
ValueError: Name socket4 already used

2019-09-03 21:41:52,015: [17] m.c.plugin:INFO: Instructions executed: 0
2019-09-03 21:41:52,023: [17] m.c.plugin:INFO: Coverage: 0 different instructions executed
2019-09-03 21:41:52,087: [17] m.c.manticore:INFO: Generated testcase No. 0 - test
2019-09-03 21:41:52,551: [17] m.c.manticore:INFO: Results in /test/mcore_1yzdufl9
2019-09-03 21:41:52,552: [17] m.n.manticore:INFO: Total time: 62.54935026168823
```

You can run it natively and interact with it by just issuing `netcat 127.0.0.1 8081` a couple times to pass the `accept`s and it should print:
```
root@b217dbd5198c:/test# ./a.out
Listening on port 8081...
Listening on fd 3...
Accepting on fd 4...
Accepting on fd 4...
```

ehennenfent: Oh, this is a problem with smtlib. We free up FD 4 for the creation of a new socket, but we don't delete the backing symbolic array "socket4" from the constraint set, so when we try to create a new socket, it yells at us. Previously this was even worse, as we'd use the name "socket" for every backing array so you couldn't make more than one. 

If you want to fix it, you could do one of the following:
* Count the symbols we've created with a given name and deduplicate them by postfixing the counter
* Replace the file descriptor number in the symbol name with a long-ish random number
* Change the name of the symbol in the constraint set as soon as the FD gets closed.

ekilmer: Ah. That makes more sense. Thank you!

Alos, if it's okay with you, I think a different PR to fix this overlapping symbolic array issue would be more appropriate than including a fix in this PR.

I've opened #1521 to keep track of this.

manticore/platforms/linux.py
ekilmer: Good call. I also see other areas in the code where only `buf not in self.current.memory`. Shall I fix those instances too? There even seems to be a comment about checking `buf+count` (line 1338), but not actually doing it, here:
https://github.com/trailofbits/manticore/blob/7a9533a7a17531a07f0683f472e2c39a68c3271f/manticore/platforms/linux.py#L1334-L1340
Is there a reason for that?

disconnect3d:
What about testing? The first option - unit tests - seems like an efficient and minimal approach but a small C program test would also be awesome.

If we go with the C program: please make sure we have sources for it in the repo too (I would assume we push binary as well, not to depend e.g. on a compiler change or not to waste time compiling it on CI).

manticore/platforms/linux.py
disconnect3d: We should probably have such check for each byte that is being written OR for the whole slice.

Something similar to `max_exec_size` or whatever name it ends up with, suggested in https://github.com/trailofbits/manticore/pull/1522, might be handy here.

disconnect3d: Regarding:
> I also see other areas in the code where only buf not in self.current.memory. Shall I fix those instances too? 

That would be awesome. But please, use another PR for that :).

ekilmer: Add my change in https://github.com/trailofbits/manticore/pull/1514/commits/b66400090465243abc8a94e4bf1e5b41f76de012

ekilmer:
> We need to handle the situation where buf+X where x<count properly.

I don't exactly understand what you mean by this, but does https://github.com/trailofbits/manticore/pull/1514/commits/b66400090465243abc8a94e4bf1e5b41f76de012 address it?

manticore/platforms/linux.py
```diff
@@ -2115,21 +2115,36 @@ def sys_accept4(self, sockfd, addr, addrlen, flags):
...
+            if buf not in self.current.memory or (buf + count) not in self.current.memory:
+                logger.info("buf or (buf + count) points to invalid address. Returning EFAULT")
+                return -errno.EFAULT
 
-        if not isinstance(sock, Socket):
-            return -errno.ENOTSOCK
+            try:
+                sock = self._get_fd(sockfd)
+            except FdError:
+                return -errno.EINVAL
```
ehennenfent: I notice that we only get to this case if `count != 0`. Does that mirror the real-life behavior of the `recv` syscall? Or should it be as it was in the previous version, where this call might return an error even if `count == 0`?

ekilmer: Good catch.

According to the Linux kernel source, a check is done first on whether the whole buffer points to a valid place in memory, and then checks if the `fd` is valid. https://elixir.bootlin.com/linux/v5.2.13/source/net/socket.c#L2001

disconnect3d:
> > We need to handle the situation where buf+X where x<count properly.
> 
> I don't exactly understand what you mean by this, but does [b664000](https://github.com/trailofbits/manticore/commit/b66400090465243abc8a94e4bf1e5b41f76de012) address it?

Yeah, I believe this is a good-enough implementation of it.

An ideal one would also ensure a recv that would partially recv and then fault e.g. due to memory permissions on another page, but we probably don't go into such cases for now.

ekilmer:
> An ideal one would also ensure a recv that would partially recv and then fault e.g. due to memory permissions on another page, but we probably don't go into such cases for now.

Ah. I understand what you mean now. I think https://github.com/trailofbits/manticore/pull/1514/commits/eda51bfddfceedf54a97633f67114630e968c8af should fix this.

disconnect3d:
> > An ideal one would also ensure a recv that would partially recv and then fault e.g. due to memory permissions on another page, but we probably don't go into such cases for now.
> 
> Ah. I understand what you mean now. I think [eda51bf](https://github.com/trailofbits/manticore/commit/eda51bfddfceedf54a97633f67114630e968c8af) should fix this.

That's perfect!

PR #1520, opened 2019-09-03
Clean-up log

smoelius:
This commit does two things: (1) repair log "paths" that appear to be
non-existent/out-of-date, and (2) add the "manticore.utils.helper",
logging.INFO "path".

I added the new "helper" path to level 2, but I am open to changing
this.


ehennenfent:
LGTM. In the long run we still probably want a much more involved logging overhaul, but any improvement is welcome.

PR #1522, opened 2019-09-04
perf: optimize decode_instruction

catenacyber:
By avoiding multiple calls to memory `access_ok`

Tested on multiple_styles :
There are 387008 calls to `access_ok` from `decode_instruction for 3.5 seconds

With the PR, there are
25268 calls to `access_ok` from `decode_instruction for 0.3 seconds
25268 calls to `max_exec_size` from `decode_instruction for 0.3 seconds

### DIFF
--- a/manticore/native/cpu/abstractcpu.py
+++ b/manticore/native/cpu/abstractcpu.py
@@ -889,11 +889,10 @@ class Cpu(Eventful):
         text = b""
 
         # Read Instruction from memory
-        for address in range(pc, pc + self.max_instr_width):
+        maxexecsize = self.memory.max_exec_size(pc, self.max_instr_width)
+        for address in range(pc, pc + maxexecsize):
             # This reads a byte from memory ignoring permissions
             # and concretize it if symbolic
-            if not self.memory.access_ok(address, "x"):
-                break
 
             c = self.memory[address]
 
--- a/manticore/native/memory.py
+++ b/manticore/native/memory.py
@@ -892,6 +892,19 @@ class Memory(object, metaclass=ABCMeta):
         else:
             return self.map_containing(index).perms
 
+    def max_exec_size(self, index, size):
+        r = 0
+        addr = index
+        while addr < index + size:
+            if addr not in self:
+                return r
+            m = self.map_containing(addr)
+            if not m.access_ok("x"):
+                return r
+            r += m.end - addr
+            addr = m.end
+        return size
+
     def access_ok(self, index, access, force=False):
         if isinstance(index, slice):
             assert index.stop - index.start >= 0


manticore/native/memory.py
```diff
@@ -892,6 +892,19 @@ def perms(self, index):
         else:
             return self.map_containing(index).perms
 
+    def max_exec_size(self, index, size):
```
disconnect3d: This needs a docstring and I we should probably change the argument names to `addr` and maybe `offset`?

My reasoning is that we pass here _some address_ and we ask _what is the maximum offset from that address that is executable_. This being said maybe the function name should be `max_exec_offset`? I am not sure here.

manticore/native/memory.py
```diff
@@ -892,6 +892,19 @@ def perms(self, index):
...
+        r = 0
```
disconnect3d: Maybe:
```suggestion
        exec_offset = 0
```
and later on

manticore/native/memory.py
```diff
@@ -892,6 +892,19 @@ def perms(self, index):
...
+        addr = index
+        while addr < index + size:
+            if addr not in self:
+                return r
+            m = self.map_containing(addr)
+            if not m.access_ok("x"):
```
disconnect3d: Heh, this uses `.perm` attribute under the hood which is public... Maybe this attribute should be private? Or maybe we should use it instead and do `if 'x' in m.perm`. We should probably leave it as it is for now and keep that in mind/possibly get back to it in the future.

manticore/native/memory.py
disconnect3d: Btw this function might be handy for optimizing other memory accesses - read/writes - so maybe it would be best to pass the permission flag here too?

catenacyber: > maybe offset

It is a size, not an offset

> Btw this function might be handy for optimizing other memory accesses

Which ones ?
Should this be part of this PR ?

manticore/native/memory.py
catenacyber: I replaced it with `max_size`

manticore/native/memory.py
catenacyber: ok, adding a comment

catenacyber:
> This idea is great and we could probably reuse it in other parts of the code too :).

Thanks, do you know where ?
And what is exactly the idea ? not calling `access_ok` too often ?

catenacyber:
I added another optimization having only one call to `self.memory.__getitem__` 
It looks effective

manticore/native/memory.py
disconnect3d: > Which ones ?

At first I thought about all usages of `memory.access_ok` but it turns out it accepts a `slice` too!

Anyway, this: https://github.com/trailofbits/manticore/blob/master/manticore/native/memory.py#L1201-L1202 could probably be fixed/optimized out to do the check on a slice instead of single bytes.

At least maybe, because that is symbolic and can lead to some other issues? I am not sure here.

> Should this be part of this PR ?

Not really. I was thinking that we could change the function to accept the permission flag as argument but since `accept_ok` accepts a `slice` and is used like this in some places already, then maybe not.

manticore/native/cpu/abstractcpu.py
```diff
@@ -889,14 +889,13 @@ def decode_instruction(self, pc):
         text = b""
 
         # Read Instruction from memory
-        for address in range(pc, pc + self.max_instr_width):
+        maxexecsize = self.memory.max_access_size(pc, self.max_instr_width, "x")
```
disconnect3d: Maybe:
```suggestion
        exec_size = self.memory.max_access_size(pc, self.max_instr_width, "x")
```
and further too?

manticore/native/memory.py
```diff
@@ -892,6 +892,28 @@ def perms(self, index):
...
+        """
+        Finds maximum executable memory size
+        starting from addr and up to max_size.
+
+        :param addr_start: the starting address.
+        :param size: the maximum size.
+        :param access: the wished access.
+        """
+        size = 0
+        addr = addr_start
```
disconnect3d: The `addr_start` is not used later on, so we could have it as the argument name, reuse it and assign the ending address instead:

```python
        max_addr = addr + max_size
        while addr < max_addr:
```

...my reasoning is that it is Python and the `addr_start + max_size` won't be optimized out (at least not in or by CPython VM):

image

manticore/native/memory.py
```diff
@@ -892,6 +892,28 @@ def perms(self, index):
         else:
             return self.map_containing(index).perms
 
+    def max_access_size(self, addr_start, max_size, access):
+        """
+        Finds maximum executable memory size
```
disconnect3d: > Finds maximum executable memory size

Since we don't reuse the function for other permissions yet and I am unsure if it is needed for now (since `.access_ok` accepts a slice too) we should probably go back to `max_exec_size`? :P

disconnect3d:
> > This idea is great and we could probably reuse it in other parts of the code too :).
> 
> Thanks, do you know where ?

See one of the comments to the code; unfortunately Github marks it as outdated :f.

> And what is exactly the idea ? not calling `access_ok` too often ?

Yeah.

manticore/native/cpu/abstractcpu.py
catenacyber: ok

manticore/native/memory.py
catenacyber: ok

catenacyber:
So, here is a commit with the code review comments taken into account.

For other spots, I propose to do this first as they may get more complicated...

disconnect3d:
Hmm, I already merged it, but can we also add a unit test or two for the added function? @catenacyber

PR #1525, opened 2019-09-13
performance: faster program load with faster memory write

catenacyber:
I experienced it while loading a big program (200Mb)
After more than 10 minutes waiting, I saw that I was slowly looping in `write_bytes` in abstractcpu.py as the memory type is `FileMap` and not `AnonMap`

With this simple patch, it takes less than a minute.

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -1047,7 +1047,7 @@ class Linux(Platform):
             logger.debug(
                 f"Zeroing main elf fractional pages. From bss({elf_bss:x}) to brk({elf_brk:x}), {bytes_to_clear} bytes."
             )
-            cpu.write_bytes(elf_bss, "\x00" * bytes_to_clear, force=True)
+            cpu.memory.write(elf_bss, "\x00" * bytes_to_clear, force=True)
 
         stack_size = 0x21000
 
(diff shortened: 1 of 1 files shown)


catenacyber:
No ideas why 2 tests fail... Any help ?

ehennenfent:
The Unicorn tests are failing because only `cpu.write_bytes` emits the `did_write_memory` event, for which the Unicorn module has a subscribed callback that keeps memory states synced between Manticore and Unicorn. We do need to make sure that callback gets emitted anytime we write to memory, even if bypassing the CPU while doing so.
Mind you, `cpu.write_bytes` has a special "fast write" behavior specifically intended for copying memory maps into Unicorn, so if you bypass it to write directly to the memory map while loading a Linux binary, it'll make the regular emulation faster, but make the "fast forwarding" capabilities of the Unicorn module significantly less effective.

catenacyber:
Thank you Eric.
Here is another proposal to achieve the same speed-up for `load` while not bypassing  `cpu.write_bytes` but rather relaxing the condition to go on the fast path.
But I am not sure if it has more side effects...

ehennenfent:
Closing this & the other PR's that haven't been updated since 2019. Feel free to reopen if you're planning further updates.

PR #1526, opened 2019-09-16
Unsound symbolication - A general approach to handle symbolic imprecisions 

feliam:
This PR provides a way to procrastinate the solving of complicated, unknown or irreversible functions, like sha3. We call this unsound symbolication (for now).

We simply **replace** the result of applying one of this unknown functions over symbolic data **with a free symbol**. And **record the input-output** (domain-range) relation of every application in a table. Then normal exploration continues potentially reaching unfeasible states.  And finally we provide a method to **solve/search concrete sustitutions for the generated free symbols**. This is done in a way that the solutions match the provided concrete complicated function.

A down to earth implementation of Godefroid's higher level test case generation for symbolic execution. It can generate testcases for stuff like..
```
contract Sha3_Multiple_tx{
            event Log(string);
            bytes32 val;
            function foo(uint x) public {
                if (x == 12345){
                    val = keccak256(keccak256(uint(6789)));
                }
                else{
                    if (keccak256(val) == keccak256(keccak256(keccak256(x)))){
                        emit Log("bug");
                    }
                }
            }
        }
```


manticore/ethereum/manticore.py
```diff
@@ -43,8 +43,28 @@
...
+    def from_string(cls, name):
+        return cls.__members__[name]
+
+
+consts = config.get_group("evm")
+consts.add("defaultgas", 3000000, "Default gas value for ethereum transactions.")
+consts.add(
+    "sha3",
+    default=Sha3Type.concretize,
+    description="concretize(*): sound simple concretization\nsymbolicate: unsound symbolication with gout of cycle FP killing",
```
ehennenfent: s/gout/out

manticore/ethereum/manticore.py
```diff
@@ -169,10 +189,11 @@ def make_symbolic_address(self, name=None, select="both"):
         """
         if select not in ("both", "normal", "contract"):
             raise EthereumError("Wrong selection type")
-        if select in ("normal", "contract"):
-            # FIXME need to select contracts or normal accounts
-            raise NotImplemented
-
+        # if select in ("normal", "contract"):
+        # FIXME need to select contracts or normal accounts
```
ehennenfent: May want to remove the select arg until it supports anything other than "both"

manticore/ethereum/manticore.py
```diff
@@ -1625,17 +1590,8 @@ def _emit_trace_file(filestream, trace):
...
-        global_findings = set()
-        for detector in self.detectors.values():
-            for address, pc, finding, at_init in detector.global_findings:
-                if (address, pc, finding, at_init) not in global_findings:
-                    global_findings.add((address, pc, finding, at_init))
-        return global_findings
-
     @ManticoreBase.at_not_running
-    def finalize(self, procs=10):
+    def finalize(self, procs=45):
```
ehennenfent: Should we pull the `procs` argument from `config` instead of hardcoding it? 45 may cause some trouble on most machines.

manticore/platforms/evm.py
```diff
@@ -2221,12 +2220,11 @@ class EVMWorld(Platform):
         "evm_read_storage",
         "evm_write_storage",
         "evm_read_code",
-        "decode_instruction",
+        "evm_write_code" "decode_instruction",
```
ehennenfent: Should be comma separated

manticore/platforms/evm.py
```diff
@@ -2299,6 +2297,42 @@ def __setstate__(self, state):
...
+        """
+        Get an unsound symbolication for function `func`
+
+        """
+        data = self.try_simplify_to_constant(data)
+        try:
+            result = []
+            self._publish(
+                "on_symbolic_function", func, data, result
+            )  # This updates the local copy of result
```
ehennenfent: Not sure I like using event publishing to populate result like this, but it's probably the simplest way to accomplish this. We could, in theory, resurrect event callbacks being able to return values to the publisher if we wanted to make this a bit more understandable.

manticore/platforms/evm.py
```diff
@@ -3021,8 +3063,9 @@ def dump(self, stream, state, mevm, message):
                 temp_cs.add(storage.get(index) != 0)
 
                 try:
+                    # FIXME do this only if it has symbolic writes?
```
ehennenfent: Lingering FIXME - can probably just stick an `if issymbolic` in there, right?

manticore/ethereum/manticore.py
```diff
@@ -1455,6 +1410,13 @@ def unregister_detector(self, d):
     def workspace(self):
         return self._workspace._store.uri
 
+    @property
+    def global_findings(self):
+        try:
+            return self._global_findings
+        except AttributeError:
+            return set()
```
ehennenfent: Should you also set `self._global_findings` here?

examples/script/concolic.py
```diff
@@ -72,6 +72,8 @@ def flip(constraint):
     assert len(equal.operands) == 2
     # assume they are the equal -> ite form that we produce on standard branches
     ite, forcepc = equal.operands
+    if not (isinstance(ite, BitVecITE) and isinstance(forcepc, BitVecConstant)):
```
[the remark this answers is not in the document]
feliam: Something else.

manticore/utils/event.py
```diff
@@ -150,7 +150,14 @@ def _publish_impl(self, _name, *args, **kwargs):
...
             for callback in methods:
-                callback(robj(), *args, **kwargs)
+                try:
+                    callback(robj(), *args, **kwargs)
+                except Exception as e:
+                    import traceback
+
+                    # traceback.print_stack()
+                    traceback.print_last()
+                    print("Exception", e, callback)
```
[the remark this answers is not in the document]
feliam: I think that Concretize and TerminateState should be the only ones allowed.
Exceptions vs callbacks is something we should design and document.

manticore/core/smtlib/operators.py
```diff
@@ -263,12 +259,6 @@ def UREM(a, b):
     return a % b
 
 
-def simplify(value):
```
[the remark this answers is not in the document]
feliam: Yes it was unused

tests/ethereum/test_consensys_benchmark.py
```diff
@@ -20,8 +20,7 @@ class EthBenchmark(unittest.TestCase):
 
     def setUp(self):
         self.mevm = ManticoreEVM()
-        self.mevm.register_plugin(KeepOnlyIfStorageChanges())
```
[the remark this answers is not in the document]
feliam: I did not break anything. Already re-added.
KOISC plugin test here https://github.com/trailofbits/manticore/blob/a958d718ce1f89cc367b0618d3e0345c683098e1/tests/ethereum/test_plugins.py#L23-L34

manticore/core/smtlib/solver.py
```diff
@@ -429,9 +429,10 @@ def get_all_values(self, constraints, expression, maxcnt=None, silent=False):
                     f"get_all_values only implemented for {type(expression)} expression type."
                 )
 
+            from manticore.core.smtlib import translate_to_smtlib
```
feliam: Is this used?

manticore/core/smtlib/visitors.py
```diff
@@ -275,12 +285,17 @@ def __init__(self, **kw):
...
         LessThan: operator.__lt__,
         LessOrEqual: operator.__le__,
-        Equal: operator.__eq__,
+        BoolEqual: operator.__eq__,
         GreaterThan: operator.__gt__,
         GreaterOrEqual: operator.__ge__,
         BoolAnd: operator.__and__,
         BoolOr: operator.__or__,
         BoolNot: operator.__not__,
+        BitVecUnsignedDiv: lambda x, y: (x & (1 << 256) - 1) // (y & (1 << 256) - 1),
```
feliam: Slice it out to a diff PR and make a unittest for this guy

manticore/ethereum/manticore.py
feliam: Detectors and findings are an issue.  This should be available only after finalization. I'm considering adding a manticore "State" for this. 
 
Before every time a Detector finds a tentative issue it used to add it in the global list immediately. So the findings were reported even if the state in which they have been found later REVERTED of did not reach a final condition.

manticore/core/smtlib/expression.py
```diff
@@ -1040,6 +1032,9 @@ def __getitem__(self, index):
             for k, v in self._concrete_cache.items():
                 if k >= start and k < start + size:
                     array_proxy_slice._concrete_cache[k - start] = v
+
+            for i in self.written:
```
feliam: So this was wrong before? We may need a regression test.

manticore/core/smtlib/expression.py
```diff
@@ -1084,9 +1079,9 @@ def written(self):
             # take out Proxy sleve
             array = self._array
             offset = 0
-            if isinstance(array, ArraySlice):
+            while isinstance(array, ArraySlice):
                 # if it is a proxy over a slice take out the slice too
-                offset = array._slice_offset
+                offset += array._slice_offset
```
feliam: So this was wrong before? We may need a regression test.

manticore/ethereum/manticore.py
```diff
@@ -464,9 +399,13 @@ def __init__(self, workspace_url: str = None, policy: str = "random"):
         super().__init__(initial_state, workspace_url=workspace_url, policy=policy)
         self.subscribe("will_terminate_state", self._terminate_state_callback)
         self.subscribe("did_evm_execute_instruction", self._did_evm_execute_instruction_callback)
-        self.subscribe("did_read_code", self._did_evm_read_code)
-        self.subscribe("on_symbolic_sha3", self._on_symbolic_sha3_callback)
-        self.subscribe("on_concrete_sha3", self._on_concrete_sha3_callback)
+        consts.sha3 = consts.sha3.symbolicate
```
ehennenfent: Is this meant to be hardcoded here? With it in place, it looks like the next `if` statement will always be false and the one following it will always be true.

manticore/ethereum/manticore.py
```diff
@@ -1276,78 +1105,214 @@ def run(self, **kwargs):
...
+                # With the current x,y pair being one of the known sha3 pairs try
+                # to match the following symbolic pairs
+                with state as temp_state:
+                    temp_state.constrain(seen)
+                    if match(temp_state, func, new_symbolic_pairs, new_concrete_pairs, start=start):
+                        concrete_pairs.update(new_concrete_pairs)
+                        return True
+            return False
+
+        # Save concrete unction
```
ehennenfent:
```suggestion
        # Save concrete function
```

ehennenfent: This has moved to line 1159

Arachnid:
This PR removed support for loading contracts from JSON; any clue why?

montyly:
Hi @Arachnid, can you give more context on your usage?

Manticore relies now on [`crytic-compile`](https://github.com/crytic/crytic-compile) to load contracts, and the JSONs from the compiler are directly handled by the library.

Arachnid:
@montyly I just mean that this PR removed the `json_create_contract` function, and I can't see why - it's not mentioned in the description.

Using `crytic-compile` requires having a system solc that matches the version needed for the code I'm analyzing, and requires that paths for includes etc all be configured correctly - but systems like truffle and hardhat already output JSON artefacts that contain all the necessary information without having to recompile from scratch. It'd be nice to be able to use those instead of duplicating work.

montyly:
Hi @Arachnid, indeed we need to improve our documentation on how to analyze bytecode-only contracts.

We have an example here: https://gist.github.com/feliam/ec8b9aa102055948524838ffe0a09769. Does that help you?

Related: https://github.com/crytic/building-secure-contracts/issues/23

Arachnid:
@montyly I'm not trying to analyze a bytecode-only contract. I have the source - but configuring Manticore and a local solc instance to compile it exactly the same way as the Truffle or Hardhat project I already have working is a pain, and seems unnecessary. Manticore should be able to load Truffle or Hardhat's build artefacts and get everything it needs from there, bypassing any of that setup. It seems like you once had that functionality, but it was removed in this PR for some reason.

montyly:
@Arachnid: I am not sure to understand. Manticore already works with truffle/hardhat, you can pass the top-level directory of the project as the target. 

Feel free to join our [slack](https://empireslacking.herokuapp.com/) (#ethereum) and ping me directly, we might be able to help better through slack

Arachnid:
Oh, I see! Is that documented anywhere?

montyly:
[Our documentation](https://github.com/crytic/building-secure-contracts/blob/master/program-analysis/manticore/running-under-manticore.md#run-a-standalone-exploration) mentions that Manticore can be run a the project directory, through we could be more clear about truffle etc.

Overall our blockchain tools rely on [crytic-compile](https://github.com/crytic/crytic-compile/), so any platform supported by crytic-compile will work with Manticore

PR #1527, opened 2019-09-20
perf: avoid multiple calls to read same self.PC

catenacyber:
Tested on example hxp2018_angrme

x86.py:696(read) drops from 900 000 calls to 500 000 calls
This results in a cumulative gain of about one second in abstractcpu.py:602(__getattr__)

For the last one, we must read again `self.PC` as `self._publish("will_execute_instruction", curpc, insn)` calls the hooks which may change the value of `self.PC`

### DIFF
--- a/manticore/native/cpu/abstractcpu.py
+++ b/manticore/native/cpu/abstractcpu.py
@@ -950,19 +950,21 @@ class Cpu(Eventful):
         """
         Decode, and execute one instruction pointed by register PC
         """
-        if issymbolic(self.PC):
+        curpc = self.PC
+        if issymbolic(curpc):
             raise ConcretizeRegister(self, "PC", policy="ALL")
-        if not self.memory.access_ok(self.PC, "x"):
-            raise InvalidMemoryAccess(self.PC, "x")
+        if not self.memory.access_ok(curpc, "x"):
+            raise InvalidMemoryAccess(curpc, "x")
 
-        self._publish("will_decode_instruction", self.PC)
+        self._publish("will_decode_instruction", curpc)
 
-        insn = self.decode_instruction(self.PC)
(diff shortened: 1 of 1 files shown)


manticore/native/cpu/abstractcpu.py
```diff
@@ -950,19 +950,21 @@ def execute(self):
...
+        curpc = self.PC
+        if issymbolic(curpc):
             raise ConcretizeRegister(self, "PC", policy="ALL")
-        if not self.memory.access_ok(self.PC, "x"):
-            raise InvalidMemoryAccess(self.PC, "x")
+        if not self.memory.access_ok(curpc, "x"):
+            raise InvalidMemoryAccess(curpc, "x")
 
-        self._publish("will_decode_instruction", self.PC)
+        self._publish("will_decode_instruction", curpc)
```
disconnect3d: I suppose `will_decode_instruction` can also change the state, so we need to "invalidate" the cached value again

catenacyber: Is there a use case for this ?
Or is this a general principle when publishing an event ?

disconnect3d: I think we should not rely that somebody won't change any state during any event.

Regarding use case... What event is used when `@hook` on an address is used?

If that is not `will_decode_instruction` and is `will_execute_instruction` then maybe it should be the first one :thinking:. Or at least someone might use that as an optimization.

catenacyber: For `@hook`,  it is `will_execute_instruction`

After thinking, I agree that wa cannot assume change state during event `will_decode_instruction` So, I refresh PC value once.

Optimization is a bit less effective, but still good to have ;-)

Issue #1528, opened 2019-09-24
Mutexes for global context are not reliable


pythonpsycho1337:
### OS / Environment

No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.3 LTS
Release:	18.04
Codename:	bionic
user@user-Y700-15ISK:~/Document

### Manticore version

Version: 0.3.0

### Python version

Python 3.6.8

### Dependencies

alabaster==0.7.12
apsw==3.16.2.post1
asn1crypto==0.24.0
attrs==18.2.0
Automat==0.7.0
awscli==1.16.107
Babel==2.6.0
backports.functools-lru-cache==1.4
backports.ssl-match-hostname==3.5.0.1
beautifulsoup4==4.6.0
boto==2.44.0
botocore==1.12.97
cached-property==1.3.1
capstone==4.0.1
Cerealizer==0.8.2
certifi==2018.11.29
chardet==3.0.4
CherryPy==8.9.1
cliapp==1.20170827
click==6.7
cmdtest==0.32
colorama==0.3.9
constantly==15.1.0
coverage==4.5.2
crcmod==1.7
cryptography==2.1.4
cssselect==1.0.3
cssutils==1.0.2
cycler==0.10.0
Cython==0.27.3
deluge==1.3.15
dnspython==1.15.0
docker==2.5.1
docker-compose==1.17.1
docker-pycreds==0.2.1
dockerpty==0.4.1
docopt==0.6.2
docutils==0.14
enum34==1.1.6
feedparser==5.2.1
filelock==3.0.10
fretwork==0.3.0
funcsigs==1.0.2
functools32==3.2.3.post2
futures==3.2.0
google-compute-engine==2.8.16
html5-parser==0.4.4
html5lib==0.999999999
hyperlink==18.0.0
idna==2.8
imagesize==1.1.0
incremental==17.5.0
ipaddress==1.0.17
Jinja2==2.10
jmespath==0.9.3
jsonschema==2.6.0
keyring==10.6.0
keyrings.alt==3.0
Lab==3.0
lizard==1.16.3
lxml==4.2.1
Markdown==2.6.9
MarkupSafe==1.1.0
matplotlib==2.1.1
mechanize==0.2.5
mercurial==4.5.3
mock==2.0.0
msgpack==0.5.6
netifaces==0.10.4
numpy==1.15.3
olefile==0.44
packaging==19.0
PAM==0.4.2
pbr==3.1.1
[...]

PR #1529, opened 2019-09-26
Improvements in event publishing 

feliam:
Sometimes the "did_execute_instruction" failed to be fired after an int80 (native/x86)

### DIFF
--- a/manticore/native/cpu/abstractcpu.py
+++ b/manticore/native/cpu/abstractcpu.py
@@ -953,10 +953,14 @@ class Cpu(Eventful):
         """
         Decode, and execute one instruction pointed by register PC
         """
-        if self._delayed_event :
+        if self._delayed_event:
             self._icount += 1
-            self._publish("did_execute_instruction", self._last_pc, self.PC,
-                          self.decode_instruction(self._last_pc))
+            self._publish(
+                "did_execute_instruction",
+                self._last_pc,
+                self.PC,
+                self.decode_instruction(self._last_pc),
+            )
             self._delayed_event = False
 
         if issymbolic(self.PC):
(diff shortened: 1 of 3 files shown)


manticore/native/cpu/abstractcpu.py
```diff
@@ -999,7 +1012,7 @@ def execute(self):
                     )
                     self.backup_emulate(insn)
         except (Interruption, Syscall) as e:
-            e.on_handled = lambda: self._publish_instruction_as_executed(insn)
+            self._delayed_event = True
```
ehennenfent: So `on_handled` never gets called, eh? Could there be another bug? Seems like it's supposed to be handled right here: https://github.com/trailofbits/manticore/blob/master/manticore/platforms/linux.py#L2428-L2429

pythonpsycho1337:
issue #1528 is related to this PR.

manticore/native/cpu/abstractcpu.py
```diff
@@ -951,6 +954,16 @@ def execute(self):
         Decode, and execute one instruction pointed by register PC
         """
         curpc = self.PC
+        if self._delayed_event:
+            self._icount += 1
+            self._publish(
+                "did_execute_instruction",
+                self._last_pc,
+                curpc,
+                self.decode_instruction(self._last_pc),
```
feliam: Iff the previous syscall just removed an executable page this may be tricky.

manticore/native/cpu/abstractcpu.py
feliam: When the syscall wants to fork or raises the on_handled gets ignored?
image

ehennenfent:
Now that you've switched to `delayed_event`, we can remove this, right?
https://github.com/trailofbits/manticore/blob/dev-syscall-event/manticore/platforms/linux.py#L2428-L2429

PR #1530, opened 2019-09-27
Unsigned div constant folding simplification

feliam:


manticore/core/smtlib/visitors.py
```diff
@@ -281,6 +281,7 @@ def __init__(self, **kw):
         BoolAnd: operator.__and__,
         BoolOr: operator.__or__,
         BoolNot: operator.__not__,
+        BitVecUnsignedDiv: lambda x, y: (x & (1 << 256) - 1) // (y & (1 << 256) - 1),
```
disconnect3d:
```suggestion
        BitVecUnsignedDiv: lambda x, y: (x & ((1 << 256) - 1)) // (y & ((1 << 256) - 1)),
```

disconnect3d: This should also be moved to sth like `MASK_256BIT = 1<<256 - 1`.

Also, why do we do `& MASK_256BIT` here and we don't do for e.g. `BitVecAdd` etc?

Is it redundant?

Issue #1428, opened 2019-05-06
Remove Generic Exceptions


ehennenfent:
We use `raise Exception` in 47 different places. Not all of those are in unit tests. As a general rule, we should change these to more-specific exceptions so that it's easier to catch specific issues.

disconnect3d:
There are only 5 places left, not taking into account examples/tests/tests generators:
```
➜  manticore git:(master) ✗ git grep 'raise Exception' | grep -v examples | grep -v test_ | grep -v make_tests | grep 'raise Exception'
manticore/core/manticore.py:973:                    raise Exception("Deadlock: Waiting for CTRL+C")
manticore/core/plugin.py:129:            raise Exception
manticore/ethereum/cli.py:61:                    raise Exception(
manticore/platforms/decree.py:283:                raise Exception("Set_brk can never work. avoid overflows")
manticore/utils/helpers.py:88:                raise Exception(
```

And taking everything else into account its 15:
```
➜  manticore git:(master) ✗ git grep 'raise Exception' | wc -l
      15
```

PR #1537, opened 2019-10-17
Remove generic exceptions (#1428)

Boyan-MILANOV:
Remove generic exceptions and replace them by more specific expressions so that it is easier to catch some issues. Related to issue #1428

### DIFF
--- a/manticore/binary/binary.py
+++ b/manticore/binary/binary.py
@@ -29,6 +29,13 @@ class Binary:
         pass
 
 
+class BinaryException(Exception):
+    """
+    Binary file exception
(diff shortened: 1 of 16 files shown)


disconnect3d:
Hey, thanks for the PR!

The PR fails on black check:
```
git diff --name-only $TRAVIS_COMMIT_RANGE | python3 scripts/pyfile_exists.py | xargs black --check
```

Can you run black to reformat the changed files?

Boyan-MILANOV:
You're welcome!
All tests are passing now :)

PR #1538, opened 2019-10-18
Fix ResourceWarning for open file stream leaks

ekilmer:
Keep track of open files and close them when done.

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -950,12 +950,22 @@ class Linux(Platform):
 
         # Get interpreter elf
         interpreter = None
+
+        # Need to clean up when we are done
+        def _clean_interp_stream():
+            if interpreter is not None:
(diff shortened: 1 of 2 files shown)


manticore/platforms/linux.py
```diff
@@ -950,12 +950,21 @@ def load(self, filename, env):
 
         # Get interpreter elf
         interpreter = None
+        # Need to clean up when we are done
+        def _clean_interp_stream():
+            if interpreter is not None:
+                try:
+                    interpreter.stream.close()
+                except IOError as e:
+                    logger.error(str(e))
```
ekilmer: This is based on what was done here https://github.com/trailofbits/manticore/blob/a605cbad393e894d7dafe52934ee30fe78a84379/manticore/platforms/linux.py#L488-L493

PR #1539, opened 2019-10-24
Emit profiling.bin in workspace directory

artemdinaburg:
Save profiling data into the workspace directory and not script directory.

### DIFF
--- a/manticore/core/manticore.py
+++ b/manticore/core/manticore.py
@@ -132,6 +132,7 @@ class ManticoreBase(Eventful):
         "terminate_state",
         "kill_state",
         "execute_instruction",
+        "terminate_execution",
     }
 
     def __init__(self, initial_state, workspace_url=None, policy="random", **kwargs):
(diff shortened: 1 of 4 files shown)


ehennenfent:
Per discussion on Slack, for compatibility's sake, add a `generate_testcase` callback to `Profiler` that calls `testcase.open_stream` and passes the resulting stream to `save_profiling_data`

artemdinaburg:
> Per discussion on Slack, for compatibility's sake, add a `generate_testcase` callback to `Profiler` that calls `testcase.open_stream` and passes the resulting stream to `save_profiling_data`

Commenting on this before I forget: there doesn't seem to be a good "manticore is exiting" callback to register with the plugin, and i can't get a reference to a stream from just a `manticore` object in `cli.py`.

Is the preferred solution to add some kind of new callback, or to make streams accessible from something that is not a `testcase` ?

ehennenfent:
I added a `did_terminate_execution` callback that's fired via `manticore.kill`. It won't fire if you just call `manticore.run` and let it fall off the end after that, but the CLI scripts explicitly call either `finalize` or `kill`, so they shouldn't have any problems.

PR #1544, opened 2019-10-29
Mypy support

ekilmer:
In this PR, [mypy](https://github.com/python/mypy) checks the manticore package for inconsistencies in both existing type hints in the Manticore source code and inferred type hints from base Python packages. Currently, this mypy configuration does not enforce any kind of increased coverage of type hints in new code. It will only produce errors if the new code is inconsistent with pre-existing type-annotated code. 

The changes made here are close to the minimum set of changes to complete a full run of mypy over the manticore package. In cases where there were bugs, I tried to make the smallest fix to pass the checks by mypy. Let me know if you would like me to create separate PRs for these fixes.

Furthermore, there are a few instances where I've added `# type: ignore` to statements that I have been unable to find good examples or documentation on how to fix. I believe I have mentioned the errors or reasons why I added those `ignore` comments in each commit (related to the file changed). There are a few known issues/shortcomings of mypy, mainly related to decorators that are tracked in these issues: 
* [support function decorators excellently](https://github.com/python/mypy/issues/3157)
* [decorator as class member raises "self-argument missing"](https://github.com/python/mypy/issues/7778)

By accepting this PR, the Manticore project can begin to gradually introduce type-hints into the code base with confidence that everything is staying consistent. Eventually, we may want to start turning on more strict requirements such that new code must not decrease type coverage of the existing code base, similar to CodeClimate checks.

Other changes/comments made in this PR:
* Pinned versions of `mypy` and `black`:
    * Consistency is nice to have so that `master` doesn't magically break if there's an incompatible version bump on the linting tools.
    * On the other hand, if a contributor doesn't read the `CONTRIBUTING.md` or look at the versions used in CI, or if they don't know how to set up a virtual environment, then their linting passes could be wrong (assuming they install the latest version of which we haven't upgraded to yet)
* Currently mypy is set to check for Python3.6 style type hints and typing errors. I'm not sure if these are fully compatible with Python3.7, but since we only test Python3.6 in CI, I think this is okay. Wether we should have support for both is another discussion, I think.

## Next Steps

Since this PR only scratches the surface of what mypy can do for us, I want to list some thoughts here on how to gradually introduce stronger/stricter rules with mypy, until this gets merged (whereafter it will be transcribed to an issue):

* Turn on  [`check_untyped_defs`](https://mypy.readthedocs.io/en/stable/config_file.html#untyped-definitions-and-calls) in `mypy.ini`, which "type-checks the interior of functions without type annotations" and produces errors based on inconsistencies. 
* Enable more strict options for `smtlib` to catch errors in this core part of Manticore.
* Enable mypy on example scripts
* Add type-hints to commonly used API functions to benefit end-users (initially based on functions used in example scripts)
* Enable mypy on tests
* Generate type-hints using [MonkeyType](https://monkeytype.readthedocs.io/en/latest/index.html) or [PyAnnotate](https://github.com/dropbox/pyannotate)
* Track coverage of type-hints and enforce only positive changes to coverage percentage for new PRs


manticore/core/smtlib/solver.py
```diff
@@ -496,18 +497,24 @@ def optimize(self, constraints: ConstraintSet, x: BitVec, goal: str, M=10000):
...
                         if maybe_sat == "sat":
                             m = RE_MIN_MAX_OBJECTIVE_EXPR_VALUE.match(_status)
-                            expr, value = m.group("expr"), m.group("value")
-                            assert expr == aux.name
-                            return int(value)
+                            if m:
+                                expr, value = m.group("expr"), m.group("value")
+                                assert expr == aux.name
+                                return int(value)
+                            else:
```
ehennenfent: Looks fine, but I'm guessing this is unrelated to the rest of the PR?

ekilmer: Actually, it's something that mypy caught, since the return type of `regex.match()` could be `None` (type `Optional`), so in order to pass mypy, I added this change and the similar one a couple lines down.

I can make a separate PR for these changes, if you'd like.

ehennenfent:
So, when this branch is merged, mypy detects:
* 8 errors in the dev-unsound-symb branch
* 4 errors in #1543
* 443 errors in the WASM branch (high score)
Other branches may also have issues, but I expect them to be more limited in scope.

Overall, doesn't look like this will cause major problems for outstanding PR's. Let's discuss your specific needs on Slack.

ehennenfent:
In updating this to `master`, I ran into some issues with mypy. The first was that it didn't recognize that `issymbolic` is effectively a transparent alias for `isinstance(<x>, Expression)`, causing it to report several false positives. Second, it complained about one variable name being used to represent values of two different types in isolated branches of a conditional, which is perhaps bad coding practice, but definitely not a bug. 

Both of these are relatively minor annoyances, but I don't want them to slow down development. We might consider looking at [pytype](https://github.com/google/pytype) in the future as a possible alternative, although currently it can't evaluate Manticore without crashing. Let's merge this for now and see if it causes any significant problems for us. If so, we can explore different solutions in the next release cycle.

PR #1549, opened 2019-11-07
Symbol friendly fake hash replacement for evm sha3

feliam:
This implements a fake hash model for evm sha3 so it is solvable. 
This results in a different solidity based storage layout and will produce irreproducible tx traces if any hash result is used as part of the calldata in any transaction.


ehennenfent:
Not sure what's going on with `solc` in the failing tests, but at the very least, let's fix line 627 in ethereum/manticore.py, which is missing a `%s` or other format string so that we can correctly print the contract names
`logger.info("Failed to compile contract", contract_names)`

tests/ethereum/test_sha3.py
```diff
@@ -477,5 +479,17 @@ def test_example_concrete_1(self):
...
 
+class EthSha3TestFake(EthSha3TestSymbolicate):
+    def setUp(self):
+        evm_consts = config.get_group("evm")
+        evm_consts.sha3 = evm_consts.sha3.fake
+
+        self.mevm = ManticoreEVM()
+        self.worksp = self.mevm.workspace
+
+    def test_example1(self):
```
ehennenfent: Overall, looks good, but was there supposed to be something here?

feliam: This test is skipped intentionally.
-- long response
I'm importing all the same tests for "symbolicate" strategy. Though this one (test_example_1) is like 
```
            function foo(uint input) payable public{
                if (sha3(input) == 0x12341234){
                    emit Log("Found a bug");
                }
            }
``` 
It is supposed not to be solved. And with current fakehash implementation the solver simply find a way to invert it. And that should not be the case. I think. So skipped.
Maybe here is a more elegant way to inherit all the others but this one

ehennenfent: Okay, so, this is ready-to-merge then?

ehennenfent: Excellent

PR #1551, opened 2019-11-13
Fix Z3 version parsing

ekilmer:
* Fix for long format Z3 version with git hash
* Fix for unparsable Z3 version to assume newer and compatible

### DIFF
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -198,7 +198,12 @@ class Z3Solver(Solver):
             self._send("(get-info :version)")
             self._received_version = self._recv()
         key, version = shlex.split(self._received_version[1:-1])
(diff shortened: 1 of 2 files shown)


ehennenfent:
Any reason you used `float("inf")` instead of, say, -1?

ekilmer:
> Any reason you used `float("inf")` instead of, say, -1?

It was mainly so that the logic for checking whether the Z3 version was in-bounds (higher than version 4.4.1) wouldn't need to change.

I also saw this post that says Python3 has arbitrarily sized integers now: https://stackoverflow.com/a/22948228

> Python 3's int does not have a maximum size, but if you just need something that's guaranteed to be higher than any other int value, then you can use the float value for Infinity, which you can get with float("inf").

Unless I'm missing something.

PR #1554, opened 2019-11-19
Add flag to only generate alive states when finalizing Manticore

montyly:
It is common to only want to see the states alive when finalizing Manticore. This PR adds this feature at the cli and script level.

- Add `--only-alive-testscases` flag
- Add `only_alive_states` to `m.finliaze` (default False)

### DIFF
--- a/manticore/ethereum/manticore.py
+++ b/manticore/ethereum/manticore.py
@@ -1608,6 +1608,7 @@ class ManticoreEVM(ManticoreBase):
         transaction).
 
         :param procs: force the number of local processes to use in the reporting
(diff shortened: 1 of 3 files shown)


manticore/ethereum/manticore.py
```diff
@@ -1601,7 +1601,7 @@ def _emit_trace_file(filestream, trace):
             filestream.write(ln)
 
     @ManticoreBase.at_not_running
-    def finalize(self, procs=None):
+    def finalize(self, procs=None, only_alive_states=False):
```
ehennenfent: Can you add `only_alive_states` to the docstring for this function?

montyly:
@ehennenfent good catch! I added the docstring

PR #1559, opened 2019-11-22
reorg some deps a bit, usee env markers for dataclasses

reaperhulk:
Moves `black` to the dev extra and makes `dataclasses` install only on Python < 3.7. This will make it so the `dataclasses` backport properly installs even when using a wheel.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,4 @@
 import os
-import sys
 from setuptools import setup, find_packages
 
 on_rtd = os.environ.get("READTHEDOCS") == "True"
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -13,16 +12,12 @@ def rtd_dependent_deps():
...
-dataclass_deps = ["dataclasses"] if version.major == 3 and version.minor < 7 else []
-
-
 # If you update native_deps please update the `REQUIREMENTS_TO_IMPORTS` dict in `utils/install_helper.py`
 # (we need to know how to import a given native dependency so we can check if native dependencies are installed)
 native_deps = ["capstone==4.0.1", "pyelftools", "unicorn==1.0.2rc1"]
 
 # Development dependencies without keystone
-dev_noks = native_deps + ["coverage", "nose", "Sphinx"]
+dev_noks = native_deps + ["coverage", "nose", "Sphinx", "black"]
```
ekilmer: I think this section from the travis script should be removed for this PR. https://github.com/trailofbits/manticore/blob/537483699ef1797fd8f4769dc717529d537bb39b/scripts/travis_install.sh#L29-L32

I would also suggest pinning `black` and `mypy` so we don't run into trouble if they update underneath us and change their formatting and type-checking rules, respectively.

reaperhulk: Good idea. I'll go ahead and pin them and then look at whether we can update the install scripts to use this extra. If that proves to be too invasive I'll move it to a separate PR.

ekilmer: Actually, after looking more at the travis script, it might not be as easy as I initially thought... Hmmmm

reaperhulk: Yeah I updated the PR with the pins, but there's some deeper changes needed to make the linting phase in CI use it as an extra. A future PR :)

ekilmer: Sounds good!

PR #1561, opened 2019-11-22
try pytest xdist

reaperhulk:
This may not help much yet, but let's see what CI has to say.


ehennenfent:
Looks like this shaved off:
* Nine minutes from ethereum
* Two minutes from ethereum_vm
* Three minutes from native
* Twelve minutes from WASM

In other words, nothing huge, but hey, I'll take it!

PR #1562, opened 2019-11-22
Prevent Pytest from capturing stdout

ehennenfent:
Should prevent the issue we're seeing where the WASM tests timeout due to limited output


reaperhulk:
The timeout issue is resolved for the moment via #1563, but using `pytest-xdist`'s has a side effect: `-s` to not capture stdout has no effect.

Ideally we'll be able to remove `pytest-xdist` soon as we explore other options to speed up the tests.

PR #1564, opened 2019-11-24
Enable z3 parallel mode

disconnect3d:
Let's give it a try.

By the way, `z3 -p` lists all parameters and regarding the `parallel`
module, there are also some other options we might want to try tweaking:
```
root@f74c65a4f976:/# z3 -p | grep paral -A50
[module] parallel, description: parameters for parallel solver
    conquer.backtrack_frequency (unsigned int) (default: 10)
    conquer.batch_size (unsigned int) (default: 100)
    conquer.delay (unsigned int) (default: 10)
    conquer.restart.max (unsigned int) (default: 5)
    enable (bool) (default: false)
    simplify.exp (double) (default: 1)
    simplify.inprocess.max (unsigned int) (default: 2)
    simplify.max_conflicts (unsigned int) (default: 4294967295)
    simplify.restart.max (unsigned int) (default: 5000)
    threads.max (unsigned int) (default: 10000)
[module] lp
```


ehennenfent:
Interesting, so this looks like it's actually _slower_ than the default. Have you tried running this on a machine with more cores than Travis?

disconnect3d:
> Have you tried running this on a machine with more cores than Travis?

Not really. It would be nice to see some comparisons...

ehennenfent:
Closing this & the other PR's that haven't been updated since 2019. Feel free to reopen if you're planning further updates.

PR #1566, opened 2019-11-24
pytest-xdist prevents disturbing stdout, revert to travis_wait

reaperhulk:
This prevents us from seeing realtime output but in exchange I've made pytest fast-fail so jobs won't run past the first failing test. Not ideal but we'll keep improving

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ install:
   - scripts/travis_install.sh $TEST_TYPE
 
 script:
-  - scripts/travis_test.sh $TEST_TYPE
+  - travis_wait 70 scripts/travis_test.sh $TEST_TYPE
 
 after_success:
   - ./cc-test-reporter format-coverage -t coverage.py -o "coverage/codeclimate.$TEST_TYPE.json"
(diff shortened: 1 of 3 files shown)


ehennenfent:
While I'm glad you've gotten the tests passing, I'm unsure about killing pytest on the first failing test. Usually we want to at least _know_ about all of the failures so that we can effectively debug them (especially if there are multiple bugs in one PR). If we kill the tests after the first failure, then the first step for debugging from now on will be to re-run the full test suite somewhere other than Travis. On a smaller project, I'd say that's fine, but I don't think it's a constraint we should impose on Manticore because not only can our tests take hours, but they also require one to properly set up `keystone` and `solc`, which is unfortunately nontrivial.

reaperhulk:
I'm happy to do whatever the actual people doing the real work on this project want! I'll go ahead and remove the fast fail behavior.

ehennenfent:
Definitely appreciate that you're messing around with potential improvements to the CI, so feel free to keep filing PR's. Depending on how things with pytest go, we may want to revert #1561 and work on consolidating these changes into one big PR (assuming we don't find another way to get it working in the next couple of days).

PR #1567, opened 2019-12-01
switch to github actions

reaperhulk:
This mostly just lifts and shifts us to GitHub Actions, but there are a few other changes:

* I made a physical file for the C binary that gets compiled for `run_simple.py` because `EOF` echoing requires hard tabs and that doesn't play nicely with yaml files.
* The tests will now run on a daily basis even if nothing is merged (this is useful to detect changes that result in failure on unpinned deps)
* We only run CI on master branch now, so if you want to see if your branch works (and you're not operating off a fork) you'll need to submit a PR. But really, just develop on a branch on your own fork 😊 
* There is now a `lint` extra in setup.py
* Try out codecov forr coverage data. Code Climate's other features aren't particularly useful and it doesn't surface coverage data easily. Plus that s3 upload thing, boo hiss.

We can probably significantly simplify the bash for this, but I want to keep the changes reviewable so that's a challenge for another day.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,248 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  schedule:
+    # run CI every day even if no PRs/merges occur
+    - cron:  '0 12 * * *'
+
+jobs:
+  # needs to run only on pull_request
+  lint:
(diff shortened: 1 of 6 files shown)


reaperhulk:
Amusingly it appears that when first **creating** an action you can't have it run on a PR unless it's also a branch on the primary repo. I'll resubmit this that way soon.

PR #1568, opened 2019-12-02
switch to github actions (take two)

reaperhulk:
This mostly just lifts and shifts us to GitHub Actions, but there are a few other changes:

* I made a physical file for the C binary that gets compiled for run_simple.py because EOF echoing requires hard tabs and that doesn't play nicely with yaml files.
* The tests will now run on a daily basis even if nothing is merged (this is useful to detect changes that result in failure on unpinned deps)
* We only run CI on master branch now, so if you want to see if your branch works (and you're not operating off a fork) you'll need to submit a PR. But really, just develop on a branch on your own fork 😊
* There is now a lint extra in setup.py
* Try out codecov forr coverage data. Code Climate's other features aren't particularly useful and it doesn't surface coverage data easily. Plus that s3 upload thing, boo hiss.
* We can probably significantly simplify the bash for this, but I want to keep the changes reviewable so that's a challenge for another day.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,248 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  schedule:
+    # run CI every day even if no PRs/merges occur
+    - cron:  '0 12 * * *'
+
+jobs:
+  # needs to run only on pull_request
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.6
+    - name: Lint
+      if: github.event_name == 'pull_request'
+      env:
+        BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+      run: |
+        pip install -e .[lint]
+        black --version
+        # not entirely sure what's going to happen here on master builds
+        git diff --name-only $BASE_SHA..$HEAD_SHA | python scripts/pyfile_exists.py | xargs black --diff --check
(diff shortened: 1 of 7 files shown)


ehennenfent:
Like:
- Switching away from CodeClimate is fine with me. Black and mypy catch almost all of the issues that it would flag, and not being able to see or download coverage files for individual branches has been very frustrating. 
- It should be very simple to split the symbolic WASM tests from the concrete tests, which should shave ~8 minutes off of the total time taken by the WASM tests (See: [generate_tests.sh](https://github.com/trailofbits/manticore/blob/master/tests/wasm/generate_tests.sh#L36-L37))

No like:
- Since we're hard-coding the versions of black and mypy in `setup.py`, we should have a separate `lint_dependencies` list that we copy into the appropriate places rather than duplicating those entries.

ehennenfent:
I'd also suggest updating the CI badge in the README

[Codecov](https://codecov.io/github/trailofbits/manticore)
`[Codecov](https://codecov.io/github/trailofbits/manticore)`

reaperhulk:
Thanks for the feedback @ehennenfent! I updated the PR to add the badges and fix the lint deps. codecov **should** automatically merge coverage, but it's possible that it's misbehaving since it normally wants information about coverage on the master branch before it starts providing statistics.

Once this is merged I'll look at splitting up the wasm tests to get a bit more speed as well :)

ehennenfent:
One more thing: Looking over the workflow file, I think these are branch builds, not merged builds? If we're using these to check PR's, the first step should probably be to merge into master and confirm there are no conflicts (and that the tests will pass once merged).

reaperhulk:
The PR builds check against a merge commit, not the PR head. You can see the way it does this in the test output here: https://github.com/trailofbits/manticore/pull/1568/checks?check_run_id=330553477#step:2:727

PR #1572, opened 2019-12-10
removes travis entirely

reaperhulk:
We'll also want to disable the travis and codeclimate integrations for this repository now that we've completed the switch.

next up, faster wallclock tests!


reaperhulk:
This shouldn't be merged until everyone agrees we're happy with GitHub Actions. I can remove the integrations then and rebase this and we should be ready to merge at that point.

ehennenfent:
> This shouldn't be merged until everyone agrees we're happy with GitHub Actions. I can remove the integrations then and rebase this and we should be ready to merge at that point.

I've updated the branch protections for `master` to use the GH Actions test resultsI instead of Travis. I think this should be good to release now. Unless there's something else you wanted to change first?

reaperhulk:
LGTM! Merge away

PR #1573, opened 2019-12-13
Run mypy on manticore directory by default

ekilmer:
Use explicit file path in mypy.ini for default runs


ekilmer:
Successful run output here https://github.com/trailofbits/manticore/runs/347897545#step:4:92

PR #1577, opened 2019-12-17
Fix Regression(s)

ekilmer:
I've noticed that the `pwnable_collision` challenge in the manticore-examples repo is failing, as referenced in https://github.com/trailofbits/manticore-examples/issues/34

This PR is attempting to make minimal changes to the current code-base to try getting that test to pass.

## Current findings:

* ~~`arithmetic_simplify` makes test_pwnable_collision in manticore-examples
repo hang. I'm not exactly sure what is happening, but there's probably
infinite recursion/walking/"simplification" happening somewhere. It still fails with this commented, but the next commit should make it work again...~~ ~~Looks like it was something to do with checking the wrong stuff for `visit_BitVecITE` in `arithmetic_simplify`~~ An inconsistency with using `expression.operands` and `*operands`?

* `visit` in `Visitor` and `_method` in `Translator` handle `ArrayProxy` wrong

### DIFF
--- a/manticore/core/smtlib/visitors.py
+++ b/manticore/core/smtlib/visitors.py
@@ -379,19 +379,17 @@ class ArithmeticSimplifier(Visitor):
             return expression
 
     def visit_BoolAnd(self, expression, *operands):
-        if isinstance(expression.operands[0], Constant) and expression.operands[0].value:
-            return expression.operands[1]
-        if isinstance(expression.operands[1], Constant) and expression.operands[1].value:
-            return expression.operands[0]
+        if isinstance(operands[0], Constant) and operands[0].value:
+            return operands[1]
+        if isinstance(operands[1], Constant) and operands[1].value:
+            return operands[0]
 
         # AND ( EQ (EXTRACT(0,8, a), EXTRACT(0,8, b)),  EQ (EXTRACT(8,16, a), EXTRACT(8,16 b)) ->
         # EQ(EXTRACT(0,16, a), EXTRACT(0,16, b))
-        if isinstance(expression.operands[0], BoolEqual) and isinstance(
-            expression.operands[1], BoolEqual
-        ):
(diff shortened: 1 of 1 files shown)


ekilmer:
I'm not sure how to write tests for these changes, other than proving this PR fixes https://github.com/trailofbits/manticore-examples/issues/34

manticore/core/smtlib/visitors.py
```diff
@@ -488,12 +479,12 @@ def visit_BoolOr(self, expression, a, b):
             return a
 
     def visit_BitVecITE(self, expression, *operands):
```
feliam: I think *operands are the potentially modified expression.operands.
operand[0] should be something like simplify(expression.operand[0])

ekilmer: Thank you for taking a look at this @feliam. I don't quite understand your comment. Is that a suggestion to change the code or a statement about what the code is doing currently?

I can do some more experimentation if you think the changes in this PR aren't completely correct. 

I'm also open to you committing directly to this branch if you have a fix that has everything in trailofbits/manticore-examples#35 passing.

ehennenfent: I think Felipe's comment is in regards to the current behavior. Visit methods get called recursively, so the `*operands` argument will contain the results of running `visit_<x>` on each of the operands to the BitVecITE. Whereas, if you access the operands via `expression.operands[n]`, you'll get a raw version of  each operand itself instead of the output of the visitor.

ehennenfent:
🎉

ekilmer:
I also tested reverting the `ArrayProxy` handling commit (d60e0f7), but there's still a failure in the `pwnable_collision` challenge

ekilmer:
Also from @feliam on Slack:

> I like all the expression.oprerands ->operands changes

> [....] ArrayProxy should not even be an Array. Expressions should be unmutable. So we should not visit an ArrayProxy in the first place as it is mutable and hence it should not be an Expression

> so I like that PR

PR #1579, opened 2019-12-18
Add option for hooking after instruction executes

ekilmer:
This makes it easier to direct Manticore at specified locations. We can
now add constraints that we want to be true before executing the next
instruction.

TODO:
- [x] Add tests


ekilmer:
Not sure why I'm getting `HTTP 400` errors on the code coverage uploads

PR #1580, opened 2020-01-03
Connect to eth node

### LINKED ISSUE #1755: Support executing on a live EVM node
When using Manticore, it's often a major hassle to replicate the EVM testing environment. We should allow Manticore to connect to a live EVM node corresponding to a testnet so that setting up contracts for exploration will be much easier.

smoelius:
This PR enables Manticore to connect to an Ethereum node and generate symbolic transactions for a deployed contract using the contract's on-chain storage. Manticore obtains the contract's storage values using the node's JSON RPC interface.

This PR specifically adds two command line options:
* `--rpc URL`: specify the URL of an Ethereum node to connect to. Must be of the form `IP:PORT`
* ` --txtarget TXTARGET`: specify the address of a deployed contract to target

This PR further adds an optional `world_state` parameter to `ManticoreEVM`'s constructor. One can connect to an Ethereum node from within a Manticore script as follows:
```
from manticore.platforms.evm_world_state import RemoteWorldState

m = ManticoreEVM(world_state=RemoteWorldState(URL))
```


manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,405 @@
+import logging
+import os
+from abc import ABC, abstractmethod
+from typing import Dict, Optional, Set, Union
+from urllib.parse import ParseResult, urlparse
+from web3 import Web3  # type: ignore
```
ekilmer: I think it would be better to add this to [`mypy.ini`](https://github.com/trailofbits/manticore/blob/master/mypy.ini)

Should look something like
```
[mypy-web3.*]
ignore_missing_imports = True
```

smoelius: Thanks!

ehennenfent:
Hey Sam, is this still a WIP? I notice that most of the checks are passing; It's just codecov that's unhappy.

smoelius:
Thanks for asking, @ehennenfent, but it is still a WIP.  I have some additional tests which are only half-finished.  If all goes well, I should get to work on this around 10 Feb.

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,431 @@
...
+        expected = ParseResult(scheme="", netloc="", path=url, params="", query="", fragment="")
+        if actual != expected:
+            raise EthereumError("URL must be of the form 'IP:PORT': " + url)
+        self._url = url
+
+    def _web3(self) -> Web3:
+        # sam.moelius: Force WebsocketProvider.__init__ to call _get_threaded_loop.  The existing
+        # "threaded loop" could be leftover from a fork, in which case it will not work.
+        Web3.WebsocketProvider._loop = None
+        web3 = Web3(Web3.WebsocketProvider(URI("ws://" + self._url)))
```
ehennenfent, line 196: Does this handle the case of a valid URL that doesn't point to an ethereum node?

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,431 @@
...
+    def get_storage(self, address: int) -> Optional[Storage]:
+        storage = None
+        try:
+            storage = self._underlay.get_storage(address)
+        except NotImplementedError:
+            pass
+        # sam.moelius: Rightfully, the overlay's storage should be merged into the underlay's
+        # storage.  However, this is not currently implemented.
+        if storage is not None:
+            raise NotImplementedError
```
ehennenfent, line 324: The comment is helpful, but as it's on a different line, let's add a message to this error so that users will have some idea of what's missing if they encounter this error.

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,431 @@
...
+            return self._underlay.get_gaslimit()
+
+    def get_coinbase(self) -> Union[int, BitVec]:
+        if self._coinbase is not None:
+            return self._coinbase
+        else:
+            return self._underlay.get_coinbase()
+
+    def delete_account(self, constraints: ConstraintSet, address: int):
+        self._nonce[address] = DefaultWorldState().get_nonce(address)
```
ehennenfent, line 380: Nit: we don't to create `DefaultWorldState` three times.

ehennenfent (changes requested):
Would like to get some eyes on this from someone who spends more time working with EVM than me, but overall it's looking pretty good. Left some comments; mostly nits and requests for clarification. I like how the new `Storage` API provides reasonable names for things, rather than just the dict API. Thanks for your hard work!

manticore/__main__.py
```diff
@@ -165,6 +165,8 @@ def positive(value):
         "--txnoether", action="store_true", help="Do not attempt to send ether to contract"
     )
 
+    eth_flags.add_argument("--txvictim", type=str, help="Address of a deployed contract to attack")
```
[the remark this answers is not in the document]
smoelius: `--txcontract`?

manticore/core/smtlib/expression.py
```diff
@@ -663,7 +663,7 @@ def __init__(
         self, index_bits: int, index_max: Optional[int], value_bits: int, *operands, **kwargs
     ):
         assert index_bits in (32, 64, 256)
-        assert value_bits in (8, 16, 32, 64, 256)
+        assert value_bits in (1, 8, 16, 32, 64, 256)
```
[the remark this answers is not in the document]
smoelius: This is the reason for that change:
https://github.com/trailofbits/manticore/blob/2ff301ebedc960ab975468f7615afb4f888876ee/manticore/platforms/evm_world_state.py#L23-L29
`map` records which (symbolic) offsets have been written to.  `map` is an `ArrayProxy`, which inherits from `Array`.  True, "written to" is a Boolean condition.  But if we had an `ArrayBool`, we would also require an `ArrayBoolProxy` to make this approach work, right?  Is there a better approach?

manticore/__main__.py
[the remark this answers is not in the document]
ehennenfent: "Contract" or "target" both seem okay to me

manticore/core/smtlib/expression.py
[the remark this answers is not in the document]
ehennenfent: > If we had an ArrayBool, we would also require an ArrayBoolProxy to make this approach work, right?

Makes sense to me. 

> Is there a better approach?

I'm fine with this, but perhaps @feliam will have a stronger opinion?

manticore/__main__.py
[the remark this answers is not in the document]
feliam: Note: We are trying to move out of argparse in favor of config.py.

manticore/core/smtlib/expression.py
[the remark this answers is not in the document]
smoelius: Sorry, I actually meant the requirement for an `ArrayBoolProxy` class to be a negative.  I think it would create unnecessary code duplication.

I guess what I should ask, what would be the "win" from using such `Bool` classes?

manticore/__main__.py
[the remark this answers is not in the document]
smoelius: > "Contract" or "target" both seem okay to me

I like "target" and there less room for confusion with `--contract`.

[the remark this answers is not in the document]
smoelius: > Note: We are trying to move out of argparse in favor of config.py.

@feliam Is there an action required here?

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,431 @@
...
+        return self._web3().eth.getBlock("latest")["gasLimit"]
+
+    def get_coinbase(self) -> int:
+        return int(self._web3().eth.coinbase)
+
+
+####################################################################################################
+
+
+# sam.moelius: If we decide to cache results returned from a RemoteWorldState, then they should NOT
```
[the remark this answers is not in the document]
smoelius: Something like this?
```
class OverlayWorldState(WorldState):
    """
    If we decide to cache results returned from a RemoteWorldState, then they should NOT be cached
    within an overlay.  The reason is that this could affect the results of subsequent operations.
    Consider a call to get_storage_data followed by a call to has_storage.  If nothing was written
    to storage within the overlay, then the call to has_storage will throw an exception.  But if the
    result of the call to get_storage_data was cached in the overlay, then no exception would be
    thrown.
    """
    ...
```

manticore/platforms/evm_world_state.py
smoelius: Not sure. What do you think Manticore should do in that case?

manticore/core/smtlib/expression.py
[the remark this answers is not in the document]
ehennenfent: Apologies, I think my comment was ambiguous. I'm fine with leaving it as-is because I don't see any harm that allowing 1-bit BitVectors would cause, but Felipe might be aware of some hidden problems there.

manticore/platforms/evm_world_state.py
ehennenfent: As it is, I'd expect that feeding this a bad URL would throw an exception and make Manticore crash. Rather than crashing, I think Manticore should catch that exception, do any necessary cleanup, and then raise an `EthereumError`, similar to what it does for an improperly formatted URL. 


> What if a valid URL pointed to something that spoke the protocol, but returned bogus results?

Hmm, I don't know of a way to handle this, but it doesn't sound to me like we need to. On the WASM side, though there's a lot of static validation Manticore could do on the binaries, I take the stance of "If you give me a bad binary, that's _your_ problem." Somewhat similarly, I think handling the easy failure cases relatively gracefully is enough here.

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,445 @@
...
+            index_bits=256,
+            value_bits=256,
+            name=f"STORAGE_DATA_{address:x}",
+            avoid_collisions=True,
+            default=0,
+        )
+        self.dirty = False
+
+    def set(self, offset: Union[int, BitVec], value: Union[int, BitVec]):
+        self.map[offset] = 1
```
feliam, line 40: Maybe  `value != 0` instead of  `1` ?

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,445 @@
...
+
+class DefaultWorldState(WorldState):
+    def is_remote(self) -> bool:
+        return False
+
+    def accounts(self) -> Set[int]:
+        return set()
+
+    def get_nonce(self, address: int) -> int:
+        return 0
```
feliam, line 130: Maybe this should raise not implemented? 
I think some defaults are different?

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,445 @@
...
+    def get_storage_data(self, address: int, offset: Union[int, BitVec]) -> int:
+        if not isinstance(offset, int):
+            raise NotImplementedError
+        return int.from_bytes(self._web3().eth.getStorageAt(_web3_address(address), offset), "big")
+
+    def get_code(self, address: int) -> bytes:
+        return self._web3().eth.getCode(_web3_address(address))
+
+    def get_blocknumber(self) -> int:
+        return self._web3().eth.blockNumber
```
feliam, line 248: I wonder if we should check somewhere that the blocknumber/fork is supported by us.

setup.py
```diff
@@ -59,6 +59,7 @@ def rtd_dependent_deps():
         "ply",
         "crytic-compile>=0.1.1",
         "wasm",
+        "web3",
```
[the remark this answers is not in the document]
feliam: :degrassehands:

feliam (commented):
This PR is awesome and other than add node connectivity walks toward a cleaner inner workings.

manticore/platforms/evm.py
```diff
@@ -2553,10 +2552,12 @@ def get_storage_data(self, storage_address, offset):
         :return: the value
         :rtype: int or BitVec
         """
-        value = self._world_state[storage_address]["storage"].get(offset, 0)
+        value = self._world_state.get_storage_data(storage_address, offset)
```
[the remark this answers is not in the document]
smoelius: I think this behavior should be preserved. `get_storage_data` will be called on the overlay. which will in turn call the underlay here:
https://github.com/trailofbits/manticore/blob/06da949259349ed847f213abe23bd56c65d237d5/manticore/platforms/evm_world_state.py#L346
If the underlay is a `DefaultWorldState`, it will return 0:
https://github.com/trailofbits/manticore/blob/06da949259349ed847f213abe23bd56c65d237d5/manticore/platforms/evm_world_state.py#L141-L142

manticore/platforms/evm.py
```diff
@@ -2566,102 +2567,63 @@ def set_storage_data(self, storage_address, offset, value):
...
         """
-        storage = self._world_state[address]["storage"]
-        array = storage.array
-        while not isinstance(array, ArrayVariable):
-            if isinstance(array, ArrayStore):
-                return True
-            array = array.array
-        return False
-
-    def get_storage(self, address):
```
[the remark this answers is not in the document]
smoelius: The concern is that the lack of this method could break external scripts?

I ask because the old `get_storage` method returned a `Dict` or `ArrayProxy` (I think). I am afraid that supporting that API faithfully could be a heavy lift.

manticore/platforms/evm_world_state.py
smoelius: A bad URL now causes Manticore to raise an `EthereumError`.

manticore/platforms/evm_world_state.py
smoelius: I think it should be 1.  If the underlay is a `RemoteWorldState` and the value at that offset is non-zero, then the `map` should reflect that the overlay differs from the underlay at that offset.

manticore/platforms/evm_world_state.py
smoelius: When would the default be different?

manticore/platforms/evm.py
```diff
@@ -2566,102 +2567,63 @@ def set_storage_data(self, storage_address, offset, value):
...
-        :param address: account address
-        :return: account storage
-        :rtype: bytearray or ArrayProxy
-        """
-        return self._world_state[address]["storage"]
+    def _get_storage(self, constraints: ConstraintSet, address: int) -> Storage:
+        """Private auxiliary function to retrieve the storage"""
+        storage = self._world_state.get_storage(address)
+        if storage is None:
+            storage = Storage(constraints, address)
```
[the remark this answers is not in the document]
smoelius: FWIW, I hate this method.  But I agree that the behavior you suggest is more consistent.

manticore/platforms/evm.py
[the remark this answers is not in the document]
ehennenfent: > The concern is that the lack of this method could break external scripts?

Yeah, since the old method was reasonably well-documented and appeared on a fairly prominent object, I'm worried that removing it might break user scripts. Then again, I'm not entirely sure what the purpose of this function is. It looks like it was called in a few places internally, which you've replaced with calls to `_get_storage`. 

Ideally we'd have a deprecation period before removing any APIs, but this change should be okay so long as we document it, and there's some kind of workaround available to achieve the same functionality as we had before. Would it make sense to have `get_storage` as a this wrapper around `_get_storage` and instruct users that they'll need to process the returned data somewhat differently? Or should we direct them to something like `get_storage_data` or `get_storage_items`?

manticore/platforms/evm_world_state.py
smoelius: I added such a message.

manticore/platforms/evm.py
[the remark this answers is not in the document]
smoelius: I re-added the method with return type `Storage`:
https://github.com/trailofbits/manticore/blob/954e92f08067745f8d5a004bc97acc770517ddad/manticore/platforms/evm.py#L2580
And I implemented `__getitem__` for `Storage`:
https://github.com/trailofbits/manticore/blob/954e92f08067745f8d5a004bc97acc770517ddad/manticore/platforms/evm_world_state.py#L48
So, this should satisfy any users of this method, regardless of whether they were expecting a dictionary or symbolic array.

manticore/ethereum/manticore.py
```diff
@@ -997,28 +1000,31 @@ def multi_tx_analysis(
...
-            create_value = 0
-        contract_account = self.solidity_create_contract(
-            solidity_filename,
-            contract_name=contract_name,
-            owner=owner_account,
-            args=args,
-            compile_args=compile_args,
-            balance=create_value,
-        )
+        if contract_account is None:
```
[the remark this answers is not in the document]
smoelius: I added these checks:
https://github.com/trailofbits/manticore/blob/1d26902093080cbf236e3b6db8ed351eae09cab7/manticore/ethereum/cli.py#L146-L147
https://github.com/trailofbits/manticore/blob/1d26902093080cbf236e3b6db8ed351eae09cab7/manticore/ethereum/manticore.py#L1028-L1029

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,445 @@
...
+        self.dirty = False
+
+    def set(self, offset: Union[int, BitVec], value: Union[int, BitVec]):
+        self.map[offset] = 1
+        self.data[offset] = value
+        self.dirty = True
+
+    def dump(self, stream: TextIOBase, state: State):
+        concrete_indexes = set()
+        for sindex in self.map.written:
```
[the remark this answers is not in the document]
smoelius: As we discussed offline, what we are really using from `map` is its `written` field.  So, I think this code should be okay.

manticore/platforms/evm.py
```diff
@@ -2566,102 +2567,63 @@ def set_storage_data(self, storage_address, offset, value):
...
+    def add_to_balance(self, address: int, value: Union[int, BitVec]):
+        new_balance = self._world_state.get_balance(address) + value
+        self.set_balance(address, new_balance)
+        return new_balance
 
-    def send_funds(self, sender, recipient, value):
-        self._world_state[sender]["balance"] -= value
-        self._world_state[recipient]["balance"] += value
+    def send_funds(self, sender: int, recipient: int, value: Union[int, BitVec]):
+        self.add_to_balance(sender, -value)
```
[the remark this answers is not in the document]
smoelius: `send_funds` is used in several places where the balance can be symbolic, e.g., here:
https://github.com/trailofbits/manticore/blob/ee206c47ceb6508a897c728f9705fe7822cbc9ff/manticore/platforms/evm.py#L2429
So I think this should be okay.

manticore/platforms/evm.py
```diff
@@ -2984,7 +2934,7 @@ def _process_pending_transaction(self):
                 f"Caller account {hex(caller)} does not exist; valid accounts: {list(map(hex, self.accounts))}"
             )
 
-        if address not in self.accounts:
+        if not self._world_state.is_remote() and address not in self.accounts:
```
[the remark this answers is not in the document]
smoelius: > We could move this to self.**contains** ? (or remove self.**containts**)

Sorry, I assume you mean `self.constraints`, but I don't understand the relevance here.

> Can not we ask the node if the account exists?

Ask the node if the account exists and, if not, create it?  Maybe, but that causes me some concern.  The `owner` and `attacker` accounts should have already been created:
https://github.com/trailofbits/manticore/blob/ee206c47ceb6508a897c728f9705fe7822cbc9ff/manticore/ethereum/manticore.py#L1009-L1010
So, I am afraid that creating an account here could hide errors, either in user scripts or in Manticore itself.

> Maybe self.accounts should be a generator so it works with both emulated and remote?

`self.accounts` is an established method that returns a dictionary:
https://github.com/trailofbits/manticore/blob/522896264c731a440f616ac1c9a5d5726dd595b0/manticore/ethereum/manticore.py#L364-L366
Also, it's not clear to me how making `self.accounts` a generator would facilitate working with a remote node. There is an RPC method to return the accounts owned by a node:
https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_accounts
But I suspect this is less than what you want(?).

manticore/platforms/evm.py
```diff
@@ -3063,43 +3013,32 @@ def dump(self, stream, state, mevm, message):
...
                 all_used_indexes = []
                 with state.constraints as temp_cs:
                     # make a free symbolic idex that could address any storage slot
                     index = temp_cs.new_bitvec(256)
                     # get the storage for account_address
-                    storage = blockchain.get_storage(account_address)
+                    storage = blockchain._get_storage(temp_cs, account_address)
                     # we are interested only in used slots
-                    # temp_cs.add(storage.get(index) != 0)
-                    temp_cs.add(storage.is_known(index))
```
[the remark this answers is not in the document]
smoelius: As we discussed offline, written and is_known are still needed.  What would be ideal would be if the written/is_known functionality could be separated out from `ArrayProxy` and put into its own class.

manticore/platforms/evm_world_state.py
smoelius: I don't think that should be hard to add. What specifically should be the test?

manticore/platforms/evm.py
```diff
@@ -2566,102 +2567,63 @@ def set_storage_data(self, storage_address, offset, value):
         :param value: the value to write
         :type value: int or BitVec
         """
-        self._world_state[storage_address]["storage"][offset] = value
-
-    def get_storage_items(self, address):
```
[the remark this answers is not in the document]
smoelius: I added the method back.

manticore/platforms/evm_world_state.py
feliam: So...
`self.data.is_known(offset)`  is already doing map's work then?
And `len(self.data.written)` the work of self.dirty ?

smoelius: > So...
> `self.data.is_known(offset)` is already doing map's work then?

Sort of. I think `self.data.is_known(offset)` should hold iff `self.map.is_known(offset)`. So, in that sense, map and data's `written` fields do the same work.

The advantage to splitting them is that it allows to construct this conditional:
https://github.com/trailofbits/manticore/blob/1d26902093080cbf236e3b6db8ed351eae09cab7/manticore/platforms/evm_world_state.py#L54
It might be possible to do away with map and use `self.data.written` or `self.data.is_known` directly. I am trying to respect the encapsulation properties of `self.data`.

> And `len(self.data.written)` the work of self.dirty ?

I think this may be true.

feliam: https://github.com/trailofbits/manticore/blob/1d26902093080cbf236e3b6db8ed351eae09cab7/manticore/platforms/evm_world_state.py#L54
`BitVecITE(256, self.map[offset] != 0, self.data[offset], default) ` !=?  `self.data.get(offset, default) `

manticore/platforms/evm_world_state.py
feliam: There was an EIP that defined different initial nonces for contracts vs. normal accounts but that should be enforced upstream probably

manticore/platforms/evm_world_state.py
feliam: Not sure if it is a valid concern. But when blockNumber is <9999 the transactions are supposed to do one thing and that thing changes with the block. 
If the block is pre istanbul the results may no be reproducible?

No idea how this works in test nets.

smoelius: I mean, what should the check be specifically?  Maybe generate a warning if the blocknumber is less than 4370000, to be consistent with this?
https://github.com/trailofbits/manticore/blob/88ddfbca9efcaa47f736183bee4fd4c6d3ef9b66/manticore/platforms/evm_world_state.py#L169-L171

manticore/platforms/evm_world_state.py
smoelius: It looks like we are doing the right thing:
https://github.com/trailofbits/manticore/blob/88ddfbca9efcaa47f736183bee4fd4c6d3ef9b66/manticore/platforms/evm.py#L2779-L2781
Nonetheless, I added some checks to `test_general.py`:
https://github.com/trailofbits/manticore/pull/1580/commits/88ddfbca9efcaa47f736183bee4fd4c6d3ef9b66

manticore/platforms/evm_world_state.py
smoelius: Indeed, it looks like you are right!  I eliminated the use of `self.map`.

ehennenfent (dismissed):

ehennenfent:
Whoops, sorry about that. Apparently "Approve changes" doesn't just dismiss my request for changes, it approves the whole PR. 

Anyway, the changes you've made in response to my comments are looking good to me. I've dismissed my change request, but as Felipe still has some comments open, I'll defer to him before we approve & merge.

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,478 @@
...
+        self.data = constraints.new_array(
+            index_bits=256,
+            value_bits=256,
+            name=f"STORAGE_DATA_{address:x}",
+            avoid_collisions=True,
+            # sam.moelius: The use of default here creates unnecessary if-then-elses.  See
+            # ArrayProxy.get in expression.py.
+            # default=0,
+        )
+        self.dirty = False
```
feliam, line 36: `self.dirty != ( len(self.data.written) !=0 )` ?

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,478 @@
...
+        # storage.  However, this is not currently implemented.
+        if storage is not None:
+            # sam.moelius: At present, the get_storage methods of both DefaultWorldState and
+            # RemoteWorldState raise NotImplementedError.  So this exception should be unreachable.
+            raise NotImplementedError("Merging of storage is not implemented")
+        storage = self._storage.get(address)
+        return storage
+
+    def get_storage_data(
+        self, constraints: ConstraintSet, address: int, offset: Union[int, BitVec]
```
feliam, line 373: note to self. Why do we need the constraints here

manticore/platforms/evm_world_state.py
smoelius: I suspect they are are equal.

Accessing an object's data members feels wrong to me, but I take it you prefer that to having `dirty`?

Let me know.  I am happy to make the change.

manticore/platforms/evm_world_state.py
smoelius: Technically, no, but: https://github.com/trailofbits/manticore/blob/eecd9b5a67b25e2cabe55ffc783e7a0ea65d7bde/manticore/platforms/evm_world_state.py#L106-L107

smoelius: At one point I was toying with the idea of creating a new `Storage` here:
https://github.com/trailofbits/manticore/blob/eecd9b5a67b25e2cabe55ffc783e7a0ea65d7bde/manticore/platforms/evm_world_state.py#L383-L384

manticore/platforms/evm_world_state.py
feliam: I am operating on the idea that if the member is not prepended with an underscore it is good for general use. :shrug: 
But other than that the question I'm trying to answer is why do we need Storage instead of ArrayProxy. (Other than its better naming)

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,478 @@
...
+# written.
+
+
+class Storage:
+    def __init__(self, constraints: ConstraintSet, address: int):
+        """
+        :param constraints: the ConstraintSet with which this Storage object is associated
+        :param address: the address that owns this storage
+        """
+        self.constraints = constraints
```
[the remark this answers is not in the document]
smoelius: Probably not.

Seeing the constraints change in subsequent calls was making me nervous, but that is probably okay.

Lat me look at it one more time tomorrow with fresh eyes.  If it still looks okay, I'll make the change.

manticore/platforms/evm_world_state.py
feliam: Why get_storage_data is responsible of creating new "Storage" ?

We learn that the account need an Storage when it is actually read?

smoelius: My concern is: will there be cases where we want to make assertions about a contract's Storage before it has been written to?  By "make assertions," I mean something like this:
https://github.com/trailofbits/manticore/blob/55e2486f4791757b77fd16b0dfe65d4ecb6e3e80/tests/other/test_smtlibv2.py#L146-L148
Or, more concretely, would we want to allow something like this?
```
    cs.add(world_state.get_storage_data(address, offset) == ord('A'))
```
Right now, this won't work if `address`'s Storage has not been written to.

I *think* we want this to not work. We currently take a "closed world" approach in the sense that the only addresses we acknowledge in the overlay are those to which we have observed writes. While this may sound restrictive, it actually facilitates solving certain types of constraint sets. An example is here:
https://github.com/trailofbits/manticore/blob/eecd9b5a67b25e2cabe55ffc783e7a0ea65d7bde/tests/ethereum/test_storage.py#L235-L238
That `assertFalse` would fail if we did not have this "closed world" assumption.

Having said that, there is a tiny voice in my head that says this may not be the right approach, i.e., there may be situations where we want to allow assertions on a contract's Storage, even though we have not observed writes to it.

Adding to this, `set_storage_data` requires a `constraints` parameter. So requiring it also for `get_storage_data` does not seem that jarring.

manticore/platforms/evm_world_state.py
smoelius: At this point, it looks likes Storage is just a thin wrapper around `ArrayProxy`. It adds a few methods and set some defaults in the constructor, but that's about it.

manticore/platforms/evm.py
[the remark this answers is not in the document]
feliam: > > We could move this to self.**contains** ? (or remove self.**containts**)
> 
> Sorry, I assume you mean `self.constraints`, but I don't understand the relevance here.
> 
`address in self` resolves to `self.__contains__(address)` I think. So we can move the implementation of it to each flavor of  WorldState. Each WorlState should be able to know if it contains certain account, no?

> > Can not we ask the node if the account exists?
> 
> Ask the node if the account exists and, if not, create it? 
No, I was not trying to find where to create them. I just want the WorldState to have the responsability of knowing which accounts it has.

Though I would create the account when the user wants to create them AND I'd bring them from the node when it is references as a target of a transaction.

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,468 @@
...
+
+    def get(self, offset: Union[int, BitVec], default: Union[int, BitVec]) -> Union[int, BitVec]:
+        if not isinstance(default, BitVec):
+            default = BitVecConstant(256, default)
+        return self._data.get(offset, default)
+
+    def set(self, offset: Union[int, BitVec], value: Union[int, BitVec]):
+        self._data[offset] = value
+
+    def get_items(self) -> List[Tuple[Union[int, BitVec], Union[int, BitVec]]]:
```
feliam, line 46: Looks like an Array method? 
_Need to check where it is used_

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,468 @@
...
+            stream.write(
+                f"storage[{index:x}] = {state.solve_one(self._data[index], constrain=True):x}\n"
+            )
+
+
+####################################################################################################
+
+
+class WorldState:
+    def new_storage(
```
feliam, line 69: What would be the docstring of this method? Is this a convenience method?

I'm struggling to think when do we want a storage without an account.
I expect a world state to have add_account(), delete_account(), accounts()

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,468 @@
...
+        self, constraints: ConstraintSet, address: int, items: Optional[Dict[int, int]] = None
+    ) -> Storage:
+        storage = Storage(constraints, address)
+        if items is not None:
+            for key, value in items.items():
+                storage.set(key, value)
+        return storage
+
+    @abstractmethod
+    def is_remote(self) -> bool:
```
feliam, line 79: I'm guessing I'll be better the less we actually use this method.

manticore/platforms/evm_world_state.py
```diff
@@ -0,0 +1,468 @@
...
+    @abstractmethod
+    def has_storage(self, address: int) -> bool:
+        pass
+
+    @abstractmethod
+    def get_storage(self, address: int) -> Optional[Storage]:
+        pass
+
+    # sam.moelius: The addition of the constraints parameter is future-proofing.  We might want
+    # to create a new Storage object in OverlayWorldState when self._storage.get(address) is None.
```
feliam, line 103: In which case a Storage of a known address will be None?

manticore/platforms/evm_world_state.py
feliam: 3 subcases here:
    ** We know all the already used in the node
    ** We know some of the used slots in the node
    ** We do not know any of the already used slots in the node


Lets assume it is not possible to get all known slots from a storage in the node. :( Booo
So the known slots could start to be just all non zero storage items from offsets 0 to 100 ?

```known_slots = (0,1,2,3)```

This will account for the static variables. Maybe that's enough?
Other static analysis over the bytecode can be done. We get all those and write them to our model.


So initially when an unknown account is referenced (as a tx target) we bring it 
from the node and build an OverlayedRemoteWorld and storage. We create the overlay 
account with an overlayed array storage and write the known values to it (The ones from 0 to 100 for now?).

```
class OStorage:
    array :ArrayProxy
    node :Node
```

READ
====
Now suppose that storage is then read at a free symbolic index: index

oStorage.read(index) -> ?????

I can imagine 2 things to happen here. And it may even request a fork if both are possible at the same time.

* index is known.

The ArrayProxy put in front of the node knows about the index. `array.is_known(index) -> True`.
The result of reading here is simply `array[index]`. We do not need to use a default as we know the index is known.
This returns a symbol representing one value read from the full _known_ storage.


* index is not known 

The ArrayProxy put in front of the node does not knows about the index. `array.is_known(index) -> False`.
We need to look for the value in the node.
But where? The index is symbolic. Which actual storage slot do we bring down from the node?
This sucks.
Suck 1: Ask the solver and sample solutions for index. (note there are no _known_ yet)
        Get the corresponding values from the node. 
        Print a warning about incompletness (if not ALL solutions where retrieved)
        Write them to the array
        Constrain index to be one of the solutions
        Index is now _known_ then go to the previous case

Suck 2:
        Assume all storage in the node above 100 are blank
        Go to previous case

WRITE
=====
Writing to a storage slot requires reading it first. So it will resolve itself when solving reading

manticore/platforms/evm_world_state.py
smoelius: Yes, totally agree.

manticore/platforms/evm_world_state.py
smoelius: > Looks like an Array method?

Sorry, I don't understand the question.  What is mean by "an Array method"?

> _Need to check where it is used_

It is currently used only in `EVMWorld.get_storage_items` (which I added back in abe52bc).

manticore/platforms/evm_world_state.py
smoelius: > What would be the docstring of this method? Is this a convenience method?

I added the docstring. It is a convenience method.
https://github.com/trailofbits/manticore/blob/235d0b5796324956b332cceb69778f88d74688f2/manticore/platforms/evm_world_state.py#L80-L85

> I'm struggling to think when do we want a storage without an account.
> I expect a world state to have add_account(), delete_account(), accounts()

I don't think those methods make sense for the remote model, because they presume that an account "exists" or "does not exist". What is more accurate is there are things that we know about an address and things that we do not, which could include nonce, balance, storage, and code. That is why I espouse the removal of the `accounts` method.

manticore/platforms/evm_world_state.py
smoelius: That sounds right.

manticore/platforms/evm_world_state.py
feliam: I mean that it looks like it would be better suited as an member method of the type Array.
It does not use anything of Storage object and it pips into Array's secret-ish elements.

Maybe we should consider having this method as a member method of Array. (If that's not already there?)

manticore/platforms/evm_world_state.py
feliam: They do not make sense for the special case of a remote world. I guess those will not be implemented there.  And the layer on top of it will _need_ to handle that functionality. 

World State looks like something that needs a  way to add, delete and list account (or to check if it contains an account)

manticore/platforms/evm_world_state.py
feliam: When the storage of an account that exists will be None? Is this for the special case of a remote world?

manticore/platforms/evm_world_state.py
smoelius: > And the layer on top of it will _need_ to handle that functionality.

Why do you think those methods are needed?

manticore/platforms/evm_world_state.py
smoelius: Essentially, yes.

manticore/platforms/evm_world_state.py
feliam: A method to add an account? So the user can add accounts to the ethereum world.
A method to del an account? I guess selfdestruct wants it.
A method to list accounts?  To concretize target addresses for example.
A method to check if an account exists? You need to know if the account is already there to calculate gas in tx instructions sometimes.

ehennenfent:
Now that #1676 is merged, we should update this PR from master so that all of those conflicts will be fixed before final review.

PR #1583, opened 2020-01-07
Improve Pickle and Multiprocessing compatibility

ehennenfent:
~Minor improvements to Manticore's compatibility with pickle, made while experimenting with running it on OSX.~

~Changing the pickle protocol version will break the copying of states between instances of Manticore running on different minor versions of Python 3. I'm assuming this won't be a problem for anyone, but let me know.~

* Bumps the Pickle protocol version to use the highest available protocol on any given platform.
* Makes the signal handler for the multiprocessing manager serializable
* Adds sneaky fallback to multiprocessing.threading on platforms that don't support multiprocessing.processes. Not implying you should ever use this...


ehennenfent:
Incidentally, with this PR in place and some tweaks to the multiprocessing settings, you can run Manticore on OSX. The [if_check](https://github.com/trailofbits/manticore/blob/master/examples/wasm/if_check/solve.py) example becomes the following:

```python
from manticore.wasm import ManticoreWASM
from manticore.core.plugin import Plugin
from manticore.utils import config

# No `fork` support outside of linux, and Manticore doesn't work with `spawn`,
# so we need to use threads instead of processes. https://bugs.python.org/issue33725
consts = config.get_group("core")
consts.mprocessing = consts.mprocessing.threading


def getchar(state):
    """ Symbolic `getchar` implementation. Returns an arbitrary single byte """
    res = state.new_symbolic_value(32, "getchar_res")
    state.constrain(0 < res)
    state.constrain(res < 256)
    return [res]


class PrintRetPlugin(Plugin):
    """ A plugin that looks for states that returned zero and solves for their inputs """

    def will_terminate_state_callback(self, state, *args):
        retval = state.stack.peek()
        if retval == 0:
            print("Solution found!")
            for sym in state.input_symbols:
                solved = state.solve_one(sym)
                print(f"{sym.name}: {chr(solved)} --> Return {retval}")


def solve():
    # Pass our symbolic implementation of the `getchar` function into the WASM environment
    # as an import.
    m = ManticoreWASM("if_check.wasm", env={"getchar": getchar})

    # Register our state termination callback
    m.register_plugin(PrintRetPlugin())

    # Run the main function, which will call getchar
    m.main()

    # Save a copy of the inputs to the disk
    m.finalize()


# Multiprocessing on OSX wants the code to be called in this manner 
if __name__ == "__main__":
    solve()
```

PR #1589, opened 2020-01-11
Improve WASM Memory Ergonomics

ehennenfent:
Adds several helper methods to the WASM memory so as to conform to the standard laid out by native modules. Also lazily creates memory pages so that binaries can request large amounts of memory without Manticore needing to fill an appropriately large region with 0's.

### DIFF
--- a/manticore/wasm/structure.py
+++ b/manticore/wasm/structure.py
@@ -760,6 +760,10 @@ class ModuleInstance(Eventful):
     export_map: typing.Dict[str, int]
     #: Contains instruction implementations for all non-control-flow instructions
     executor: Executor
+    #: Stores names of store functions, if available
+    function_names: typing.Dict[FuncAddr, str] = {}
+    #: Stores names of local variables, if available
+    local_names: typing.Dict[FuncAddr, typing.Dict[int, str]] = {}
     #: Stores the unpacked sequence of instructions in the order they should be executed
     _instruction_queue: typing.Deque[Instruction]
     #: Keeps track of the current call depth, both for functions and for code blocks. Each function call is represented
@@ -782,8 +786,8 @@ class ModuleInstance(Eventful):
         self.exports = []
         self.export_map = {}
         self.executor = Executor(constraints)
-        self.function_names: typing.Dict[FuncAddr, str] = {}
-        self.local_names: typing.Dict[FuncAddr, typing.Dict[int, str]] = {}
+        self.function_names = {}
+        self.local_names = {}
         self._instruction_queue = deque()
         self._block_depths = [0]
         self._state = None
--- a/manticore/wasm/structure.py
+++ b/manticore/wasm/structure.py
@@ -1064,7 +1064,7 @@ class ModuleInstance(Eventful):
 
         name = self.function_names.get(funcaddr, f"Func{funcaddr}")
         buffer = " | " * (len(self._block_depths) - 1)
-        logger.debug(buffer + "%s(%s)" % (name, ", ".join(str(i) for i in local_vars)))
+        print(buffer + "%s(%s)" % (name, ", ".join(str(i) for i in local_vars)))
         if isinstance(f, HostFunc):  # Call native function
             self._publish("will_call_hostfunc", f, local_vars)
             res = list(f.hostcode(self._state, *local_vars))
--- a/manticore/wasm/structure.py
+++ b/manticore/wasm/structure.py
@@ -1064,7 +1064,7 @@ class ModuleInstance(Eventful):
 
         name = self.function_names.get(funcaddr, f"Func{funcaddr}")
         buffer = " | " * (len(self._block_depths) - 1)
-        print(buffer + "%s(%s)" % (name, ", ".join(str(i) for i in local_vars)))
+        logger.debug(buffer + "%s(%s)" % (name, ", ".join(str(i) for i in local_vars)))
         if isinstance(f, HostFunc):  # Call native function
             self._publish("will_call_hostfunc", f, local_vars)
             res = list(f.hostcode(self._state, *local_vars))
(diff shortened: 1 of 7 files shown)


manticore/wasm/structure.py
```diff
@@ -442,7 +454,31 @@ def load(cls, filename: str):
...
-            # TODO - custom sections (https://www.w3.org/TR/wasm-core-1/#custom-section%E2%91%A0)
+            elif sec_id == SEC_UNK:
+                # WASM module renders all types as `GeneratedStructureData` so `isinstance` doesn't work :(
+                if (
+                    hasattr(section, "name_type")
+                    and hasattr(section, "payload_len")
+                    and hasattr(section, "payload")
+                ):
+                    name_type = section_data.name_type
+                    if name_type == 0:  # module name
```
ekilmer: nit: Should we make an `enum` for the values of `name_type`?

manticore/wasm/structure.py
```diff
@@ -442,7 +454,31 @@ def load(cls, filename: str):
...
+                            )
+                    elif name_type == 2:  # local variable names
+                        for func in section_data.payload.funcs:
+                            func_idx = func.index
+                            for n in func.local_map.names:
+                                ty = n.get_decoder_meta()["types"]["name_str"]
+                                m.local_names.setdefault(FuncAddr(func_idx), {})[
+                                    n.index
+                                ] = strip_quotes(ty.to_string(n.name_str))
+                # TODO - other custom sections (https://www.w3.org/TR/wasm-core-1/#custom-section%E2%91%A0)
```
ekilmer: Unsupported sections should probably log a warning or debug message

manticore/wasm/structure.py
```diff
@@ -488,8 +523,60 @@ class MemInst:
     https://www.w3.org/TR/wasm-core-1/#memory-instances%E2%91%A0
     """
 
-    data: typing.List[int]  #: The backing array for this memory
+    _published_events = {"write_memory", "read_memory"}
```
ekilmer: I'm surprised `mypy` isn't unhappy. Usually it yells at me for missing class fields with type annotations.

Should be something like 

```suggestion
    _published_events: typing.Set[str] = {"write_memory", "read_memory"}
```

manticore/wasm/structure.py
```diff
@@ -488,8 +523,60 @@ class MemInst:
...
+
+    _pages: typing.Dict[int, typing.List[int]]
     max: typing.Optional[U32]  #: Optional maximum number of pages the memory can contain
+    _current_size: int  # Tracks the theoretical size of the memory instance, including unmapped pages
+
+    def __init__(self, starting_data, max=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        pagesize = 2 ** 16
```
ekilmer: Can this value change? Should it be an exposed constant or class member?

manticore/wasm/structure.py
```diff
@@ -500,16 +587,68 @@ def grow(self, n: int) -> bool:
...
+
+        :param base: Address to read the int from
+        :param size: Size of the int (in bits)
+        :return: The int in question
+        """
+        return Operators.CONCAT(
+            size, *map(Operators.ORD, reversed(self.read_bytes(base, size // 8)))
+        )
+
+    def read_bytes(self, base, size):
```
ekilmer: Nit: Possible to add type annotations to these functions?

manticore/wasm/structure.py
```diff
@@ -488,8 +523,60 @@ class MemInst:
...
+    def __contains__(self, item):
+        return item in range(self.npages * (2 ** 16))
+
+    def _check_initialize_index(self, memidx):
+        page = memidx // (2 ** 16)
+        if page not in range(self.npages):
+            raise OutOfBoundsMemoryTrap(memidx)
+        if page not in self._pages:
+            self._pages[page] = [0x0] * (2 ** 16)
+        return divmod(memidx, 2 ** 16)
```
ekilmer: Looks like the `pagesize` value is used in a few places here again.

tests/wasm/generate_tests.sh
```diff
@@ -40,4 +40,6 @@ chmod +x gen.sh
 cat modules.txt | xargs -n1 -P"$cores" ./gen.sh
 rm gen.sh
 
+mv test_callbacks.skip test_callbacks.py
```
ekilmer: Why is this needed?

ehennenfent: The WASM examples don't exhibit all the behaviors that Manticore can fire callbacks for, so `test_callbacks.skip` draws on examples from the spec tests for things like global variable usage, traps, and memory usage. Since those tests won't exist unless you've run generate_tests.sh, I disabled `test_callbacks` until you've done so by renaming the file.

ekilmer: Fair enough! My main concern was that the `test_callbacks.skip` might not be formatted or linted (if we ever eventually lint tests) by tools that look for the `*.py` extension.

manticore/wasm/structure.py
ehennenfent: It's because it inherits from Eventful, which already specifies the type.
`_published_events: Set[str] = set()`

manticore/wasm/structure.py
ehennenfent: Done. The type hints only describe concrete argument, which is fine since that's all we support right now. If we add a symbolic memory model we'll need to update them.

manticore/wasm/structure.py
ehennenfent: There are hardcoded constants in the `wasm` module, but it looks like they're out of date relative to the spec. I added a comment pointing to the spec itself; hopefully that will suffice.

PR #1592, opened 2020-01-21
Use dill instead of pickle

ekilmer:
> [`dill`](https://pypi.org/project/dill/) extends python’s `pickle` module for serializing and de-serializing python objects to the majority of the built-in python types.

> dill can also pickle more ‘exotic’ standard types:
> * functions with yields, nested functions, lambdas,
> * cell, method, unboundmethod, module, code, methodwrapper,
> * dictproxy, methoddescriptor, getsetdescriptor, memberdescriptor,
> * wrapperdescriptor, xrange, slice,
> * notimplemented, ellipsis, quit

### DIFF
--- a/manticore/utils/helpers.py
+++ b/manticore/utils/helpers.py
@@ -1,5 +1,5 @@
 import logging
-import pickle
+import dill as pickle
 import sys
 from collections import OrderedDict
 
--- a/scripts/stats.py
+++ b/scripts/stats.py
@@ -1,4 +1,5 @@
-import sys, os, pickle, argparse
+import sys, os, argparse
+import dill as pickle
 
 
 # Process the data of a workspace and prints out some global information
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,7 @@ setup(
     packages=find_packages(exclude=["tests", "tests.*"]),
     python_requires=">=3.6",
     install_requires=[
+        "dill",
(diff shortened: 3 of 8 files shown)


ehennenfent:
Is there something you're having trouble serializing with pickle? Looks like dill incurs a fairly significant performance hit: https://stackoverflow.com/questions/37906154/dill-vs-cpickle-speed-difference

ekilmer:
>  Looks like dill incurs a fairly significant performance hit

Oh, interesting. @woodruffw any thoughts? 

> Is there something you're having trouble serializing with pickle?

Not exactly, since I was able to work around it, but there were some partial and nested functions that I tried to serialize and would fail, but I ended up being able to refactor, which might have been for the better, now that I think about it.

woodruffw:
> > Looks like dill incurs a fairly significant performance hit
> 
> Oh, interesting. @woodruffw any thoughts?

Didn't know this! I had assumed that dill would be faster, but looks like that was wrong. In either case: is the performance of our serialization method of choice a dominating factor in Manticore's execution time? My intuition is that even with dill's slowdown we're still being dominated by Z3, followed by context switches when doing anything with Unicorn. But I could be wrong.

If we've fixed this on the consumer side @ekilmer then I'm inclined to say that we can punt here, and keep it in mind should we run into more problems.

PR #1596, opened 2020-01-24
Streamline State Serialization Interface

ehennenfent:
Adds a `SerializeState` exception that (ab)uses the concretization system to save a copy of the current state to the disk. Also adds a helper method to ManticoreBase that makes resuming execution from a saved state a bit cleaner. Overall, this should allow you to pause a concrete execution halfway through and then resume your analysis from there as many times as desired. 

Currently, this works for WASM and native. I haven't tried it with EVM. On WASM, the `run` method will happily start executing whatever's in the instruction queue without trying to invoke any functions. Likewise, the native `run` method starts executing wherever the PC is pointing - the call to `_start` actually takes place during the initial state creation.

This isn't a full-blown checkpoint system. From the tests, you'll note that plugins and the shared context are not carried over. Also, since it only serializes a single state, if you want to resume execution after a point where you've forked on symbolic data, you'll need to rebuild the state lists yourself.

### DIFF
--- a/manticore/core/manticore.py
+++ b/manticore/core/manticore.py
@@ -311,6 +311,24 @@ class ManticoreBase(Eventful):
     def __str__(self):
         return f"<{str(type(self))[8:-2]}| Alive States: {self.count_ready_states()}; Running States: {self.count_busy_states()} Terminated States: {self.count_terminated_states()} Killed States: {self.count_killed_states()} Started: {self._running.value} Killed: {self._killed.value}>"
 
+    @classmethod
+    def from_saved_state(cls, filename: str, *args, **kwargs):
+        """
+        Creates a Manticore object starting from a serialized state on the disk.
+
+        :param filename: File to load the state from
+        :param args: Arguments forwarded to the Manticore object
+        :param kwargs: Keyword args forwarded to the Manticore object
+        :return: An instance of a subclass of ManticoreBase with the given initial state
+        """
+        from ..utils.helpers import PickleSerializer
+
+        fd = open(filename, "rb")
+        deserialized = PickleSerializer().deserialize(fd)
+        fd.close()
+
+        return cls(deserialized, *args, **kwargs)
+
     def _fork(self, state, expression, policy="ALL", setstate=None):
         """
         Fork state on expression concretizations.
--- a/manticore/core/manticore.py
+++ b/manticore/core/manticore.py
@@ -323,9 +323,8 @@ class ManticoreBase(Eventful):
         """
         from ..utils.helpers import PickleSerializer
 
-        fd = open(filename, "rb")
-        deserialized = PickleSerializer().deserialize(fd)
-        fd.close()
+        with open(filename, "rb") as fd:
+            deserialized = PickleSerializer().deserialize(fd)
 
         return cls(deserialized, *args, **kwargs)
 
--- a/tests/native/test_resume.py
+++ b/tests/native/test_resume.py
@@ -1,0 +1,62 @@
+import unittest
+from manticore.native import Manticore
+from manticore.core.plugin import Plugin
+from manticore.core.state import SerializeState, TerminateState
+from pathlib import Path
+
+
+class InsnCounterPlugin(Plugin):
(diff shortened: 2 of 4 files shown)


manticore/core/manticore.py
```diff
@@ -311,6 +311,24 @@ def __init__(self, initial_state, workspace_url=None, policy="random", **kwargs)
...
+        :param filename: File to load the state from
+        :param args: Arguments forwarded to the Manticore object
+        :param kwargs: Keyword args forwarded to the Manticore object
+        :return: An instance of a subclass of ManticoreBase with the given initial state
+        """
+        from ..utils.helpers import PickleSerializer
+
+        fd = open(filename, "rb")
+        deserialized = PickleSerializer().deserialize(fd)
+        fd.close()
```
ekilmer: Nit: Why not use `with open(filename, "rb") as f:`?

tests/native/test_resume.py
```diff
@@ -0,0 +1,62 @@
...
+from manticore.core.plugin import Plugin
+from manticore.core.state import SerializeState, TerminateState
+from pathlib import Path
+
+
+class InsnCounterPlugin(Plugin):
+    def did_execute_instruction_callback(self, state, last_pc, pc, instruction):
+        with self.locked_context("counter", dict) as ctx:
+            val = ctx.setdefault(instruction.mnemonic, 0)
+            ctx[instruction.mnemonic] = val + 1
```
ekilmer: I'm curious about the `with self.locked_context`.  This seems to be a context shared by all states, correct?

When saving and loading a single state, there could be other states generated by forking with program exploration, so I think these counts may be off in the tests after saving and loading only one state and comparing it to a full run (`counts_canonical`). Would keying the counts with `state.id` or something be better?

ekilmer:
This features sounds super awesome!

manticore/core/manticore.py
ehennenfent: Force of habit. I try not to create variables in indented blocks, but since `with` doesn't create a new scope, it'd be fine here. Better, actually, since Python would still close the file descriptor in the event of a deserialization exception.

tests/native/test_resume.py
ehennenfent: The instruction I put the hook at should be before any forking takes place, and the total counts across all states should be consistent. At least, I think they should. This test works just fine on OSX (cheating, I know) but fails on linux, with most of the instruction counts being slightly off. They're not _gone_ in either case, so it doesn't seem like a multiprocessing issue as I had initially expected. Perhaps we are missing a state somewhere? I'll keep looking into it...

ekilmer: > The instruction I put the hook at should be before any forking takes place

Maybe a comment in the code about this at the hooked address would help, but this makes sense otherwise.

manticore/core/manticore.py
```diff
@@ -311,6 +311,24 @@ def __init__(self, initial_state, workspace_url=None, policy="random", **kwargs)
...
+        Creates a Manticore object starting from a serialized state on the disk.
+
+        :param filename: File to load the state from
+        :param args: Arguments forwarded to the Manticore object
+        :param kwargs: Keyword args forwarded to the Manticore object
+        :return: An instance of a subclass of ManticoreBase with the given initial state
+        """
+        from ..utils.helpers import PickleSerializer
+
+        fd = open(filename, "rb")
```
feliam:
```
with open(filename, "rb") as fd:
    deserialized = PickleSerializer().deserialize(fd)
```

tests/native/test_resume.py
ehennenfent: Hmm, for some reason, instruction counts on Linux are not consistent across runs. That would explain why the test is failing. Perhaps I need a better metric for measuring native success.

manticore/core/manticore.py
ehennenfent: Fixed in f3c2fd1

tests/native/test_resume.py
ehennenfent: Fixed, kinda. Now I check the state counts at different points in execution. Not quite as precise as the WASM tests, but close enough.

Issue #1272, opened 2018-11-19
Design and make a Text User Interface that shows progress realtime


feliam:
Design and make a TUI that shows all interesting aspects of the exploration real time. It should be a plugins that hooks on all interesting events. New events may be necessary. Ideally it should not require new python dependencies

* It should have a global "page" that shows
* It should have different "tabs" for the different states
* Each state should show _where_ it is at any time EVM and or Solidity
* Storage, memory, stack, gas should be accessible
* List of internal tx should be accessible
* Logs should be accessible
* Coverage and general statistics should be accesible
* path constraint should be accessible
* if stuff is symbolic a '?' or similar placeholder should be shown instead.
* A way to produce a concrete example of symbolic values should exists

...

Hints follows..

```
0xafb6d63079413d167770de9c3f50db6477811bdb: 0x0004: MSTORE  Save word to memory.

0xafb6d63079413d167770de9c3f50db6477811bdb: Stack                                                                           Memory
0xafb6d63079413d167770de9c3f50db6477811bdb:      address 0x0000000000000000000000000000000000000000000000000000000000000040 0000  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0xafb6d63079413d167770de9c3f50db6477811bdb:        value 0x0000000000000000000000000000000000000000000000000000000000000080 0010  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0xafb6d63079413d167770de9c3f50db6477811bdb:                                                                                 0020  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0xafb6d63079413d167770de9c3f50db6477811bdb:                                                                                 0030  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0xafb6d63079413d167770de9c3f50db6477811bdb:                                                                                 0040  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0xafb6d63079413d167770de9c3f50db6477811bdb:                                                                                 0050  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0xafb6d63079413d167770de9c3f50db6477811bdb:                                                                                 0060  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
[...]

PR #1605, opened 2020-01-30
Support for TUI

pwang00:
Added an additional thread in manticore/core/worker.py to set up socket server and communicate with TUI via serialized protobuf messages.  Currently more of a proof-of-concept than a robust implementation, but hopefully adding additional functionality won't be too difficult.

Current setup has hardcoded host and port (127.0.0.1, 1337), and successfully sends serialized states to the client TUI.  In the future, the host and port should be saved by ansible upon provisioning a DO droplet.


ehennenfent:
Closes #1272

ekilmer:
As a general rule, generated files shouldn't be checked into the repo. They should be generated when the project is built. This way, there is no chance of forgetting to update them manually.

pwang00:
> As a general rule, generated files shouldn't be checked into the repo. They should be generated when the project is built. This way, there is no chance of forgetting to update them manually.

Yeah point taken.  Atm I think the Manticore dependency list has to be updated to account for protobuf and npyscreen.

ekilmer:
Regarding the generation of protobuf files during build: Unfortunately, I don't know the best way of doing this for a python project, but I think taking a look at Google's own [`setup.py`](https://github.com/protocolbuffers/protobuf/blob/master/python/setup.py) might be a good place to start.

I can also help with taking a closer look if nothing helpful is turning up 👍

pwang00:
Actually looking into this right now, seems like a possible way to do this is to just run a post-install script via setuptools.  I'll test and push if this ends up working.

setup.py
```diff
@@ -34,6 +36,7 @@ def rtd_dependent_deps():
     "lint": lint_deps,
 }
 
+subprocess.Popen(protoc_cmd)
```
ekilmer: We should use `subprocess.checkout_output` for this https://docs.python.org/3/library/subprocess.html#subprocess.check_output

setup.py
```diff
@@ -36,7 +36,7 @@ def rtd_dependent_deps():
     "lint": lint_deps,
 }
 
-subprocess.Popen(protoc_cmd)
+output = subprocess.check_output(protoc_cmd)
```
ekilmer: Nit: If we don't use the `output` variable, then let's just get rid of it. The `check_output` function will throw a `CalledProcessError` exception when the exit code is non-zero, which should be fine in this case

ehennenfent:
> As a general rule, generated files shouldn't be checked into the repo. They should be generated when the project is built. This way, there is no chance of forgetting to update them manually.

Looks like `pip install protobuf` doesn't install the `protoc` compiler. That leaves us with two options:
* Add `protoc` as an external dependency that must be installed via `apt` before installing Manticore
* Add the generated `state_pb2.py` file back to the repository and rely on our CI to tell us when to update it

Personally I'd vote for the second option.

pwang00:
> > As a general rule, generated files shouldn't be checked into the repo. They should be generated when the project is built. This way, there is no chance of forgetting to update them manually.
> 
> Looks like `pip install protobuf` doesn't install the `protoc` compiler. That leaves us with two options:
> 
>     * Add `protoc` as an external dependency that must be installed via `apt` before installing Manticore
> 
>     * Add the generated `state_pb2.py` file back to the repository and rely on our CI to tell us when to update it
> 
> 
> Personally I'd vote for the second option.

Yeah, the user would have to (on ubuntu) `sudo apt-get install protobuf-compiler` and I suppose depending on the distribution (e.g. Amazon AMI) even finding the right package name to install could be a pain. I don't think there's an efficient way to automatically install protoc for any given Linux distro.

I would probably go with the second option as well since I think doing so would save the user a step in the installation process, and that the protocol shouldn't change *that* often across builds.

manticore/core/manticore.py
```diff
@@ -191,7 +191,7 @@ def __init__(self, initial_state, workspace_url=None, policy="random", **kwargs)
 
         During exploration Manticore spawns a number of temporary states that are
         maintained in different lists:
-
+
```
ekilmer: Looks like some dangling whitespace made it in here somehow?

PR #1604, opened 2020-01-30
Make sure std files are not None before assigning peer

ekilmer:
@bradlarsen Check if this works please

closes #1602

PR #1606, opened 2020-01-31
Add a native example program & test case for #1604

bradlarsen:


tests/native/test_integration_native.py
```diff
@@ -260,6 +260,24 @@ def test_basic_arm(self):
...
+            PYTHON_BIN,
+            "-m",
+            "manticore",
+            "--no-color",
+            "--workspace",
+            workspace,
+            filename,
+            "+++++++",
+        ]
+        subprocess.check_call(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
ekilmer: I know the other tests are written using `os.path`, but I think we should try to modernize where/when possible and use `pathlib.Path`.

The only clunky thing might be that we have to convert the `Path` object to a `str` in the `cmd` list.

ekilmer: I think we should omit the `stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL` arguments because `pytest` will automatically hide the output unless there's a failure. However, even in the case of a failure, with the `DEVNULL` redirections, we won't get _any_ output.

bradlarsen: Good catch!

ehennenfent: For a (rather clunky) example, you can look at the WASM example tests: https://github.com/trailofbits/manticore/blob/934a4f0e332dfdbbe225a3f978a508429d5472a9/tests/wasm/test_examples.py#L35-L37

bradlarsen: P.S. the dev guide for Manticore suggests running `nosetests`, which does _not_ silence the output.

ekilmer: That dev guide needs to be changed, it seems. We run `pytest` in CI, which means we should also be telling people to run `pytest` locally. Also, the last release for `nose` was in 2015, which is another good reason to use `pytest`.

bradlarsen: The rest of the file uses `os.path` exclusively.  Switching just my newly-added test to use a different path manipulation API is going to make it more difficult for future readers -- they'll have to reason about 2 APIs instead of 1, and also wonder why the existing convention was broken in just one place in a file.

If we're gonna switch to a different API for path manipulation, I suggest we do it file-at-a-time as the smallest granularity.

ekilmer: Good points 👍

bradlarsen: Let me see about updating that wiki page.

tests/native/test_integration_native.py
```diff
@@ -10,31 +10,34 @@
 from manticore.binary import Elf, CGCElf
 from manticore.native.mappings import mmap, munmap
 
-DIRPATH = os.path.dirname(__file__)
+from typing import List, Set
```
ekilmer: I don't believe `mypy` checks the type annotations in the `tests` directory. If we're going to use type annotations here, could you update `mypy.ini` also, please?

https://github.com/trailofbits/manticore/blob/934a4f0e332dfdbbe225a3f978a508429d5472a9/mypy.ini#L3

bradlarsen: You're absolutely right. (My editor was running mypy in the file, regardless of the ini) Updating mypy.ini to include `tests` reveals a handful of issues:
```
tests/other/utils/test_config.py:130: error: Name 'test_parse' already defined on line 113
tests/ethereum/test_detectors.py:92: error: Incompatible types in assignment (expression has type "Type[DetectUnusedRetVal]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:112: error: Incompatible types in assignment (expression has type "Type[DetectSuicidal]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:134: error: Incompatible types in assignment (expression has type "Type[DetectExternalCallAndLeak]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:242: error: Incompatible types in assignment (expression has type "Type[DetectEnvInstruction]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:265: error: Incompatible types in assignment (expression has type "Type[DetectDelegatecall]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:304: error: Incompatible types in assignment (expression has type "Type[DetectRaceCondition]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:403: error: Incompatible types in assignment (expression has type "Type[DetectManipulableBalance]", base class "EthDetectorTest" defined the type as "None")
tests/native/test_state.py:213: error: Name 'testContextSerialization' already defined on line 160
tests/native/test_memory.py:1431: error: Not all arguments converted during string formatting
tests/native/test_cpu_manual.py:11: error: Incompatible import of "Memory" (imported name has type "Type[tests.native.mockmem.Memory]", local name has type "Type[manticore.native.memory.Memory]")
tests/native/test_cpu_automatic.py:6: error: Cannot resolve name "solver" (possible cyclic definition)
Found 12 errors in 6 files (checked 171 source files)
```
Let me take a look at these.

bradlarsen: I opened #1608 for fixing the existing typing issues in `tests`, and having mypy check that directory by default.

ekilmer:
Thank you for adding the tests, comments, and type annotations!!

PR #1609, opened 2020-02-04
Symbolic function strategies.

feliam:
Currently used just for sha3. This PR implements 3 ways to deal with symbolic imprecisions.
Concretize
Uninstantiated functions + delayed fix
Fake: Let the solver choose a fake similar function


manticore/ethereum/manticore.py
```diff
@@ -1129,75 +1144,34 @@ def _on_concretize(self, state, func, data, result):
...
-                    state.constrain(constraint)  # bijective
-                symbolic_pairs.append((data, value))
-                state.context[f"symbolic_func_sym_{name}"] = symbolic_pairs
-
-        else:
+        value = None
+        try:
             value = func(data)
+        except Exception as e:
+            pass  # function is unknown
```
ehennenfent: Is it always okay to silence the exception here? Are there cases where we might want at least a warning instead?

feliam: Maybe a better API would be to make func return None when it's not defined

PR #1610, opened 2020-02-04
Use `mypy` to check the `tests` dir; fix some typing errors

### LINKED ISSUE #1608: mypy reports issues in the `tests` directory
If you modify `mypy.ini` so that the `files = manticore` line is changed to `files = manticore, tests`, then mypy reports a number of issues within the `tests` directory:
```
tests/other/utils/test_config.py:130: error: Name 'test_parse' already defined on line 113
tests/ethereum/test_detectors.py:92: error: Incompatible types in assignment (expression has type "Type[DetectUnusedRetVal]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:112: error: Incompatible types in assignment (expression has type "Type[DetectSuicidal]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:134: error: Incompatible types in assignment (expression has type "Type[DetectExternalCallAndLeak]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:242: error: Incompatible types in assignment (expression has type "Type[DetectEnvInstruction]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:265: error: Incompatible types in assignment (expression has type "Type[DetectDelegatecall]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:304: error: Incompatible types in assignment (expression has type "Type[DetectRaceCondition]", base class "EthDetectorTest" defined the type as "None")
tests/ethereum/test_detectors.py:403: error: Incompatible types in assignment (expression has type "Type[DetectManipulableBalance]", base class "EthDetectorTest" defined the type as "None")
tests/native/test_state.py:213: error: Name 'testContextSerialization' already defined on line 160
tests/native/test_memory.py:1431: error: Not all arguments converted during string formatting
tests/native/test_cpu_manual.py:11: error: Incompatible import of "Memory" (imported name has type "Type[tests.native.mockmem.Memory]", local name has type "Type[manticore.native.memory.Memory]")
tests/native/test_cpu_automatic.py:6: error: Cannot resolve name "solver" (possible cyclic definition)
Found 12 errors in 6 files (checked 171 source files)
```

bradlarsen:
Fixes #1608.

Additionally, this includes a few type annotations in a couple other places, which I added as I was looking there.

### DIFF
--- a/manticore/utils/log.py
+++ b/manticore/utils/log.py
@@ -1,9 +1,11 @@
 import logging
 import sys
 
+from typing import Set
+
 manticore_verbosity = 0
 DEFAULT_LOG_LEVEL = logging.WARNING
-all_loggers = set()
+all_loggers: Set[str] = set()
 default_factory = logging.getLogRecordFactory()
 logfmt = "%(asctime)s: [%(process)d] %(name)s:%(levelname)s %(message)s"
 handler = logging.StreamHandler(sys.stdout)
--- a/manticore/utils/log.py
+++ b/manticore/utils/log.py
@@ -1,7 +1,7 @@
 import logging
 import sys
 
-from typing import Set
+from typing import List, Set, Tuple
 
 manticore_verbosity = 0
 DEFAULT_LOG_LEVEL = logging.WARNING
@@ -18,7 +18,7 @@ class ContextFilter(logging.Filter):
     This is a filter which injects contextual information into the log.
     """
 
-    def summarized_name(self, name):
+    def summarized_name(self, name: str) -> str:
         """
         Produce a summarized record name
           i.e. manticore.core.executor -> m.c.executor
@@ -44,7 +44,7 @@ class ContextFilter(logging.Filter):
     colored_levelname_format = "\x1b[{}m{}:\x1b[0m"
     plain_levelname_format = "{}:"
 
-    def colored_level_name(self, levelname):
+    def colored_level_name(self, levelname: str) -> str:
         """
         Colors the logging level in the logging record
         """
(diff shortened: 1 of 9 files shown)


ehennenfent:
Speculating from the CI output, but it looks like `MockMem` and `Memory` don't behave quite the same way, which is causing the native tests to fail? 

Also, I've consolidated the `testing` and `tests` labels into the former.

bradlarsen:
> Speculating from the CI output, but it looks like `MockMem` and `Memory` don't behave quite the same way, which is causing the native tests to fail?

Previously, the use of `Memory` already referred to `mockmem.MockMemory`.
The previous change (`Memory` -> `MockMemory`) shouldn't make a runtime difference; it merely avoids the shadowed import that `mypy` didn't like.  

The test failure, in this case, was actually due to me incorrectly fixing this bad format string the first time around:
```
diff --git a/tests/native/test_memory.py b/tests/native/test_memory.py
index ac1f6d3..c74254b 100644
--- a/tests/native/test_memory.py
+++ b/tests/native/test_memory.py
@@ -1428,7 +1428,7 @@ class MemoryTest(unittest.TestCase):
         # No Access Reading <4160741376>
         # self.assertRaisesRegexp(MemoryException, r"No access reading.*", mem.__getitem__, x)
         with self.assertRaisesRegex(
-            InvalidSymbolicMemoryAccess, "Invalid symbolic memory access.*".format(addr)
+            InvalidSymbolicMemoryAccess, f"Invalid symbolic memory access.*<{addr:x}>"
         ):
             _ = mem[x]
             # mem[addr] = 'a'
```

> Also, I've consolidated the `testing` and `tests` labels into the former.

Thanks!

PR #1612, opened 2020-02-13
Fix for returndatasize

feliam:
This adds/fix support for RETURNDATASIZE instruction.


manticore/platforms/evm.py
```diff
@@ -2030,12 +2033,10 @@ def CALL(self, gas, address, value, in_offset, in_size, out_offset, out_size):
 
     @CALL.pos  # type: ignore
     def CALL(self, gas, address, value, in_offset, in_size, out_offset, out_size):
-        data = self.world.last_transaction.return_data
-        if data is not None:
-            data_size = len(data)
-            size = Operators.ITEBV(256, Operators.ULT(out_size, data_size), out_size, data_size)
-            self.write_buffer(out_offset, data[:size])
-
+        data = self._return_data
```
feliam: is the same code in every CALL* maybe we should find a name for it and make it a priv8 method

feliam:
> The code looks good to me. Was wondering though - do we have a test that covers `RETURNDATASIZE`? I couldn't find one while grepping the source.

No. There is not RETURNDATASIZE instruction in Frontier. The test should actually test that the opcode under RETURNDATASIZE behaves like an INVALID here (As specified in Frontier).

manticore/platforms/evm.py
ehennenfent: Ah, good point. `_save_old_return_data` or something?

ehennenfent:
Mmm, our frankenstein version of solidity strikes again. Maybe we need a separate branch with the Istanbul tests implemented so we have something to check PRs like this against...

PR #1614, opened 2020-02-19
Experiment with testing on OSX

ehennenfent:
Not committing to front-lining OSX support as a feature because there are some fundamental performance concerns, but I would like to see how the tests behave.


ekilmer:
I believe one of the limiting factors with Manticore running Linux binaries on Mac is dynamic linking and loading the linked Linux libraries, right? Static compilations should be fine though, and it seems all of our test binaries are static 👍, so things _should_ work okay for the tests.

One idea for supporting dynamic-linked Linux binaries on Mac would be to include directions to setup a mirror of some "standard" Linux library distribution and pass an option to Manticore to tell it where to look. But, even then, if the program is anything more complex than a standard libc-compiled program, the additional compiled libraries will need to be copied (or built) and placed in the correct place on Mac (within the standard Linux library distrib)

ekilmer:
Also, looks like the failure is due to an outdated version of `bash` on Mac according to this StackOverflow https://stackoverflow.com/a/14612719

ehennenfent:
> I believe one of the limiting factors with Manticore running Linux binaries on Mac is dynamic linking and loading the linked Linux libraries, right? 

Right. From slack:

Since Manticore is a python project it should be possible to run it on any platform in theory, but there are going to be some problems.
On the native side, several of the linux system calls rely on functions from the Python standard library that behave differently across platforms or don’t work at all outside of linux. We also load dynamic libraries from the disk instead of shipping them with Manticore, so trying to run it on OSX will load the darwin versions instead of the linux versions.
...
The most overarching issue is that the multiprocessing library Python uses for distributing tasks over processes or threads works best on Linux. OSX’s fork system call is broken so Python can’t create new processes as quickly as it can on Linux. Using the spawn API instead would require us to make ManticoreBase pickle-able, which I attempted in the past but had some trouble with. Instead, Manticore falls back to using threads if you try to run it on not-linux. Access to shared memory works differently depending on which model you use, so you might have to completely rewrite your script if you switch platforms. There's also the GIL to think about, so you won't get as much performance out of the threaded version.

ekilmer:
>> I believe one of the limiting factors with Manticore running Linux binaries on Mac is dynamic linking and loading the linked Linux libraries, right?

> Right. From slack:

I had not seen this!! Great summary 👍

bradlarsen:
One issue with testing on macOS is that we haven't been able to successfully build `keystone` there in GitHub Actions.

Just leaving a note here, that it ought to be possible to get `keystone` installed on their macOS runners -- keystone definitely builds on macOS.  However, figuring out how to do it in GitHub CI will likely involve finding out the right incantation to install whatever needed dependencies there are on the macOS runners.

tests/wasm/generate_tests.sh
```diff
@@ -3,16 +3,20 @@
...
+else
+  cp "$(command -v wast2json)" .
+fi
+
 wget -nc -nv -O spec.zip -c https://github.com/WebAssembly/spec/archive/opam-1.1.zip
 
-yes | unzip -q -j spec.zip 'spec-*/test/core/*' -d .
-rm run.py README.md
+unzip -q -j spec.zip 'spec-*/test/core/*' -d .
+rm run.py README.md
```
ekilmer, line 17: Nit: Remove trailing whitespace
```suggestion
rm run.py README.md
```

tests/wasm_sym/generate_symbolic_tests.sh
```diff
@@ -3,16 +3,20 @@
...
+else
+  cp "$(command -v wast2json)" .
+fi
+
 wget -nc -nv -O spec.zip -c https://github.com/WebAssembly/spec/archive/opam-1.1.zip
 
-yes | unzip -q -j spec.zip 'spec-*/test/core/*' -d .
-rm run.py README.md
+unzip -q -j spec.zip 'spec-*/test/core/*' -d .
+rm run.py README.md
```
ekilmer, line 17: Nit: Remove trailing whitespace
```suggestion
rm run.py README.md
```

PR #1615, opened 2020-02-20
Improve balance related exploration

feliam:
Add a config constant to enable exploration with symbolic balances

### DIFF
--- a/manticore/ethereum/manticore.py
+++ b/manticore/ethereum/manticore.py
@@ -29,6 +29,7 @@ from ..core.smtlib import (
     Expression,
     issymbolic,
     simplify,
+    Z3Solver,
 )
 from ..core.state import TerminateState, AbandonState
 from .account import EVMContract, EVMAccount, ABI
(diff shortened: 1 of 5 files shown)


manticore/ethereum/manticore.py
```diff
@@ -995,17 +1001,22 @@ def multi_tx_analysis(
         args=None,
         compile_args=None,
     ):
-        owner_account = self.create_account(balance=1000, name="owner")
-        attacker_account = self.create_account(balance=1000, name="attacker")
+        owner_account = self.create_account(balance=10000000000000000000, name="owner")
+        attacker_account = self.create_account(balance=10000000000000000000, name="attacker")
```
ehennenfent: Are we changing a default behavior here? Worried this could break assumptions made by some user scripts.

feliam: Yes. More ether to default accounts in the "multi_tx" canned analysis (normally only used from the cli)

We were not even trying to send ether in the construction of a contract before. 1000 was not enough to pay for the gas in some examples.

PR #1618, opened 2020-02-21
Add feature for SymbolicSocket through sys_accept

ekilmer:
SymbolicSocket will generate symbolic bytes and variables on-demand
through interaction with the relevant system calls


ekilmer:
As it is now, I don't think Manticore will write out a file containing the concretized bytes that were sent to/recv'd from the `accept`ed socket.

Do we want to add that here? Otherwise, I believe Manticore will just list the concretized array variable names at the very end?

ehennenfent:
>  I don't think Manticore will write out a file containing the concretized bytes that were sent to/recv'd from the accepted socket. Do we want to add that here? 

I don't have strong feelings on the matter. It would be a minor usability improvement to do so. Since your team works with sockets more than I do, I'll leave it up to your judgement as to whether it's helpful to split the sockets into separate files or if it's sufficient to have them listed alongside other symbols in the summary.

ekilmer:
> > I don't think Manticore will write out a file containing the concretized bytes that were sent to/recv'd from the accepted socket. Do we want to add that here?
> 
> I don't have strong feelings on the matter. It would be a minor usability improvement to do so. Since your team works with sockets more than I do, I'll leave it up to your judgement as to whether it's helpful to split the sockets into separate files or if it's sufficient to have them listed alongside other symbols in the summary.

Okay, great. I'll leave it as it is now, then. We interact with Manticore programmatically through its API more than the files it generates (for now), so I'll keep this how it is.

PR #1624, opened 2020-03-06
Transparently compress & decompress persisted state files

### LINKED ISSUE #1622: Transparently compress Manticore persisted state files
In recent versions of Manticore, it looks like each state is persisted to disk as a Python pickle file.  When using Manticore on a more complicated binary, where at least thousands of states are explored, these persisted states take up lots of space!  (A VM I was using had its 60GB disk fill up entirely when analyzing a realistic program.)

I noticed that if these states are compressed with gzip, they shrink by 10x or more.

It would be friendlier if Manticore would transparently compress & decompress its persisted states using gzip.  This might possibly not even slow Manticore down at all, depending on how expensive I/O operations are.

bradlarsen:
Fixes #1622.

### DIFF
--- a/manticore/utils/helpers.py
+++ b/manticore/utils/helpers.py
@@ -122,7 +122,7 @@ class PickleSerializer(StateSerializer):
 
     def deserialize(self, f):
         logger.info("Deserializing %s", f.name if hasattr(f, "name") else "<unknown>")
-        return pickle.load(GzipFile(fileobj=f, mode="rb"))
+        return pickle.load(GzipFile(fileobj=f, mode="rb") if consts.compress_states else f)
 
 
 def pickle_dumps(obj: Any) -> bytes:
--- a/manticore/utils/helpers.py
+++ b/manticore/utils/helpers.py
@@ -80,8 +80,13 @@ class StateSerializer:
 
 
 class PickleSerializer(StateSerializer):
+    """
+    A StateSerializer that uses a gzip-based Python pickle format.
+    """
+
     DEFAULT_RECURSION: int = 0x10000  # 1M
     MAX_RECURSION: int = 0x1000000  # 16.7M
+    COMPRESSION_LEVEL: int = 5
 
     def __init__(self):
         super().__init__()
@@ -90,7 +95,10 @@ class PickleSerializer(StateSerializer):
     def serialize(self, state, f):
         logger.info("Serializing %s", f.name if hasattr(f, "name") else "<unknown>")
         try:
-            pickle_dump(state, GzipFile(fileobj=f, mode="wb", compresslevel=5))
+            pickle_dump(
+                state,
+                GzipFile(fileobj=f, mode="wb", compresslevel=PickleSerializer.COMPRESSION_LEVEL),
+            )
         except RuntimeError:
             new_limit = sys.getrecursionlimit() * 2
             if new_limit > PickleSerializer.MAX_RECURSION:
--- a/manticore/utils/helpers.py
+++ b/manticore/utils/helpers.py
@@ -7,8 +7,17 @@ from io import BytesIO
 
 from typing import Any, IO
 
+from .config import get_group
 
 logger = logging.getLogger(__name__)
+
+consts = get_group("core")
+consts.add(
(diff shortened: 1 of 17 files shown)


manticore/utils/helpers.py
```diff
@@ -75,33 +79,45 @@ def deserialize(self, f):
         raise NotImplementedError
 
 
-class PickleSerializer(StateSerializer):
+class GzipPickleSerializer(StateSerializer):
```
ehennenfent: As long as we're not creating side-by-side `PickleSerializer` and `GZipPickleSerializer` options, why not leave the name the same so that the diff will be smaller?

ekilmer: If you know of a way to have Github Actions to test each commit in a PR and every commit passed, we could merge this PR without squashing and review each commit individually.

Here's the commit for the transparent `PickleSerializer` change without renmaing https://github.com/trailofbits/manticore/pull/1624/commits/4409c00c91f8c6573b3fe0638be31858f12f026b

bradlarsen: Sure, I'm fine with leaving `PickleSerializer` named as-is. :)

bradlarsen:
I still plan to run a few small experiments to be able to get a feel for what the performance implications are of having the state files be compressed.

ehennenfent:
@bradlarsen Did you find other bugs in the workspace implementation? I notice you tweaked some of the tests.

bradlarsen:
@ehennenfent I didn't see bugs in there, in particular -- it was just a couple tweaks to the tests to avoid noise from pytest and to avoid a deprecation warning.

ehennenfent:
One more thing - can you add `.gz` to the end of the state file names to reflect the format change?

bradlarsen:
## Benchmarking

To better understand the performance implications, I did some benchmarking before & after the changes in this pull request.

### Wall-clock time on `multiple-styles`

I ran Manticore from the command-line on the `multiple-styles` example 20x on both versions of the code, using the following command in an Ubuntu 18.04 VM on an otherwise idle machine:
```
for i in `seq 20`; do
    rm -rf benchmark.log mcore_*
    TIMEFORMAT='%1R, '
    time python -mmanticore -v --core.procs=1 --core.mprocessing=single examples/linux/binaries/multiple-styles --core.seed 42 >benchmark.log
done
```
This collects the wall-clock time from each attempt, formatted for easy pasting into a Python REPL.

Note, I controlled for Manticore's random seed via the `--core.seed 42` arguments.

The results for unmodified Manticore, as a Python list: `[12.4, 12.5, 12.5, 12.6, 13.1, 13.0, 13.7, 12.8, 12.9, 12.8, 12.9, 12.6, 12.9, 13.0, 12.8, 13.0, 12.7, 13.3, 13.8, 13.2,]`

The results for Manticore from this pull request: `[12.7, 12.9, 12.9, 12.8, 12.8, 12.7, 12.9, 12.6, 12.8, 13.8, 12.8, 14.2, 12.8, 12.8, 12.8, 12.8, 12.7, 12.9, 12.6, 12.9,]`

Are these different?  A t-test via [scipy.stats.ttest_ind](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.ttest_ind.html) doesn't give us confidence in a difference in wall clock time (P-value of 0.90).

### Fine-grained sampling with speedscope
I also collected sampling profiling data from running on the `multiple-styles` examples using [py-spy](https://github.com/benfred/py-spy), having it dump results in [speedscope](https://github.com/jlfwong/speedscope) format.  In those results, it shows that call stack samples with `PickleSerializer.serialize` or `PickleSerializer.deserialize` in them comprise about 2% of all samples in unmodified Manticore, and about 4% of all samples with the changes from this pull request.

We would need a much more serialization-heavy benchmark to demonstrate that this difference is meaningful in practice.

### Space required for serialized states on `multiple-styles`
The state files from unmodified Manticore from the above experiments total 7.8MB (~440KB each).  In contrast, with the changes from this pull request, the state files total 792KB (~44KB per state).  So even with the lowest level of gzip compression, serialized states are ~10x smaller, at least in this benchmark.

PR #1631, opened 2020-03-12
Repatriate pyevmasm

feliam:
Remove pyevmasm as a dependency

### DIFF
--- a/manticore/ethereum/evmasm.py
+++ b/manticore/ethereum/evmasm.py
@@ -1,7 +1,7 @@
 from bisect import bisect
 from binascii import hexlify, unhexlify
 from builtins import map, next, range, object
-from future.builtins import next, bytes
+#from future.builtins import next, bytes  # type: ignore
 import copy
 
 DEFAULT_FORK = "petersburg"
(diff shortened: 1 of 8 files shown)


ehennenfent:
This looks like mostly a straight inlining of the `pyevmasm` code, yeah? Unless there are some changes I'm missing?

What's the motivation? I assume you're trying to avoid any issues caused by upstream upgrades to `pyevmasm`, but in that case, why not just lock the dependency in setup.py to one specific version?

ehennenfent:
@computerality Do you have any thoughts on this? Felipe has been experiencing frustration lately because making fixes to `pyevmasm` requires modifying it upstream, merging to master, and pushing a release to pypi. Ideally it'd be nice to shorten that cycle substantially without having to inline `pyevmasm` and risk forgetting to push improvements back upstream.

computerality:
That cycle does sound annoying, especially since pyevmasm isn’t really under any other active development other than for manticore. Before last week, the most recent update was this past summer. There are only a couple people who are using the library, and only two with releases in the last 6 months. http://github.com/crytic/pyevmasm/network/dependents

After a while, we can just review the changes and merge them back into pyevmasm

ehennenfent:
Per the meeting yesterday, we can go ahead and close this PR. Felipe has upstream access to PyEVMASM now, so if we need to make changes, what we'll do is:
* Push those changes to a branch on crytic/pyevmasm
* Update setup.py to pull pyevmasm from the branch instead of PyPI

If we need to release, we'll do the following:
* Merge changes to crytic/pyevmasm into the master branch
* Push a release of pyevmasm to PyPI
* Revert changes to Manticore's setup.py to make it pull pyevmasm from PyPI

This is because PyPI won't let us upload packages that pull dependencies directly from GitHub, even though pip is happy to allow it when installing packages from the filesystem.

PR #1632, opened 2020-03-13
Include invalid transaction type in exception message

smoelius:


manticore/platforms/evm.py
```diff
@@ -310,7 +310,7 @@ def sort(self):
     @sort.setter
     def sort(self, sort):
         if sort not in {"CREATE", "CALL", "DELEGATECALL"}:
-            raise EVMException("Invalid transaction type")
+            raise EVMException("Invalid transaction type: " + sort)
```
feliam: f"Invalid transaction type: {sort}"  
:man_shrugging:

PR #1634, opened 2020-03-13
Deduct for gas

smoelius:
This is just the start of what must be done to account for gas costs.


smoelius:
While I knew this PR would not address all concerns, it's more broken than I thought. Something like this is needed:
https://github.com/trailofbits/manticore/blob/a635cb2c508040f4b13a4b02d96dcc8a4cceb6c4/manticore/platforms/evm.py#L2999-L3003
Closing for now.

PR #1635, opened 2020-03-17
Add __slots__ to expressions

ehennenfent:
This PR modifies SMTLib's inheritance structure to avoid using multiple inheritance. This allows us to use `__slots__` to improve memory usage and performance with SMTLib expressions.

## Old inheritance diagram:
Old Diagram

## New Inheritance diagram:
Constant, Variable, and Operation have been inlined in their subclasses

The `Constant`, `Variable`, and `Operation` classes from the old diagram have been inlined into their respective subclasses. I've added a test to make sure that the subclasses still expose the same methods. I've also added tuples containing all the subclasses (eg `Constant = (BitVecConstant, BoolConstant)`) so that `isinstance(x, Constant)` will still work. 

Running this on `multiple_styles`, I saw a 3.5% increase in speed and a 4.5% decrease in overall memory usage. But of course, `multiple-styles` is a very lightweight example.

### DIFF
--- a/manticore/core/smtlib/constraints.py
+++ b/manticore/core/smtlib/constraints.py
@@ -11,8 +11,8 @@ from .expression import (
     Bool,
     BitVec,
     BoolConstant,
-    ConstantType,
-    VariableType,
+    Constant,
+    Variable,
     ArrayProxy,
     BoolEqual,
 )
@@ -157,8 +157,8 @@ class ConstraintSet:
             for expression in related_constraints:
                 if (
                     isinstance(expression, BoolEqual)
-                    and isinstance(expression.operands[0], VariableType)
-                    and isinstance(expression.operands[1], (*VariableType, *ConstantType))
+                    and isinstance(expression.operands[0], Variable)
+                    and isinstance(expression.operands[1], (*Variable, *Constant))
                 ):
                     constant_bindings[expression.operands[0]] = expression.operands[1]
 
@@ -177,7 +177,7 @@ class ConstraintSet:
             if replace_constants:
                 constraint = simplify(replace(constraint, constant_bindings))
                 # if no variables then it is a constant
-                if isinstance(constraint, ConstantType) and constraint.value == True:
+                if isinstance(constraint, Constant) and constraint.value == True:
                     continue
 
             translator.visit(constraint)
@@ -265,7 +265,7 @@ class ConstraintSet:
 
     def is_declared(self, expression_var):
         """ True if expression_var is declared in this constraint set """
-        if not isinstance(expression_var, VariableType):
+        if not isinstance(expression_var, Variable):
             raise ValueError(f"Expression must be a Variable (not a {type(expression_var)})")
         return any(expression_var is x for x in self.get_declared_variables())
 
--- a/manticore/core/smtlib/constraints.py
+++ b/manticore/core/smtlib/constraints.py
@@ -11,10 +11,10 @@ from .expression import (
     Bool,
     BitVec,
     BoolConstant,
+    ConstantType,
+    VariableType,
     ArrayProxy,
     BoolEqual,
-    Variable,
-    Constant,
 )
 from .visitors import GetDeclarations, TranslatorSmtlib, get_variables, simplify, replace
 import logging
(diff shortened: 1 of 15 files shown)


manticore/core/smtlib/expression.py
```diff
@@ -215,26 +162,62 @@ def __bool__(self):
...
-        super().__init__(value, *args, **kwargs)
+        self._value = value
+        super().__init__(*args, **kwargs)
 
     def __bool__(self):
         return self.value
 
+    @property
+    def value(self):
+        return self._value
```
disconnect3d: Ideally, the property should be there only in a debug mode :<

manticore/core/smtlib/expression.py
```diff
@@ -215,26 +162,62 @@ def __bool__(self):
...
+    __slots__ = ["_operands"]
 
-class BoolOperation(Operation, Bool):
     def __init__(self, *operands, **kwargs):
-        super().__init__(*operands, **kwargs)
+        self._operands = operands
+
+        # If taint was not forced by a keyword argument, calculate default
+        if "taint" not in kwargs:
+            kwargs["taint"] = reduce(lambda x, y: x.union(y.taint), operands, frozenset())
```
disconnect3d: Could be...
```python
kwargs.setdefault('taint', reduce(lambda x, y: x.union(y.taint), operands, frozenset()))
```

manticore/core/smtlib/expression.py
disconnect3d: Random thought: maybe we should use the private version in internal code? I *believe* (not confirmed) those attributes might be somehow *hot* and we don't want to use a call internally to get their value if possible

manticore/core/smtlib/expression.py
```diff
@@ -465,18 +450,38 @@ def Bool(self):
...
+        return self._name
+
+    def __copy__(self, memo):
+        raise ExpressionException("Copying of Variables is not allowed.")
+
+    def __deepcopy__(self, memo):
+        raise ExpressionException("Copying of Variables is not allowed.")
+
+    def __repr__(self):
+        return "<{:s}({:s}) at {:x}>".format(type(self).__name__, self.name, id(self))
```
disconnect3d: I wonder if there is much difference between `type(self)` and `self.__class__`

manticore/core/smtlib/expression.py
```diff
@@ -1204,3 +1266,8 @@ def __init__(
         assert true_value.size == size
         assert false_value.size == size
         super().__init__(size, condition, true_value, false_value, *args, **kwargs)
+
+
+Constant = (BitVecConstant, BoolConstant)
+Variable = (BitVecVariable, BoolVariable, ArrayVariable)
+Operation = (BitVecOperation, BoolOperation, ArrayOperation, ArraySelect)
```
disconnect3d: Maybe `Constants`, `Variables`, `Operations`?

manticore/core/smtlib/constraints.py
```diff
@@ -158,7 +158,7 @@ def to_string(self, related_to=None, replace_constants=False):
                 if (
                     isinstance(expression, BoolEqual)
                     and isinstance(expression.operands[0], Variable)
-                    and isinstance(expression.operands[1], (Variable, Constant))
+                    and isinstance(expression.operands[1], (*Variable, *Constant))
```
disconnect3d: or `Variables + Constants`, heh

manticore/core/smtlib/expression.py
ehennenfent: Previous maintainers agree with you :
https://github.com/trailofbits/manticore/blob/abb191a3a562f30aba42ec4b4b4b4604c283130f/manticore/core/smtlib/visitors.py#L779
We should strive to present the same API to users, but we could probably replace a lot of internal property uses with direct access.

manticore/core/smtlib/expression.py
ehennenfent: 97647b33dbad014cbe914afdc610046794a585e7

manticore/core/smtlib/visitors.py
```diff
@@ -861,17 +930,26 @@ def visit_BitVecConstant(self, expression):
     def visit_BoolConstant(self, expression):
         return expression.value and "true" or "false"
 
-    def visit_Variable(self, expression):
+    def _visit_variable(self, expression):
         return expression.name
 
+    def visit_ArrayVariable(self, expression):
```
disconnect3d: Could also be:
```python
In [10]: class X:
    ...:     def _priv(self, x):
    ...:         return x
    ...:     public = _priv
    ...:

In [11]: X().public(3)
Out[11]: 3

In [12]: X().public.__name__
Out[12]: '_priv'
```

but names :<

PR #1636, opened 2020-03-19
Fix config.py so it is reproducible

### LINKED ISSUE #1751: Reproducible Tiered Configuration Strategy
Our configuration system is currently ad-hoc, and not preserved across runs. Manticore should provide an API for encoding configuration settings at several granularities so that we can get the exact same results when running it in different environments.

feliam:
All config variables should exist via config.py.
Attempt to erradicate argparse.
Attempt to separate EVM, WASM, NATIVE configs
Make it save the full config at mcore.yml towards a reproducible exploration.


bradlarsen:
Related to this, I've noticed that if you run the existing unit test suite with pytest and the [`pytest-random-order`](https://pypi.org/project/pytest-random-order/) plugin, several Ethereum tests sporadically fail due to global configuration not being initialized fully when test modules are run in an unexpected order.

bradlarsen:
Yes, trying to run the tests in random order via `pytest-randomly` reveals path-dependent troubles in our test suite due to config.py stuff:  https://github.com/trailofbits/manticore/pull/1660

PR #1637, opened 2020-03-20
Attempt to fix Truffle tests

ehennenfent:
All of our pull request builds are failing right now due to (apparently) upstream changes to Truffle. Perhaps bumping the version number will be enough to fix them?


ehennenfent:
There may be further issues, but the ethereum_vm tests (which were killing the other builds) have passed, so I think this is ready to review. All I've done is replace our manual installation of NPM with GitHub's.

PR #1639, opened 2020-03-26
Added support for proc/self/maps

sschriner:


manticore/platforms/linux.py
```diff
@@ -151,6 +151,78 @@ def sync(self):
...
+
+    def __getstate__(self):
+        state = {"name": self.name, "mode": self.mode, "closed": self.closed}
+        state["pos"] = None if self.closed else self.tell()
+        return state
+
+    def __setstate__(self, state):
+        name = state["name"]
+        mode = state["mode"]
+        closed = state["closed"]
```
feliam: I guess you can calculate `closed` given `pos`?
If used then closed could be a property? (One of those @property getter things)

manticore/platforms/linux.py
```diff
@@ -151,6 +151,78 @@ def sync(self):
...
+    def name(self):
+        return self.file.name
+
+    @property
+    def mode(self):
+        return self.file.mode
+
+    @property
+    def closed(self):
+        return self.file.closed
```
feliam: This. I wonder if `self.file.closed == (self.pos is None) ` 
Dismiss comment if not.

sschriner: I think actually I can remove that and several of the other properties as I used the definition that was already defined in the File object. (I copy pasted properties with the intention of deleting what I did not modify).  Should I look at this property in the file object?

manticore/platforms/linux.py
sschriner: Yes I should be able to calculate closed with pos

manticore/platforms/linux.py
```diff
@@ -152,6 +152,50 @@ def sync(self):
         return
 
 
+class ProcSelfMaps(File):
+    def __init__(self, flags, linux):
+        self.filename = "_proc_self_maps_"
+        with open(self.filename, "w") as mapsFile:
+            print(linux.current.memory.__proc_self__, file=mapsFile)
```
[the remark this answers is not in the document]
sschriner: Not particularly. I've coded in C++ much more than python. I don't remember my train of logic with this. I will swap this out.

manticore/platforms/linux.py
```diff
@@ -152,6 +152,50 @@ def sync(self):
...
+            self.file = open(name, mode)
+            if closed:
+                self.file.close()
+        except IOError:
+            # If the file can't be opened anymore (should not typically happen)
+            self.file = None
+        if pos is not None:
+            self.seek(pos)
+
+    def stat(self):
```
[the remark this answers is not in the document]
sschriner: Yes that's what it looks like will remove

manticore/platforms/linux.py
```diff
@@ -152,6 +152,50 @@ def sync(self):
...
+            raise EnvironmentError("/proc/self/maps is only supported in read only mode")
+        self.file = open(self.filename, mode)
+
+    def __getstate__(self):
+        state = {"name": self.name, "mode": self.mode}
+        state["pos"] = None if self.closed else self.tell()
+        return state
+
+    def __setstate__(self, state):
+        name = state["name"]
```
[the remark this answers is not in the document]
sschriner: No it's essentially the same. Originally I had the .name property returning '/proc/self/maps', in which case I would've needed to modify the __getstate__, __setstate__, but I realized that probably wasn't necessary. It's probably best to just remove these and rely on the superclass.

manticore/native/memory.py
```diff
@@ -814,6 +814,64 @@ def __str__(self):
             ]
         )
 
+    def proc_self_mappings(self):
+        """
+        Returns a sorted list of all the mappings for this memory for /proc/self/maps.
```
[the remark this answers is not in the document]
sschriner: Added a more verbose comment

sschriner:
Working on the test. I'm adding to the test_syscalls.py file as suggested via Slack earlier

manticore/native/memory.py
```diff
@@ -814,6 +814,69 @@ def __str__(self):
...
+                        (
+                            m.start,
+                            m.end,
+                            m.perms + private_shared_perms,
+                            0,
+                            device,
+                            inode,
+                            "[" + m.name + "]",
+                        )
+                    )
```
ekilmer: Could we make this into something more well-defined?

Perhaps a `@dataclass` class that has useful names.

Something like
```python
@dataclass
class ProcSelfMapInfo
    start_addr: int
    end_addr: int
    map_perms: ....

[...]

result.append(ProcSelfMapInfo(m.start, m.end, m.perms + private_shared_perms, ...))
```
More info about dataclasses https://pypi.org/project/dataclasses/ (PyPI link since we support Python3.6)

sschriner: Will take a look at this :)

ehennenfent:
Code itself LGTM, would just like to see a test for this. We could throw a Rust "hello world" binary in tests/native/binaries, add a test to make sure that it runs, and then call it a day.

PR #1640, opened 2020-03-26
Add a syscall implementation for `llseek`

bradlarsen:


ehennenfent:
While you're in there, do you mind adding tests for `sys_lseek` and `sys_llseek` to tests/native/test_syscalls.py? CodeCov won't be happy about this otherwise since we didn't previously have tests covering those calls.

bradlarsen:
> While you're in there, do you mind adding tests for `sys_lseek` and `sys_llseek` to tests/native/test_syscalls.py? CodeCov won't be happy about this otherwise since we didn't previously have tests covering those calls.

Sure; let me take a look.  I had checked the coverage earlier and it looks like we don't currently have any unit tests exercising the existing `sys_lseek`.

ehennenfent:
Looks pretty good to me! I notice codecov is still grumpy about patch coverage, probably because it looks like that unit test only covers `sys_lseek`, not `sys_llseek`. Is it possible to just copy and paste it for the latter?

bradlarsen:
> Looks pretty good to me! I notice codecov is still grumpy about patch coverage, probably because it looks like that unit test only covers sys_lseek, not sys_llseek. Is it possible to just copy and paste it for the latter?

I'm still working on that.  I'm not certain that the implementation of `sys_llseek` is actually correct in all cases... the man page for `llseek` indicates that unlike `lseek`, it isn't exposed in glibc, and is a raw system call that writes its return value to an output argument.

bradlarsen:
Okay, I think the `sys_llseek` implementation mostly works now, and I've added tests for that, and some extra tests for `sys_lseek`.

Both `sys_llseek` and `sys_lseek` don't seem to behave quite properly when given certain invalid arguments that would position a file descriptor before the start of the file.  Instead of fixing that right now, I've added a couple expected failure tests at least documenting that infidelity.

manticore/platforms/linux.py
```diff
@@ -1333,19 +1330,57 @@ def sys_lseek(self, fd, offset, whence):
...
         try:
             return self._get_fd(fd).seek(signed_offset, whence)
         except FdError as e:
             logger.info(
-                ("LSEEK: Not valid file descriptor on lseek." "Fd not seekable. Returning EBADF")
+                f"LSEEK: Not valid file descriptor on lseek. Fd not seekable. Returning {-e.err}"
+            )
+            return -e.err
+
+    def sys_llseek(self, fd, offset_high, offset_low, resultp, whence):
```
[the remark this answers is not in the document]
bradlarsen: Done! I think the signatures I added are even correct.

manticore/platforms/linux.py
```diff
@@ -1345,7 +1345,9 @@ def sys_lseek(self, fd, offset, whence):
             )
             return -e.err
 
-    def sys_llseek(self, fd, offset_high, offset_low, resultp, whence):
+    def sys_llseek(
+        self, fd: File, offset_high: int, offset_low: int, resultp: int, whence: int
```
ekilmer: I think `fd` should just be an `int`, right? It's used as an index into an array here https://github.com/trailofbits/manticore/blob/bd28600ef35cdef06ab9522d4911a07350df3777/manticore/platforms/linux.py#L1253

And `self.files` is accessed using `int`, as shown when setting up the `stdio` `stdout` and `stderr` https://github.com/trailofbits/manticore/blob/bd28600ef35cdef06ab9522d4911a07350df3777/manticore/platforms/linux.py#L666-L671

bradlarsen: Yeah, you're right...

ekilmer: Not sure why `mypy` isn't yelling at us though :(

bradlarsen: mypy by default doesn't do any checking within functions that don't have any type annotations.  Additionally, throughout the Linux platform code, there are very few type signatures, so most code in there gets an inferred type that allows anything.

I added several more type hints to things in the Linux platform code, for things that were obvious.  With those additions, mypy _did_ gripe about a few things, such as in `sys_write()`.  I fixed that.

ekilmer: Awesome!! Thank you!!

ehennenfent:
Tests are green and codecov is (mostly) happy. Let's get this merged before it gets any bigger 😎

PR #1644, opened 2020-03-31
New instructions

feliam:
New istanbul instructions.
CHAINID
SELFBALANCE
EXTCODEHASH


manticore/platforms/evm.py
```diff
@@ -1832,6 +1841,10 @@ def GASLIMIT(self):
         """Get the block's gas limit"""
         return self.world.block_gaslimit()
 
+    def CHAINID(self):
+        """Get current chainid."""
+        return 0
```
montyly: Should we return `1` by default? It's the value of mainet: https://chainid.network/

feliam: fixed

montyly:
Looks good to me. 

I think `setup.py` needs to update `pyevmasm` to `0.2.2`, so `istanbul` can be used (currently the CI is crashing because of that)

ehennenfent:
Will this work if we bump the pyevmasm version to 0.2.3 (now on PyPI)?

setup.py
```diff
@@ -50,15 +50,15 @@ def rtd_dependent_deps():
     python_requires=">=3.6",
     install_requires=[
         "pyyaml",
-        # evm dependencies
+        "wrapt",
```
ehennenfent, line 53: We removed wrapt in #1672. Merge conflict?

Issue #1584, opened 2020-01-07
Incorporate typeguard pytest plugin


smoelius:
https://typeguard.readthedocs.io/en/latest/#using-the-pytest-plugin

I did some preliminary experiments using typeguard's [import hook](https://typeguard.readthedocs.io/en/latest/#using-the-import-hook) directly.  Specifically, I inserted these lines at the start of [manticore.py](https://github.com/trailofbits/manticore/blob/master/manticore/ethereum/manticore.py):
```
from typeguard.importhook import install_import_hook

install_import_hook("manticore")
```
Those experiments suggested that there are bugs present within the codebase that typeguard exposes.

bradlarsen:
This is an excellent issue.  Indeed, running our test suite with the pytest plugin for Typeguard, there are many issues detected with incorrect type hints.

This is relevant to #1645, where I've been adding a bunch more type hints to the Linux platform emulation code.

PR #1645, opened 2020-03-31
Add a bunch of type hints & fix a few issues with Linux platform emulation

bradlarsen:


manticore/platforms/linux.py
```diff
@@ -1744,12 +1760,14 @@ def sys_dup2(self, fd, newfd):
         try:
             file = self._get_fd(fd)
         except FdError as e:
-            logger.info("DUP2: fd ({fd}) is not open. Returning {-e.err}")
+            logger.info("sys_dup2: fd ({fd}) is not open. Returning -{errorcode(e.err)}")
             return -e.err
 
-        soft_max, hard_max = self._rlimits[self.RLIMIT_NOFILE]
+        soft_max, hard_max = self._rlimits[resource.RLIMIT_NOFILE]
```
bradlarsen, line 1766: Note, there was a typo here previously, and the code would never have worked...  After I had added some type hints, and mypy started looking at this function, the typo was revealed.  Guess we don't have tests for this at present.

manticore/native/memory.py
```diff
@@ -278,7 +278,7 @@ def __init__(self, start, size, perms, data_init=None, name=None, **kwargs):
         :param perms: the access permissions of the map.
         :param data_init: the data to initialize the map.
         """
-        super().__init__(start, size, perms, name, **kwargs)
+        super().__init__(start, size, perms, name)
```
bradlarsen, line 281: Note, previously, the `**kwargs` passed here would have never worked properly if there were any key/value pairs there.  mypy revealed this problem after adding some type hints to this constructor.

manticore/platforms/linux.py
```diff
@@ -793,21 +809,21 @@ def update(address, code):
...
+    #     with open(f"vdso{bits:2d}.dump") as f:
+    #         vdso_size = len(f.read())
+    #     vdso_addr = self.current.memory.mmapFile(
+    #         self.current.memory._floor(vdso_top - vdso_size),
+    #         vdso_size,
+    #         "r x",
+    #         {32: "vdso32.dump", 64: "vdso64.dump"}[bits],
+    #         0,
+    #     )
+    #     return vdso_addr
```
bradlarsen, line 824: Note: `load_vdso` was dead code previously, and had type error in it also.

manticore/platforms/linux.py
```diff
@@ -1228,43 +1244,40 @@ def _open(self, f) -> int:
...
-        :return: C{0} on success.
         """
         try:
-            self.files[fd].close()
-            self._closed_files.append(
-                self.files[fd]
-            )  # Keep track for SymbolicFile testcase generation
+            f = self.files[fd]
+            if f is None:
+                raise FdError(f"Bad file descriptor ({fd})", errno.EBADF)
```
bradlarsen, line 1256: This error-handling case was missed previously, and we don't seem to have tests that exercise it.

manticore/platforms/linux.py
```diff
@@ -1228,43 +1244,40 @@ def _open(self, f) -> int:
...
 
-    def _dup(self, fd: int) -> int:
-        """
-        Duplicates a file descriptor
-        :rtype: int
-        :param fd: the file descriptor to duplicate.
-        :return: C{0} on success.
-        """
-        return self._open(self.files[fd])
-
```
bradlarsen, line 1255: Note: I got rid of this because it was used in just a single place; simpler to inline it and delete code.

manticore/native/memory.py
ekilmer: Good catch!

ekilmer: Also, it looks like there is no docstring for `name`, which was probably added at a different time than the other arguments, and `name` was probably meant to be caught in the `**kwargs`.

Not sure what we actually want to do to resolve this. I'm not a huge fan of using `kwargs` here, since we know all of the arguments of `Map`. As long as we're consistent in the order the arguments are written in the function signature, then we should list them all explicitly. However, we would also have be copying the parameter docstring(s) as well, which can desynchronize easily :/ . Maybe there is a way to link to `Map`'s docstring for certain parameters?

manticore/platforms/linux.py
```diff
@@ -449,25 +461,32 @@ class Linux(Platform):
...
         :ivar files: List of active file descriptors
-        :type files: list[Socket] or list[File]
+        :type files: list[Socket or File or None]
         """
         super().__init__(path=program, **kwargs)
 
         self.program = program
         self.clocks = 0
-        self.files = []
+        self.files: List[Union[None, File, Socket]] = []
```
ekilmer, line 485: Could this be
```suggestion
        self.files: List[Optional[Union[File, Socket]]] = []
```

manticore/platforms/linux.py
```diff
@@ -449,25 +461,32 @@ class Linux(Platform):
...
         # A cache for keeping state when reading directories { fd: dent_iter }
-        self._getdents_c = {}
-        self._closed_files = []
-        self.syscall_trace = []
+        self._getdents_c: Dict[int, Any] = {}
+        self._closed_files: List[Union[File, Socket]] = []
+        self.syscall_trace: List = []
```
ekilmer, line 489: Huh. I thought mypy didn't like a `List` without brackets for the type. Must default to `Any` then. 

Here's at least one line populates `syscall_trace`
https://github.com/trailofbits/manticore/blob/7dfaf9300dba9f76280b5edd69c5e1ade084594c/manticore/platforms/linux.py#L1432

So it should be something like 
```suggestion
        self.syscall_trace: List[Tuple[str, int, bytes]] = []
```

manticore/platforms/linux.py
ekilmer: From slack by @feliam 

> AUXV were selected to avoid vdso

If it's dead code, we should probably just delete 🤷‍♂

manticore/platforms/linux.py
```diff
@@ -1228,43 +1244,40 @@ def _open(self, f) -> int:
             self.files.append(f)
         return fd
 
-    def _close(self, fd: int) -> None:
+    def _close(self, fd):
```
ekilmer, line 1247: Why were the type-hints removed here?

manticore/platforms/linux.py
ekilmer: Good catch!

manticore/native/cpu/abstractcpu.py
```diff
@@ -795,7 +797,7 @@ def write_string(self, where, string, max_length=None, force=False):
 
         self.write_bytes(where, string + "\x00", force)
 
-    def read_string(self, where, max_length=None, force=False):
+    def read_string(self, where: int, max_length: int = None, force: bool = False) -> str:
```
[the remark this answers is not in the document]
bradlarsen: Good catch!

manticore/native/cpu/abstractcpu.py
```diff
@@ -776,7 +776,9 @@ def read_bytes(self, where, size, force=False):
             result.append(Operators.CHR(self.read_int(where + i, 8, force)))
         return result
 
-    def write_string(self, where, string, max_length=None, force=False):
+    def write_string(
+        self, where: int, string: str, max_length: int = None, force: bool = False
```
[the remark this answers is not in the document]
bradlarsen: Good catch!

manticore/platforms/linux.py
bradlarsen: Yes, in general, `Optional[T]` is a synonym for `Union[None, T]` in mypy.

manticore/platforms/linux.py
bradlarsen: Mistake!  Good catch.

manticore/platforms/linux.py
bradlarsen: Yeah, in mypy, a `List` without any parameter is shorthand for `List[Any]`.  Similarly, `Dict` without any parameters is shorthand for `Dict[Any, Any]`:

> Using a generic class without specifying type parameters assumes Any for each position.

From here: https://docs.python.org/3/library/typing.html

Thanks for the more precise type for `syscall_trace` :)

bradlarsen:
I'm gonna run tests with the typeguard pytest integration (see #1584) before we should consider merging this.

bradlarsen:
> I'm gonna run tests with the typeguard pytest integration (see #1584) before we should consider merging this.

It turns out there are several lies in our type hints throughout Manticore that typeguard + pytest can identify.  I looked at those results for a while, and I don't think I've _introduced_ any new problems there.

I'm gonna do the work of fixing up what typeguard + pytest find in another PR from this.

bradlarsen:
@ehennenfent I think this is ready to go now.

PR #1646, opened 2020-03-31
CI: have pytest report 100 slowest tests

bradlarsen:
This change merely passes an extra argument to `pytest` when it's invoked via GitHub Actions, so that the 100 slowest tests are reported.  This information is useful to identify possible places to tighten up the test suite.


reaperhulk:
codecov delivering the nonsense with coverage (why is merged coverage such a challenge for these stupid services) so ignoring and merging.

PR #1647, opened 2020-03-31
Add a badge to README.md for LGTM

bradlarsen:


ehennenfent:
LGTM apparently only polls repositories once every 24 hours unless you integrate it with the PR review process. I'd like to fix some of the more serious problems before we merge this, but it's still showing code analysis results from March 31st. If it updates by tomorrow I'll go through and patch what I can before we merge this.

PR #1651, opened 2020-04-02
Remove unused arg ans separate output from workspace

feliam:


manticore/core/manticore.py
```diff
@@ -254,7 +254,6 @@ def __init__(self, initial_state, workspace_url=None, policy="random", **kwargs)
         :param initial_state: the initial root `State` object
         :type state: State
         :param workspace_url: workspace folder name
-        :param policy: scheduling policy
         :param kwargs: other kwargs, e.g.
```
ehennenfent, line 257: Can we get a docstring for `outspace_url` as well?

PR #1655, opened 2020-04-06
Use typeguard in CI testing; fix many typing issues

### LINKED ISSUE #1584: Incorporate typeguard pytest plugin

bradlarsen:
**THIS IS A DRAFT PULL REQUEST; NOT READY YET**

Fixes #1584.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,7 +176,7 @@ jobs:
 
         run_tests_from_dir() {
             DIR=$1
-            pytest --durations=100 --cov=manticore -n auto "tests/$DIR"
+            pytest --typeguard-packages=manticore --durations=100 --cov=manticore -n auto "tests/$DIR"
             coverage xml
         }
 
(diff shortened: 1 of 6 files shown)


ehennenfent:
One potential option for this is to copy `ci.yml` to `typeguard.yml` and then replace the first few lines of the file with the following:

```
name: Typeguard Checks
on:
  # Delete the following line before merging:
  pull_request:
  schedule:
    # Run every day at 1 AM
    - cron:  '0 1 * * *'
```

Then you can revert ci.yml to its original state. I think this may be a good compromise between not having _any_ runtime type checking and having the tests for each PR take several hours.

bradlarsen:
The initial pull request here had a lot of changes in it, and I ended up splitting it into several smaller pieces, getting those merged into master (e.g. #1672 and #1673), and rebasing and force-pushing what was left here.

The changes left in this pull request that haven't been incorporated are primarily about actually enabling the typeguard pytest plugin when testing.

It would be good to revisit this one day!  But other priorities take precedence now.

PR #1657, opened 2020-04-07
Fix 2 problems in Linux `sys_open` support & add type hints

bradlarsen:
One problem: `sys_open` had a type error, in which Manticore tried to write `str` data to a file opened in binary mode.  (Adding type hints to the function gets mypy to complain about the problem.)

Another problem: Manticore's `FilesystemStore` workspace implementation didn't support characters such as `/` within key names; if such keys are seen, as in the new `examples/linux/fileio` testcase, Manticore would crash.

This pull request fixes both issues.  The fix for the second issue is an expedient fix, simply replacing `/` characters in keys with `_`.  However, a more correct fix for this problem would involve specifying at the `Store` class level exactly which characters are allowed in keys, maximum lengths, etc.  Without such a fix, `Store` is a leaky abstraction, where limitations of the various subclasses reveal themselves to calling code.


bradlarsen:
Ha, it looks like the example program I added doesn't have the same assumed command-line invocation as the rest of them in `examples/linux`.  I'll have to fix that.

bradlarsen:
This one is ready for review now, and the CI tests pass.

manticore/core/workspace.py
```diff
@@ -570,7 +578,7 @@ def save_stream(self, key, *rest, **kwargs):
     @contextmanager
     def _named_stream(self, name, binary=False, lock=False):
         """
-        Create an indexed output stream i.e. 'test_00000001.name'
+        Create an indexed output stream e.g. 'test_00000001.name'
```
ehennenfent, line 581: I'm gonna get paranoid about this now haha

PR #1659, opened 2020-04-07
Work on "Model is not Available" errors in tests

ehennenfent:
It's gotten annoying how they spontaneously kill our builds. I'd like to find out what's going on, especially if they're genuinely due to resource constraints.


ehennenfent:
I've built this several times but haven't been able to get it to exhibit the "Model is not available" error. Perhaps we should merge it so we'll be able to see what's going on next time the error happens?

ehennenfent:
@ekilmer thanks for reviewing. I don't know if we want to add another dependency to the main build, but for now I'm going to merge as I can't seem to trigger the "Model not available" error on this branch, despite it happening frequently on `master`. Before our next release, we can consider removing it.

PR #1660, opened 2020-04-09
Randomize test suite order when using pytest

bradlarsen:
The idea behind running tests in a random order is that it will help to identify hidden ordering dependencies among test cases.  The principle is that test cases (or perhaps groups of related test cases) shouldn't be order-dependent; they should succeed regardless of the order that they're run in.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,11 +23,11 @@ dev_noks = (
     native_deps
     + [
         "coverage",
-        "Sphinx",
-        "pytest==5.3.0",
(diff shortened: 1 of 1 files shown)


bradlarsen:
Hey, looks like issues with config.py in #1636 are revealed here by the addition of `pytest-randomly`.

bradlarsen:
Abandoning for now.  This would be a good thing to enable eventually, but we first need to fix the ordering-dependence issues already present in Manticore tests.

PR #1661, opened 2020-04-09
Make CodeCov wait for reports from all CI jobs

ehennenfent:
This should prevent the red 'X' from appearing as soon as the first job finishes.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,1 +1,6 @@
 comment: false
+
+codecov:
+  notify:
+    after_n_builds: 8
+    wait_for_ci: yes
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,5 +2,7 @@ comment: false
 
 codecov:
   notify:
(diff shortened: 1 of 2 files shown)


codecov.yml
```diff
@@ -1 +1,6 @@
 comment: false
+
+codecov:
+  notify:
+    after_n_builds: 8
```
bradlarsen, line 5: The `after_n_builds` setting was set to 8 here because that's how many `coverage` jobs there are in CI?  That sounds brittle — it would be good to include a short comment about this.

ehennenfent: Yep, no way for codecov to detect this by itself, unfortunately. Added a comment.

ehennenfent:
I've removed the `fail_ci_if_error` flag for now as I think codecov's current level of stability is going to hinder our development process too much. I'll keep an eye out for missing coverage measurements on PRs moving forward.

PR #1667, opened 2020-04-14
Add type hints to several parts of Manticore

bradlarsen:
These added type hints get us three main benefits:

1. Static type checkers such as `mypy` can help identify inconsistencies in the code
2. Code understanding tools such as `jedi` (used in many editors for go-to-definition functionality) better understand the Manticore codebase
3. Humans may also have an easier time understanding Manticore APIs

### DIFF
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -411,7 +411,7 @@ class Z3Solver(Solver):
         """Recall the last pushed constraint store and state."""
         self._send("(pop 1)")
 
-    def can_be_true(self, constraints: ConstraintSet, expression=True):
+    def can_be_true(self, constraints: ConstraintSet, expression: Union[bool, Bool] = True):
         """Check if two potentially symbolic values can be equal"""
         if isinstance(expression, bool):
             if not expression:
@@ -592,7 +592,7 @@ class Z3Solver(Solver):
                 return last_value
             raise SolverError("Optimizing error, unsat or unknown core")
 
-    def get_value(self, constraints, *expressions):
+    def get_value(self, constraints: ConstraintSet, *expressions):
         """
         Ask the solver for one possible result of given expressions using
         given set of constraints.
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -416,7 +416,7 @@ class Z3Solver(Solver):
         """Recall the last pushed constraint store and state."""
         self._send("(pop 1)")
 
-    def can_be_true(self, constraints: ConstraintSet, expression: Union[bool, Bool] = True):
+    def can_be_true(self, constraints: ConstraintSet, expression: Union[bool, Bool] = True) -> bool:
         """Check if two potentially symbolic values can be equal"""
         if isinstance(expression, bool):
             if not expression:
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -330,8 +330,13 @@ class Z3Solver(Solver):
 
         return buf
 
-    def __readline_and_count(self):
-        buf = self._proc.stdout.readline()
+    def __readline_and_count(self) -> Tuple[str, int, int]:
+        stdout = self._proc.stdout
+        if stdout is None:
+            raise SolverError("Could not read from stdout: file descriptor is None")
+        buf = stdout.readline()
+        if buf is None:
+            raise SolverError("Could not read from stdout")
         return buf, buf.count("("), buf.count(")")
 
     # UTILS: check-sat get-value
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -292,7 +292,7 @@ class Z3Solver(Solver):
         if constraints is not None:
             self._send(constraints)
 
-    def _send(self, cmd: Union[str, ConstraintSet]):
+    def _send(self, cmd: str) -> None:
         """
         Send a string to the solver.
 
--- a/manticore/native/cpu/abstractcpu.py
+++ b/manticore/native/cpu/abstractcpu.py
@@ -264,9 +264,9 @@ class Abi:
     Used for function call and system call models.
     """
 
-    def __init__(self, cpu):
+    def __init__(self, cpu: "Cpu"):
         """
-        :param manticore.core.cpu.Cpu cpu: CPU to initialize with
+        :param CPU to initialize with
         """
         self._cpu = cpu
 
--- a/manticore/native/cpu/abstractcpu.py
+++ b/manticore/native/cpu/abstractcpu.py
@@ -392,7 +392,7 @@ class Abi:
 platform_logger = logging.getLogger("manticore.platforms.platform")
 
 
-def unsigned_hexlify(i):
+def unsigned_hexlify(i: Any) -> Any:
     if type(i) is int:
         if i < 0:
             return hex((1 << 64) + i)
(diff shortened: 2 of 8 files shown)


manticore/core/smtlib/solver.py
```diff
@@ -411,7 +416,7 @@ def _pop(self):
         """Recall the last pushed constraint store and state."""
         self._send("(pop 1)")
 
-    def can_be_true(self, constraints: ConstraintSet, expression=True):
+    def can_be_true(self, constraints: ConstraintSet, expression: Union[bool, Bool] = True):
```
ekilmer, line 419: This is missing a return statement
```suggestion
    def can_be_true(self, constraints: ConstraintSet, expression: Union[bool, Bool] = True) -> bool:
```

manticore/core/smtlib/solver.py
```diff
@@ -592,7 +597,7 @@ def optimize(self, constraints: ConstraintSet, x: BitVec, goal: str, max_iter=10
                 return last_value
             raise SolverError("Optimizing error, unsat or unknown core")
 
-    def get_value(self, constraints, *expressions):
+    def get_value(self, constraints: ConstraintSet, *expressions):
```
ekilmer, line 600: This one is also missing a return type hint

manticore/core/smtlib/solver.py
bradlarsen: Thanks; that looks good!  Without a return type annotation, `mypy` assumes `Any`, so indeed, the `bool` return type here is more specific (and matches the superclass).

manticore/native/cpu/abstractcpu.py
```diff
@@ -264,9 +264,9 @@ class Abi:
     Used for function call and system call models.
     """
 
-    def __init__(self, cpu):
+    def __init__(self, cpu: "Cpu"):
```
ekilmer, line 267: Does we need the quotes around `Cpu`? Is it because `Cpu` is defined later?

manticore/native/memory.py
```diff
@@ -577,18 +576,18 @@ class Memory(object, metaclass=ABCMeta):
     This class handles all virtual memory mappings and symbolic chunks.
     """
 
-    def __init__(self, maps=None, cpu=StubCPU()):
+    def __init__(self, maps: Optional[Iterable[Map]] = None, cpu=StubCPU()):
```
ekilmer, line 579: I wonder... I know you didn't touch this, but shouldn't the `cpu=StubCPU()` be `cpu=None` and then do
```
if not cpu:
    cpu = StubCPU()
```

manticore/platforms/linux.py
```diff
@@ -1043,13 +1043,13 @@ def set_entry(self, entryPC):
         self.current.PC = elf_entry
         logger.debug(f"Entry point updated: {elf_entry:016x}")
 
-    def load(self, filename: str, env) -> None:
+    def load(self, filename: str, env_list: List) -> None:
```
ekilmer, line 1046: Minor nit: Isn't the `_list` suffix superfluous now that there is a type hint?

manticore/native/cpu/abstractcpu.py
bradlarsen: Exactly!  In those circumstances, you need to use a string to annotate the type as a "forward reference".

ekilmer: Yep! Okay, just making sure we aren't putting things in quotes that don't need to be 👍

manticore/platforms/linux.py
bradlarsen: I renamed the variable to address a type hint error later in the code -- on line 1063, `env` (the previous name) is reassigned to a different type, which `mypy` doesn't like.

An alternative, I suppose, would be to keep the `env` parameter named the same, and rename the variable in the body instead.  Let me know if you want that.

manticore/core/smtlib/solver.py
bradlarsen: What's the return type?  I was having trouble figuring out what it is.

`mypy` treats unannotated return types as `Any` return types, so the lack of an explicit annotation there _shouldn't_ be a problem.  Though if we have a more accurate type, that can help both readers & tools.

manticore/platforms/linux.py
ekilmer: Ahh okay. Yeah, I'm not sure I like how `mypy` handles that. I think how it is here is okay. 

Do you have any input @ehennenfent ?

manticore/core/smtlib/solver.py
ekilmer: Hmmm. I was hoping typeguard would be able to tell us. We can get to it another time, I guess.

manticore/platforms/linux.py
bradlarsen: Yeah mypy is pretty strict: it doesn't permit a variable to be reassigned with a different type.  So it is.

manticore/core/smtlib/solver.py
ehennenfent: You can pass any type in `expressions` and it'll get returned from the values if it's not symbolic.

manticore/native/memory.py
```diff
@@ -843,15 +838,15 @@ def mappings(self):
 
         return sorted(result)
 
-    def __str__(self):
+    def __str__(self) -> str:
```
ehennenfent, line 841: Does `mypy` not automatically expect `__str__` to return a string?

manticore/platforms/linux.py
ehennenfent: I was grumpy about this at first but have now been converted to the cult of "don't reuse variable for data of different types" 

https://twitter.com/conectado2/status/1248032863238393857

manticore/native/memory.py
bradlarsen: Six of one, half-dozen of another?

manticore/native/memory.py
bradlarsen: No, `mypy` complains if you claim the return type of a `__str__` method is something other than `str`!

But, with the way we have `mypy` configured, only methods that have any type hints on them at all will be inspected by `mypy`.  So by adding the explicit return type annotation here, `mypy` will type-check the body of the method (which it otherwise wouldn't do).

ehennenfent: > the way we have mypy configured, only methods that have any type hints on them at all will be inspected by mypy

Barring a major refactor I think this works well for us, but I'm somewhat surprised that mypy doesn't override that for dunder methods, which really really should conform to the expected behavior even if it's not required of them.

manticore/native/cpu/abstractcpu.py
```diff
@@ -392,7 +392,7 @@ def invoke(self, model, prefix_args=None):
 platform_logger = logging.getLogger("manticore.platforms.platform")
 
 
-def unsigned_hexlify(i):
+def unsigned_hexlify(i: Any) -> Any:
```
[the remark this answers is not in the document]
bradlarsen: Yeah, why no `SupportsStr` type?  Probably because _everything_ in Python can be converted to a string — `__str__` is a method on `object`.

Additionally, the type of `unsigned_hexlify` is difficult to pin down precisely — if it is given a non-`int` argument, `unsigned_hexlify` is the identify function.  Otherwise, `unsigned_hexlify` returns a `str`.

I added the explicit signature again there, silly as it is, so that `mypy` would check the body of function, which it wouldn't otherwise do in the absence of type hints.

PR #1668, opened 2020-04-14
Fix several incorrect type hints

bradlarsen:
Using typeguard in local test runs (see #1584), I found several issues where type hints in Manticore are in fact not accurate at runtime.

This pull request fixes nearly all of those, and additionally fixes a couple bugs.


manticore/core/smtlib/solver.py
```diff
@@ -276,7 +276,7 @@ def __del__(self):
         except Exception as e:
             logger.error(str(e))
 
-    def _reset(self, constraints=None):
+    def _reset(self, constraints: Optional[str] = None) -> None:
```
ekilmer, line 279: Why is this not
```suggestion
    def _reset(self, constraints: Optional[Union[str, ConstraintSet]] = None) -> None:
```
like `_send`?

I see that you call `.to_string()` on the `constraints` to send to `_reset`. If it's the case that `_send` should always receive a `str`, then I think that type annotation should be changed/checked as well, at least for consistency. From my cursory search, it seems we only send `str`s to `_send`.

I'm generally in favor of more restricted types (i.e., not `Union`), if possible.

bradlarsen: So, you're asking why `_reset` doesn't have the same signature as `_send`?  Good question!

The change here _adds_ a type hint to `_reset`, giving it a tighter signature than what `_send` currently has.  The other half of this — tightening the type of `_send` to `def _send(self, cmd: str)` — ended up in my other open pull request (#1667).  Let me add that to _this_ pull request instead, for coherence.

I was trying to keep _this_ pull request focused on fixing type hints that are manifestly wrong at runtime (actually shown dynamically to not be accurate type hints).

ekilmer: > I was trying to keep this pull request focused on fixing type hints that are manifestly wrong at runtime (actually shown dynamically to not be accurate type hints).

Ah. Okay, that makes sense. There's like 3 open PRs for type hints, and I couldn't keep the differences straight in my head haha

bradlarsen:
Note: the `CI / tests (ethereum_bench)` check failed a couple times here before it succeeded.  I don't think it's related to changes made in this PR, and I was also unable to reproduce the failure locally.

manticore/platforms/platform.py
```diff
@@ -10,14 +12,14 @@ class OSException(Exception):
 
 
 @wrapt.decorator
-def unimplemented(wrapped, _instance, args, kwargs):
+def unimplemented(wrapped: Callable, _instance, args: Iterable, kwargs: Mapping):
```
ehennenfent, line 15: Did Mypy/typeguard complain about these? I would've expected to see `tuple` and `dict` as the types for these instead, but perhaps it's more general than I thought.

manticore/core/workspace.py
```diff
@@ -28,12 +28,12 @@ def __exit__(self, *excinfo):
 import errno
 import threading
 from ..utils import config
-from ..utils.helpers import PickleSerializer
+from ..utils.helpers import PickleSerializer, StateSerializer
```
ehennenfent, line 31: StateSerializer doesn't seem to be used elsewhere in this file

bradlarsen: Good catch!  That must have been forgotten from an earlier edit.  We don't have a linter on unused imports, so here it is.

```suggestion
from ..utils.helpers import PickleSerializer
```

manticore/platforms/platform.py
bradlarsen: Hmmm, indeed, `Iterable` and `Mapping` may indeed be more general than needed here.  I _thought_ writing `args: Tuple` and `kwargs: Dict` had been problematic, but now I'm having trouble figuring out exactly what.  `mypy` definitely doesn't complain.

Let me change this to the more specific types (`Tuple` and `Dict`).

Note, `Iterable` and `Mapping` are used for expressing certain "duck types" in Python: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#standard-duck-types

manticore/core/smtlib/constraints.py
```diff
@@ -73,13 +73,11 @@ def __len__(self):
             return len(self._constraints) + len(self._parent)
         return len(self._constraints)
 
-    def add(self, constraint, check=False):
```
ehennenfent, line 76: Strictly speaking, one is supposed to access constraint sets via `state.constrain`, but in many places our code uses `state.constraints.add`. We should leave this `check` parameter in place just in case any user code relies on it, and update the comment to explain what it actually does.

manticore/core/smtlib/constraints.py
```diff
@@ -101,12 +99,6 @@ def add(self, constraint, check=False):
 
         self._constraints.append(constraint)
 
-        if check:
-            from ...core.smtlib import solver
-
-            if not solver.check(self):
```
feliam, line 107: Nobody was using it. "solver.check" does not exists. This should be Z3Solver.instance().check

ehennenfent: Whoops, in that case, I stand corrected. Nonetheless, if you've fixed it in another branch and are using it to debug, we should leave it in.

manticore/core/smtlib/constraints.py
bradlarsen: At present, as @feliam pointed out [in another comment](https://github.com/trailofbits/manticore/pull/1668/files/fdb4265b759b55d55958d9800aa4edad50920de3#r409885491), `check` is unused throughout the codebase.  Also, the current bit of code in this method that attempted to do anything with `check` was broken and would not possibly work.

@feliam @ehennenfent, would you guys like me to fix this `check` implementation, or delete the currently unused parameter entirely?

manticore/platforms/linux.py
```diff
@@ -3204,7 +3201,7 @@ def sys_accept(self, sockfd, addr, addrlen):
         # TODO: Make a concrete connection actually an option
         # return super().sys_accept(sockfd, addr, addrlen)
 
-    def sys_open(self, buf: int, flags: int, mode: int) -> int:
+    def sys_open(self, buf: int, flags: int, mode: Optional[int]) -> int:
```
ehennenfent, line 3204: We don't appear to actually use `mode` for anything. It gets passed around a few times but ultimately ignored in favor of `flags`. Probably safe to strip it out if you want.

manticore/core/smtlib/solver.py
```diff
@@ -418,7 +418,7 @@ def can_be_true(self, constraints: ConstraintSet, expression=True):
                 return expression
             else:
                 # if True check if constraints are feasible
-                self._reset(constraints)
+                self._reset(constraints.to_string())
```
ehennenfent, line 421: Previously, we could pass `ConstraintSet`s around without needing to explicitly render them to strings as the f-string in `_send` would implicitly call `__str__`. Just offhand I can't see why one way of doing this would be inherently better than the other, but I'm guessing that this way is more convenient from a type standpoint. Can you elaborate?

manticore/core/smtlib/constraints.py
bradlarsen: @feliam says in Slack:
> It is not tested and it is pretty easy to emulate   in debug situations.  Kill it
> We'll re add it (with it corresponding tests) if we find a clear need in the future

manticore/platforms/linux.py
bradlarsen: If you check `man 2 open` on Linux, you'll see that 2 forms of the `open` function exist, one that takes `mode`.  I found that in testing / experimentation with larger binaries, that if `mode` is removed, Manticore gets runtime errors when it hits that form of syscall.  Additionally, if its type is listed just as `int`, `typeguard` complains at times.

The syscall invocation mechanism in Manticore is actually pretty subtle; I wonder now if `mode` should _actually_ be defined via varargs... but I'll leave that investigation for future work.

manticore/core/workspace.py
```diff
@@ -132,7 +136,7 @@ def load_value(self, key: str, binary: bool = False):
             return s.read()
 
     @contextmanager
-    def save_stream(self, key: str, binary: bool = False) -> Generator[Any, None, None]:
```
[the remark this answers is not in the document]
bradlarsen: So, the lowdown on generators in Python...

You're surely familiar with using `yield` to write iterator-type functions, but `yield` is actually far more flexible in Python: generator objects (typically created from functions that have `yield` in them) can additionally be _sent_ values, and also _return_ a value when they are done generating.

Generators in Python can be used to express [coroutines](https://en.wikipedia.org/wiki/Coroutine); see [PEP342](https://www.python.org/dev/peps/pep-0342/) for more details.

Anyway, to answer your questions:
- The return type of a function with `yield` in it [is indeed `Generator[yield_type, send_type, return_type]`](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#generators).
- The return type that was written there previously _is_ correct, but there are currently incompatibilities that arise when you try to use that `Generator` return type annotation, `mypy`, `typeguard`, and the `@contextlib` decorator — it's not possible at present to write a return type hint there that satisfies both `mypy` and `typeguard`.

manticore/core/smtlib/solver.py
bradlarsen: Certainly!  [Elsewhere in this pull request](https://github.com/trailofbits/manticore/pull/1668/files#r408833413), the type of the `Z3Solver._reset()` method was annotated explicitly as _only_ accepting an optional string argument of constraints, and `Z3Solver._send()` had its type annotated as taking a string (rather than _either a string or ConstraintSet_).

My thinking is that explicit `Union` types are more difficult to reason about that single types or interface types — more cases to worry about in the implementation!

In light of the `_send()` and `_reset()` changes, the call sites of those methods need to be made consistent.  Hence the addition of the explicit `.to_string()` calls.  `mypy` will let you know if you've crossed the streams.

ehennenfent: > My thinking is that explicit Union types are more difficult to reason about that single types or interface types

Okay, I agree. Since `_send` and `_reset` aren't user-facing this shouldn't break anything either.

PR #1669, opened 2020-04-14
Fix type confusion in manual CPU tests; delete dead testing code

bradlarsen:
It turns out that most of `tests/native/mockmem.py` was dead code to begin with.  Additionally, using `typeguard` during testing (see #1584), it is revealed that the use of `tests.native.mockmem.Memory` turns some type hints elsewhere into lies (since the mocked version isn't a subclass of `Memory`).

The simplest fix: massage the single file that uses `tests.native.mockmem.Memory`, and then delete that whole file.


tests/native/test_cpu_manual.py
```diff
@@ -8,33 +8,13 @@
...
-
-
-class RWOperand(ROOperand):
-    """ Mocking class for operand rw """
-
-    def write(self, value):
-        self.value = value & ((1 << self.size) - 1)
-        return self.value
-
-
```
bradlarsen, line 37: Note, this was already dead code, prior to these changes.

tests/native/test_cpu_manual.py
```diff
@@ -170,31 +150,6 @@ class SymCPUTest(unittest.TestCase):
...
-        def read(self):
-            return self.value & ((1 << self.size) - 1)
-
-    class RWOperand(ROOperand):
-        """ Mocking class for operand rw """
-
-        def write(self, value):
-            self.value = value & ((1 << self.size) - 1)
-            return self.value
```
bradlarsen, line 188: Note, this was already dead code, prior to these changes.

tests/native/test_cpu_manual.py
```diff
@@ -170,31 +150,6 @@ class SymCPUTest(unittest.TestCase):
...
-
-    def setUp(self):
-        mem = MockMemory()
-        self.cpu = I386Cpu(mem)  # TODO reset cpu in between tests...
-        # TODO mock getchar/putchar in case the instruction accesses memory directly
-
-    def tearDown(self):
-        self.cpu = None
-
```
bradlarsen, line 197: Note: the `self.cpu` parameter was only used in a few tests in this file previously.  So instead of creating & destroying it around each test, I changed the code to explicitly create a `Cpu` object only within the test cases that use it.

PR #1671, opened 2020-04-17
printable_bytes

smoelius:
Extract printable bytes from return data in the same way as it is
currently done for logs.


manticore/platforms/evm.py
```diff
@@ -880,6 +883,12 @@ def world(self):
     def check256int(value):
         assert True
 
+    @staticmethod
```
feliam, line 886: Should not this be a core.util auxiliar method? Not really an EVM functionality.

smoelius: Fine with me.  I'll make the change.

PR #1672, opened 2020-04-17
Rework syscall invocation for proper behavior under typeguard

bradlarsen:
Previously, using Typeguard with Manticore would break several emulated syscalls.  With this commit, it does not.

Some background information:

When a syscall is made from an emulated binary, Manticore uses the syscall number to look up the appropriate Python method that models that syscall, and then uses Python introspection to massage arguments to that syscall model function as deemed appropriate.

Previously, this mechanism used the deprecated `inspect.getfullargspec` to determine the number of arguments to the model function, and whether or not it takes varargs.

However, `inspect.getfullargspec` doesn't look through wrapper functions; it looks only at exactly the function object it is given. This is an issue, however, when trying to inspect _decorated_ functions. (The use of a decorator in Python introduces a _new_ function object that wraps the decorated item.)

How did that break Manticore when using Typeguard?  It turns out that when using Typeguard via the `--typeguard-packages=manticore` option to `pytest`, the Typeguard plugin implicitly adds a `@typeguard.check_types` decorator on _every_ function & method in the `manticore` package.

In this way, each syscall implementation function in Manticore ends up with a wrapper around it, and the syscall invocation mechanism based on `inspect.getfullargspec` would somewhat quietly cause syscalls to break.

Now, instead of using `inspect.getfullargspec`, Manticore's syscall invocation mechansim uses the non-deprecated `inspect.signature` API to get the needed information.  This API _does_ look through wrapper functions. Additionally, it allows us to get rid of some conditional logic, about whether `self` appears in a function's parameter list or not.

### DIFF
--- a/manticore/native/cpu/abstractcpu.py
+++ b/manticore/native/cpu/abstractcpu.py
@@ -320,8 +320,8 @@ class Abi:
         Extract arguments for model from the environment and return as a tuple that
         is ready to be passed to the model.
 
-        :param callable model: Python model of the function
-        :param tuple prefix_args: Parameters to pass to model before actual ones
+        :param model: Python model of the function
+        :param prefix_args: Parameters to pass to model before actual ones
         :return: Arguments to be passed to the model
         """
         sig = inspect.signature(model)
--- a/manticore/native/cpu/abstractcpu.py
+++ b/manticore/native/cpu/abstractcpu.py
@@ -23,10 +23,16 @@ from capstone.arm64 import ARM64_REG_ENDING
 from capstone.x86 import X86_REG_ENDING
 from capstone.arm import ARM_REG_ENDING
 
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, Optional, Tuple
 
 logger = logging.getLogger(__name__)
 register_logger = logging.getLogger(f"{__name__}.registers")
+
+
+def _sig_is_varargs(sig: inspect.Signature) -> bool:
+    VAR_POSITIONAL = inspect.Parameter.VAR_POSITIONAL
+    return any(p.kind == VAR_POSITIONAL for p in sig.parameters.values())
+
 
 ###################################################################################
 # Exceptions
@@ -309,7 +315,7 @@ class Abi:
             yield base
             base += word_bytes
 
-    def get_argument_values(self, model, prefix_args):
+    def get_argument_values(self, model: Callable, prefix_args: Tuple) -> Tuple:
         """
         Extract arguments for model from the environment and return as a tuple that
         is ready to be passed to the model.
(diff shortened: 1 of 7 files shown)


manticore/native/cpu/abstractcpu.py
```diff
@@ -309,26 +315,21 @@ def values_from(self, base):
...
             base += word_bytes
 
-    def get_argument_values(self, model, prefix_args):
+    def get_argument_values(self, model: Callable, prefix_args: Tuple) -> Tuple:
         """
         Extract arguments for model from the environment and return as a tuple that
         is ready to be passed to the model.
 
         :param callable model: Python model of the function
         :param tuple prefix_args: Parameters to pass to model before actual ones
```
ekilmer, line 324: Nit: Remove the types in the docstring

ehennenfent:
> Now, instead of using inspect.getfullargspec, Manticore's syscall invocation mechansim uses the non-deprecated inspect.signature API to get the needed information. This API does look through wrapper functions.

Wondering: Do we still need `wrapt`? We added it in #1384 to help deal with this problem. It's supposed to improve type checking, so it may be providing some other useful feature that I'm not aware of.

manticore/native/cpu/abstractcpu.py
bradlarsen: Like this?  ;)
```suggestion
        :param model: Python model of the function
        :param prefix_args: Parameters to pass to model before actual ones
```

bradlarsen:
> Do we still need `wrapt`?

@ehennenfent I looked at this a bit closer, and I don't think it's necessary any more!  So I've rephrased the single use of `wrapt.decorator` & eliminated that dependency in this branch.  Let's see how the tests look.

manticore/platforms/platform.py
```diff
@@ -11,17 +11,20 @@ class OSException(Exception):
...
-        f"Unimplemented system call: %s: %s(%s)",
-        addr_str,
-        wrapped.__name__,
-        ", ".join(hex(a) if isinstance(a, int) else str(a) for a in args),
-    )
-    return wrapped(*args, **kwargs)
+def unimplemented(wrapped: Callable) -> Callable:
+    @wraps(wrapped)
+    def new_wrapped(*args, **kwargs):
+        cpu = getattr(getattr(_instance, "parent", None), "current", None)
```
ehennenfent, line 17: `_instance` is probably going to be undefined here, right?

bradlarsen: yes, absolutely.  I didn't read that code closely enough on my first couple attempts :D

bradlarsen:
`wrapt` is gone, there is more thorough testing, and the CI tests pass now.  I think this is ready to get merged; just needs to be approved.

manticore/platforms/linux.py
```diff
@@ -33,10 +32,13 @@
 from ..native.state import State
 from ..platforms.platform import Platform, SyscallNotImplemented, unimplemented
 
-from typing import Any, Dict, IO, List, Optional, Set, Tuple, Union
+from functools import wraps
+from typing import Any, Callable, cast, Deque, Dict, IO, List, Optional, Set, Tuple, TypeVar, Union
```
ekilmer, line 36: Some of these are unnecessary, since there are no other changes in this file

bradlarsen: Good eye!  Whoops :|

One of these days we should consider turning on something like `isort`, or an unused import checker, as a lint step in our CI.

PR #1675, opened 2020-04-22
Model of basic_string::append(char *)

sschriner:


sschriner:
Pausing work on append to implement capacity() and reserve(n) which will be used in append

PR #1676, opened 2020-04-23
VMTests tests for istanbul

feliam:
Old upstream **VMtests** were recently refilled as BlockchainTests and now support **Istanbul**. https://github.com/ethereum/tests/tree/develop/BlockchainTests/ValidBlocks/VMTests

This PR:
* Balances/gas/tx related istanbul support
* removes VMTests for Frontier 
* updates make_VMTest.py to handle new format in concrete mode
* passes all Istanbug VMTest 
* adds new refund logic to SSTORE
* adds transaction fee logic 

Removes:
* make VMTest in symbolic mode  (TODO)


manticore/core/state.py
```diff
@@ -296,6 +296,20 @@ def concretize(self, symbolic, policy, maxcount=7):
...
+                vals = (True,)
+            else:
+                #We assume the path constraint was feasible to begin with
+                vals = (False,)
+        elif policy == "PESI":
+            logger.info("Pesimistic case when forking")
+            if self._solver.can_be_false(self._constraints, symbolic):
+                vals = (False,)
+            else:
+                #We assume the path constraint was feasible to begin with
```
ehennenfent, line 311: The comments for both cases are the same. Copy & Paste gaffe?

tests/ethereum/test_general.py
```diff
@@ -1076,7 +1075,7 @@ def test_event_forwarding_after_state_fork_during_message_call(self):
 
         contract Lib {
            function isSeven(uint a) public pure returns (bool) {
-               if (a == 7) {
+               if (a == 0x414243444546) {
```
ehennenfent, line 1078: The function is still called `isSeven` but the value is now compared to 0x414243444546

manticore/ethereum/manticore.py
```diff
@@ -390,7 +390,7 @@ def contract_accounts(self):
     def get_account(self, name):
         return self._accounts[name]
 
-    def __init__(self, workspace_url: str = None, policy: str = "random"):
+    def __init__(self, plugins=None, **kwargs):
```
ehennenfent, line 393: This is definitely a user-facing API change, but it might be okay because I can't think of a time I've ever seen anybody use the workspace_url or policy parameters, for EVM or otherwise. If anyone has objections, we can, of course, tweak this so that workspace_url and policy will still be in their correct position. 

If nothing else, the `param` docstring items are still outdated and should be changed

manticore/platforms/evm.py
```diff
@@ -80,6 +81,17 @@ def globalfakesha3(data):
     ),
 )
 
+consts.add(
+    "txfail",
+    default="optimistic",
+    description=(
+        "Default behavior for tansaction failing because not enough funds."
```
ehennenfent, line 88:
```suggestion
        "Default behavior for transaction failing because not enough funds."
```

manticore/platforms/evm.py
```diff
@@ -901,6 +977,13 @@ def disassemble(self):
     def PC(self):
         return self.pc
 
+    def _getcode(self, pc):
+        bytecode = self.bytecode
+        for pc_i in range(pc, len(bytecode)):
+            yield simplify(bytecode[pc_i]).value
+        while True:
+            yield 0
```
ehennenfent, line 985:
```suggestion
            yield 0  # STOP opcode
```

manticore/native/memory.py
```diff
@@ -1320,14 +1320,14 @@ def _try_get_solutions(self, address, size, access, max_solutions=0x1000, force=
...
-        solutions = solver.get_all_values(self.constraints, address, maxcnt=max_solutions)
+        solutions = solver.get_all_values(self.constraints, address, maxcnt=max_solutions, silent=True)
 
         crashing_condition = False
         for base in solutions:
             if not self.access_ok(slice(base, base + size), access, force):
                 crashing_condition = Operators.OR(address == base, crashing_condition)
 
-        crash_or_not = solver.get_all_values(self.constraints, crashing_condition, maxcnt=3)
+        crash_or_not = solver.get_all_values(self.constraints, crashing_condition, maxcnt=2, silent=True)
```
ehennenfent, line 1330: Still confused about what's going on here. We want either exactly one or exactly two solutions from `get_all_values`, and given that it checks if `len(results) >= maxcnt`, it seems like it should be three in this scenario, with `silent` set to `False`. Since this is a native change and this is an EVM PR, perhaps it would be worth extracting this so we can test it in isolation? Let's discuss on Slack tomorrow.

manticore/core/smtlib/expression.py
```diff
@@ -1150,11 +1158,21 @@ def is_known(self, index):
...
+                #check if the concrete index is explicitly in written
+                if isinstance(i, Constant) and index.value == i.value:
                     return BoolConstant(True)
+
+                #Build an expression to check if our concrete index could be the
+                # solution for anyof the used symbolic indexes
+                is_known_index = BoolOr(is_known_index.cast(index == i), is_known_index)
+            return is_known_index
+
+        # The index is symbolic we need to compare it agains it all
```
smoelius, line 1172:
```suggestion
        # The index is symbolic we need to compare it against all
```

manticore/core/smtlib/visitors.py
```diff
@@ -325,12 +325,12 @@ def visit_BitVecSignExtend(self, expression, *operands):
...
     def visit_BitVecExtract(self, expression, *operands):
-        if all(isinstance(o, Constant) for o in expression.operands):
-            value = expression.operands[0].value
+        if all(isinstance(o, Constant) for o in operands):
+            value = operands[0].value
             begining = expression.begining
             end = expression.end
             value = value >> begining
-            mask = 2 ** (end - begining + 1) - 1
+            mask = (1 << (end - begining)) - 1
```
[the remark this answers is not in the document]
feliam: `end` and `beginning` must be constants here so this change is meh.
In the middle of the debugging I started doubting and changed the mast to something I thought it was more readable at the time.

manticore/core/state.py
feliam: No it is actually what I wanted to write there. 
This is forking and concretizing `symbolic`. If possible, "PESI" ignores the `True` option and "OPTI" ignores the `False` option. 

Preconditions & assumptions 
* `symbolic` is a Bool. 
* self._constraints are feasible (symbolic has at least one solution)
* solutions/vals can be: true, false or true|false

manticore/ethereum/manticore.py
feliam: Also this is walking toward removing "register_plugin" et.al. and forcing it to be registered at Manticore creation.

`policy` is actually ignored and `workspace_url` is more a parameter of the parent class

manticore/native/memory.py
feliam: `crashing_condition` being a Bool can have at the most 2 solutions. 
It's an optimization that avoids the third check-sat

crashing_condition != True
crashing_condition != False
check-sat
obviously unsat

setup.py
```diff
@@ -52,13 +55,14 @@ def rtd_dependent_deps():
...
         "prettytable",
-        "rlp",
         "ply",
+        "rlp",
         "crytic-compile>=0.1.1",
         "wasm",
         "dataclasses; python_version < '3.7'",
         "pyevmasm>=0.2.3",
         "psutil",
+        "pyevmasm>=0.2.2",
```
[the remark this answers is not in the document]
feliam: Also duplicated!

manticore/core/state.py
```diff
@@ -296,6 +296,20 @@ def concretize(self, symbolic, policy, maxcount=7):
...
             vals = [self._solver.get_value(self._constraints, symbolic)]
+        elif policy == "OPTI":
+            logger.info("Optimistic case when forking")
+            if self._solver.can_be_true(self._constraints, symbolic):
+                vals = (True,)
+            else:
+                #We assume the path constraint was feasible to begin with
+                vals = (False,)
+        elif policy == "PESI":
+            logger.info("Pesimistic case when forking")
```
smoelius, line 307:
```suggestion
            logger.info("Pessimistic case when forking")
```

manticore/core/state.py
```diff
@@ -40,7 +40,7 @@ class Concretize(StateException):
 
     """
 
-    _ValidPolicies = ["MIN", "MAX", "MINMAX", "ALL", "SAMPLED", "ONE"]
+    _ValidPolicies = ["MIN", "MAX", "MINMAX", "ALL", "SAMPLED", "ONE", "PESI", "OPTI"]
```
smoelius, line 43:
```suggestion
    _ValidPolicies = ["MIN", "MAX", "MINMAX", "ALL", "SAMPLED", "ONE", "PESSI", "OPTI"]
```
Also, maybe explain what "PESSI" and "OPTI" mean?  They're a little opaque.

manticore/ethereum/manticore.py
```diff
@@ -615,11 +615,12 @@ def solidity_create_contract(
                     )
                 else:
                     contract_account = self.create_contract(
-                        owner=owner, init=md._init_bytecode, balance=balance
+                        owner=owner, init=md._init_bytecode, balance=0, gas=gas
                     )
-
                 if contract_account is None:
+                    return None
                     raise EthereumError("Failed to build contract %s" % contract_name_i)
```
smoelius, line 622: Isn't this `raise` unreachable?  Is that intentional?

tests/ethereum/test_regressions.py
```diff
@@ -273,7 +273,7 @@ def test_addmod(self):
         caller = 0x42424242424242424242
         value = 0
         bytecode = ""
-        vm = evm.EVM(constraints, address, data, caller, value, bytecode)
+        vm = evm.EVM(constraints, address, data, caller, value, bytecode, gas=23000)
```
smoelius, line 276: 2300?

Same question applies to line 333 below.

manticore/ethereum/manticore.py
```diff
@@ -1019,22 +1036,24 @@ def multi_tx_analysis(
...
             create_value = 0
+
         contract_account = self.solidity_create_contract(
             solidity_filename,
             contract_name=contract_name,
             owner=owner_account,
             args=args,
             compile_args=compile_args,
             balance=create_value,
+            gas=230000,
```
smoelius, line 1056: 2300?
 
Same question applies to line 1099 below.

manticore/ethereum/manticore.py
```diff
@@ -1421,7 +1442,7 @@ def _terminate_state_callback(self, state, e):
             logger.info(
                 "Manticore exception: state should be terminated only at the end of the human transaction"
             )
-
+        s"""
```
smoelius, line 1445:
```suggestion
        """
```

manticore/platforms/evm.py
```diff
@@ -80,6 +81,17 @@ def globalfakesha3(data):
     ),
 )
 
+consts.add(
+    "txfail",
+    default="optimistic",
+    description=(
+        "Default behavior for transaction failing because not enough funds."
+        "optimistic: Assume there is always enough funds to pay for value and gas. Wont fork"
+        "pesimistic: Assume the balance is never enough for paying fo a transaction. Wont fork"
```
smoelius, line 90:
```suggestion
        "pessimistic: Assume the balance is never enough for paying fo a transaction. Wont fork"
```

manticore/platforms/evm.py
```diff
@@ -986,84 +1067,48 @@ def _consume(self, fee):
...
-                self.constraints.add(Operators.UGT(self.gas, fee))
-            else:
-                logger.debug(
-                    f"Not enough gas for instruction {self.instruction.name} at 0x{self.pc:x}"
-                )
+            self.constraints.add(self._failed == False)
+            if self.is_failed():
                 raise NotEnoughGas()
+
         elif consts.oog == "pesimistic":
```
smoelius, line 1101:
```suggestion
        elif consts.oog == "pessimistic":
```

manticore/ethereum/manticore.py
feliam: It got important now. Because the gas got debited from the server and all that.
I just wanted to give a generic amount that is normally enough to run through the creation. 
Maybe we need a conf?  idk

tests/ethereum/test_regressions.py
feliam: I think just wanted the execution to have enough gas.

manticore/ethereum/manticore.py
feliam: Unreachable/unintentional. Was in the middle of deciding what a failed contract creation should do.

PR #1677, opened 2020-04-23
Sort Constraint Sets during Translation

ehennenfent:
This is an attempt to make Z3 performance more predictable by making the constraint order fixed. It does a kinda-topological sort on the expressions by sorting them based on the order in which they were created. The code for doing this is not pretty, but in general, it should make the order of constraints in the smt files dependent on the order in which states are explored, not the pseudo-random order we draw the constraints out of a `set`.


manticore/core/smtlib/expression.py
```diff
@@ -18,12 +18,20 @@ class ExpressionException(SmtlibError):
 class Expression:
     """ Abstract taintable Expression. """
 
-    __slots__ = ["_taint"]
+    # Static data bad, I know. Since the creation order is only used to order the expressions
+    # consistently, we don't need to worry about race conditions here. Collisions will introduce
+    # a little bit of nondeterminism, but should be rare. UNLESS state merging gets involved,
+    # in which case I haven't thought enough about what might happen.
+    _nth: int = 0  #: Total number of existing Expressions
```
ehennenfent, line 25: Though I think it's ultimately fine (provided no one else tries to use this field), this is an ugly solution and we can discuss potential alternatives. As-is, it's much cheaper than grabbing a timestamp or using a lock to ensure consistency.

ehennenfent:
TODO: This will need some tests come Monday

ehennenfent:
Closing this as we addressed the `get_all_values` bug through other means, and we have an independent rewrite of the expression system in progress.

PR #1681, opened 2020-04-24
Symbolic model for strcpy

sschriner:
Draft for work on strcpy. Working on adding symbolic support.


manticore/native/models.py
```diff
@@ -140,3 +140,38 @@ def strlen(state, s):
...
+    Algorithm: Copy every byte from the src to dst until finding a byte that will is or must be '\000'
+
+    :param State state: current program state
+    :param int dst: destination string address
+    :param int src: source string address
+    :return: pointer to the dst
+    :rtype: int
+    """
+    cpu = state.cpu
+    ret = dst
```
ekilmer, line 158: I think this needs to be after we concretize `dst`, in the case that it's symbolic

ekilmer: I'm also not sure if that matters, since we `raise` the exception anyway and divert control-flow. Idk.

manticore/native/models.py
```diff
@@ -140,3 +140,38 @@ def strlen(state, s):
...
+    """
+    strcpy symbolic model
+
+    Algorithm: Copy every byte from the src to dst until finding a byte that will is or must be '\000'
+
+    :param State state: current program state
+    :param int dst: destination string address
+    :param int src: source string address
+    :return: pointer to the dst
+    :rtype: int
```
ekilmer, line 155: Could you add the type hints to the function itself, please, then remove them here

manticore/native/models.py
sschriner: Ah I don't know either but rather be safe than sorry lol

manticore/native/models.py
sschriner: done :)

manticore/native/models.py
```diff
@@ -57,7 +59,37 @@ def _find_zero(cpu, constrs, ptr):
...
-def strcmp(state, s1, s2):
+def _find_zeros(cpu, constrs, ptr):
+    """
+    Helper for finding the closest NULL or, effectively NULL byte from a starting address.
+
+    :param Cpu cpu:
+    :param ConstraintSet constrs: Constraints for current `State`
+    :param int ptr: Address to start searching for a zero from
+    :return: Offset from `ptr` to first byte that is 0 or an `Expression` that must be zero
+    """
```
sschriner, line 70: Need to modify this since it was copied from _find_zero

manticore/native/models.py
```diff
@@ -140,3 +172,56 @@ def strlen(state, s):
...
+    ret = dst
+    c = cpu.read_int(src, 8)
+    # Copy until '\000' is reached or symbolic memory that can be '\000'
+    while (issymbolic(c) and not Z3Solver.instance().can_be_true(constrs, c == 0)) or c != 0:
+        cpu.write_int(dst, c, 8)
+        src += 1
+        dst += 1
+        c = cpu.read_int(src, 8)
+
+    # Even if the byte was symbolic and constrained to '\000' write a concrete '\000'
```
sschriner, line 205: This comment should be if the byte was symbolic and constrained to '\000' or '\000' write a concrete value and return

manticore/native/models.py
```diff
@@ -57,7 +59,37 @@ def _find_zero(cpu, constrs, ptr):
...
+    offset = 0
+    zeros = []
+    while True:
+        byt = cpu.read_int(ptr + offset, 8)
+
+        if issymbolic(byt):
+            if Z3Solver.instance().can_be_true(constrs, byt == 0):
+                zeros.append(offset)
+            if not Z3Solver.instance().can_be_true(constrs, byt != 0):
+                break
```
ekilmer, line 81: This is taking me a while to comprehend. Could you add an explanation to this, please

There are too many `can_be_true`, `not`, `==`, `!=` combinations and it's making my head spin, specifically the `not Z3Solver.instance().can_be_true(constrs, byt != 0)` expression 🙃

manticore/native/models.py
```diff
@@ -140,3 +172,56 @@ def strlen(state, s):
             ret = ITEBV(cpu.address_bit_size, byt == 0, offset, ret)
 
     return ret
+
+
+def strcpy(state: State, dst: Union[int, Expression], src: [int, Expression]) -> int:
+    """
+    strcpy symbolic model
+
+    Algorithm: Copy every byte from the src to dst until finding a byte that will is or must be '\000'
```
ekilmer, line 181:
```suggestion
    Algorithm: Copy every byte from the src to dst until finding a byte that will or must be '\000'
```

manticore/native/models.py
ekilmer: Ohhh this comment makes more sense.

manticore/native/models.py
```diff
@@ -140,3 +172,56 @@ def strlen(state, s):
...
+        c = cpu.read_int(src, 8)
+
+    # Even if the byte was symbolic and constrained to '\000' write a concrete '\000'
+    if (issymbolic(c) and not Z3Solver.instance().can_be_true(constrs, c != 0)) or c == 0:
+        cpu.write_int(dst, 0, 8)
+        return ret
+
+    zeros = _find_zeros(cpu, constrs, src)
+    null = zeros[-1]
+    # If the symmbolic byte was not constrained to '\000' write the appropriate symbolic bytes
```
ekilmer, line 212:
```suggestion
    # If the symbolic byte was not constrained to '\000' write the appropriate symbolic bytes
```

manticore/native/models.py
```diff
@@ -140,3 +172,56 @@ def strlen(state, s):
...
+    for offset in range(null, -1, -1):
+        src_val = cpu.read_int(src + offset, 8)
+        dst_val = cpu.read_int(dst + offset, 8)
+        if zeros[-1] == offset:
+            # Make sure last byte of the copy is always a concrete '\000'
+            true_val = ITEBV(cpu.address_bit_size, src_val == 0, 0, src_val)
+            zeros.pop()
+
+        # For every byte that could be null before the current byte add an if then else case to the bitvec tree to set the value to the src or dst byte accordingly
```
ekilmer, line 221: Nit: This comment is pretty long. `black` doesn't touch comments, but I think we should try to stick to max 100 line column length

manticore/native/models.py
sschriner: Yeah I will add a comment to this. Everything I added also made my head spin and I needed a break when I was done lol, but the general idea is 

If the given byte can be 0:
   then add it to the list of zeros
If the byte is constrained to 0 (there is not a case where it can be anything but 0):
   break

manticore/native/models.py
sschriner: I will fix this I forget to update it :)

manticore/native/models.py
sschriner: Will watch for this

manticore/native/models.py
```diff
@@ -140,3 +142,113 @@ def strlen(state, s):
...
+    Checks if a given byte read from memory is not NULL or cannot be NULL
+
+    :param byte: byte read from memory to be examined
+    :param constrs: state constraints
+    :return: whether a given byte is not NULL or cannot be NULL
+    """
+    if issymbolic(byte):
+        return not Z3Solver.instance().can_be_true(constrs, byte == 0)
+    else:
+        return byte != 0
```
ekilmer, line 172: Are these functions the inverse of each other?

Is it true that `not is_NULL()` is the same as `cant_be_NULL`?

ekilmer: Forgot about the `can_be_NULL`... Let me take a closer look

ekilmer: Actually, these do make sense to me now 👍

sschriner:
I plan to test this on some binaries before merging.

manticore/native/models.py
```diff
@@ -140,3 +142,123 @@ def strlen(state, s):
...
+def is_NULL(byte, constrs) -> bool:
+    """
+    Checks if a given byte read from memory is NULL or effectively NULL
+
+    :param byte: byte read from memory to be examined
+    :param constrs: state constraints
+    :return: whether a given byte is NULL or constrained to NULL
+    """
+    if issymbolic(byte):
+        return not Z3Solver.instance().can_be_true(constrs, byte != 0)
```
bradlarsen, line 156: Note, you might be able to rephrase this using `Z3Solver.instance().must_be_true`.  But I haven't used that API, so I can't comment further on performance implications, behavioral differences, etc :)

manticore/native/models.py
```diff
@@ -140,3 +142,123 @@ def strlen(state, s):
             ret = ITEBV(cpu.address_bit_size, byt == 0, offset, ret)
 
     return ret
+
+
+def is_NULL(byte, constrs) -> bool:
+    """
+    Checks if a given byte read from memory is NULL or effectively NULL
```
bradlarsen, line 149:
```suggestion
    Checks if a given byte read from memory is NULL.
    This supports both concrete & symbolic byte values.
```

manticore/native/models.py
```diff
@@ -140,3 +142,123 @@ def strlen(state, s):
...
+
+    # Copy until '\000' is reached or symbolic memory that can be '\000'
+    c = cpu.read_int(src, 8)
+    while cant_be_NULL(c, constrs):
+        cpu.write_int(dst, c, 8)
+        src += 1
+        dst += 1
+        c = cpu.read_int(src, 8)
+
+    # If the byte is symbolic and constrained to '\000' or is '\000' write concrete val and return
```
bradlarsen, line 222:
```suggestion
    # If the byte is symbolic and constrained to '\000', or is concrete and '\000', write concrete val and return
```

manticore/native/models.py
```diff
@@ -140,3 +142,123 @@ def strlen(state, s):
             ret = ITEBV(cpu.address_bit_size, byt == 0, offset, ret)
 
     return ret
+
+
+def is_NULL(byte, constrs) -> bool:
```
bradlarsen, line 147: You might consider renaming this to something like `is_definitely_NULL`, which might be a bit clearer as to what it's testing.

manticore/native/models.py
```diff
@@ -140,3 +142,123 @@ def strlen(state, s):
...
+    src: Union[int, BitVec],
+    dst: Union[int, BitVec],
+    offset: int,
+    src_val: Union[int, BitVec],
+) -> None:
+    """
+    Builds ITE tree for each symbolic dst byte
+    """
+    dst_val = cpu.read_int(dst + offset, 8)
+    for zero in reversed(zeros):
```
bradlarsen, line 261: Suggestion: if you used a `collections.deque` (or `typing.Deque[int]`) instead of a list, you could avoid having to reverse the list every time, as you could add or remove in amortized constant time from either end with a `deque`.

tests/native/test_models.py
```diff
@@ -198,3 +225,125 @@ def test_symbolic_mixed(self):
...
+    def _test_strcpy(self, string, dst_len=None):
+        # Create src and dsty strings
+        if dst_len is None:
+            dst_len = len(string)
+        cpu = self.state.cpu
+        s = self._push_string(string)
+        d = self._push_string_space(dst_len)
+        dst_vals = [None] * dst_len
+        for i in range(dst_len):
+            # Set each dst byte to a random char to simplify equal compairisons
```
bradlarsen, line 301:
```suggestion
            # Set each dst byte to a random char to simplify equal comparisons
```

tests/native/test_models.py
```diff
@@ -198,3 +225,125 @@ def test_symbolic_mixed(self):
...
+
+                offset += 1
+                src = cpu.read_int(s + offset, 8)
+                dst = cpu.read_int(d + offset, 8)
+
+            # Check last sym dst byte
+            dst = self._check_BitVecITE(dst, org_dest_val[offset])
+            self.assertEqual(dst.true_value, 0)
+
+    def _test_strcpy(self, string, dst_len=None):
```
bradlarsen, line 292: Can you add a brief comment (1 or 2 sentences) as to what this method is testing, at a high level?

manticore/native/models.py
sschriner: Taking a look at this. I'm also not super well versed with the API.

sschriner: Yes that cut the python test time in half! Good catch!

manticore/native/models.py
```diff
@@ -140,3 +143,124 @@ def strlen(state, s):
...
+    src_val: Union[int, BitVec],
+) -> None:
+    """
+    Builds ITE tree for each symbolic dst byte
+    """
+    dst_val = cpu.read_int(dst + offset, 8)
+    for zero in zeros:
+        c = cpu.read_int(src + zero, 8)
+        src_val = ITEBV(8, c != 0, src_val, dst_val)
+    cpu.write_int(dst + offset, src_val, 8)
```
ehennenfent, line 266: Since `_build_ITE` is a helper function, it makes me a little bit uncomfortable that it has side effects. Instead, can we return `src_val` and pass it to `cpu.write_int` in `strcpy`?

sschriner: Will do :)

manticore/native/models.py
```diff
@@ -140,3 +144,138 @@ def strlen(state, s):
...
+    :param state: current program state
+    :param dst: destination string address
+    :param src: source string address
+    :return: pointer to the dst
+    """
+    if issymbolic(src):
+        raise ConcretizeArgument(state.cpu, 1)
+
+    if issymbolic(dst):
+        raise ConcretizeArgument(state.cpu, 2)
```
ekilmer, line 211: These argument numbers seem to be flipped. Also, the argument numbering should start at `0`

sschriner: Done

manticore/native/models.py
```diff
@@ -113,14 +117,14 @@ def strcmp(state, s1, s2):
...
     """
     strlen symbolic model.
 
     Algorithm: Walks from end of string not including NULL building ITE tree when current byte is symbolic.
 
     :param State state: current program state
-    :param int s: Address of string
+    :param s: Address of string
     :return: Symbolic strlen result
     :rtype: Expression or int
```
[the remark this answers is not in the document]
sschriner: I removed the int and added the return type hint to both strcpy and strcmp. But I used Union[int, BitVec] instead of Union[int, Expression] because the linter complained about Expression when I tried to use that before

tests/native/test_models.py
```diff
@@ -198,3 +227,123 @@ def test_symbolic_mixed(self):
...
+
+        # Check the number of input states matches the number of regexes
+        self.assertEqual(len(stdins), len(expected))
+
+        # Assert that every regex has a matching input
+        for e in expected:
+            match = False
+            for s in stdins:
+                if re.fullmatch(e, s) == None:
+                    match = True
```
ekilmer, line 348: For efficiency sake, we should `break` after setting `match = True` to short-circuit after we find a match.

We could also remove the matched stdin string from `stdins` to cut down the complexity, but I'm not sure if the removal from the list would save us any time 🤷  Up to you if you want to do this part.

manticore/native/models.py
```diff
@@ -140,3 +144,96 @@ def strlen(state, s):
...
+    :param constrs: state constraints
+    :return: whether a given byte is not NULL or cannot be NULL
+    """
+    if issymbolic(byte):
+        return Z3Solver.instance().must_be_true(constrs, byte != 0)
+    else:
+        return byte != 0
+
+
+def can_be_NULL(byte, constrs) -> bool:
```
woodruffw, line 178: Nitpick, but maybe `cannot_be_NULL`? Unless `cant` is used idiomatically in the rest of Manticore.

tests/native/test_models.py
```diff
@@ -198,3 +227,123 @@ def test_symbolic_mixed(self):
...
+
+        m.run()
+        m.finalize()
+
+        # Approximate regexes for expected testcase output
+        # Example Match above each regex
+        # Manticore varies the hex output slightly per run
+        expected = [
+            # STDIN: b'\x00AAAAAAAAA'
+            "STDIN: b\\'\\\x00A.{8,32}\\'",
```
woodruffw, line 311: JFYI: You should be able to use raw strings here to make these regexps more intelligible:

```python
r"STDIN: b'\xffA\x00\xff\xff\xff\xff\xff\xff\xff'"
```

tests/native/test_models.py
sschriner: Yea I thought about doing the list removal but I didn't think that would lower the time complexity because removal is O(n). So the break is probably better 😄

manticore/native/models.py
```diff
@@ -140,3 +144,96 @@ def strlen(state, s):
...
+        offset = state.context["strcpy"]
+
+    # Copy until a src_byte is symbolic and constrained to '\000', or is concrete and '\000'
+    src_val = cpu.read_int(src + offset, 8)
+    while not is_definitely_NULL(src_val, constrs):
+        cpu.write_int(dst + offset, src_val, 8)
+
+        # If a byte can be NULL set the src_val for concretize and fork states
+        if can_be_NULL(src_val, constrs):
+            state.context["strcpy"] = offset
```
ehennenfent, line 228: As a note for myself, this is okay because state contexts aren't locked like plugin contexts are. It's one context per state, and each state's context isn't accessible from other threads, so we shouldn't get weird issues with states stepping on one another.

PR #1683, opened 2020-04-27
RFC [do not merge]: SerializeState interface changes

smoelius:
I am looking for feedback on these changes to the `SerializeState` interface.

The idea is that, from within a (`will`|`did`)`_execute_instruction_callback`, you can `preserve` a state. Then, later, you can use a state's `parent_id`s to recover the state's provenance in terms of those preserved states.

Note that I am trying, in part, to retain the current behavior whereby forked states are deleted.

### DIFF
--- a/manticore/core/manticore.py
+++ b/manticore/core/manticore.py
@@ -265,8 +265,9 @@ class ManticoreBase(Eventful):
                 "_killed",
                 "_ready_states",
                 "_terminated_states",
-                "_killed_states",
                 "_busy_states",
+                "_killed_states",
+                "_preserved_states",
                 "_shared_context",
             )
         ):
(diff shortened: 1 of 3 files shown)


ehennenfent:
Hey Sam, can you provide more context about the intended use case for this feature? It looks neat, but I'm not sure how it compares to the existing `SerializeState` exceptions.

smoelius:
Existing `SerializeState` exceptions save the state to a file. However, that state is not assigned a state id. So, for example, such a saved state cannot be returned to the ready queue, at least not without doing more work.

This PR was my attempt to work within the existing framework and assign state ids to saved states so that, e.g., they could be returned to the ready queue.

The addition of the `parent_id` was so that you could recover the saved state from whence your current state came, i.e., walk back up the "tree."

My impression was that others desired a similar feature, so I thought I was providing a public service. :P

Having said that, I have since I received some largely negative feedback offline. (Don't worry, no offense taken.) So I full well expect to close this. I am just going to leave it here for reference for a short while, if that's okay.

PR #1686, opened 2020-04-30
Improve concretize log messages

smoelius:
Before this change:
```
m.c.worker:INFO: [0] Debug Concretize()
```
After this change:
```
m.c.worker:INFO: [0] Debug 'Concretize: Concretize PC (Policy: ALL)'
```

### DIFF
--- a/manticore/core/worker.py
+++ b/manticore/core/worker.py
@@ -132,7 +132,7 @@ class Worker:
                         assert current_state is None
                     # Handling Forking and terminating exceptions
                     except Concretize as exc:
-                        logger.info("[%r] Debug %r", self.id, exc.message)
+                        logger.info("[%r] Performing %r", self.id, exc.message)
                         # The fork() method can decides which state to keep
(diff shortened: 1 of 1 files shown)


bradlarsen:
This is great!

manticore/core/worker.py
```diff
@@ -132,7 +132,7 @@ def run(self, *args):
                         assert current_state is None
                     # Handling Forking and terminating exceptions
                     except Concretize as exc:
-                        logger.info("[%r] Debug %r", self.id, exc)
+                        logger.info("[%r] Debug %r", self.id, exc.message)
```
ehennenfent, line 135:
```suggestion
                        logger.info("[%r] Performing %r", self.id, exc.message)
```
Appreciate you making this fix! While we're in here, I feel like we may as well rethink the "Debug" part of this message, since 1) it doesn't add any information, and 2) it's an info message, not a debug message.

PR #1688, opened 2020-04-30
Add set_verbosity to the documentation

langston-barrett:
Setting `verbosity` on `ManticoreBase` appears deprecated, but the preferred method is not yet documented.


ehennenfent:
Needed to update from `master` to fix an issue with our Ethereum tests, but now everything is looking good! Thanks for the PR!

PR #1689, opened 2020-04-30
Revert default gas limit

ehennenfent:
Reverts one of the [changes](https://github.com/trailofbits/manticore/pull/1652/files#diff-e4836efb704ed46c30e5cfb02496a419R725) from #1652 to fix the sporadically-failing `test_integer_overflow_multitx_onefunc_feasible` test in `ethereum_bench`

### DIFF
--- a/manticore/ethereum/manticore.py
+++ b/manticore/ethereum/manticore.py
@@ -728,7 +728,7 @@ class ManticoreEVM(ManticoreBase):
             if new_address not in all_addresses:
                 return new_address
 
-    def transaction(self, caller, address, value, data, gas=21000, price=1):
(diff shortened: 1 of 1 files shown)


ehennenfent:
@feliam Do you have thoughts on this? Is removing the gas limit a sane default behavior, or should we raise it up from what it was?

feliam:
> @feliam Do you have thoughts on this? Is removing the gas limit a sane default behavior, or should we raise it up from what it was?
I wonder Where and why a unittest is relying on the default. I move that to None to try detecting those. 
This is going to be a conflict soon enough. LGTM

PR #1693, opened 2020-05-04
Move truffle to its own CI type

feliam:
This simply moves the ethereum truffle related smoke test to its own CI entity.


ehennenfent:
Sorry I didn't get to this earlier, but we need to increment `after_n_builds` in codecov.yml

PR #1694, opened 2020-05-04
Try getting more coverage from truffle-test

feliam:


.github/workflows/ci.yml
```diff
@@ -157,7 +157,7 @@ jobs:
             mkdir truffle_tests
             cd truffle_tests
             truffle unbox metacoin
-            manticore . --contract MetaCoin --workspace output
+            coverage -m manticore . --contract MetaCoin --workspace output
```
ehennenfent, line 160:
```suggestion
            coverage run -m manticore . --contract MetaCoin --workspace output
```
I always forget that `coverage` isn't a drop-in replacement for `python`, but `coverage run` usually works that way.

codecov.yml
```diff
@@ -4,5 +4,5 @@ codecov:
   notify:
     # We have 8 test steps that produce coverage data.
```
ehennenfent, line 5:
```suggestion
    # We have 9 test steps that produce coverage data.
```
Comment

PR #1695, opened 2020-05-05
Reduce concrete hash log messages

smoelius:
Manticore produces many log messages of the following form, and many of them are redundant:
```
m.e.manticore:INFO: Found a concrete globalsha3 ... -> ...
```
This PR makes it so that only one log message is generated for each new find.


smoelius:
Sorry, please pause on this one.  It looks like there's a bug.

smoelius:
I think it's fixed now.  Sorry for my confusion (esp. @feliam).

PR #1699, opened 2020-05-07
Move states that selfdestruct to the ready queue

smoelius:


ehennenfent:
To note from the meeting this morning, previously we were treating self destructs the same way as reverts - ie, they're boring and we don't care about them. That's not actually the case, so this PR enables us to continue exploration afterwards.

Issue #1679, opened 2020-04-24
Flaw in `__get_related` core logic


smoelius:
There is a flaw in how [`__get_related`](https://github.com/trailofbits/manticore/blob/52f215e2f0cd61dee023379082113c491523f39d/manticore/core/smtlib/constraints.py#L108) works: when called on certain unsatisfiable sets, it can return a satisfiable one. The flaw arises when:
* `self` consists of a single constraint C
* C is the value of the `related_to` parameter
* C contains no variables
* C is unsatisfiable

Since C contains no variables, it is not considered "related to" itself and is thrown out by `__get_related`. Since C was the sole element of `self`, `__get_related` returns the empty set. Thus, `__get_related` was called on an unsatisfiable set, {C}, but it returned a satisfiable one, {}.

I suspect this is the reason for #1678.

ehennenfent:
It looks like the changes introduced in #1674 introduce some problems of their own. In #1677, I discovered that we end up with constraint sets where constraints from the related set have the same name (but a different size) than constraints from the original set, which could probably lead to incorrect results from Z3. The old `get_related` logic seemed to handle some edge cases that the default case doesn't, so we should try to fix this soon. 

Sam, do you have an existing example of the flaw above that you're able to share? If not, no worries - it should be easy enough to replicate from your description.

PR #1703, opened 2020-05-12
Add hashing to constraintSets

feliam:
Add __hash__ capabilities to ConstraintSet so we can use simple lru caching.
This improves commons case of re-doing the exact same query twice.
-baby steps towards query cache.

### DIFF
--- a/manticore/core/smtlib/constraints.py
+++ b/manticore/core/smtlib/constraints.py
@@ -17,10 +17,15 @@ from .expression import (
     Constant,
 )
 from .visitors import GetDeclarations, TranslatorSmtlib, get_variables, simplify, replace
+from ...utils import config
 import logging
 
 logger = logging.getLogger(__name__)
 
+consts = config.get_group("smt")
+consts.add(
+    "related_constraints", default=True, description="Try slicing the current path constraint to contain only related items"
+)
 
 class ConstraintException(SmtlibError):
     """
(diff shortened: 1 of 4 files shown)


manticore/core/smtlib/constraints.py
```diff
@@ -118,7 +126,8 @@ def __get_related(self, related_to=None):
         # satisfiable one, {}.
         #   In light of the above, the core __get_related logic is currently disabled.
         # if related_to is not None:
-        if False:
+        # feliam: This assumes the previous constraints are already SAT (normal SE forking)
+        if consts.related_constraints and related_to is not None:
```
ehennenfent, line 130: As you noted on slack, this will effectively revert #1674 without fixing #1679. Maybe we can set the default to `False` until that's taken care of?

manticore/core/smtlib/constraints.py
```diff
@@ -17,10 +17,16 @@
...
 )
 from .visitors import GetDeclarations, TranslatorSmtlib, get_variables, simplify, replace
+from ...utils import config
 import logging
 
 logger = logging.getLogger(__name__)
 
+consts = config.get_group("smt")
+consts.add(
+    "related_constraints", default=False, description="Try slicing the current path constraint to contain only related items"
```
feliam, line 27: Default false.

PR #1705, opened 2020-05-12
Experimental coverage changes

ehennenfent:
Tells coverage.py to respect the top-level coveragerc file, and makes CodeCov tolerate a small drop in overall coverage percentage to account for nondeterministic gas tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,7 +170,7 @@ jobs:
         run_tests_from_dir() {
             DIR=$1
             echo "Running only the tests from 'tests/$DIR' directory"
-            pytest --durations=100 --cov=manticore -n auto "tests/$DIR"
+            pytest --durations=100 --cov=manticore --cov-config=../../.coveragerc -n auto "tests/$DIR"
             coverage xml
         }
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -169,8 +171,9 @@ jobs:
 
         run_tests_from_dir() {
             DIR=$1
+            COVERAGE_RCFILE=$GITHUB_WORKSPACE/.coveragerc
             echo "Running only the tests from 'tests/$DIR' directory"
-            pytest --durations=100 --cov=manticore -n auto "tests/$DIR"
+            pytest --durations=100 --cov=manticore --cov-config=$GITHUB_WORKSPACE/.coveragerc -n auto "tests/$DIR"
```
bradlarsen, line 176: Is the `COVERAGE_RCFILE` variable actually needed?  Its value is passed explicitly as an argument to pytest.

ehennenfent: I think it should be used any time that the bare `coverage` command is evaluated. So, the pytest line should be fine, but the `coverage xml` further down may need it. I say "may" because afaik, our xml file generation doesn't currently rely on any values set in the rcfile, but it certainly could if we decided to update it in the future.

PR #1706, opened 2020-05-13
Fix constant folding & constraint set slicing

feliam:
* Improved constant folding simplifier ( several bugfixes related with signed ints vs. bitvectors)
* This PR contains a single testcase that reproduces https://github.com/trailofbits/manticore/issues/1679 And its "fix"
* ConstraintSet slicing is optional and disabled by default.


manticore/core/smtlib/constraints.py
```diff
@@ -146,7 +146,7 @@ def __get_related(self, related_to=None):
                             break
 
                     variables = get_variables(constraint)
-                    if related_variables & variables:
+                    if related_variables & variables or not variables:
```
feliam, line 149: With this added "this" test passes with both get_related enabled and disabled

ehennenfent: As long as both are passing, let's set the default back to `True`

feliam: Well I'm not 100% This single testcase representes all possible fails explained in issue https://github.com/trailofbits/manticore/issues/1679

smoelius: Devils advocate, here...  What if `related_to` is `A` and `constraint` is something like `B && !B`, i.e., `constraint` is unsatisfiable and has variables, but they do not overlap with those of `related_to`?

tests/ethereum/test_regressions.py
```diff
@@ -353,6 +353,40 @@ def test_mulmod(self):
         # 0x8000000000000000000000000000000000000000000000000000000082000011
         self.assertEqual(Z3Solver.instance().get_all_values(constraints, result), [2423129])
 
+    def test_related_to(self):
+        import gzip
+        from manticore import config
+        from manticore.core.smtlib.visitors import translate_to_smtlib
+        from manticore.core.smtlib import ConstraintSet, Z3Solver, Operators, BitVecConstant
+        import pickle, sys
+        filename = os.path.abspath(os.path.join(DIRPATH, "data", "ErrRelated.pkl.gz"))
```
smoelius, line 362: What's in this file?

smoelius: Also, should this test be in `ethereum` directory?

manticore/core/smtlib/constraints.py
feliam: Yes get_related should work only on previously feasible path contraints.
If you have feasible `A & B & C` and want to check if `B \/ C` can be True then you do not need to bother the solver with the A.

smoelius: > get_related should work only on previously feasible path contraints.
In other words, you consider it a precondition that the constraint set on which `__get_related` is called is feasible?

feliam: Yes. I know it is not enforced.
We should make it a pre-condition or attempt to solve the full constraint set and trust that some cache magic prevents it to do the solving over the whole thing again and again.

smoelius: > Yes. I know it is not enforced.
> We should make it a pre-condition or attempt to solve the full constraint set and trust that some cache magic prevents it to do the solving over the whole thing again and again.

I don't know how reasonable this approach is. `__get_related` is called by `to_string`, and `to_string` is something that one might reasonably call to debug a bad constraint set.

Moreover, `to_string` is called by `can_be_true`:
https://github.com/trailofbits/manticore/blob/2c90ee07c9e691304db97f573bff61ee220ffbb6/manticore/core/smtlib/solver.py#L433
So there's a little bit of a chicken-and-egg problem here.

If it were up to me, I would not insist that `__get_related` be called on only satisfiable constraint sets.

feliam: Many constraints do not overlap in terms of the memory they reference. Constraint independence divides constraint sets into disjoint independent subsets based on the symbolic variables and memory cells they reference. By explicitly tracking these subsets, it can frequently eliminate irrelevant constraints prior to sending it to the solver.
For example, given the constraint set {i < j, j < 20, k > 0}, a query of whether i = 20 just requires the first two constraints.  This was first explained in [EXE] https://web.stanford.edu/~engler/exe-ccs-06.pdf

We can move the optimizations/transformations that are controlled via kwargs in the `to_string()` to a different `ConstraintSet` method. Like `slice_related_to(..)` and `make_bindings` so it is done more explicit.

If the problem is to print a ConstraintSet for debugging you can just just do not slice the ConstraintSet.

``` print (str(cs)) ```

There is also a forgotten `pretty_print` for expressions.

Comment (ehennenfent) [reply manticore/core/smtlib/constraints.py:149]: Fwiw, I use the `pretty_print` all the time while debugging

File: tests/ethereum/test_regressions.py (hunk 2 above)
Comment (ehennenfent) [reply tests/ethereum/test_regressions.py:362]: My guess is that it's a state that exhibits the `get_related` bug, but pickled and gzipped so that we don't need to construct it from the source code.

Comment (feliam) [reply tests/ethereum/test_regressions.py:362]: Yes. Not the state but the ConstraintSet that showed the problem. This is not really ethereum specific.

Comment (smoelius) [reply tests/ethereum/test_regressions.py:362]: Is it expensive to construct from source? A binary isn't easy to peruse over the web, for example.

Comment (feliam) [reply tests/ethereum/test_regressions.py:362]: I saved the constraint and the constraintset that found to have trouble with old get_related.
I re build the constraint but the constraintset was big ish so I left it all binary and obscure.
Its a hack. I'll be much better to produce the testcases ourselves.
As it is still disabled by default I could remove this unintelligible test.

File: tests/ethereum/test_regressions.py [hunk 3]
```diff
@@ -353,6 +353,40 @@ def test_mulmod(self):
...
+
+        constraints, constraint = pickle.loads(gzip.open(filename,'rb').read())
+
+        consts = config.get_group("smt")
```
feliam, line 366: (There is a contextmanager in config.py to test with different config vars and then restore them to original values)

manticore/core/smtlib/constraints.py
feliam: Moved the disabled get_related as a special method of a ConstraintSet to slice it when the user wants it.
Also added "some" tests for it.

manticore/core/smtlib/constraints.py
```diff
@@ -115,7 +116,7 @@ def _get_sid(self) -> int:
         self._sid += 1
         return self._sid
 
-    def __get_related(self, related_to=None):
+    def related_to(self, related_to: Optional[Expression] = None):
```
ekilmer, line 119: Does adding this still work?
```suggestion
    def related_to(self, related_to: Optional[Expression] = None) -> ConstraintSet:
```

manticore/core/smtlib/constraints.py
```diff
@@ -127,49 +128,57 @@ def __get_related(self, related_to=None):
...
+
+        # related_variables, related_constraints
+        cs = ConstraintSet()
+        for var in related_variables:
+            cs._declare(var)
+        for constraint in related_constraints:
+            cs.add(constraint)
+        return cs
+
+    def to_string(self, replace_constants=True):
```
ekilmer, line 176:
```suggestion
    def to_string(self, replace_constants: bool =True) -> str:
```

manticore/core/smtlib/solver.py
```diff
@@ -516,8 +525,8 @@ def optimize(self, constraints: ConstraintSet, x: BitVec, goal: str, max_iter=10
             X = temp_cs.new_bitvec(x.size)
             temp_cs.add(X == x)
             aux = temp_cs.new_bitvec(X.size, name="optimized_")
-            self._reset(temp_cs.to_string(related_to=X))
-            self._send(aux.declaration)
+            self._reset(temp_cs.to_string())
+            # self._send(aux.declaration)
```
ekilmer, line 529: Remove?
```suggestion
```

manticore/core/smtlib/solver.py
```diff
@@ -490,11 +490,20 @@ def get_all_values(self, constraints, expression, maxcnt=None, silent=False):
...
-                        return result
+                        return (
+                            result
+                            / home
+                            / felipe
+                            / Projects
+                            / manticore
+                            / tests
+                            / other
+                            / test_smtlibv2.py
```
ekilmer, line 501: This seems oddly specific.

manticore/core/smtlib/visitors.py
```diff
@@ -301,13 +300,31 @@ def __init__(self, **kw):
...
+        BitVecUnsignedDiv: lambda x, y: 0
+        if (y & UNSIGN_MASK) == 0
+        else (x & UNSIGN_MASK) // (y & UNSIGN_MASK),
         UnsignedLessThan: lambda x, y: (x & UNSIGN_MASK) < (y & UNSIGN_MASK),
         UnsignedLessOrEqual: lambda x, y: (x & UNSIGN_MASK) <= (y & UNSIGN_MASK),
         UnsignedGreaterThan: lambda x, y: (x & UNSIGN_MASK) > (y & UNSIGN_MASK),
         UnsignedGreaterOrEqual: lambda x, y: (x & UNSIGN_MASK) >= (y & UNSIGN_MASK),
     }
 
+    def visit_BitVecDiv(self, expression, *operands):
```
ekilmer, line 312:
```suggestion
    def visit_BitVecDiv(self, expression, *operands) -> BitVecConstant:
```

tests/other/test_smtlibv2.py
```diff
@@ -11,15 +12,55 @@
...
 
+DIRPATH = os.path.dirname(__file__)
+
+
+class RegressionTest(unittest.TestCase):
+    def test_related_to(self):
+        import gzip
+        import pickle, sys
+
+        filename = os.path.abspath(os.path.join(DIRPATH, "data", "ErrRelated.pkl.gz"))
```
ekilmer, line 34: What's actually in the `ErrRelated.pkl.gz` file? How do I create it?

tests/other/test_smtlibv2.py
```diff
@@ -986,6 +1026,33 @@ def test_check_solver_undefined(self):
...
+        self.assertNotIn("AA", cs.related_to(bb1 == bb2).to_string())
+        self.assertNotIn("AA", cs.related_to(bb1 == False).to_string())
+
+        # Nothing is related to tautologies?
+        self.assertEqual("", cs.related_to(simplify(bb1 == bb1)).to_string())
+
+        # But if the tautollogy can not get simplified we have to ask the solver
+        # and send in all the other stuff
+        self.assertNotIn("AA", cs.related_to(bb1 == bb1).to_string())
+
```
ekilmer, line 1055: I think it'd be nice to check that we _do_ find what we expect in our `related_to` call.

One example
```suggestion

        related_bb1 = cs.related_to(bb1).to_string()
        self.assertIn("BB1", related_bb1)
        self.assertIn("BB2", related_bb1)
```
Since
```python
related == '(declare-fun BB2 () Bool)\n(declare-fun BB1 () Bool)\n(assert (or BB1 BB2))\n'
```

Are there any more interesting cases?

feliam: I did not want to lock in how it is generated and more in the fact that the unrelated vars are sliced out. Idk.. stuff may be optimized? no in this case.
Actually I do not need to make a string out of it to check if a var is declared.... :thinking:

manticore/core/smtlib/constraints.py
```diff
@@ -116,7 +116,7 @@ def _get_sid(self) -> int:
         self._sid += 1
         return self._sid
 
-    def related_to(self, related_to: Optional[Expression] = None):
+    def related_to(self, related_to: Optional[Expression] = None) -> ConstraintSet:
```
ekilmer, line 119: My bad. Forgot this was inside the class
```suggestion
    def related_to(self, related_to: Optional[Expression] = None) -> "ConstraintSet":
```

manticore/core/smtlib/visitors.py
feliam: Should not it be Optional[BitVecConstant] ?

ekilmer: Ah, yes, you're right

tests/other/test_smtlibv2.py
ekilmer: Hmmm. That's a fair point. I'm fine with it now then.

tests/other/test_smtlibv2.py
ekilmer: Good to know. I think we should put something like that in a comment for this test, just to let someone else know what it means.

manticore/core/smtlib/solver.py
```diff
@@ -687,7 +690,11 @@ def __autoconfig(self):
             support_reset = False
```
ehennenfent, line 690: Is it correct for `support_reset` to be False for all versions of Z3?

PR #1707, opened 2020-05-14
Add a warning when selected no methods for inclusion

feliam:


manticore/ethereum/plugins.py
```diff
@@ -78,8 +80,10 @@ def will_open_transaction_callback(self, state, tx):
                 selected_functions.append(md.fallback_function_selector)
 
             if self._include:
+                if not selected_functions:
+                    logger.warning("No functions selected, adding False to path constraint.")
                 # constrain the input so it can take only the interesting values
-                constraint = reduce(Operators.OR, (tx.data[:4] == x for x in selected_functions))
+                constraint = reduce(Operators.OR, (tx.data[:4] == x for x in selected_functions), False)
```
ehennenfent, line 86: This _unconditionally_ adds `False` to the path constraint, right? Not just if there aren't any functions selected, as the log message would seem to imply. Is the problem that `constraint` is empty otherwise?

ehennenfent: Per dev meeting, `False` is needed here so that the `reduce` isn't empty when `selected_functions` is empty.

PR #1708, opened 2020-05-14
Fix plugin enable/disable magic

feliam:


manticore/core/plugin.py
```diff
@@ -10,20 +10,32 @@
...
+
+
+class Plugin(metaclass=DecorateAllMeta):
+    __slots__ = ("manticore", "_enabled_key", "_plugin_context_name")
 
     def __init__(self):
         self.manticore = None
-        self._enabled_key = f"{str(type(self))}_enabled_{hash(self)}"
-        self._plugin_context_name = f"{str(type(self))}_context_{hash(self)}"
-        self.__decorate_callbacks()
```
feliam, line 19: The `__dict__` is not updated with the methods from the subclases here

manticore/core/plugin.py
```diff
@@ -10,20 +10,32 @@
 logger = logging.getLogger(__name__)
 
 
-class Plugin:
+class DecorateAllMeta(type):
+    @staticmethod
+    def _if_enabled(f):
```
ehennenfent, line 15: You've added `_if_enabled` here, but not removed it from the subclass. While I think it's technically fine, the code duplication will get confusing.

manticore/ethereum/plugins.py
```diff
@@ -84,9 +84,10 @@ def will_open_transaction_callback(self, state, tx):
             else:
                 # Avoid all not selected hashes
                 for func_hsh in md.function_selectors:
+                    constraint = True
                     if func_hsh in selected_functions:
-                        constraint = tx.data[:4] != func_hsh
-                        state.constrain(constraint)
+                        constraint = Operators.AND(tx.data[:4] != func_hsh, constraint)
```
ehennenfent, line 89: Why the `AND True` here? Does this affect the plugin enable/disable operation?

feliam: No, it does not affect the main PR goal. 
This builds a single constraint and add that to the constraintSet rather than add one for each function id this matches.

ehennenfent: I could be misreading this, but shouldn't `constraint=True` be outside the loop then?

feliam: YES!

PR #1710, opened 2020-05-15
Snapshots & is_main

feliam:
Duplicate and save ready state to do expensive temporary search and rollback
Fix Manticore.is_main logic


manticore/core/manticore.py
```diff
@@ -357,6 +357,53 @@ def __init__(self, initial_state, workspace_url=None, outputspace_url=None, **kw
         # Workers will use manticore __dict__ So lets spawn them last
         self._workers = [self._worker_type(id=i, manticore=self) for i in range(consts.procs)]
         self._is_main = True
```
feliam, line 359: Is the is_main logic correct?
`
take_snapshot should only work from main. If is_main works maybe we should use that instead of just `at_not_running`

feliam: It WAS not working. Fixed it and added tests for it. (is_main)

manticore/core/manticore.py
```diff
@@ -355,8 +369,65 @@ def __init__(self, initial_state, workspace_url=None, outputspace_url=None, **kw
...
 
         # Workers will use manticore __dict__ So lets spawn them last
+        self._is_main = None
         self._workers = [self._worker_type(id=i, manticore=self) for i in range(consts.procs)]
-        self._is_main = True
+        self._snapshot = None
+        self._main_id = (consts.mprocessing, os.getpid(), threading.current_thread().ident)
+
+    def is_main(self):
+        if self._main_id[0].value == 'single' and self.is_running():
```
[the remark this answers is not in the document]
feliam: Yes. I guess the problem here is the lacking docstring. 
What is `is_main`?

tests/ethereum/test_general.py
```diff
@@ -1755,6 +1755,83 @@ def test_FilterFunctions_fallback_function_matching(self):
...
+        contract C {
+            constructor() public payable {}
+            function f1(uint a) public payable {}
+            function f2(uint a) public payable {}
+        }
+        """
+        class X(Plugin):
+            def will_evm_execute_instruction_callback(self, state, instruction, args):
+                is_main = self.manticore.is_main()
+                is_running = self.manticore.is_running
```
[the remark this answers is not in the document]
feliam: If a left over from debugging, it seems. Removing..

manticore/core/manticore.py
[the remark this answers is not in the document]
feliam: Okay. Now if you want a manticore method to be only callable from main() script and not from a callback/plugin you need to add 2 decorators.
```
@at_not_running
@not_from_main
def your_method(..)
```

manticore/core/manticore.py
```diff
@@ -355,8 +369,60 @@ def __init__(self, initial_state, workspace_url=None, outputspace_url=None, **kw
         self._put_state(initial_state)
 
         # Workers will use manticore __dict__ So lets spawn them last
+        self._is_main = None
```
feliam, line 372: Where is this used?

ehennenfent: Nowhere that I'm aware of, or can find with a quick grepping

feliam:
> Thanks for the PR! CodeCov is being a little bit wonky, but I think it's because one of the uploads failed. Generally looking good, just a few minor nits.
> I wonder if in the future, we'll end up converting `_snapshot` into a stack so that one can push and pop Manticore execution states. We'll have to wait and see what people use it for!

Yes! I'm trying to keep it simple. A generalization of the state lists and transitions could occur if this is well received.  A very short stack for now.

manticore/core/manticore.py
```diff
@@ -356,7 +370,60 @@ def __init__(self, initial_state, workspace_url=None, outputspace_url=None, **kw
 
         # Workers will use manticore __dict__ So lets spawn them last
         self._workers = [self._worker_type(id=i, manticore=self) for i in range(consts.procs)]
-        self._is_main = True
+        self._snapshot = None
+        self._main_id = os.getpid(), threading.current_thread().ident
+
+    def is_main(self):
```
ehennenfent, line 376: Not a big deal, but this could be a `@property` so as to preserve the old API (not that anyone uses it...)

manticore/core/manticore.py
```diff
@@ -356,7 +370,60 @@ def __init__(self, initial_state, workspace_url=None, outputspace_url=None, **kw
...
+        ''' True if called from the main process/script
```
ehennenfent, line 377: Pretty sure `black` should be flagging these. Looking at the Lint results it seems that something might be broken. I'll see if I can fix it in another PR.

manticore/core/manticore.py
```diff
@@ -356,7 +370,60 @@ def __init__(self, initial_state, workspace_url=None, outputspace_url=None, **kw
...
+        Note: in "single" mode this is _most likelly_ True '''
```
ehennenfent, line 378:
```suggestion
        Note: in "single" mode this is _most likely_ True '''
```

tests/ethereum/test_general.py
```diff
@@ -1793,6 +1793,83 @@ def test_FilterFunctions_fallback_function_matching(self):
...
+        }
+        """
+
+        m: ManticoreEVM = ManticoreEVM()
+
+        creator_account = m.create_account(balance=10000000000)
+        contract_account = m.solidity_create_contract(source_code, owner=creator_account, balance=0)
+
+        #Can not go to unexistant snapshot
+        self.assertRaises(Exception, m.goto_snapshot)
```
ehennenfent, line 1812: We can make this `ManticoreError` to match the behavior in `only_from_main_script`

manticore/core/manticore.py
feliam: I'M going towards the rule: if the thing is an  "is_something" then it is a method and not a property.
I fear someone use something like
```
 if x.is_main:
     do_stuff()
```

When it is actually a method. Thoughts?

manticore/core/manticore.py
feliam: blkn

manticore/core/manticore.py
ehennenfent: I couldn't find great documentation on when it's "pythonic" to use properties over methods, other than [one SO commenter](https://stackoverflow.com/questions/1554546/when-and-how-to-use-the-builtin-function-property-in-python) who argues that you should use properties when you're committed to not having any side-effects in a method. Personally, `is_main` feels to me like it's tautologically describing a "property" of the Manticore object, and so it makes sense to use properties here. That said, I don't feel too strongly about it, so don't feel compelled to change it if you disagree.

PR #1714, opened 2020-05-21
Rollback to support yices again

feliam:
Adds back CVC4 and Yices. (Yices is fast!)
Mild refactor in solver.py.


manticore/core/smtlib/solver.py
```diff
@@ -45,9 +47,14 @@
     "optimize", default=True, description="Use smtlib command optimize to find min/max if available"
 )
 
+consts.add(
+    "solver", default="cvc4", description="Choose default smtlib2 solver (z3, yices, cvc4, race)"
```
smoelius, line 51: Are we sure cvc4 should be the default given that we have been using z3 for so long?

feliam: Yes that should not be cvc4 by default. 
From a quick look yices looks like it should be the default. :) 

* needs to make this an Enum

ekilmer:
The README should be updated to tell people they need to install Yices (and how), if there won't be any fallback to Z3.

I think a fallback to Z3 would be nice.

tests/other/test_smtlibv2.py
```diff
@@ -21,6 +22,42 @@
...
+
+    def test_check_solver_undefined(self):
+        self.solver._received_version = '(:version "78ed71b8de7d")'
+        self.assertTrue(
+
+            self.solver._solver_version()
+            == Version(major=float("inf"), minor=float("inf"), patch=float("inf"))
+        )
+        self.assertTrue(self.solver._solver_version() > Version(major=4, minor=4, patch=1))
+'''
```
ekilmer, line 59: Why is this commented? Do these tests not work now?

.github/workflows/ci.yml
```diff
@@ -49,6 +49,13 @@ jobs:
       env:
         TEST_TYPE: ${{ matrix.type }}
       run: |
+        #install cvc4
+        sudo wget -O /usr/bin/cvc4 https://github.com/CVC4/CVC4/releases/download/1.7/cvc4-1.7-x86_64-linux-opt
+        sudo chmod +x /usr/bin/cvc4
+        #install yices
+        sudo add-apt-repository ppa:sri-csl/formal-methods
+        sudo apt-get update
+        sudo apt-get install yices2
```
ekilmer, line 58: We should add this info to the docs/wiki also

manticore/core/smtlib/solver.py
```diff
@@ -352,34 +314,39 @@ def _is_sat(self) -> bool:
...
-
         if status == "unknown":
             raise SolverUnknown(status)
-
-        is_sat = status == "sat"
-        if not is_sat:
-            check_memory_usage()
-            check_disk_usage()
-        return is_sat
+        return status == "sat"
```
ekilmer, line 340: I thought the `check_memory` and disk usage was to provide more helpful messages? Do we want to get rid of that? @ehennenfent

manticore/core/smtlib/solver.py
```diff
@@ -352,34 +314,39 @@ def _is_sat(self) -> bool:
...
         """Auxiliary method to send an assert"""
-        assert isinstance(expression, Bool)
         smtlib = translate_to_smtlib(expression)
-        self._send("(assert %s)" % smtlib)
-
-    def _getvalue(self, expression):
+        self._smtlib.send(f"(assert {smtlib})")
+
+    # Union[Variable, int, bool, bytes]
+    def __getvalue_bv(self, expression_str):
```
ekilmer, line 348: Should that type hint be moved into the function argument?
```suggestion
    def __getvalue_bv(self, expression_str: Union[Variable, int, bool, bytes]) -> int:
```

manticore/core/smtlib/solver.py
```diff
@@ -388,39 +355,35 @@ def _getvalue(self, expression):
...
                 expression_str = translate_to_smtlib(c)
-                self._send("(get-value (%s))" % expression_str)
-                response = self._recv()
-                result.append(int("0x{:s}".format(response.split(expression_str)[1][3:-2]), 16))
+                # self._smtlib.send(f"(get-value ({expression_str}))")
+                # response = self._smtlib.recv().split(expression_str)[1].strip(")")
+                # pattern, base = self._get_value_fmt
+                # m = pattern.match(response)
+                # expr, value = m.group("expr"), m.group("value")
+                # result.append(int(value, base))
```
ekilmer, line 379: Remove?
```suggestion
```

manticore/core/smtlib/solver.py
```diff
@@ -388,39 +355,35 @@ def _getvalue(self, expression):
...
-                m = pattern.match(ret)
-                expr, value = m.group("expr"), m.group("value")
-                return int(value, base)
+            if isinstance(expression, BoolVariable):
+                return self.__getvalue_bool(expression.name)
+            elif isinstance(expression, BitVecVariable):
+                return self.__getvalue_bv(expression.name)
 
-        raise NotImplementedError("_getvalue only implemented for Bool and BitVec")
+        raise NotImplementedError("_getvalue only implemented for Bool, BitVec and Array")
```
ekilmer, line 388: Would be helpful to show the type of what was passed.
```suggestion
        raise NotImplementedError(f"_getvalue only implemented for Bool, BitVec and Array. Got {type(expression)}")
```

manticore/core/smtlib/solver.py
```diff
@@ -432,20 +395,95 @@ def can_be_true(self, constraints: ConstraintSet, expression: Union[bool, Bool]
...
+                if time.time() - start > consts.timeout:
+                    raise SolverError("Timeout")
+            if last_value is not None:
+                return last_value
+            raise SolverError("Optimizing error, unsat or unknown core")
+
     @lru_cache(maxsize=32)
-    def get_all_values(self, constraints, expression, maxcnt=None, silent=False):
+    def get_all_values(
+        self, constraints: ConstraintSet, expression, maxcnt: bool = None, silent: bool = False
```
ekilmer, line 493: `maxcnt` should be `Optional[bool]`. You'll have to import `Optional` at the top though: `from typing import Optional`
```diff
-        self, constraints: ConstraintSet, expression, maxcnt: bool = None, silent: bool = False
+        self, constraints: ConstraintSet, expression, maxcnt: Optional[bool] = None, silent: bool = False
```

manticore/core/smtlib/solver.py
```diff
@@ -509,101 +547,41 @@ def optimize(self, constraints: ConstraintSet, x: BitVec, goal: str, max_iter=10
...
-                if time.time() - start > consts.timeout:
-                    raise SolverError("Timeout")
-            if last_value is not None:
-                return last_value
-            raise SolverError("Optimizing error, unsat or unknown core")
+                            raise SolverError("Could not match MinMax objective value regex")
+                elif _status == "sat":
+                    ret = self._smtlib.recv()
+                    if not (ret.startswith("(") and ret.endswith(")")):
+                        raise SolverError("bad output on max, z3 may have been killed")
```
ekilmer, line 584:
```suggestion
                        raise SolverError("bad output on max, solver may have been killed")
```

manticore/core/smtlib/solver.py
```diff
@@ -637,13 +615,13 @@ def get_value(self, constraints: ConstraintSet, *expressions):
...
-                        pattern, base = self._get_value_fmt
-                        m = pattern.match(ret)
-                        expr, value = m.group("expr"), m.group("value")
-                        result.append(int(value, base))
+                        # self._smtlib.send("(get-value (%s))" % var[i].name)
+                        # ret = self._smtlib.recv()
+                        # pattern, base = self._get_value_fmt
+                        # m = pattern.match(ret)
+                        # expr, value = m.group("expr"), m.group("value")
+                        # result.append(int(value, base))
```
ekilmer, line 634: Delete
```suggestion
```

manticore/core/smtlib/solver.py
```diff
@@ -658,22 +636,142 @@ def get_value(self, constraints: ConstraintSet, *expressions):
...
+
+class CVC4Solver(SMTLIBSolver):
+    def __init__(self):
+        init = ["(set-logic QF_AUFBV)", "(set-option :produce-models true)"]
+        command = f"cvc4 --lang=smt2 --incremental"
+        super().__init__(command=command, value_fmt=10, init=init)
+
+
+class ddRaceSolver(SMTLIBSolver):
+    def __init__(self, *solvers):
```
ekilmer, line 757: I think this class deserves a brief explanation about what is going to happen

manticore/core/smtlib/solver.py
```diff
@@ -773,5 +784,26 @@ def thread(solver):
...
+    @classmethod
+    def instance(cls):
+        if consts.solver == consts.solver.auto:
+            if cls.choice is None:
+                if os.path.exists(consts.yices_bin):
+                    cls.choice = consts.solver.yices
+                elif os.path.exists(consts.z3_bin):
+                    cls.choice = consts.solver.z3
+                elif os.path.exists(consts.cvc4_bin):
+                    cls.choice = consts.solver.cvc4
```
ekilmer, line 799: I think what you're trying to do here is a bit more complicated than this. We want to use [`shutil.which`](https://docs.python.org/3.6/library/shutil.html#shutil.which) if it will just be the name of the executable. If it's a full path, then `os.path.exists` is correct.

manticore/core/smtlib/solver.py
```diff
@@ -791,19 +792,18 @@ class SelectedSolver:
     def instance(cls):
         if consts.solver == consts.solver.auto:
             if cls.choice is None:
-                if os.path.exists(consts.yices_bin):
+                if shutil.which(consts.yices_bin):
```
ekilmer, line 795: I think we should still support passing an exact path
```suggestion
                if os.path.exists(consts.yices_bin) or shutil.which(consts.yices_bin):
```

ekilmer: Nevermind!!

tests/other/test_smtlibv2.py
feliam: I do not know how to Mock subprocess.check_output and I do not want to add a member variable to the solver just to implement the tests like it was there.

manticore/core/smtlib/solver.py
feliam: We could guard it with self.debug!
This is global check right? not necessarily related to _this_ solver ?
In linux `dmesg` will show if OOM happened globally too.

tests/other/test_smtlibv2.py
ekilmer: Does the `setup()` function not always use a Z3 solver?
```
    def setUp(self):
        self.solver = Z3Solver.instance()
```

manticore/core/smtlib/solver.py
feliam: HAHA. Yes.

tests/other/test_smtlibv2.py
feliam: If this was not commented out this would use Z3.

.github/workflows/ci.yml
feliam: Added some to the readme. May need review.

manticore/core/smtlib/solver.py
```diff
@@ -39,15 +52,26 @@
...
+
+
 consts.add("defaultunsat", default=True, description="Consider solver timeouts as unsat core")
 consts.add(
     "optimize", default=True, description="Use smtlib command optimize to find min/max if available"
 )
 
+consts.add(
+    "solver",
+    default=SolverType.yices,
```
ehennenfent, line 67: The README says that Z3 is still the default. Which do we want to use? I'm in favor of leaving Z3 as the default since we have a long history of testing against it and it's very easy to install.

manticore/core/smtlib/solver.py
```diff
@@ -143,207 +167,162 @@ def minmax(self, constraints, x, iters=10000):
...
-                self._stop_proc()
-                self._start_proc()
-        if constraints is not None:
-            self._send(constraints)
-
-    def _send(self, cmd: str) -> None:
+            self._proc.stdin.close()
+            self._proc.stdout.close()
+            # Kill the process
+            self._proc.kill()
```
ehennenfent, line 207: As long as we're going to wait for termination, shouldn't we send SIGTERM first instead of SIGKILL? From the comment, it looks like that might have been removed. Any reason why?

manticore/core/smtlib/solver.py
```diff
@@ -143,207 +167,162 @@ def minmax(self, constraints, x, iters=10000):
...
-            if self._proc.stdin:
-                self._proc.stdin.write(f"{cmd}\n")
-            else:
-                raise SolverError("Could not write to stdin: file descriptor is None")
-        except IOError as e:
-            raise SolverError(str(e))
-
-    def _recv(self) -> str:
-        """Reads the response from the solver"""
+        assert self._proc
```
ehennenfent, line 226: Since these now raise `AssertionError` instead of `SolverError`, we should check our error handling elsewhere to make sure that this won't break uncovered code. At a glance, it looks like the linux module might rely on it.

manticore/core/smtlib/solver.py
ehennenfent: Strictly speaking those were intended to help diagnose the sporadically-failing SMT failures. We've talked about actually removing them so that we wouldn't need to ship psutil as a dependency. I'm fine with removing them from smtlib here, although if I can find a way to fix them to not need psutil I may leave them in as pre-run checks.

manticore/core/smtlib/solver.py
```diff
@@ -143,207 +167,162 @@ def minmax(self, constraints, x, iters=10000):
...
             buf, l, r = self.__readline_and_count()
             bufl.append(buf)
             left += l
             right += r
-            if "(error" in bufl[0]:
-                raise SolverException(f"Error in smtlib: {bufl[0]}")
 
         buf = "".join(bufl).strip()
-        if "(error" in bufl[0]:
-            raise SolverException(f"Error in smtlib: {bufl[0]}")
```
ehennenfent, line 335: Are we expecting this error to not be able to occur anymore? It doesn't look like we check for either this case or the similar case on line 330. 

Also: why do we need both `SolverError` and `SolverException`?

manticore/core/smtlib/solver.py
```diff
@@ -490,7 +538,7 @@ def get_all_values(self, constraints, expression, maxcnt=None, silent=False):
                     raise SolverError("Timeout")
             return result
 
-    def optimize(self, constraints: ConstraintSet, x: BitVec, goal: str, max_iter=10000):
+    def _optimize_fancy(self, constraints: ConstraintSet, x: BitVec, goal: str, max_iter=10000):
```
ehennenfent, line 541: Do we need to move this to a private method? It looks like it's only used internally, but there could be external scripts that depend on `optimize`. If it's solver-specific, perhaps we could make `optimize` an abstract method and implement it on the subclasses?

manticore/core/smtlib/solver.py
feliam: I set it back to Z3.  (I hope it passes tests. :joy: )

manticore/core/smtlib/solver.py
feliam: You mean that the _proc.wait() is unnecessary after a _proc.kill()?
I do not trust that the .kill() translates to a specific signal.
I want to destroy the solver companion the easiest, fastest way.

manticore/core/smtlib/solver.py
feliam: SolverError and SolverException? I do not think we need both. Let's choose one. ??

Ah!!  Good catch. Tried to ignore it in my refactor it seems. The problem there was that the thing after the "error" keyword could be a text containing unbalanced parenthesis. Need to think more about it.

Why tests pass and this does not occur normally?
We control the stmlib and this `error` only happen when we change something on the wrong side in the smtlib generation. And the smtlib parser of the solver complains.

manticore/core/smtlib/solver.py
feliam: Some solvers support for the `optimize` keyword. It is configured via super().__init__
https://github.com/trailofbits/manticore/pull/1714/files#diff-a32b8a365208c1e023fa207b83c84067R302-R305

Tried to have some configurable code rather than a few classes.

manticore/core/smtlib/solver.py
feliam: I was hoping not to add an IF here and let it fail by other means in such case.
Then added the asserts to silence mypy.  
So..  
option1: Re add the IF
option2: # ignore: type
option3: leave the asserts

I'm very gently in favor of option2

manticore/core/smtlib/solver.py
ehennenfent: From the docs, on Linux/Mac, [`.terminate`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.terminate) sends SIGTERM, which gives the process a chance to clean up first, and [`.kill`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.kill) sends a SIGKILL, which yanks the rug out from under the process and kills it instantly. So, if we're using `.kill`, there's no need for `.wait`. That said, it would be slightly safer to do something like:
```
self._proc.terminate()
try:
    self._proc.wait(2)
except TimeoutExpired:
    self._proc.kill()
```

manticore/core/smtlib/solver.py
feliam: I setting this as resolved. It may be better suited to print that stuff when the solver returns "unknown" rather than unsat. Unsat is an expected result in a bunch of places.

manticore/core/smtlib/solver.py
feliam: Removed the assertions. Still need to think what's the exception game of solver.py. Should we try to catch and  translate the exceptions to a certain type?

manticore/core/smtlib/solver.py
feliam: Added the _syntax error_ check elsewhere. Though only enabled when debugging.

manticore/core/smtlib/solver.py
feliam: Removed the .wait. As supposedly kill() does it all. If we start seeing zombies we'll add some code here.

PR #1717, opened 2020-05-27
Manticore verifier

feliam:
This check the provided properties on a contract until some termination condition is met.


manticore/core/manticore.py
```diff
@@ -428,6 +428,16 @@ def clear_terminated_states(self):
             self._remove(state_id)
         assert self.count_terminated_states() == 0
 
+    @sync
+    @at_not_running
+    def clear_ready_states(self):
```
ehennenfent, line 433: This function is called `clear_ready_states`, but it's a duplicate of `clear_terminated_states` just above it.

setup.py
```diff
@@ -65,6 +65,11 @@ def rtd_dependent_deps():
     ]
     + rtd_dependent_deps(),
     extras_require=extra_require,
-    entry_points={"console_scripts": ["manticore = manticore.__main__:main"]},
+    entry_points={
+        "console_scripts": [
+            "manticore = manticore.__main__:main",
+            "manticore_verifier = manticore.ethereum.verifier:main",
```
ehennenfent, line 71: We have a few integration tests that call `manticore` via subprocess, which in effect confirm that the console script install was successful. Let's add something similar for `manticore_verifier`, since it looks like a lot of the uncovered code is in the argument parsing and setup for the verifier CLI target.

feliam: https://github.com/trailofbits/manticore/pull/1717/files#diff-9a2242beaf0823573028164bdf320ba7R89-R94

manticore/ethereum/manticore.py
```diff
@@ -1552,29 +1549,38 @@ def generate_testcase(self, state, message="", only_if=None, name="user"):
...
-        """
-        Create a serialized description of a given state.
-        :param state: The state to generate information about
-        :param message: Accompanying message
-        """
+        # Refuse to generate a testcase from an unsound state
         if not self.fix_unsound_symbolication(state):
-            return False
-
-        if only_if is not None:
```
ehennenfent, line 1571: Removing this will change the `only_if` behavior that we discussed w/ @montyly on slack, yeah? If so, we should document that in the release notes.

tests/ethereum/test_general.py
```diff
@@ -62,6 +69,33 @@ def test_int_ovf(self):
...
+            smtcfg.solver = smtcfg.solver.yices
+
+            filename = os.path.join(THIS_DIR, "contracts/prop_verifier.sol")
+            f = io.StringIO()
+            with contextlib.redirect_stdout(f):
+                verifier.manticore_verifier(filename, "TestToken")
+            output = f.getvalue()
+            self.assertIsNotNone(
+                re.compile(
+                    r".*crytic_test_balance\s*\|\s*failed\s*\([0-9a-f]+\).*", re.DOTALL
```
ehennenfent, line 86: Weirdly enough, [CodeCov is saying that the line of code that prints out this test case is not covered](https://codecov.io/gh/trailofbits/manticore/compare/61909a157bd621b273bd5eabd685aef4d7c26af9...8c80a1d984f9f6ecd8887dc5cb7f9764d8938845/diff#D1-332
). However, running the verifier on this contract myself, the output looks fine. :shrug:

The upload seems to have succeeded, so I don't know what to make of it.

manticore/ethereum/manticore.py
feliam: Supposedly nothing is changed from the user perspective.
Added `generate_testcase_ex`  and kept `generate_testcase` with the original behaivor.

The annoying-ish thing about the current behavior is that `mcoreEVM.generate_testcase`: 
It does a bunch of unexpected operations over a state mostly to filter out states that are not feasible. So it is more a "try_generate_testcase".  
Trying to do the same thing by hand via state.solve_xxxx would not give same result unless you replicate all the secret sauce of `mcoreEVM.generte_testcase`.

 `https://github.com/trailofbits/manticore/pull/1717/files#diff-e4836efb704ed46c30e5cfb02496a419R1538

ehennenfent: Oh my bad, this is an entirely different function now. Sounds good!

PR #1724, opened 2020-06-08
Fix default output space when workspace is in mem:

feliam:


manticore/core/manticore.py
```diff
@@ -353,6 +353,8 @@ def __init__(self, initial_state, workspace_url=None, outputspace_url=None, **kw
                 raise TypeError(f"Invalid workspace type: {type(workspace_url).__name__}")
         self._workspace = Workspace(workspace_url)
         # reuse the same workspace if not specified
+        if outputspace_url is None:
+            outputspace_url = workspace_url
         if outputspace_url is None:
             outputspace_url = f"fs:{self._workspace.uri}"
```
ekilmer, line 359: This line
```
if outputspace_url is None:
    outputspace_url = f"fs:{self._workspace.uri}"
```
will never execute.

Delete it?

ekilmer: Oh wait. workspace_url might be `None`. Nevermind.

PR #1725, opened 2020-06-11
New working model of strlen

sschriner:
Updates to the strlen model based on work from strcpy. It seems to be working on concrete and symbolic values. The symbolic testing for strlen will need to be updated.


tests/native/test_models.py
```diff
@@ -208,6 +203,51 @@ def test_symbolic(self):
...
+        self.assertEqual(len(stdouts), len(expected))
+
+        # Assert that every expected stdout has a matching output
+        for e in expected:
+            match = False
+            for s in stdouts:
+                if e == s:
+                    match = True
+                    break
+            self.assertTrue(match)
```
ekilmer, line 249: Nit: Since we're comparing simple strings this time (as opposed to matching regex in the `strcpy` test), we can use sets to easily and efficiently compare `expected` vs. `output`

sschriner: Updated :)

manticore/native/models.py
```diff
@@ -116,33 +116,41 @@ def strcmp(state: State, s1: Union[int, BitVec], s2: Union[int, BitVec]):
...
-    :param State state: current program state
+    :param state: current program state
     :param s: Address of string
     :return: Symbolic strlen result
-    :rtype: Expression or int
     """
 
-    cpu = state.cpu
+    if issymbolic(addr):
+        raise ConcretizeArgument(state.cpu, 0)
```
ehennenfent, line 132: Have you confirmed that this works as expected? We previously raised `ConcretizeArgument(state.cpu, 1)` here. I thought that the argument index was supposed to be that of the model function, not the _modeled_ function, but I know @ekilmer was working on that recently, so it might have changed.

manticore/native/models.py
```diff
@@ -116,33 +116,41 @@ def strcmp(state: State, s1: Union[int, BitVec], s2: Union[int, BitVec]):
...
+    else:
+        offset = state.context["strlen"]
 
-    ret = zero_idx
+    c = cpu.read_int(addr + offset, 8)
+    while not is_definitely_NULL(c, constrs):
+        # If the byte can be NULL concretize and fork states
+        if can_be_NULL(c, constrs):
+            state.context["strlen"] = offset
+            raise Concretize("Forking on possible NULL strlen", expression=(c == 0), policy="ALL")
```
ehennenfent, line 148: If you don't want to risk aggressively forking on every single byte, you can sometimes build up a big bitvec If-Then-Else that will accurately capture all the possible values you could concretize to. You can see a similar technique in [ARM's CLZ](https://github.com/trailofbits/manticore/blob/master/manticore/native/cpu/arm.py#L1375-L1387), and also in the previous implementation of `strlen` right here. There are pros and cons to each approach, and I notice that the old implementation used `_find_zero`, which stops at the first byte that _can_ be null, not the first byte that _must_ be null as yours does. Are there any points of interest you can share that drove you toward one approach or the other?

manticore/native/models.py
ekilmer: I suggested the change from `1` to `0` based on how we do system call modeling, but now I see they're a little bit different. Will test this.

manticore/native/models.py
```diff
@@ -210,10 +210,10 @@ def strcpy(state: State, dst: Union[int, BitVec], src: Union[int, BitVec]) -> Un
     :return: pointer to the dst
     """
     if issymbolic(src):
-        raise ConcretizeArgument(state.cpu, 1)
+        raise ConcretizeArgument(state.cpu, 2)
 
     if issymbolic(dst):
-        raise ConcretizeArgument(state.cpu, 0)
+        raise ConcretizeArgument(state.cpu, 1)
```
sschriner, line 216: Thanks @ekilmer for catching this in both spots!! 😃

sschriner:
@ekilmer @ehennenfent I opted to preserve both model strategies and note the difference in them. I'm thinking this is probably the best solution for allowing the most user control over strategies and tradeoffs based on the program being analyzed.

PR #1727, opened 2020-06-12
Remove/procrastinate solver query in ether leak detector

feliam:
Removes online solver queries fro ethleak detector.

### DIFF
--- a/manticore/ethereum/detectors.py
+++ b/manticore/ethereum/detectors.py
@@ -182,12 +182,12 @@ class DetectExternalCallAndLeak(Detector):
                 self.add_finding_here(
                     state,
                     f"Reachable ether leak to sender via argument",
-                    constraint=AND(msg_sender == dest_address, sent_value != 0)
+                    constraint=AND(msg_sender == dest_address, sent_value != 0),
                 )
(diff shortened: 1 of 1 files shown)


manticore/ethereum/detectors.py
```diff
@@ -177,34 +177,47 @@ def will_evm_execute_instruction_callback(self, state, instruction, arguments):
...
-                            constraint=msg_sender != dest_address,
-                        )
+                    self.add_finding_here(
+                        state,
+                        f"Reachable external call to user controlled address via argument",
+                        constraint=AND(msg_sender != dest_address, sent_value == 0),
+                    )
+
             else:
                 if msg_sender == dest_address:
```
smoelius, line 214: This won't freak out if `msg_sender` is symbolic, right (because `dest_address` is not symbolic in this case)?

Everything else about this PR looks good to me.

feliam: Yes, msg.sender does not get too far being symbolic as it is right now... https://github.com/trailofbits/manticore/blob/737f751ab4c1be34f8e435e592242a4b52a6a501/manticore/platforms/evm.py#L3329

PR #1729, opened 2020-06-15
Slottify expressions

### LINKED ISSUE #1731: Use `__slots__` throughout the expression system
### Why
As Manticore users, we want Manticore to use RAM as efficiently as possible. One can minimize the amount of memory a Python object uses by using the `__slots__` property. Since the expressions system involves the creation of thousands of nested Python objects, using `__slots__` here would provide noticeable memory usage improvements. 

### Acceptance Criteria
All subclasses of `Expression` use the `__slots__` property

### Dev Notes
We added limited support for this in #1635, but found out that subclasses of slotted classes need to also have `__slots__` explicitly defined, so this implementation is not complete.

feliam:
* Added a metaclass magic to compile the `__slots__` from parent classes automatically-ish. (xslots)
* Refactored Expression constructor arguments to be all kwargs based as suggested by the "Super Considered Super" thing.
* Added tests to check 100% of the Expressions are slotted. It uses `__subclasses__` to ensure all are accounted for.
* ArrayProxy renamed to MutableArray
* Array's index_max index_bits value_bits -> length, index_size, value_size
* Added an ArrayConstant to hold tainted concrete "bytes" and other concrete arrays.
* Added a way to read the full model from a solved core so 2 consecutives solver.get_value wont solve the CS twice
* BitVec to Bitvec
.... (tbc)


manticore/core/smtlib/expression.py
```diff
@@ -400,555 +467,749 @@ def __ne__(self, other):
...
+
+    @property
+    def written(self):
+        # Calculate only first time
+        # This can have repeated and reused written indexes.
+        if self._written is None:
+            written = set()
+            for offset, value in get_items(self):
+                written.add(offset)
+            self._written = written
```
ekilmer, line 1162: The following should also work
```suggestion
            self._written = {offset for offset, _ in get_items(self)}
```

manticore/core/smtlib/expression.py
```diff
@@ -966,279 +1227,249 @@ def index(self):
...
+        if default is None:
+            # No default. Returns normal array select
+            return ArraySelect(self, index)
+
+        # if a default is defined we need to check if the index was previously written
+        return BitvecITE(self.is_known(index), ArraySelect(self, index), self.cast_value(default))
+
+        # build a big ITE expression
+        array, offset, items = self, 0, []
+        while not isinstance(array, ArrayVariable):
```
ekilmer, line 1258: The `return` statement creates dead code. I can't figure out what the correct thing to do is, though.

Is everything after the `return` supposed to be deleted?

manticore/core/smtlib/expression.py
```diff
@@ -1324,6 +1552,72 @@ def false_value(self):
         return self.operands[2]
 
 
-Constant = (BitVecConstant, BoolConstant)
-Variable = (BitVecVariable, BoolVariable, ArrayVariable)
-Operation = (BitVecOperation, BoolOperation, ArrayOperation, ArraySelect)
+# auxiliar functions maybe move to operators
```
ekilmer, line 1555:
```suggestion
# auxiliary functions. Maybe move to operators
```

manticore/core/smtlib/solver.py
```diff
@@ -396,6 +397,52 @@ def _pop(self):
...
+
+                        class A:
+                            def __init__(self, d, default):
+                                self._d = d
+                                self._default = default
+
+                            def __getitem__(self, index):
+                                return self._d.get(index, self._default)
+
+                        value = A(valued, variable.default)
```
ekilmer, line 439: Why is this class here? Can you add a comment?

manticore/core/smtlib/solver.py
```diff
@@ -761,4 +833,5 @@ def instance(cls):
             cls.choice = consts.solver
 
         SelectedSolver = {"cvc4": CVC4Solver, "yices": YicesSolver, "z3": Z3Solver}[cls.choice.name]
+        return YicesSolver.instance()
         return SelectedSolver.instance()
```
ekilmer, line 837: Delete one of these lines

ekilmer (changes requested):
Looking good!

Made it through 10 files. 26 more to go.

Bear with me as I try to understand what's going on in the code and this large changeset 😅

manticore/core/smtlib/constraints.py
```diff
@@ -178,58 +175,33 @@ def related_to(self, *related_to) -> "ConstraintSet":
...
         variables, constraints = self.get_declared_variables(), self.constraints
-
         if replace_constants:
             constant_bindings = {}
             for expression in constraints:
                 if (
                     isinstance(expression, BoolEqual)
                     and isinstance(expression.operands[0], Variable)
-                    and isinstance(expression.operands[1], (*Variable, *Constant))
+                    and not isinstance(expression.operands[1], (Variable, Constant))
```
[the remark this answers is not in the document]
feliam: Actually it looks like we want `and  isinstance(expression.operands[1], Constant)` no ?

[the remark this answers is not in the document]
ekilmer: Yes, that sounds more correct 👍

manticore/core/smtlib/solver.py
feliam: `get_model` needs doc if it going to stay.

The goal of this method is to get an assigment for all variables defined in the CS.
There is a `problem` with Array variables. Because the SMT solver finds assigments for the "used" slots and assume a value for all the unused slots or something like that. Need to paste an example of what smtlib produces as a assigment for an array variable but I think the result constant array assignment is:

* unbounded
* have a default value for unused slots

So, not that the A class is close to a good solution but we can not simply return a "bytes" instance here.

tests/other/test_smtlibv2.py
```diff
@@ -26,82 +27,362 @@
...
+        SelectedSolver.instance().must_be_true(cs, a[3] == 30)
+        # SelectedSolver.instance().must_be_true(cs, a[4] == 0) #undefined!
+
+        b = a[:]
+        # b := 0 10 20 30 0 0 x x x x      (x undefined)
+        SelectedSolver.instance().must_be_true(cs, b.default == 0)
+        SelectedSolver.instance().must_be_true(cs, b[0] == 0)
+        SelectedSolver.instance().must_be_true(cs, b[1] == 10)
+        SelectedSolver.instance().must_be_true(cs, b[2] == 20)
+        SelectedSolver.instance().must_be_true(cs, b[3] == 30)
```
ekilmer, line 161: I think I'm confused about what's actually being tested here.

`SelectedSolver.instance().must_be_true(cs, expr)` returns a boolean telling us whether the expression _must be true_, right? We don't ever seem to check the return value of these expressions. 

Wouldn't doing `self.assertTrue(SelectedSolver.instance().must_be_true(cs, expr))` be what we actually want?

Am I missing something?

tests/other/test_smtlibv2.py
```diff
@@ -26,82 +27,362 @@
...
+            value_size=8,
+            length=100,
+            name="name",
+            pickle_size=156,
+            sizeof=80,
         )
+        check(BitvecConstant, size=32, value=10, pickle_size=107, sizeof=56)
+        check(BoolConstant, value=False, pickle_size=94, sizeof=48)
+
+        # TODO! But you can instantiate an ArraConstant
```
ekilmer, line 218:
```suggestion
        # TODO! But you can instantiate an ArrayConstant
```

tests/other/test_smtlibv2.py
```diff
@@ -26,82 +27,362 @@
...
+        check(BitvecITE, condition=x, true_value=bvx, false_value=bvy, pickle_size=161, sizeof=56)
+
+        a = ArrayVariable(index_size=32, value_size=32, length=324, name="name")
+        check(ArrayConstant, index_size=32, value_size=8, value=b"A", pickle_size=136, sizeof=64)
+        check(ArraySlice, array=a, offset=0, size=10, pickle_size=136, sizeof=48)
+        check(ArraySelect, array=a, index=bvx, pickle_size=161, sizeof=56)
+        check(ArrayStore, array=a, index=bvx, value=bvy, pickle_size=188, sizeof=80)
+
+        def all_subclasses(cls):
+            return set((Expression,)).union(
```
ekilmer, line 283: Nit (as suggested by Pycharm)
```suggestion
            return {Expression}.union(
```

tests/other/test_smtlibv2.py
```diff
@@ -26,82 +27,362 @@
...
+        x = BoolConstant(value=True, taint=("red",))
+        y = BoolConstant(value=False, taint=("blue",))
+        z = BoolOr(x, y)
+        self.assertIn("red", x.taint)
+        self.assertIn("blue", y.taint)
+        self.assertIn("red", z.taint)
+        self.assertIn("blue", z.taint)
+
+    def test_Expression_BitvecTaint(self):
+        # Bool can not be instantiaated
```
ekilmer, line 311:
```suggestion
        # Bitvec can not be instantiated
```

tests/other/test_smtlibv2.py
```diff
@@ -26,82 +27,362 @@
...
-"""
-class Z3Specific(unittest.TestCase):
+        all_types = all_subclasses(Expression)
+        self.assertSetEqual(checked, all_types)
+
+    def test_Expression_BitvecOp(self):
+        a = BitvecConstant(size=32, value=100)
+        b = BitvecConstant(size=32, value=101)
+        x = a + b
+        self.assertTrue(isinstance(x, Bitvec))
```
ekilmer, line 296:
```suggestion
        self.assertTrue(isinstance(x, BitvecOperation))
        self.assertEqual(type(x), BitvecAdd))
```

tests/other/test_smtlibv2.py
```diff
@@ -26,82 +27,362 @@
...
+        x = BitvecConstant(size=32, value=123, taint=("red",))
+        y = BitvecConstant(size=32, value=124, taint=("blue",))
+        z = BoolGreaterOrEqualThan(x, y)
+        self.assertIn("red", x.taint)
+        self.assertIn("blue", y.taint)
+        self.assertIn("red", z.taint)
+        self.assertIn("blue", z.taint)
+
+    def test_Expression_Array(self):
+        # Bool can not be instantiaated
```
ekilmer, line 323:
```suggestion
        # Array can not be instantiated
```

tests/other/test_smtlibv2.py
```diff
@@ -26,82 +27,362 @@
...
+        self.assertSetEqual(checked, all_types)
+
+    def test_Expression_BitvecOp(self):
+        a = BitvecConstant(size=32, value=100)
+        b = BitvecConstant(size=32, value=101)
+        x = a + b
+        self.assertTrue(isinstance(x, Bitvec))
+
+    def test_Expression_BoolTaint(self):
+        # Bool can not be instantiaated
```
ekilmer, line 299:
```suggestion
        # Bool can not be instantiated
```

tests/other/test_smtlibv2.py
```diff
@@ -26,82 +27,362 @@
...
+        self.assertIn("blue", y.taint)
+        self.assertIn("red", z.taint)
+        self.assertIn("blue", z.taint)
+
+    def test_Expression_Array(self):
+        # Bool can not be instantiaated
+        self.assertRaises(Exception, Array, ())
+
+        a = ArrayConstant(index_size=32, value_size=8, value=b"ABCDE")
+        a[0] == ord("A")
```
ekilmer, line 327: This statement has no effect? Are we trying to assert that `a[0]` is equal to `ord("A")`?

tests/other/test_smtlibv2.py
```diff
@@ -26,82 +27,362 @@
...
-        print (self.solver._solver_version())
-        self.assertTrue(self.solver._solver_version() == Version(major=4, minor=4, patch=1))
+    def tearDown(self):
+        del self.solver
 
-    def test_check_solver_newer(self):
-        self.solver._received_version = '(:version "4.5.0")'
-        self.assertTrue(self.solver._solver_version() > Version(major=4, minor=4, patch=1))
+    def test_signed_unsigned_LT_(self):
+        mask = (1 << 32) - 1
```
ekilmer, line 354: `mask` is unused. Remove?

tests/other/test_smtlibv2.py
```diff
@@ -1080,6 +1386,7 @@ def testRelated(self):
         # and send in all the other stuff
         self.assertNotIn("AA", cs.related_to(bb1 == bb1).to_string())
 
+    @unittest.skip("FIXME")
```
ekilmer, line 1389: Is this going to be fixed in this PR?

tests/other/test_smtlibv2.py
```diff
@@ -446,20 +742,29 @@ def testBasicArrayProxySymbIdx2(self):
...
-            self.solver.can_be_true(cs, array[1] == 12345)
-        )  # no default so it can be anything
+    def testBasicConstatArray(self):
+        cs = ConstraintSet()
+        array1 = MutableArray(
+            cs.new_array(index_size=32, value_size=32, length=10, name="array1", default=0)
+        )
+        array2 = MutableArray(
+            cs.new_array(index_size=32, value_size=32, length=10, name="array2", default=0)
+        )
```
ekilmer, line 762: `array2` is unused. Do we want to compare it with array1 somehow?

tests/other/test_smtlibv2.py
```diff
@@ -446,20 +742,29 @@ def testBasicArrayProxySymbIdx2(self):
...
+        array1[0:10] = range(10)
+        self.assertTrue(array1[0] == 0)
```
ekilmer, line 764:
```suggestion
        self.assertEqual(array1[0], 0)
```

tests/other/test_smtlibv2.py
```diff
@@ -446,20 +742,29 @@ def testBasicArrayProxySymbIdx2(self):
...
+        # yeah right self.assertTrue(array1[0:10] == range(10))
+        array_slice = array1[0:10]
+        self.assertTrue(array_slice[0] == 0)
```
ekilmer, line 767:
```suggestion
        self.assertEqual(array_slice[0], 0)
```

ekilmer, line 767: We also want to add another test to this, right?
```suggestion
        self.assertEqual(array_slice[0], 0)
        self.assertEqual(array_slice[9], 9)
```

ekilmer (changes requested):
Finished reviewing all files! Thank you again for this work. Excited to see it land soon!! 💯

manticore/core/smtlib/expression.py
```diff
@@ -1,33 +1,117 @@
...
+        attrs["__xslots__"]: Tuple[str] = tuple(xslots)
+        if abstract:
+            attrs["__slots__"] = ()
+        else:
+            attrs["__slots__"]: Tuple[str] = tuple(map(cls._remove_mod, attrs["__xslots__"]))
+
+        return super().__new__(cls, clsname, bases, attrs)
+
+
+class Expression(object, metaclass=XSlotted, abstract=True):
```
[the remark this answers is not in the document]
feliam: It's been a while since I do not entertain that idea. Originally my concern was about it doing some slow dynamic checks.
I'd like to see how an ABC version of this may look and run.
We still would need a way for handle slots in multi-inherited class thing.

manticore/core/smtlib/expression.py
```diff
@@ -1,33 +1,117 @@
...
     """ Abstract taintable Expression. """
 
-    __slots__ = ["_taint"]
+    __xslots__: Tuple[str, ...] = ("_taint",)
 
     def __init__(self, taint: Union[tuple, frozenset] = ()):
```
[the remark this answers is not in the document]
feliam: Everything should only use kwargs. No positional args. we need a  "*" there.

manticore/core/smtlib/expression.py
```diff
@@ -159,105 +249,80 @@ def __rxor__(self, other):
...
 
     def __bool__(self):
-        # try to be forgiving. Allow user to use Bool in an IF sometimes
-        from .visitors import simplify
-
-        x = simplify(self)
-        if isinstance(x, Constant):
-            return x.value
-        raise NotImplementedError("__bool__ for Bool")
+        raise NotImplementedError
```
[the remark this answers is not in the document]
feliam: "You tried to use a Bool Expression as a boolean constant. Expressions could represent a set of concrete values." ?

We can try to be forgiving in  BoolContant an BoolOperation maybe. I tried to cut the `simplify` abuse a bit.

manticore/core/smtlib/expression.py
```diff
@@ -400,555 +467,749 @@ def __ne__(self, other):
...
-
-class GreaterOrEqual(BoolOperation):
-    def __init__(self, a, b, *args, **kwargs):
-        assert a.size == b.size
-        super().__init__(a, b, *args, **kwargs)
+    def __bool__(self):
+        simplified = local_simplify(self)
+        if isinstance(simplified, Constant):
+            return simplified.value
+        raise NotImplementedError
```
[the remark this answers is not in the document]
feliam: Yes....
on the subject... Every use of local_simplify looks hacky. We are potentially silently masking the taint values of the expression
```
 flag: Bool = $someTaintedBoolExpresion
 if flag:
   dosomething()
 else:
   dosomethingelse()
```
Whatever expresion is made in `dosomething.*` may need to keep the taint of flag.
(so this is something we may not want to induce from the Expressions??)

manticore/core/smtlib/expression.py
```diff
@@ -400,555 +467,749 @@ def __ne__(self, other):
...
+        return ArrayStore(array=self, index=index, value=value)
 
     @property
-    def operands(self):
-        return self._operands
+    def written(self):
+        return set()
+
+    def is_known(self, index):
+        return False
```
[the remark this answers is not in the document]
feliam: this need to be designed well and documented.

`is_known(index)` supposedly is true if the index has been written too `written` is the set of touched indexes.

If not known the value should be the default. If the default is None then a free variable (that is the default smt behavior)

manticore/core/smtlib/expression.py
```diff
@@ -400,555 +467,749 @@ def __ne__(self, other):
...
+
+    @property
+    def underlying_variable(self):
+        array = self
+        while not isinstance(array, ArrayVariable):
+            array = array.array
+        return array
```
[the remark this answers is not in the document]
feliam: This thing assumes an array is build over an single initial ArrayVariable on top of which the values are set.
This digs out that inner variable. It is probably just for printing some nicer outputs?
Candidate for removal.

manticore/core/smtlib/expression.py
```diff
@@ -400,555 +467,749 @@ def __ne__(self, other):
...
+    )
+
+    @property
+    def concrete_cache(self):
+        if self._concrete_cache is not None:
+            return self._concrete_cache
+        self._concrete_cache = {}
+        for index, value in get_items(self):
+            if not isinstance(index, Constant):
+                break
```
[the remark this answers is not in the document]
feliam: Because.....
Once you find a symbolic write. Or once you found a write to a symbolic index this symbolic index could overwrite any(*) of the following Constants indexes that may come next.


* actually it could potentially overwrite some and not others. You could get smart here and try to partition the array somehow.

manticore/core/smtlib/expression.py
feliam: Yes. But I'm not super comfortable with what happens before the return.

feliam: Oh so the thing after the return tries not to use Arrays. An build all the array logic as a big ITE.

manticore/core/smtlib/expression.py
```diff
@@ -159,105 +238,89 @@ def __rxor__(self, other):
...
 
 
 class BoolITE(BoolOperation):
-    def __init__(self, cond: "Bool", true: "Bool", false: "Bool", **kwargs):
-        super().__init__(cond, true, false, **kwargs)
+    def __init__(self, cond: Bool, true: Bool, false: Bool, **kwargs):
+        super().__init__(operands=(cond, true, false), **kwargs)
+
 
+class Bitvec(Expression, abstract=True):
```
ehennenfent, line 308: I'm torn on this. On the one hand, "Bitvec" feels better than "BitVec" to me since we rarely (if ever) actually refer to it as the expanded "Bit Vector." On the other, doing a global find-and-replace from "Bitvec" to "BitVec" reduces the size of this diff by about 20%, and the name change may break external scripts that use the BitVec types directly. Yes, they should probably use the ConstraintSet helper methods and Operators instead, but I know I've done it a few times, so others may have too. 

I'm leaning toward the latter. Thoughts?

feliam: Thank you for diving into this MPR!
I do not have too much stake in the naming part but I think we should consider making all the arguments of the constructor of the multi-inheritance expression system to have _only_ keyword arguments. As prescribed in this old blog post: https://rhettinger.wordpress.com/2011/05/26/super-considered-super/.

If we go that way eventually all uses of those will be changed.

ekilmer: >  I think we should consider making all the arguments of the constructor of the multi-inheritance expression system to have only keyword arguments.

I agree with this 👍

feliam: BTW currently (in this PR) only the basic types Bool, Bitvec, Expression, Operation, Constant... force you to use keyword args. But the user facing types (the leafs in the inheritance tree ) still use positional args. Mostly to keep the diff "smaller".

manticore/core/smtlib/expression.py
```diff
@@ -159,126 +260,101 @@ def __rxor__(self, other):
...
-    @property
-    def declaration(self):
-        return f"(declare-fun {self.name} () Bool)"
-
-
-class BoolConstant(Bool):
-    __slots__ = ["_value"]
+        raise ExpressionError(
+            "You tried to use a Bool Expression as a boolean constant. Expressions could represent a set of concrete values."
+        )
```
ekilmer, line 265: This statement is causing issues. I've replaced it with just 
```suggestion
        return self
```
and it seems to be passing for me natively on both Python 3.6 and 3.9 🤷 

Maybe we should revert this to how it was.

CHANGELOG.md
```diff
@@ -440,7 +440,7 @@ Thanks to our external contributors!
 - Ethereum: Support for Solidity `bytesM` and `bytes` types
 - Ethereum: Beta API for preconstraining inputs (`ManticoreEVM.constrain`)
 - Improved performance for smtlib module
-- Ability to transparently operate on bytearray and symbolic buffer (ArrayProxy) types (e.g: concatenate, slice)
+- Ability to transparently operate on bytearray and symbolic buffer (MutableArray) types (e.g: concatenate, slice)
```
ehennenfent, line 443: Nit: This looks like a find & replace change. I think we should probably avoid rewriting the changelog for previous versions. We can add a line in the next release noting that we renamed "ArrayProxy" to "MutableArray"

manticore/core/smtlib/expression.py
```diff
@@ -455,500 +532,713 @@ def Bool(self):
...
-    def __init__(self, array: "Array", index: "BitVec", value: "BitVec", *args, **kwargs):
-        assert index.size == array.index_bits
-        assert value.size == array.value_bits
-        super().__init__(array, index, value, *args, **kwargs)
+    __xslots__: Tuple[str, ...] = (
+        "_written#v",
+        "_concrete_cache#v",
+        "_length#v",
+        "_default#v",
+    )
```
ekilmer, line 1175: What does the `#v` suffix mean here?

PR #1757, opened 2020-06-27
Enable nightly uploads to PyPI

ehennenfent:
This will upload nightly scheduled builds of Manticore to PyPI under the extension `.devYYMMDD`


ehennenfent:
As this is meant to be schedule-only, the most recent run of the tests won't reflect the upload. However, on [the run before that](https://github.com/trailofbits/manticore/pull/1757/checks?check_run_id=819831836), it finished successfully.

You can see the released version here: https://pypi.org/project/manticore/0.3.4.dev200629/

.github/workflows/ci.yml
```diff
@@ -75,5 +75,25 @@ jobs:
...
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.6
+    - name: Build Dist
+      run: |
+        python3 -m pip install wheel
+        python3 setup.py --dev_release sdist bdist_wheel
+    - name: Upload to PyPI
+      uses: pypa/gh-action-pypi-publish@master
```
ekilmer, line 95: We should use a stable tag/release.

Looks like `v1.3.0` is the latest

.github/workflows/ci.yml
```diff
@@ -75,5 +75,25 @@ jobs:
...
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.6
+    - name: Build Dist
+      run: |
+        python3 -m pip install wheel
+        python3 setup.py --dev_release sdist bdist_wheel
+    - name: Upload to PyPI
+      uses: pypa/gh-action-pypi-publish@v1
```
ekilmer, line 95:
```suggestion
      uses: pypa/gh-action-pypi-publish@v1.3.0
```

ehennenfent: You sure about this? Standard practice for other actions seems to be to only pin the major version.

Also, though 1.3.0 is the latest tag, 1.2.2 is the latest version available in the marketplace.

ekilmer: > only pin the major version

Oh, I didn't realize that. I guess I just assumed that they did a more simple lookup based on exact tag/branch

Ignore this then :)

.github/workflows/ci.yml
```diff
@@ -75,5 +75,25 @@ jobs:
         file: ./coverage.xml
         yml: ./codecov.yml
 # Disabled this line because Codecov has been extra flaky lately, and having to re-run the entire
-# test suite until every coverage upload step succeeds is more of a hassle than it's worth. 
+# test suite until every coverage upload step succeeds is more of a hassle than it's worth.
 #        fail_ci_if_error: true
+  upload:
+    runs-on: ubuntu-latest
+#    if: github.event_name == 'schedule'
```
ekilmer, line 82: I just thought of something else regarding this conditional.

Do we need to specify the branch name (`master`)?

ekilmer: Based on the other `schedule` workflow runs, it seems to use the "default" working branch, so I think this is probably fine.

Issue #1771, opened 2020-07-09
Examples all fail silently with raised default gas


wolflo:
### Summary of the problem
The majority of manticore examples create contracts as follows:
```
user_account = m.create_account(balance=1000)
contract_account = m.solidity_create_contract(contract_src, owner=user_account, balance=0)
```
The default transaction gas is now 3000000, so `solidity_create_contract` fails silently (returns `None`) because the sender can't cover the gas.

### Manticore version
0.3.4

PR #1772, opened 2020-07-09
Update EVM usage example

wolflo:
Addresses #1771 (at least for the example in the readme). Increases the balance of the created account so the user has enough eth to send the contract creation transaction with the default transaction gas.

It's possible it would be more helpful to specify the gas for `solidity_create_contract` instead, but this keeps the example as simple as possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ contract Adder {
 """
 m = ManticoreEVM()
 
-user_account = m.create_account(balance=1000)
+user_account = m.create_account(balance=10000000)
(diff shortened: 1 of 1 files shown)


wolflo:
@feliam that's great! Should I close this then if the examples won't run into this issue anymore?

feliam:
No no. Old examples are failing because by default now the transaction fee and gas are correctly debited from the caller's balance. 

We need a specific configuration constant that disables all ethereum balance and gas calculations. So it is easier for the user that is interested in a different aspect of the contract

Merging this PR in 3 2 1...

Issue #1773, opened 2020-07-12
Incorrect gas metering for CALL 


wolflo:
### Summary of the problem
Gas metering for `CALL` instructions that do not cause accounts to be added to the state trie still deducts the 25000 gas account creation cost in some cases.

### Manticore version
0.3.4

### Python version
3.8.3

### Step to reproduce the behavior
A script that makes some relevant calls and logs the deducted gas can be found [here](https://github.com/wolflo/manticore-scratch/blob/master/call_gas_mcore.py), and should result in the output below:
```
user: 0xdc50f787714f7694b8191991269e2cb684fce07b
existing account: 0xffbb4f64add523a9ac621b32c0f3173184b3e1c9
non-existant account: 0x137665008fce9edef4456904d5401c71c71c71c7

call to existing account with value
----------
call from 0x21a6b66ca8cc804bfccfe2902a121f1c5493e207 to 0xffbb4f64add523a9ac621b32c0f3173184b3e1c9
value: 1
gas consumed: 32400

call to non-existant account without value
----------
call from 0x21a6b66ca8cc804bfccfe2902a121f1c5493e207 to 0x137665008fce9edef4456904d5401c71c71c71c7
value: 0
gas consumed: 25700

call to existing account without value
----------
call from 0x21a6b66ca8cc804bfccfe2902a121f1c5493e207 to 0xffbb4f64add523a9ac621b32c0f3173184b3e1c9
value: 0
gas consumed: 700
```

### Expected behavior
Disclaimer: This is from my own [reference](https://github.com/wolflo/evm-opcodes/blob/master/gas.md), so it would be good to check against another source
```python
base_gas = 700 + mem_expansion_cost
if call_value > 0: # sending value with call
    base_gas += 9000
    if is_empty(target_addr): # forcing a new account to be created in the state trie by sending value to empty account
        base_gas += 25000
# In addition, any call with `call_value > 0` has a 2300 gas stipend added to the gas sent with the call, which is then refunded to the caller if unspent.
```

- A call to an existing account with `call_value > 0` should deduct `700 + 9000 - 2300 stipend = 7400 gas`
- A call to an non-existant account with `call_value == 0` should deduct `700 gas`

### Actual behavior
- A call to an existing account with `call_value > 0` deducts `700 + 9000 + 25000 - 2300 = 32400 gas`
- A call to an non-existant account with `call_value == 0` deducts `700 + 25000 = 25700 gas`

[...]

PR #1774, opened 2020-07-12
Update gas metering for calls to empty accounts

wolflo:
Addresses #1773, but is in need of some tests still.

Also, it brings up the related question of determining account existence. [EIP-161](https://eips.ethereum.org/EIPS/eip-161) defines an account as "*empty* when it has **no code** and **zero nonce** and **zero balance**." I'm not sure if existence in `EVM.world.accounts` ends up having the same meaning or not.


ehennenfent:
Thanks for the PR! Just as an update, we're interested in merging this, but we've introduced a test regression that we're trying to eliminate before merging other PRs. You can track the progress of that here: #1776

wolflo:
@ehennenfent thanks for the update. I can write some tests for it in the meantime, but it looks like most of the evm tests rely on the Ethereum VM test suite. My assumption is this isn't covered because the vm tests replace call execution with the `callcreates` item.

So, are python tests preferred for EVM changes, or additional GST-style tests?

ehennenfent:
The particular style tests are written in doesn't matter so much as the time it takes to run them - in particular, we don't want to push the time for one build step above ~40m. Assuming that Python tests are easier, you could put them in `tests/ethereum`, which already has several miscellaneous EVM tests that fall under it.

ehennenfent:
Hey there! We've fixed #1776 so I wanted to check in again. It looks like all this needs is a small regression test to ensure that gas is calculated correctly for the `CALL` instruction. I think `ethereum/test_general.py` and `ethereum/test_regressions.py` both have some similar tests that should be easily modifiable.

wolflo:
Sorry I lost track of this a bit, but will add a test this week.

wolflo:
@ehennenfent happy to make any changes or updates you can suggest. It doesn't seem like this branch is the source of the CI failures, but I can rebase or dig into the issue as well.

PR #1775, opened 2020-07-16
State Introspection API

### LINKED ISSUE #1736: State Introspection API
### Why
As a Manticore user, I'm frustrated by Manticore's apparent inactivity while it's running. If no warnings or debug messages are generated, I can't tell whether Manticore is making progress, or simply spinning its wheels. 

To fix this, I need an APi that will let me see query what Manticore is doing during execution.

### Acceptance Criteria
At any time from a script, a user can:

- [x] List the states and their IDs
- [x] View the current status of a state (Waiting for solver / Waiting for worker / Running / Killed / Terminated)
- [x] View the current program counter for a state and number of instructions executed
- [x] View the amount of time since the last status transition
*Stretch goals:*
- [x] Provide a means to subclass the plugin that captures internal state
- [x] Register a callback that's called every `n` seconds when Manticore is executing

### Dev Notes
We'll need to add state transition callbacks to the event system. Then we'll have a plugin running on the main thread that subscribes to all these events and maintains a model of the current state lists that doesn't require jumping across thread/process boundaries.

ehennenfent:
Adds an API to query Manticore's state lists. Closes #1736 

TODO:
- [x] Items from #1736 
- [x] Docs
- [x] Tests


ehennenfent:
This is ready for for a round of review. 

The most important API functions are the following items on `ManticoreBase`:
* [`introspect`](https://manticore.readthedocs.io/en/dev-introspection-api/base.html#manticore.core.manticore.ManticoreBase.introspect) - returns a dict of [State Descriptors](https://manticore.readthedocs.io/en/dev-introspection-api/states.html#manticore.core.plugin.StateDescriptor), which provide read-only descriptions of the states w/o needing the lock
* [`register_daemon`](https://manticore.readthedocs.io/en/dev-introspection-api/base.html#manticore.core.manticore.ManticoreBase.register_daemon) - allows registering of a thread that runs in the main process until Manticore exits
* [`pretty_print_states`](https://manticore.readthedocs.io/en/dev-introspection-api/base.html#manticore.core.manticore.ManticoreBase.pretty_print_states) - Prints a formatted state descriptor dict to stdout

And for a quick API usage example, see [test_state_introspection.py](https://github.com/trailofbits/manticore/blob/ae01ebf7d82ff7ea12841f9e835020f5c999f3ed/tests/other/test_state_introspection.py)

Test coverage is  `¯\_(ツ)_/¯`. I think the integration tests get us pretty close to 85%, but I've yet to have a successful upload to CodeCov. Anecdotal evidence from their issue tracker suggests that their API has been _extra_ flaky lately. 

There's one feature I would have liked to add, but don't currently have the hours for. Specifically, setting up a signal handler so that Manticore will call `pretty_print_states` on SIGUSR1. Turns out this is unexpectedly tricky in a multiprocessing environment, and I don't have the available hours for it right now.

manticore/core/state.py
```diff
@@ -184,8 +269,22 @@ def constraints(self, constraints):
         self._constraints = constraints
         self.platform.constraints = constraints
 
+    def _update_state_descriptor(self, descriptor: StateDescriptor, *args, **kwargs):
+        """
+         Called on execution_intermittent to update the descriptor for this state. This is intended for information
+         like the PC or instruction count, where updating after each instruction would be a waste of cycles.
+         This one updates the execution counts
+         :param descriptor: StateDescriptor for this state
```
ekilmer, line 277: I think a newline is required for docs
```suggestion

         :param descriptor: StateDescriptor for this state
```

manticore/core/worker.py
```diff
@@ -229,3 +230,24 @@ def start(self):
...
+class DaemonThread(WorkerThread):
+    """
+    Special case of WorkerThread that will exit whenever the main Manticore process exits.
+    """
+
+    def start(self, target: typing.Optional[typing.Callable] = None):
+        """
+        Function that starts the thread. Can take an optional callable to be invoked at the start, or can be subclassed,
+        in which case `target` should be None and the the `run` method will be invoked at the start.
+        :param target: an optional callable that will be invoked to start the thread. The callable should accept this
```
ekilmer, line 244: I think a newline is required for docs
```suggestion

        :param target: an optional callable that will be invoked to start the thread. The callable should accept this
```

manticore/ethereum/state.py
```diff
@@ -1,6 +1,17 @@
...
 class State(StateBase):
     def execute(self):
+        super().execute()
         return self._platform.execute()
+
+    def _update_state_descriptor(self, descriptor: StateDescriptor, *args, **kwargs):
+        """
+        Called on execution_intermittent to update the descriptor for this state.
+        This one should apply any EVM-specific information to the descriptor.
+        :param descriptor: StateDescriptor for this state
```
ekilmer, line 14:
```suggestion

        :param descriptor: StateDescriptor for this state
```

manticore/native/state.py
```diff
@@ -96,3 +97,12 @@ def invoke_model(self, model):
         :param model: callable, model to invoke
         """
         self._platform.invoke_model(model, prefix_args=(self,))
+
+    def _update_state_descriptor(self, descriptor, *args, **kwargs):
+        """
+        Called on execution_intermittent to update the descriptor for this state.
+        This one should apply any native-specific information to the descriptor. Right now, that's just the PC
+        :param descriptor: StateDescriptor for this state
```
ekilmer, line 105:
```suggestion

        :param descriptor: StateDescriptor for this state
```

manticore/ethereum/state.py
```diff
@@ -1,6 +1,18 @@
...
+
+    def _update_state_descriptor(self, descriptor: StateDescriptor, *args, **kwargs):
+        """
+        Called on execution_intermittent to update the descriptor for this state.
+        This one should apply any EVM-specific information to the descriptor.
+
+        :param descriptor: StateDescriptor for this state
+        """
+        super()._update_state_descriptor(descriptor, *args, **kwargs)
+        descriptor.pc = None  # TODO - Grab whatever notion of a PC EVM has
```
ehennenfent, line 18: @feliam & @montyly - What details about the ManticoreEVM state would you be interested in having made available via the introspection API?

manticore/core/manticore.py
```diff
@@ -1065,18 +1082,33 @@ def run(self):
...
+        # Passing generators to callbacks is a bit hairy because the first callback would drain it if we didn't
+        # clone the iterator in event.py. We're preserving the old API here, but it's something to avoid in the future.
         self._publish("will_run", self.ready_states)
         self._running.value = True
         # start all the workers!
         for w in self._workers:
             w.start()
 
+        # Create each daemon thread and pass it `self`
+        if not self._daemon_threads:  # Don't recreate the threads if we call run multiple times
```
feliam, line 1104: So we have this here just because we need the workers to start before this threads?
(As opposed to have a Thread started from the main script before the run)

If the workers were stated at `Manticore.__init__()` could not we do something like the following from main():  
```
with SomeDaemonThread():
    m.run()
```
?

manticore/core/manticore.py
```diff
@@ -1139,3 +1178,41 @@ def save_run_data(self):
...
+        """
+        Allows the user to extend the functionality of the default IntrospectionAPI plugin.
+        Must be called before ManticoreBase.run()
+
+        :param new_plugin: a subclass of IntrospectionAPIPlugin
+        """
+        assert issubclass(
+            new_plugin, self._introspection_plugin
+        ), "New plugin must be a subclass of IntrospectionAPIPlugin"
+        self._introspection_plugin = new_plugin
```
feliam, line 1193: We have the reference to this plugin from outside manticore. Why do we need t keep a specific name for it if it is going to be invoked via events.

tests/other/test_state_introspection.py
```diff
@@ -0,0 +1,94 @@
...
+    def on_execution_intermittent_callback(self, state, update_cb, *args, **kwargs):
+        super().on_execution_intermittent_callback(state, update_cb, *args, **kwargs)
+        with self.locked_context("manticore_state", dict) as context:
+            context[state.id].i_am_custom = True
+
+
+class TestCustomIntrospector(unittest.TestCase):
+    def introspect_loop(self, thread):
+        while True:
+            self.history.append(thread.manticore.introspect())
```
feliam, line 79: Can not send an event instead....
manticore.publish("introspect")

tests/other/test_state_introspection.py
```diff
@@ -0,0 +1,94 @@
...
+            sleep(0.5)
+
+    def test_custom_introspector(self):
+        self.history = []
+        m = Manticore(ms_file, stdin_size=17)
+        m.set_instrospection_plugin(MyIntrospector)
+        m.register_daemon(self.introspect_loop)
+        m.run()
+
```
feliam, line 88: ... and instead keep the history in the "Introspector()"
```
class Introspector(Plugin):
     def get_history(self):
          with self.locked_context("blah", list) as l:
              return list(l)

     def at_introspection(state):
          with self.locked_context("blah", list) as l:
              l.append(state.get_status())

intro = Introspector("Specifc configuration")
m = Manticore(ms_file, stdin_size=17, plugins=(intro,))
with Thread(target=myfunc):
    m.run()
print (intro.get_history())
```

ehennenfent:
Apologies for the review spam... Thanks to the new iOS app I've invented the GitHub equivalent of a butt-dial.

ehennenfent:
I want to clear up something because I've realized I never explicitly said it anywhere - The introspection API is _not_ synchronous. The IntrospectionAPIPlugin relies on states to tell it about their current status via the event system. So, when a state is created, moves between state lists, or is destroyed, a callback fires and the plugin updates its internal model of what the state lists currently look like. For things that don't have events (or for events that are fired too frequently to be performant), there's a new `exec_intermittent` event that updates the internal models' view of things like the PC and number of instructions. All of this means that after initialization, the IntrospectionAPIPlugin never needs to touch the state lists directly, so you should be able to call `ManticoreBase.introspect` as much as you want without hurting performance. That said, it also means that "introspection" isn't 100% accurate, since the results returned from the API could lag behind any given state by as much as 1k instructions.

manticore/core/manticore.py
ehennenfent: All the internal data is updated by events, but there's one external method - `get_state_descriptors` - that's called by `ManticoreBase.introspect`. 

This particular function exists so that a user can change the type of `_introspector` before instantiating it - a feature @ekilmer requested. I wanted to avoid adding too many parameters to `__init__`, so I made it a field on the object with a function to change it. Frankly though, I'm starting to think it might be more elegant to bite the bullet and add a parameter. Thoughts?

manticore/core/manticore.py
ehennenfent: To be honest, I didn't think very hard about where to start the daemon threads. In my mind, it doesn't make much sense to have them running until Manticore is also running, but I don't think there's necessarily any problem with starting them in `__init__`.

manticore/core/manticore.py
```diff
@@ -1065,18 +1082,33 @@ def run(self):
...
+            self.register_plugin(self._introspector)
+
         # Lazy process start. At the first run() the workers are not forked.
         # This actually starts the worker procs/threads
         if self.subscribe:
             # User subscription to events is disabled from now on
             self.subscribe = None
 
+        # Passing generators to callbacks is a bit hairy because the first callback would drain it if we didn't
+        # clone the iterator in event.py. We're preserving the old API here, but it's something to avoid in the future.
```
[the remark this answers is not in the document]
ehennenfent: That sounds like it would work, although I worry a bit about user callbacks not using the locked APIs to access items on the Manticore object and causing concurrency problems.

manticore/core/manticore.py
```diff
@@ -1065,18 +1082,33 @@ def run(self):
         # different process modifies the state
         self.stcache = weakref.WeakValueDictionary()
 
+        if self._introspector is None:
+            self._introspector = self._introspection_plugin()
+            self.register_plugin(self._introspector)
```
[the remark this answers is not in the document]
ehennenfent: That's fine with me. It doesn't really make sense to run the introspector before `run` is called (since there aren't any states), but I don't think it should break anything.

manticore/core/manticore.py
```diff
@@ -405,7 +398,9 @@ def goto_snapshot(self):
             raise ManticoreError("No snapshot to go to")
         self.clear_ready_states()
         for state_id in self._snapshot:
+            self._publish("will_enqueue_state", None, can_raise=False)
```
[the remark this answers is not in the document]
ehennenfent: Yep, that checks out. I was unable to catch this event in the plugin and instead had to rely on the `will_run` callback to capture the initial states.

manticore/core/manticore.py
```diff
@@ -373,6 +361,11 @@ def __init__(self, initial_state, workspace_url=None, outputspace_url=None, **kw
 
         # Workers will use manticore __dict__ So lets spawn them last
         self._workers = [self._worker_type(id=i, manticore=self) for i in range(consts.procs)]
+
+        # We won't create the daemons until .run() is called
+        self._daemon_threads = []
```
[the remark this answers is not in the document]
ehennenfent: Frequently, when running Manticore, I find myself calling `run` in one thread, and having another thread doing something else - like intermittently printing instructions/second or communicating with another process. The daemon_threads API formalizes that and makes it easy to incorporate into Manticore without manually constructing an additional thread. 

In this PR, daemon threads are only used for testing. However, I ripped them out of the TUI PR, which uses them to run the socket server that the TUI talks to.

manticore/core/manticore.py
```diff
@@ -365,6 +351,8 @@ def __init__(self, initial_state, workspace_url=None, outputspace_url=None, **kw
         # Note that each callback will run in a worker process and that some
         # careful use of the shared context is needed.
         self.plugins = set()
+        self._introspection_plugin: type = IntrospectionAPIPlugin
+        self._introspector: typing.Optional[IntrospectionAPIPlugin] = None
```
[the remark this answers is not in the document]
ehennenfent: `_introspector` is the IntrospectionAPIPlugin instance. The only reason to separate it out is so that `ManticoreBase.introspect` can reference it [later](https://github.com/trailofbits/manticore/pull/1775/files#diff-29da5f65e4b187c1f150f160571ea968R1202)

manticore/core/manticore.py
```diff
@@ -365,6 +351,8 @@ def __init__(self, initial_state, workspace_url=None, outputspace_url=None, **kw
         # Note that each callback will run in a worker process and that some
         # careful use of the shared context is needed.
         self.plugins = set()
+        self._introspection_plugin: type = IntrospectionAPIPlugin
```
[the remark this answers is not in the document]
ehennenfent: > Maybe .plugins should be a dict[name -> plugin]

This is fine with me. It would simplify the introspection API _and_ the TUI infrastructure. ManticoreEVM access the plugins list directly in a few places, but we can refactor that.

> What IntrospectionAPIPlugin need to do that can not be done via an event via a plain Plugin

It's a normal plugin. It doesn't have any privileged APIs or use the daemon thread interface.

So we're clear though, `_introspection_plugin` is the _type_, not the plugin itself. The plugin instance is constructed later, and lives under `._introspector`

manticore/core/workspace.py
```diff
@@ -658,5 +659,6 @@ def save_constraints(testcase, state: StateBase):
     def save_input_symbols(testcase, state: StateBase):
         with testcase.open_stream("input") as f:
             for symbol in state.input_symbols:
-                buf = SelectedSolver.instance().get_value(state.constraints, symbol)
+                # TODO - constrain=False here, so the extra migration shouldn't cause problems, right?
```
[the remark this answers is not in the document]
ehennenfent: Good to know. Do we want to move it to the native side of things, or is this one of those "it's broken, but it's never used" situations?

pgoodman:
One thing that's not obvious to me: why not give a plugin or client control over the state list? That is, Manticore could implement its own state scheduling policies, but permit that to be extensible/composable, so that a plugin can choose to introspect on everything, though let Manticore's default implementation do its thing (via `super()` for via a proxy class), or go and implement things itself, taking over state scheduling.

manticore/core/manticore.py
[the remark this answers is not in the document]
feliam: So as I understand this the plugins in general will only _run_ from receiving events inside a manticore.run(). An then they can be configured or queried from the script main.

manticore/core/manticore.py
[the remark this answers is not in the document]
feliam: ~Oh right this it is a type. So why do we need to save this type here?~
So we construct it in a later time.

So why can not a normal plugin be created like any other plugin from the main script then added somehow?

manticore/core/manticore.py
```diff
@@ -1139,3 +1178,41 @@ def save_run_data(self):
...
+        self._introspection_plugin = new_plugin
+
+    def introspect(self) -> typing.Dict[int, StateDescriptor]:
+        """
+        Allows callers to view descriptors for each state
+
+        :return: the latest copy of the State Descriptor dict
+        """
+        if self._introspector is not None:
+            return self._introspector.get_state_descriptors()
```
feliam, line 1202: So if we have a way to name the plugins can not this then translate into something like:
```
...
m.plugins['introspector'].get_state_descriptor()
...
```
Also do you imagine that to be called from any place other than the main script?

manticore/core/plugin.py
```diff
@@ -402,3 +408,256 @@ def will_write_register_callback(self, state, register, value):
...
+        with self.locked_context("manticore_state", dict) as context:
+            context[state_id] = StateDescriptor(state_id=state_id, state_list=StateLists.ready)
+
+    def will_run_callback(self, ready_states: typing.Generator):
+        """
+        Called at the beginning of ManticoreBase.run(). Creates a state descriptor for each of the ready states.
+
+        :param ready_states: Generator that allows us to iterate over the ready states (and modify them if necessary)
+        """
+        for state in ready_states:
```
feliam, line 480: As to elude the generator thing (If that is drained and all that) we could remove it from the event signature and get it from
```self.manticore.ready_states```

File: manticore/core/workspace.py (hunk 16 above)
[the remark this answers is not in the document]
Comment (feliam) [reply manticore/core/workspace.py:662]: Not sure where is "input symbols" used. But it is not a generic feature. I would bet it is native only.
I think... all from here(https://github.com/trailofbits/manticore/pull/1775/files#diff-1b639c49c52cd5bbb0440a6cb598b3dbR590) downwards should not be here. 
Workspace trade secrets are not related with What to save from a specific State. I think.

File: manticore/core/manticore.py (hunk 6 above)
Comment (feliam) [reply manticore/core/manticore.py:1104]: Oh..I was confused. I thought for how it looks like that it was put after the worker.start() on purpose. 
Also it happens that a fork will not fork/copy those threads to the forked processes anyway.  :) 

So then we could in fact start the daemon threads from the main script? 
```
m = Manticore()
with SomeDaemonThread():
    m.run()
```
I still need to dig out the reason why this could be a ManticoreBase responsibility.

File: manticore/core/manticore.py (hunk 10 above)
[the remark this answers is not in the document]
Comment (feliam) [reply manticore/core/manticore.py:1096]: "I think everything is properly guarded by a lock"
There. I said it.

File: manticore/core/manticore.py (hunk 13 above)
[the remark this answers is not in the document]
Comment (feliam) [reply manticore/core/manticore.py:366]: ah! here it is that reason.
So it is a convenient thing. Maybe instead of registering an keeping "_daemon_threads" "_daemon_callbacks" 

```
m = Manticore()
with m.Daemon(callback=somefunction):
    m.run()
```
or 
```
m = Manticore()
m.run(daemon_callback=somefunction)
```

manticore/ethereum/state.py
feliam: If I understand what a StateDescriptor is correclty it should show a proyection of what happened in the state?
Instruction count? TX trace length? current PC ?

A PC in EVM would look like: `(state.platform.current_vm.address, state.platform.current_vm.pc)` 
The contract account and the pc inside that. Maybe that's enough for the initial support?

You can print `state.platform.current_vm` that would take a photo of that vm. (slow) ?

manticore/core/manticore.py
ehennenfent: Yes, that's exactly what I intend to translate this to in #1781.

> Also do you imagine that to be called from any place other than the main script?

In the TUI branch, it's called from the server thread, and I use it in several of the daemon thread examples. I don't think it makes sense to call it from a plugin or worker, but I also don't think that would hurt anything. Unless accessing the locked context from a worker could cause a deadlock?

manticore/core/manticore.py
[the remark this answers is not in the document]
ehennenfent: I tweaked the `ManticoreBase` constructor to add the `introspection_plugin_type` parameter, which eliminates the need for the `_introspection_plugin` field here (as well as the helper method that sets it)

manticore/core/manticore.py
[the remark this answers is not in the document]
ehennenfent: From a technical standpoint, it's a normal plugin, so there would be nothing wrong with creating it outside of Manticore and passing it to `register_plugin`. But since it's necessary to drive the introspection API, I've privileged it so that it's always created as part of `__init__`

manticore/core/manticore.py
ehennenfent: > Also it happens that a fork will not fork/copy those threads to the forked processes anyway 

Right, there should only ever be one copy of those threads running in the main process. 

> So then we could in fact start the daemon threads from the main script?

I think the context manager approach could work, but there are a few things to work out. The nice thing about it is that it would make it easy to notify daemon threads that a run had completed. The somewhat difficult aspect is that for things like the TU, which needs to capture log messagesI, we don't want the thread to exit until the entire program exits - not just when `run` completes. So you could still start the thread via a context manager, but it'd have to live on in the background somewhere.

manticore/core/plugin.py
ehennenfent: That would work too, although I'm wary of changing this API since our examples have used `ready_states` as an argument to `will_run_callback` for a long time.

Also, I ended up needing to change the underlying cloning mechanism a bit. Using `itertools.tee` doesn't actually work as intended. While it clones the contents of the generator, the generator is still considered "drained" after the first pass. Since we have code after the `yield`, that code only gets fired once, instead of firing after each plugin has a chance to iterate over the ready states. So, instead, `event.py` now calls `getattr` on the ManticoreBase object so it can get different copies of the `ready_states` generator, instead of just cloning the one.

tests/other/test_state_introspection.py
ehennenfent: We could, but is there a reason to keep the full history around long-term? It's nice to have for testing, but so far I've only thought of use cases for the introspection API where the most recent copy of the state descriptors matters. Would EVM benefit from this feature?

tests/other/test_state_introspection.py
ehennenfent: I'm not sure I'm following entirely. How would we trigger such an event? We'd need to create a `DaemonThread` that's also `Eventful`, right?

feliam: The introspect_loop calls a method on manticore instance on a thread on the main process and keep a list of the results -

I think I was saying that you could _just_ publish an event to manticore via `manticore_instance._publish("instrostpect_event")` instead of calling manticore.introspect(). And then let a plugin register or not to it and do something about it.

I guess I was trying to propose to move instrospect() out of Maticore??

tests/other/test_state_introspection.py
feliam: I am clearly failing to understand something. 

I was thinking about a Plugin that implements whatever introspection and an event called "introspection".
Then the plugin will do whatever at that event (`at_instrospection()`), even keep or not the full history. And have a method to _print out_ the stats/results (`get_history()`).

manticore/core/manticore.py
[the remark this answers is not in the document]
feliam: Nice! I think ALL plugins should be registered at init! And that `register_plugin` should be deprecated.

So the only question I have is why manticore needs an special internal reference to a plugin....

manticore/core/manticore.py
feliam: Need to find out why that server thread or whoever is calling `introspect()` does not have a reference to the introspector plugin already.

ehennenfent:
I think this is ready to go. I updated from `master` so we no longer need to keep `_introspector` on the Manticore object - instead, we store it in the plugin dict. I also updated the EVM update function to capture the current contract address and PC as Felipe mentioned. Merge is just pending an approving review!

ekilmer (approved):
LGTM

PR #1776, opened 2020-07-17
Capture return values in run.sh

ehennenfent:
Switching from YAML to a script means that individual steps can fail, but the tests will still pass if the return value of the entire script is okay.

### DIFF
--- a/manticore/ethereum/verifier.py
+++ b/manticore/ethereum/verifier.py
@@ -411,7 +411,7 @@ def main():
     )
     eth_flags.add_argument(
         "--propre",
-        default == r"crytic_.*",
+        default=r"crytic_.*",
         type=str,
         help="A regular expression for selecting properties",
     )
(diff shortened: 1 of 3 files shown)


ehennenfent:
@feliam It looks like the `NoAliveStates` error was caused by the timeout firing somewhere in between the first and second call to `m.transaction` in `verifier.py`. I've "fixed" it by checking `m.is_killed()` in between and adding a `break`, but I'm not sure that'll stand up as a robust solution. Thoughts?

feliam:
> @feliam It looks like the `NoAliveStates` error was caused by the timeout firing somewhere in between the first and second call to `m.transaction` in `verifier.py`. I've "fixed" it by checking `m.is_killed()` in between and adding a `break`, but I'm not sure that'll stand up as a robust solution. Thoughts?

Yes I think it may be good enough! 
I think that the states that are caught in a timeout/kill in the middle of an exploration are sent to a "killed" state list. 
My current understanding is that it does not matter here because the property was tested before any symbolic transaction.

PR #1777, opened 2020-07-17
Add Manticore native State-specific hooks

ekilmer:
Can now add hooks to specific states instead of _all_ states.

Still need to write tests.


tests/native/test_state.py
```diff
@@ -211,6 +269,52 @@ def testContextSerialization(self):
...
+
+            state.add_hook(None, do_nothing, after=False)
+            state.add_hook(None, do_nothing, after=True)
+            state.add_hook(0x400647, fin, after=True)
+            state.add_hook(0x400647, fin, after=False)
+
+        for state in self.m.ready_states:
+            self.m.add_hook(None, do_nothing, after=True, state=state)
+
+        self.m.run()
```
ehennenfent, line 315: If I'm reading it right, these tests cover whether the hooks can be registered correctly, but not whether they actually fire. Since one of the hooks here uses a `print` statement, perhaps we could wrap this line in a `contextlib.redirect_stdout` and check that "Reached fin callback" is present somewhere?

ekilmer: Ahhh good catch. Could have sworn I intended to do that in the first place. Thank you 🙂

PR #1780, opened 2020-07-29
Linux: Add stat method for FdLike

ekilmer:
This was missing previously but should be supported for anything that has a file descriptor.


manticore/platforms/linux.py
```diff
@@ -279,7 +307,7 @@ def stat(self):
         try:
             return os.fstat(self.fileno())
         except OSError as e:
-            return -e.errno
+            raise FdError(f"Cannot stat: {e.strerror}", e.errno)
```
bradlarsen, line 310: Is this right?  (It might be, but I'm not sure.)

In Manticore, the difference between "return a negative errno" and "throw a Python exception" is tricky.  I believe the negative errno approach will end up modeling the syscall failure in the emulated program, showing up as a syscall that failed.  In contrast, a Python exception, if not handled and converted into a negative errno value, could cause Manticore to blow up.

ekilmer: Good question. It should probably have some tests. Will look into this!

ekilmer: Quick experimenting:
```python
>>> import os
>>> try:
...   os.stat(444)
... except OSError as e:
...   print(e.errno)
...
9
```

is a positive number and 
```
>>> import errno
>>> print(errno.EBADF)
9
```
is positive.

Also, this particular `stat` is not the system call `sys_stat`, it's the `stat` to get the information about our `FdLike` object. In this case it's using the host's information about the object. 

For the system call `sys_stat`, we handle the return or exception thrown by `FdLike.stat()` and return what is expected for the system call, which involves negating the `FdError.err` here to return the correct value
https://github.com/trailofbits/manticore/blob/bd569944b12a890c35ff1d1e7175296c1c0d46d1/manticore/platforms/linux.py#L2806-L2810
and
https://github.com/trailofbits/manticore/blob/bd569944b12a890c35ff1d1e7175296c1c0d46d1/manticore/platforms/linux.py#L2852-L2855
and
https://github.com/trailofbits/manticore/blob/bd569944b12a890c35ff1d1e7175296c1c0d46d1/manticore/platforms/linux.py#L2894-L2898
and
https://github.com/trailofbits/manticore/blob/bd569944b12a890c35ff1d1e7175296c1c0d46d1/manticore/platforms/linux.py#L3074-L3077

That being said, I think what is here is the correct way of doing things.

I am a little confused about this comment though:

> a Python exception, if not handled and converted into a negative errno value, could cause Manticore to blow up

Could you describe a little bit more about how this applied to the changes made in this PR?

ekilmer: Ahhh, I see now. Thank _you_ for the explanation.

Definitely learning a lot about the intricacies and subtleties of all of this, so this is good to know!

I'm wondering if there is an _exception-handling_ linter or analysis we can run on Manticore to determine which and where exceptions can bubble to the top (i.e. not caught anywhere). Control-flow with exceptions is not obvious to me all the time

manticore/platforms/linux.py
```diff
@@ -375,6 +403,12 @@ def close(self):
         except OSError as e:
             return -e.errno
 
+    def stat(self):
```
ehennenfent, line 406:
```suggestion
    def stat(self) -> os.stat_result:
```

manticore/platforms/linux.py
ehennenfent: :+1: Thanks both of you for taking the time to track this down. I wasn't sure what the correct behavior was when I saw this line in the diff, but was happy to see that you'd already hashed it out.

PR #1781, opened 2020-07-30
Convert plugin list to dict

ehennenfent:
Replaces the `plugins` list on the ManticoreBase object with a dict, keyed with a unique name derived from the plugin's class and ID.


manticore/core/plugin.py
```diff
@@ -55,9 +55,13 @@ def is_enabled(self):
             return context.get(self._enabled_key, True)
 
     @property
-    def name(self):
+    def name(self) -> str:
         return str(self.__class__)
 
+    @property
+    def unique_name(self) -> str:
```
feliam, line 62: I wonder what happens if 2 instances of the same plugin class are registered into manticore.
Regarding the use of the manticore context/s.

ehennenfent: It looks like the plugin contexts are keyed off of `self._plugin_context_name`, which is set to `f"{classname}_context_{hex(hash(self))}"` at initialization. `__hash__` appears to [factor in the ID](https://stackoverflow.com/a/11324771) of the object, so as long as the two instances are really two instances and not the same instance getting registered twice, this shouldn't cause any problems. Admittedly, I can't think of a reason you'd want to register the same plugin twice, but it would technically work.

PR #1782, opened 2020-07-30
Use coveralls

ekilmer:


ekilmer:
Opened by mistake. Needs some testing first.

PR #1783, opened 2020-07-30
Use Coveralls for coverage

ekilmer:
Codecov has been extremely frustrating lately with failing to upload the results of our tests.

Since our tests take so long already, we don't want to have to re-run them multiple times to get differential coverage data

- [x] We need to allow https://coveralls.io/ to access this repo

- [x] We will need to add a Github secret token `COVERALLS_REPO_TOKEN ` related to this repo to enable it

- [x] The settings for failing a PR for bad coverage is only managed in the web interface (pic below)
Screen Shot 2020-07-30 at 4 23 21 PM


.github/workflows/ci.yml
```diff
@@ -61,22 +61,30 @@ jobs:
...
         pip install -e ".[dev-noks]"
     - name: Run Tests
       env:
         TEST_TYPE: ${{ matrix.type }}
       run: |
         cp scripts/run_tests.sh .
         ./run_tests.sh
-    - name: Coverage Upload
-      uses: codecov/codecov-action@v1
+    - name: Coveralls Parallel
```
ehennenfent, line 72: Is it necessary to manually install and run the CLI tool? Or can we use the GH action as shown in the example here: https://github.com/marketplace/actions/coveralls-github-action#complete-parallel-job-example

ekilmer: Good question. I couldn't get it to work since it expects a `.lcov` file.

Here's an open issue if you can find a better workaround https://github.com/coverallsapp/github-action/issues/30

ekilmer:
I set the decrease threshold to 0.2%, but I left the total threshold blank
Screen Shot 2020-07-31 at 4 45 31 PM

ekilmer:
Closing in favor of #1784

PR #1784, opened 2020-07-31
Coveralls Take 2

ekilmer:
This branch needs to be on the main repo.

See https://github.com/trailofbits/manticore/pull/1783 for more details

If this is merged, it will invalidate #1778


ehennenfent:
Per discussion on the previous PR, I'd prefer a solution where we don't have to manually install coveralls, but as long as their GH action isn't working for us, this looks good to me.

PR #1785, opened 2020-08-03
Use default handler for symbolic system call arguments

ekilmer:
This will only be called when there is an implementation of the system
call in the concrete Linux class and no implementation in the symbolic
SLinux.

The action is simply to concretize any found symbolic system call
arguments.

This should hopefully ease the use of Manticore when it encounters
symbolic arguments to system calls that don't have specially
implemented symbolic handling. A debug message is printed when this happens.


manticore/platforms/linux.py
```diff
@@ -2577,13 +2578,21 @@ def syscall(self):
...
         try:
             table = getattr(linux_syscalls, self.current.machine)
             name = table.get(index, None)
             if hasattr(self, name):
                 implementation = getattr(self, name)
+                # If this instance does not have an implementation for the system
+                # call, but the parent class does, use a partial function to do
+                # some processing of the unimplemented syscall before using the
+                # parent's implementation
+                owner_class = implementation.__qualname__[: -len(implementation.__name__) - 1]
```
ehennenfent, line 2593: I know it's only used on `Linux` and `SLinux`, but this parsing scares me a bit. Would checking `name in inspect.getmembers(self, inspect.is_method)` and `name in inspect.getmembers(super(), inspect.is_method)` work instead?

ekilmer: The problem with using `inspect` and `getmembers` is that it returns all of the parent's class members as well, so we still have to filter them somehow.

Screen Shot 2020-08-05 at 9 29 23 AM

I think it's difficult because we're checking for the _absence_ of a method in the derived class and it seems difficult to get the defining class for `implementation`, which we could then compare against `self.__class__`. 

After some searching, I found this stack overflow answer to get the class that defined a method, which I think would work for our situation https://stackoverflow.com/a/25959545. It uses `inspect` to try some things, but then _it also_ falls back to using `__qualname__` parsing (in a more robust way).

It would change this line to be
```python
owner_class = get_class_that_defined_method(implementation)
```

Please let me know how to proceed.

ehennenfent: Ugh, I was hoping using `inspect` would allow for a cleaner way of doing this, but that example doesn't seem to be the case. Since this is only going to be used, I'd say it's fine to continue using the string parsing approach - but maybe use `.replace` or `.split` instead of directly indexing it.

PR #1786, opened 2020-08-03
Symbolic-length reads from symbolic sockets

ekilmer:
A `read` doesn't have to always return the number of bytes provided in the `size` argument. This is especially true for network sockets that may receive only partial bytes of data at a time.

This PR adds a feature to allow Manticore to fork when a `receive` is called on a SymbolicSocket and to more precisely emulate network data transmissions that may be sent in chunks less than what is requested.

Currently, we choose symbolic values from `[1, remaining_bytes]` to avoid too much state explosion.

- [x] Check coverage and add an example binary program


ekilmer:
The uncovered paths seem acceptable to me, but I can spend more time writing tests if anyone thinks it's important enough to do so.

manticore/platforms/linux.py
```diff
@@ -692,14 +692,23 @@ def __init__(
...
         state["inputs_recvd"] = self.inputs_recvd
         state["symb_name"] = self.symb_name
         state["recv_pos"] = self.recv_pos
         state["max_recv_symbolic"] = self.max_recv_symbolic
-        state["constraints"] = self._constraints
+        state["_symb_len"] = self._symb_len
+        state["fd"] = self.fd
         return state
 
     def __setstate__(self, state):
```
sschriner, line 714:
```suggestion
    def __setstate__(self, state: State):
```

manticore/platforms/linux.py
```diff
@@ -692,14 +692,23 @@ def __init__(
...
+        # This is a meta-variable, of sorts, and it is responsible for
+        # determining the symbolic length of the array during recv/read.
+        # Initially, it is None, to indicate we haven't forked yet. After
+        # fork, each state will be assigned their respective _actual_,
+        # concretized, receive length
+        self._symb_len: Optional[int] = None
+        # Set after adding this socket to the file descriptor list
+        self.fd: Optional[int] = None
 
     def __getstate__(self):
```
sschriner, line 704:
```suggestion
    def __getstate__(self): -> State
```

manticore/platforms/linux.py
```diff
@@ -3316,34 +3363,34 @@ def sys_recv(self, sockfd, buf, count, flags, trace_str="_recv"):
             logger.debug("Submitted a symbolic flags")
             raise ConcretizeArgument(self, 3)
 
-        return super().sys_recv(sockfd, buf, count, flags)
+        return self.sys_recvfrom(sockfd, buf, count, flags, 0, 0, trace_str)
 
-    def sys_recvfrom(self, sockfd, buf, count, flags, src_addr, addrlen):
+    def sys_recvfrom(self, sockfd, buf, count, flags, src_addr, addrlen, trace_str="_recvfrom"):
```
sschriner, line 3368:
```suggestion
    def sys_recvfrom(self, sockfd: Union[int, BitVec], buf: Union[int, BitVec], count: Union[int, BitVec], flags: Union[int, BitVec], src_addr: Union[int, BitVec], addrlen: Union[int, BitVec], trace_str: str ="_recvfrom"):
```

manticore/platforms/linux.py
ehennenfent: It's confusingly named, but `state` here (and in `__setstate__`) is actually a dict. `__getstate__` and `__setstate__` are pickle helpers, not something that's unique to Manticore, so the terminology collides a bit.

sschriner: Ahh ok. Good to know. These type hints are unnecessary then

manticore/platforms/linux.py
```diff
@@ -692,14 +692,23 @@ def __init__(
...
+        # Set after adding this socket to the file descriptor list
+        self.fd: Optional[int] = None
 
     def __getstate__(self):
         state = super().__getstate__()
         state["inputs_recvd"] = self.inputs_recvd
         state["symb_name"] = self.symb_name
         state["recv_pos"] = self.recv_pos
         state["max_recv_symbolic"] = self.max_recv_symbolic
-        state["constraints"] = self._constraints
```
ehennenfent, line 702: Why'd you remove the constraint set from the pickled state? It seems to me like you'd want it to persist.

ekilmer: Good question. I believe this should be taken care of during the `SLinux` `__setstate__` here https://github.com/trailofbits/manticore/blob/898918554998c47d624d8df28ff8d806bd9f49ce/manticore/platforms/linux.py#L3278-L3282

We also initialize a SymbolicSocket with a reference to the `ConstraintSet` in `SLinux` here https://github.com/trailofbits/manticore/blob/898918554998c47d624d8df28ff8d806bd9f49ce/manticore/platforms/linux.py#L3413

I'm open to suggestions for redesigning how/where the constraints are added. However, without removing this line, I think we are duplicating serialization of the constraint set, right?

feliam: The problem with keeping links to the constraintset is forking.

This all needs a refactor/simplification. But currently we fork using a context manager: 
https://github.com/trailofbits/manticore/blob/898918554998c47d624d8df28ff8d806bd9f49ce/manticore/core/manticore.py#L504-L515

Note that there the State is spawned into a child state using the width statement. Supposedly modifying anything of the child state would not affect the parent. (This is a lie except for constraints)

Note that the child constraints reference/references are updated when spawning a child from a parent:
https://github.com/trailofbits/manticore/blob/898918554998c47d624d8df28ff8d806bd9f49ce/manticore/core/state.py#L150

If you keep a new reference to a constraintset it should be updated there.

ekilmer: @feliam Can you check out this commit https://github.com/trailofbits/manticore/pull/1786/commits/aeb4a3100be2c1e7e7aafb3b8f516abd559cb371 if it would work as a fix for this? Thank you!

feliam: Yes it looks like the new reference to the constraints are correctly propagated with that.

ekilmer: @ehennenfent Anything else you're wondering about this after the latest commits?

ehennenfent: Nope, looks good to me! All this constraint set stuff is tricky...

PR #1791, opened 2020-08-14
Add partial implementation of sendto syscall

ekilmer:
This partial implementation does not handle dest_addr and addrlen
arguments, which means this sendto acts like a regular send syscall


manticore/platforms/linux.py
```diff
@@ -2563,18 +2571,35 @@ def sys_recvfrom(
...
             return -errno.ENOTSOCK
 
-        data = self.current.read_bytes(buf, count)
+        try:
+            data = self.current.read_bytes(buf, count)
+        except InvalidMemoryAccess:
+            logger.info("SEND: buf within invalid memory. Returning EFAULT")
+            return -errno.EFAULT
+
         # XXX(yan): send(2) is currently a nop; we don't communicate yet
```
ehennenfent, line 2601: Let's remove this comment and add a docstring saying roughly the same thing

PR #1792, opened 2020-08-14
Fix BitVec with symbolic offset and fix TranslatorSmtlib.unique thread safety

geohot:
The EVM BYTE opcode with a symbolic offset argument needs the second handler or it asserts (first obviously on the offset == 0)

And I was seeing name collisions on the SMT vars, first I disabled threading to fix, but then added a lock around the unique variable.

### DIFF
--- a/manticore/core/smtlib/operators.py
+++ b/manticore/core/smtlib/operators.py
@@ -124,7 +124,7 @@ def ULE(a, b):
 
 
 def EXTRACT(x, offset, size):
-    if isinstance(x, BitVec):
+    if isinstance(x, BitVec) and not isinstance(offset, BitVec):
         if offset == 0 and size == x.size:
             return x
         return BitVecExtract(x, offset, size)
--- a/manticore/core/smtlib/operators.py
+++ b/manticore/core/smtlib/operators.py
@@ -124,7 +124,9 @@ def ULE(a, b):
 
 
 def EXTRACT(x, offset, size):
-    if isinstance(x, BitVec) and not isinstance(offset, BitVec):
+    if isinstance(x, BitVec) and isinstance(offset, BitVec):
+        return BitVecExtract(x >> offset, 0, size)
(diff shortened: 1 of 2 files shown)


manticore/core/smtlib/operators.py
```diff
@@ -124,7 +124,7 @@ def ULE(a, b):
 
 
 def EXTRACT(x, offset, size):
-    if isinstance(x, BitVec):
+    if isinstance(x, BitVec) and not isinstance(offset, BitVec):
```
feliam, line 127: This is clearly better!
But now I'm thinking that if `offset` happen to be a BitVec then the else will fail to generate a BitVec of `size` bits.

geohot: Hmm, fair point. Could call BitVecExtract on the expression with offset 0 and size size?

feliam: Hmm I do not really know if the EXTRACT symbol actually supports symbolic size. 
Maybe Operators.EXTRACT needs to force/ request a concrete size. And any user of it must concretize the size before doing the EXTRACT. ... .

geohot: Oh I don't think size can be symbolic, that will fail. But with the new change, offset should be able to.

feliam: Right! Even from the evm.BYTE size if safely concrete.  nvm that.  
https://github.com/trailofbits/manticore/blob/402de84c0686b892c7ef987882f330cad565760e/manticore/platforms/evm.py#L1603-L1606

geohot: Yea, that's the one that was failing with dynamic offset, but size is always 8.

feliam:
Thank you for this. The CI is complaining about not being able to upload coverage info but it should be fixed in master.  I'm merging this. 

Please keep them coming !

PR #1793, opened 2020-08-15
Allow Coveralls Upload to Fail

ehennenfent:
External PRs don't have access to the Coveralls token, and so can't upload coverage reports on our behalf. See: https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
         ./run_tests.sh
     - name: Coveralls Parallel
       run: |
-        coveralls
+        coveralls || true
       env:
         COVERALLS_PARALLEL: true
         COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}


.github/workflows/ci.yml
```diff
@@ -71,7 +71,7 @@ jobs:
         ./run_tests.sh
     - name: Coveralls Parallel
       run: |
-        coveralls
+        coveralls || true
```
feliam, line 74: Maybe allow it to fail just if  "GITHUB_TOKEN is not set" ??

ekilmer: Hopefully #1794 takes care of this.

`GITHUB_TOKEN` is apparently the only secret available to pull requests in a forked repo.

https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow
https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token

PR #1795, opened 2020-08-18
Improve crytic-compile support

montyly:
- crytic-compile: use latest release
- manticore-verifier: enable crytic-compile cli flags

The crytic-compile version currently used by Manticore is out of date.

Additionally, `manticore-verifier` did not use the crytic-compile cli flags

To test this:
```
$ npx truffle unbox metacoin
```
Update `contracts/MetaCoin.sol` and add
```solidity
    function crytic_test() public returns(bool){
        return balances[msg.sender] > 1;
    }
```
Then
```
$ manticore-verifier . --contract MetaCoin
$ rm truffle-config.js 
$ manticore-verifier . --contract MetaCoin --compile-force-framework Truffle --ignore-compile
```

The last command re-run Manticore, but fails without this PR

### DIFF
--- a/manticore/ethereum/manticore.py
+++ b/manticore/ethereum/manticore.py
@@ -336,7 +336,10 @@ class ManticoreEVM(ManticoreBase):
         if isinstance(source_code, io.IOBase):
             source_code = source_code.name
 
-        if isinstance(source_code, str) and not is_supported(source_code):
+        if (isinstance(source_code, str) and
+                not is_supported(source_code) and
+                # Until https://github.com/crytic/crytic-compile/issues/103 is implemented
+                "ignore_compile" not in crytic_compile_args):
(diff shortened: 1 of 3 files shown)


ggrieco-tob:
I tested this PR and seems to be working fine! :tada: However, the flag names and cli help could be improved:

```
                          ...
                          [--propconfig PROPCONFIG] [--quick-mode]
                          [--contract_name CONTRACT_NAME] [--maxfail MAXFAIL]
                          ...
```

Can we use `--contract` instead of `--contract_name` and `--config` instead of `--propconfig`?

montyly:
It's a good idea, but these flags are not related to crytic-compile; so we should probably move this recommendation to a separate issue/PR

ehennenfent:
This has been approved w/o changes for 7 days, so I'm going to merge it. Thanks for the fix!

PR #1796, opened 2020-08-19
Disable write back once we hit our stop in Unicorn emulation

ekilmer:


ehennenfent:
I've made this exact patch before, in situations that looked like the following:
1. Emulate until address a
2. At address a, fire a hook that does <something>
3. Emulate from address a to address b
4. Carry on...

What happens in a situation like this?: 
1. Emulate until address a
2. Execute a few instructions in Manticore, that write to registers and memory
3. Emulate until address b  

Does step 3 break because the updated state wasn't copied back into the emulator? I haven't tested, but think it might.

ekilmer:
Things to look into before merging:

The test program added in this PR (`rusticorn`) still takes at least 8 minutes to run under Unicorn on a fast MBP, and around 20 minutes in CI.

The long runtime occurs _within_ Unicorn control, with no syscalls talking back to Manticore, _however_ we still have callbacks from Unicorn into Manticore that are not syscalls. In particular, we have callbacks registered for the following Unicorn hooks: https://github.com/trailofbits/manticore/blob/702260c78f1a36286033b323c5e9377787c31243/manticore/utils/emulate.py#L112-L117

manticore/native/cpu/abstractcpu.py
```diff
@@ -716,7 +719,7 @@ def _raw_read(self, where: int, size=1) -> bytes:
         elif isinstance(map, AnonMap):
             data = bytes(map._data[start : start + size])
         else:
-            data = b"".join(self.memory[where : where + size])
+            data = b"".join(self.memory.read(where, size, force=force))
```
ehennenfent, line 722: For future reference: `memory.__getitem__` is just a wrapper around `memory.read`, so this change was made for the sake of getting access to the `force` parameter.

manticore/utils/emulate.py
```diff
@@ -99,88 +119,140 @@ def __init__(self, cpu):
...
+    def copy_memory(self, address: int, size: int):
         """
         Copy the bytes from address to address+size into Unicorn
         Used primarily for copying memory maps
         :param address: start of buffer to copy
         :param size: How many bytes to copy
         """
         start_time = time.time()
-        map_bytes = self._cpu._raw_read(address, size)
+        map_bytes = self._cpu._raw_read(address, size, force=True)
```
ehennenfent, line 138: Was the lack of `force=True` causing a soundness problem? It seems like it shouldn't prevent `rusticorn` from executing correctly because the old behavior would have been to raise `InvalidMemoryAccess`. Unless this is an older change meant to support one of the CHESS programs?

ekilmer: > Was the lack of force=True causing a soundness problem?

Yes. I was seeing the program change some memory maps to zero permissions, which would cause Manticore to throw `InvalidMemoryAccess` when reading from those maps. Since this is an internal implementation detail to synchronize our emulator memory, I think `force=True` is warranted here.

manticore/native/cpu/abstractcpu.py
ekilmer: Yes, that is correct.

ehennenfent (approved):

Issue #1798, opened 2020-08-22
Account Existence check


wolflo:
### Summary of the problem
Account existence is inconsistent with [EIP-161](https://eips.ethereum.org/EIPS/eip-161). Accounts can remain _existent-but-empty_, which should not be the case.

### Manticore version
0.3.4

### Python version
3.8.5

### Step to reproduce the behavior
The following test shows the issue as it relates to calculation of the gas cost for a self-destruct instruction. Account `0xffffffffffffffffffffffffffffffffffffffff` is _touched_ in the first transaction, but should remain non-existent because it has no code, zero balance, and zero nonce. The self-destruct in the second transaction should charge `21000 + 3 + 5000 + 25000 - 24000`, but instead charges `21000 + 3 + 5000 - min(gas_used / 2, 24000) = 13002` because the recipient account is considered as already existing.
```
def test_acct_existence(self):
    with disposable_mevm() as m:
        asm_acc = """   PUSH20 0xffffffffffffffffffffffffffffffffffffffff
                        SELFDESTRUCT
                  """

        # 21000 base tx + 3 push op + 5000 base selfdestruct + 25000 new acct cost - 24000 refund
        expected_gas = 27003

        m.create_account(
            address=0x111111111111111111111111111111111111111,
            code=EVMAsm.assemble(asm_acc),
            balance=1
        )
        m.create_account(
            address=0x222222222222222222222222222222222222222,
            balance=1000000000000000000
        )

        # this should not result in creation of the acct, but does
        m.transaction(
            caller=0x222222222222222222222222222222222222222,
            address=0xffffffffffffffffffffffffffffffffffffffff,
            data=b'',
            value=0
        )
        # this should result in creation of the acct, but it's already created
        m.transaction(
            caller=0x222222222222222222222222222222222222222,
            address=0x111111111111111111111111111111111111111,
            data=b'',
            value=0
        )
        self.assertEqual(m.count_ready_states(), 1)
        state = next(m.ready_states)
        txs = state.platform.transactions

        # AssertionError: 13002 != 27003
        self.assertEqual(txs[-1].used_gas, expected_gas)
```

### Expected behavior
An account that is touched but remains empty should be considered non-existent.

### Actual behavior
Touched accounts are considered existing regardless of code, nonce, and balance.

PR #1801, opened 2020-09-01
Fix EVM account existence checks for selfdestruct and call

wolflo:
Addresses #1798 and adds some tests for selfdestruct gas costs.

The changes to `test_call_gas` make sure that it covers the account existence check.


feliam (approved):
This is great! LGTM!

PR #1805, opened 2020-09-25
Heap Tracker

sschriner:


manticore/native/heap_tracking/hook_malloc_library.py
```diff
@@ -0,0 +1,236 @@
...
+    initial_state.context['mmap'] = mmap
+    initial_state.context['munmap'] = munmap
+
+
+def hook_mmap_return(state: State):
+    """ Hook to process munmap information and add a function hook to the callsite of munmap (which should 
+    be inside malloc or another function inside of malloc which calls munmap), post execution of the
+    munmap call.
+    """
+    ret_val = state.cpu.read_register("RAX")
```
ekilmer, line 64: This register should be more generic, if possible

manticore/native/heap_tracking/hook_malloc_library.py
```diff
@@ -0,0 +1,236 @@
...
+    logger.info(f"mmap ret val: {hex(ret_val)}")
+
+    state.context["malloc_lib"].process_mmap(ret_val, state.context["mmap_args"])
+    del state.context["mmap_args"]
+
+    logger.debug(f"Unhooking mmap return in malloc in state: {state.id}")
+    state.remove_hook(state.cpu.read_register("RIP"), hook_mmap_return)
```
ekilmer, line 71: I think there might be a generic `PC` register?

ekilmer (commented):
After some tests are added, I'll review it a bit more. Looking good so far!

I made a few comments on the use of specific register names, which is fine for now, but we should either try to disallow inclusion of this functionality for anything except x86-64 or make the registers more generic to work on the supported architectures

manticore/native/heap_tracking/hook_malloc_library.py
sschriner: I think there's a function that's used in the invoke_model code that I can use to replace some of the X86 specific reg names here

SonyViao (approved):

ekilmer (commented):
Looking good! Just a few comments

manticore/native/heap_tracking/hook_malloc_library.py
```diff
@@ -0,0 +1,376 @@
...
+MMAP_SYS_NUM: int
+MUNMAP_SYS_NUM: int
+
+
+def load_ret_addr(state: State) -> int:
+    """Loads the return address of a function from the stack
+    (Assuming the next instruction to be executed is the start of a function call)
+    """
+    stack_location = state.cpu.read_register("STACK")
+    ret_addr = state.cpu.read_int(stack_location, state.cpu.address_bit_size)
```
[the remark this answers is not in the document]
pgoodman: Anvill encodes return address info :-) These encode the location of the return address on entry to a function, i.e. just after a `call` instruction has been executed and brought us to the first instruction of a function.

[amd64](https://github.com/lifting-bits/anvill/blob/master/anvill/python/anvill/arch.py#L327-L328)
[x86](https://github.com/lifting-bits/anvill/blob/master/anvill/python/anvill/arch.py#L563-L564)
[aarch64](https://github.com/lifting-bits/anvill/blob/master/anvill/python/anvill/arch.py#L936-L937)

Manticore doesn't do sparc32/64, but does to armv7. My guess is it uses a link pointer register as well.

PR #1806, opened 2020-09-30
Use the PTable Python module to implement prettytable

khorben:
The original prettytable Python module was hosted on Google Code, and is
no longer maintained. PTable is a fork of prettytable, as found at
https://pypi.org/project/PTable/.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,7 @@ setup(
     install_requires=[
         "pyyaml",
         "pysha3",
-        "prettytable",
+        "PTable",
         "ply",
         "rlp",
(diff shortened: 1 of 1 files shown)


ehennenfent:
Thanks for the PR! As it happens, over the past few days, `prettytable` seems to have risen from the dead. A new release was pushed just yesterday - the first since 2013. Since it seems to be a bit [more active](https://github.com/jazzband/prettytable/pulse/monthly) than [PTable](https://github.com/kxxoling/PTable/pulse/monthly), we're going to stick with it for now unless we run into problems.

PR #1811, opened 2020-10-06
native/cpu/x86: Add support for CPUID EAX=80000000h

woodruffw:
`cpuid` with `EAX=80000000h` is meant to return the highest "extended function" ID back via `EAX`.

Both Intel's and AMD's docs are a little oblique when it comes to what exactly an "extended function" ID *is*, but it *seems* to just be another `EAX` value (i.e., `CPUID` "leaf") with the high bit (i.e. `EAX=8xxxxxxxh`) set.

Longer term, we should probably support some of these higher extended functions/leaves. In particular, it looks like recent versions of the Linux VDSO check for `CPUID EAX=80000001h` and then check if `EDX.RDTSCP[bit 27]` is high in order to use `rdtscp` (in lieu of `rdtsc` + a serializing hack?).

### DIFF
--- a/manticore/native/cpu/x86.py
+++ b/manticore/native/cpu/x86.py
@@ -961,6 +961,10 @@ class X86Cpu(Cpu):
                 0x0: (0x00000000, 0x00000000, 0x00000000, 0x00000000),
                 0x1: (0x00000000, 0x00000000, 0x00000000, 0x00000000),
             },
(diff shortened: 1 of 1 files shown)


ekilmer (approved):
This runs fine in this repo's tests and on our other binaries. Thank you!

woodruffw:
Awesome. I don't have merge perms, so I'll leave that to you or whoever else feels like it 😉

PR #1816, opened 2020-10-30
Ignore Gas Calculations by Default

ehennenfent:
As I understand it, this is necessary to get the [building secure contracts](https://github.com/crytic/building-secure-contracts) repo passing by default. It's also a frequent first step for real-world EVM analysis, so perhaps it will save some time.

The tests are performing suspiciously well locally. I would have expected some changes to be required. Let's see how GH Actions handles them.


feliam (approved):
refreshing diff

PR #1819, opened 2020-11-04
Support the examples from the building-secure-contracts repository

ehennenfent:
Previously, contract creation would ignore the `consts.oog` variable and always check that the caller had the requisite balance. This change checks that we haven't explicitly ignored gas calculations when evaluating whether the transaction has enough balance to proceed.

### DIFF
--- a/tests/ethereum/test_general.py
+++ b/tests/ethereum/test_general.py
@@ -45,6 +45,8 @@ solver = Z3Solver.instance()
 
 THIS_DIR = os.path.dirname(os.path.abspath(__file__))
 
+consts = config.get_group("evm")
+
 
 @contextmanager
 def disposable_mevm(*args, **kwargs):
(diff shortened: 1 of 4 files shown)


tests/ethereum/test_general.py
```diff
@@ -1774,6 +1778,7 @@ def test_gas_check(self):
         except TerminateState as e:
             result = str(e)
         self.assertEqual(result, "SELFDESTRUCT")
+        consts.oog = default_gas
```
feliam, line 1781: This is possible going to fail if the tests get executed concurrently. Config must go with mcore instance and be frozen and etc etc.
Also there this 
https://github.com/trailofbits/manticore/blob/c8e5e25d8e10c6235bd18fec2d3d44b56383c58a/manticore/utils/config.py#L73-L80

feliam (approved):
We need better configuration nobs all around evm.

And also config need to be able to be frzen and to be attached to each mcore instance

ehennenfent:
The building-secure-contracts repo works with #1820, so we don't need the additional gas and timeout changes present in this PR.

PR #1820, opened 2020-11-06
Detect default solver

montyly:
Try yices, if not present use z3


feliam (approved):
We'll need to change the readme's ". By default Manticore will use Z3"

PR #1823, opened 2020-12-07
Better control of gas accounting

### LINKED ISSUE #1789: Add configuration constant to ignore any gas related calculations.
User wants a way to configure manticore to avoid any cost related to calculating gas values.

The gas value must be completely ignored everywhere it may impose any slowdown. For example:
* Human transactions: sending transactions should not charge any gas fro the caller, nor decrement its balance
* Transaction instructions: CALL, DELEGATECALL, ..
* Returning from transactions or deleting storage values should not refund any gas
* Storage related instructions: SSTORE/SLOAD should not charge gas dinamically
* Memory is unbounded. No gas related limits.

feliam:
Add evm.gas configuration constant/variable/thing to control how to account gas. For example "static" will ignore all dynamic calculated fees.  (This is different from evm.oog. The latter controls how and when to check for Out Of Gas)

Change rpl -> rust-rpl dependency.


manticore/platforms/evm.py
```diff
@@ -67,6 +67,18 @@ def globalfakesha3(data):
     return None
 
 
+consts.add(
+    "gas",
+    default="static",
+    description=(
+        "Control how to keep the gas count"
+        "insane: Keep static and dynamic gas including transaction fee VERY EXPENSIVE (INSANE)."
```
ehennenfent, line 75: Nit, but can we make this "exhaustive"? Or, if you want to preserve the suggestion that it's over-the-top, perhaps "ludicrous"?

manticore/platforms/evm.py
```diff
@@ -2164,14 +2181,15 @@ def CALL_gas(self, wanted_gas, address, value, in_offset, in_size, out_offset, o
...
 
-        exception = False
         available_gas = self._gas
         available_gas -= fee
 
         exception = Operators.OR(
             Operators.UGT(fee, self._gas),
             Operators.ULT(self.safe_mul(available_gas, 63), available_gas),
         )
+        self.fail_if(exception)
```
ehennenfent, line 2191: Does this line need to check the value of `consts.gas` before deciding to raise the exception?

SonyViao (approved):

ehennenfent:
@feliam were you still hoping to merge this? I think those couple of nits are the only things blocking it.

PR #1824, opened 2020-12-07
Bump black and mypy

feliam:
Bump black version to 20.8b1
Bump mypy to 0.790


ekilmer (approved):
Looks like `black` now touches comments! Luckily it looks like the formatting is sane and only leading/trailing spaces are touched.

PR #2386, opened 2021-01-12
ekilmer fixes for dev expressions

ekilmer:
Making this as a separate branch/PR in case @feliam has uncommitted changes.


manticore/native/cpu/x86.py
```diff
@@ -2079,7 +2079,7 @@ def NEG(cpu, dest):
         :param dest: destination operand.
         """
         source = dest.read()
-        res = dest.write(-source)
+        res = dest.write((~source) + 1)
```
ekilmer, line 2082: I'd like for someone to confirm that this change is necessary/correct.

It seems very odd that this change is required to get tests to pass 🤔

ekilmer:
The only interesting parts of this PR are probably these two commits: https://github.com/trailofbits/manticore/pull/2386/commits/6a9f392fb49cb9ae1e2f4d5eff9631454a1aaab2 https://github.com/trailofbits/manticore/pull/2386/commits/1cb97beda07a89edef21e9213eba461ddbcb4e0f

PR #2387, opened 2021-01-12
Fix coveralls upload

ekilmer:
New version broke us.

Reference: https://github.com/coveralls-clients/coveralls-python/issues/251


ehennenfent (approved):
Looks great, thanks for fixing!

PR #2389, opened 2021-01-28
Syscall specific hooks

sschriner:


manticore/platforms/linux.py
```diff
@@ -2728,7 +2728,10 @@ def syscall(self):
             else:
                 raise EnvironmentError(f"Bad syscall index, {index}")
 
-        return self._syscall_abi.invoke(implementation)
+        self._syscall_abi._cpu._publish("will_invoke_syscall", index)
```
ekilmer, line 2731: I think this might need to be moved higher up somewhere before the exception handling in case base manticore doesn't have the implementation but someone still wants to hook on it.

manticore/native/manticore.py
```diff
@@ -215,54 +222,87 @@ def init(self, f):
...
         after: bool = False,
         state: Optional[State] = None,
+        syscall: bool = False,
     ):
         """
-        Add a callback to be invoked on executing a program counter. Pass `None`
-        for pc to invoke callback on every instruction. `callback` should be a callable
-        that takes one :class:`~manticore.core.state.State` argument.
+        Add a callback to be invoked on executing a program counter (or syscall). Pass `None`
+        for pc_or_sys to invoke callback on every instruction (or syscall). `callback` should
```
woodruffw, line 250: Ultra nitpick: Surround `pc_or_sys` with backticks for docgen.

woodruffw (approved):
Left one small nitpick, but otherwise LGTM from a quick pass. Someone actually on the Manticore maintenance team should do a final review, though 🙂

manticore/native/manticore.py
```diff
@@ -215,54 +222,87 @@ def init(self, f):
             self.subscribe("will_run", self._init_callback)
         return f
 
-    def hook(self, pc, after=False):
+    def hook(self, pc, after=False, syscall=False):
```
ekilmer, line 225: While we're here, can we add type hints to this? 🙂

manticore/native/manticore.py
```diff
@@ -215,54 +222,87 @@ def init(self, f):
...
+                hooks, when, hook_callback = (
+                    (self._sys_hooks, "will_invoke_syscall", self._sys_hook_callback)
+                    if not after
+                    else (
+                        self._sys_after_hooks,
+                        "did_invoke_syscall",
+                        self._sys_after_hook_callback,
+                    )
+                )
+                hooks.setdefault(pc_or_sys, set()).add(callback)
```
ekilmer, line 300: Nit: This line is repeated on line 289

manticore/platforms/linux.py
```diff
@@ -2909,13 +2909,15 @@ def execute(self):
...
+            index: int = self._syscall_abi.syscall_number()
+            self._syscall_abi._cpu._publish("will_invoke_syscall", index)
             try:
                 self.syscall()
                 if hasattr(e, "on_handled"):
                     e.on_handled()
             except RestartSyscall:
                 pass
-
+            self._syscall_abi._cpu._publish("did_invoke_syscall", index)
```
ekilmer, line 2920: This `RestartSyscall` makes me think the `did_invoke_syscall` should appear within the `try` block, but I forget when `RestartSyscall` is raised and what the state of Manticore is at that point. What are your thoughts?

tests/native/test_manticore.py
```diff
@@ -88,6 +88,59 @@ def tmp(state):
 
         assert tmp in self.m._after_hooks[entry]
 
+    def test_add_sys_hook(self):
+        def tmp(state):
+            assert state._platformn._syscall_abi.syscall_number() == index
+            self.m.kill()
+
+        name = "sys_brk"
+        index = 12
```
ekilmer, line 97: Nested functions in Python confuse me sometimes. Even though `index = 12` appears after line 93, `index` at line 93 will be `12` when it's executed? Maybe we should move `index = 12` to before the nested function like you have the test case after this one.

ekilmer (commented):
Looking good! Thank you! Just a few comments

manticore/platforms/linux.py
sschriner: I think you're right. From what I can see in Pycharm it looks like `RestartSyscall` is for `sys_write` [here](https://github.com/trailofbits/manticore/blob/5a258f499098394c0af25e2e3f00b1b603c2334d/manticore/platforms/linux.py#L1792). I will move this 🙂

manticore/native/manticore.py
```diff
@@ -39,17 +39,24 @@ def __init__(self, path_or_state, argv=None, workspace_url=None, policy="random"
             initial_state = _make_initial_state(path_or_state, argv=argv, **kwargs)
         else:
             initial_state = path_or_state
-
         super().__init__(initial_state, workspace_url=workspace_url, policy=policy, **kwargs)
 
         # Move the following into a linux plugin
         self._assertions = {}
         self.trace = None
+        self.linux_machine_arch: str
```
ehennenfent, line 47: Since this appears on the `Manticore` object let's either prefix it with an underscore to denote that it's private or add a comment explaining what it does for RtD.

manticore/native/state.py
```diff
@@ -55,56 +66,109 @@ def __enter__(self) -> "State":
...
+        hooks, when, _ = self._get_hook_context(after, syscall)
+        cbs = hooks.get(pc_or_sys, set())
         if callback in cbs:
             cbs.remove(callback)
         else:
             return False
 
-        if len(hooks.get(pc, set())) == 0:
-            del hooks[pc]
+        if len(hooks.get(pc_or_sys, set())) == 0:
```
ehennenfent, line 134: Extremely minor nit: for collections that support it, Python generally prefers `if not x` over `if len(x) == 0`

ehennenfent (commented):
This looks good! I left a few comments pointing out places where the documentation needs to be updated, but that's all. Really excited to see that we'll be able to hook syscalls by name now.

ekilmer (approved):
LGTM!

PR #2390, opened 2021-02-07
Evm platform: improve evm._valid_jumpdests

montyly:
Add the direct opcode after a JUMPI as a valid destination.

Without these addresses this was false in many cases:
https://github.com/trailofbits/manticore/blob/c70e26db26ed50ef7f55c9a6306cd8f6faaf962a/manticore/platforms/evm.py#L1248-L1250

And the solver was called when it was not required in:
https://github.com/trailofbits/manticore/blob/c70e26db26ed50ef7f55c9a6306cd8f6faaf962a/manticore/platforms/evm.py#L1256-L1266

### DIFF
--- a/manticore/platforms/evm.py
+++ b/manticore/platforms/evm.py
@@ -769,9 +769,16 @@ class EVM(Eventful):
             except Exception as e:
                 return
 
+        prev_was_jumpi = False
         for i in EVMAsm.disassemble_all(extend_with_zeroes(bytecode)):
             if i.mnemonic == "JUMPDEST":
                 self._valid_jumpdests.add(i.pc)
+            if prev_was_jumpi:
+                self._valid_jumpdests.add(i.pc)
+            prev_was_jumpi = False
+            if i.mnemonic == "JUMPI":
+                prev_was_jumpi = True
+
 
         # A no code VM is used to execute transactions to normal accounts.
         # I'll execute a STOP and close the transaction
@@ -1233,6 +1240,10 @@ class EVM(Eventful):
         already constrained to a single concrete value.
         """
         # If pc is already pointing to a JUMPDEST thre is no need to check.
+
+        if isinstance(self._need_check_jumpdest, bool) and self._need_check_jumpdest is False:
+            return
+
         pc = self.pc.value if isinstance(self.pc, Constant) else self.pc
         if pc in self._valid_jumpdests:
             self._need_check_jumpdest = False
@@ -2080,7 +2091,6 @@ class EVM(Eventful):
         """Conditionally alter the program counter"""
         # TODO(feliam) If dest is Constant we do not need to 3 queries. There would
         # be only 2 options
-
         self.pc = Operators.ITEBV(256, cond != 0, dest, self.pc + self.instruction.size)
         # This set ups a check for JMPDEST in the next instruction if cond != 0
         self._set_check_jmpdest(cond != 0)


manticore/platforms/evm.py
```diff
@@ -769,9 +769,16 @@ def extend_with_zeroes(b):
             except Exception as e:
                 return
 
+        prev_was_jumpi = False
         for i in EVMAsm.disassemble_all(extend_with_zeroes(bytecode)):
             if i.mnemonic == "JUMPDEST":
                 self._valid_jumpdests.add(i.pc)
+            if prev_was_jumpi:
```
feliam, line 776: _valid_jumpdests is supposed to be the lists of valid jumpdests. Just taking out the appearences of jumpdests iopcodes that are inside a push.

feliam:
JUMPDEST instruction existance must be checked only when the condition of a immediately previous JUMPI was True.
If you jump to the instruction after a JUMP whit the condition being True and it is not a JUMPDEST it most be INVALID. AFAIK

montyly:
That's a really good catch @feliam (I haven't considered the JUMP to the JUMPI dest). 

Maybe we can save the JUMPI false destination in a `_valid_jumpi_dests` and use it only if the opcode is a jumpi in `_valid_jumpdests`. Would that work?

feliam:
> That's a really good catch @feliam (I haven't considered the JUMP to the JUMPI dest).
> 
> Maybe we can save the JUMPI false destination in a `_valid_jumpi_dests` and use it only if the opcode is a jumpi in `_valid_jumpdests`. Would that work?

Is not all this solved in the code as it is now? 

==

Consider this code
```
10: JUMPI (cond, 11)
11: NOP
```
Should not that result in an INVALID if cond is TRUE. (And NOP is False)

==
`should_check_jumpdest` shall contain the potentially symbolic condition that makes the interpreter check for a JUMPDEST. That must be almost always trivially solvable by "implied concretization". If that thing reaches the solver and trigger a more difficult query something else needs to be simplified or cached.

montyly:
I see. So `that must be almost always trivially solvable by "implied concretization"`,  I don't think this is trivial to solve, because before asking for the value, the solver has to check if the formula is sat, which can take some time. As this is frequently called, it can have an impact on the performance for long run.

What about moving the logic of `_check_jmpdest` to be performed AFTER the PC increases? Such that Manticore already has the concrete valuation for the PC, and will never have to call the solver.

Would that work?

My current understanding is that this can be done somewhere here: https://github.com/trailofbits/manticore/blob/c70e26db26ed50ef7f55c9a6306cd8f6faaf962a/manticore/platforms/evm.py#L1287-L1288

feliam:
Idk if  "trivial to solve" was the right wording. But what I think it shall happen is that the constraintset must be already solved by the time you must check a symbolic _should_check_jumpdest. 

Here is a list of truths:

_should_check_jmpdest is False be default
_should_check_jmpdest is set to False at every instruction but JUMP/JUMPI
_should_check_jmpdest is set to True in a JUMP
_should_check_jmpdest is set to the _condition_ on a JUMPI (condition, target)

> Ideally any optimization must be done transparently and under the hood in the smtlib via the $includethethingIpastesince2010. A solution like that will not only affect this particular case but any other thing like this. I think it was called "implied concretization" in KLEE2008.

That said, IFF the target in a JUMPI does not point (also) to the next instruction we could use the trick that folows:
You can estimate the value of  _condition_ of the previous JUMPI (and _should_check_jumpdest) by comparing "current_pc" against "previous JUMPI target"

Something like this...
```prev_pc = prev_instruction.offset
prev_opcode = prev_instruction.opcode

# immediatelly previous instruction was a JUMPI
if prev_opcode is JUMPI:

    # the target code offset of the previous JUMPI was concrete
    prev_target = prev_instruction.operand_target
    if isconcrete(prev_target):

        # current_pc is the target code offset of the previous JUMPI 
        if prev_target == current_pc:
        
            #if it is not rare case in which JUMPI target can point 
            # to the following instruction
            if current_pc != prev_pc + 1:
                # then Yes check for a JUMPDEST instruction
                assert solve(_should_check_jmpdest == True)  #remove to avoid unnecesary query                
                _should_check_jmpdest = True
```

montyly:
So as this PR is not going to work, I explored another strategy: 
- When there is a fork, and the instruction is a `jumpi`, I ask the solver to give the result of the condition (in addition to the jumpi dest)
- The condition result is saved in the state
- `_check_jmpdest` can then access to the result of the condition to determine if it should perform the jumpdest check

I have a PoC in a local branch that I need to clean as its a bit hacky for now. It also requires benchmarks to evaluate if it's worth the additional complexity or not.

PR #2392, opened 2021-02-23
Attempt to Fix solc Installation

ehennenfent:
Prefer `solc-select` over `brew` for installing Solidity 0.4.x, as the latter doesn't seem to work on MacOS.


ekilmer (approved):
Tests all passing! LGTM

PR #2394, opened 2021-02-28
Use Slots on all Expression objects

### LINKED ISSUE #1731: Use `__slots__` throughout the expression system

ekilmer:

closes #1731 

Notes:

* Small refactor for easier testing of `BitVecConcat`. In an upcoming PR, we will switch all Expression objects to keyword arguments only.

* Small refactor after setting some base Expression classes as `abstract=True` 

* Had to remove a regression test that involved a Pickled file, since I don't know how to recreate that file 😞


tests/other/test_smtlibv2.py
```diff
@@ -100,6 +65,187 @@ def test_check_solver_undefined(self):
...
+        class C(A, B):
+            pass
+
+        class X(object):
+            __slots__ = ("t", "a", "b")
+
+        c = C()
+        c.a = 1
+        c.t = 10
+
```
sschriner, line 97: Nit: delete this line of whitespace 🙃

tests/other/test_smtlibv2.py
```diff
@@ -100,6 +65,187 @@ def test_check_solver_undefined(self):
...
+        c.b = 2
+        c.t = 10
```
sschriner, line 99: Is it intentional that `c.t = 10` is called twice?

sschriner (approved):
LGTM 🙂

manticore/wasm/cli.py
```diff
@@ -18,9 +18,12 @@ def wasm_main(args, _logger):
         policy=args.policy,
     )
 
-    if consts.profile:
-        profiler = Profiler()
-        m.register_plugin(profiler)
+    try:
```
ehennenfent, line 21: Is this causing an AttributeError to be thrown? The profile flag is created in ethereum/cli.py, which is definitely weird, but it's never caused problems before. Might we instead copy the following line into wasm/cli.py as well?

https://github.com/trailofbits/manticore/blob/2152023f390a7dc45cf16cd5cc283d0cfa93b8ee/manticore/ethereum/cli.py#L30

manticore/core/smtlib/expression.py
```diff
@@ -15,10 +15,60 @@ class ExpressionException(SmtlibError):
...
+    def __new__(cls, clsname, bases, attrs, abstract=False):
+        xslots = frozenset(attrs.get("__xslots__", ()))
+        # merge the xslots of all the bases with the one defined here
+        for base in bases:
+            xslots = xslots.union(getattr(base, "__xslots__", ()))
+        attrs["__xslots__"] = tuple(xslots)
+        if abstract:
+            attrs["__slots__"] = tuple()
+        else:
+            attrs["__slots__"] = tuple(map(lambda attr: attr.split("#", 1)[0], attrs["__xslots__"]))
```
ehennenfent, line 62: Where does the `#` come from?

manticore/core/smtlib/expression.py
```diff
@@ -512,8 +564,10 @@ def signed_value(self):
...
 
 
-class BitVecOperation(BitVec):
-    __slots__ = ["_operands"]
+class BitVecOperation(BitVec, abstract=True):
+    """ An operation that results in a BitVec """
+
+    __xslots__: Tuple[str, ...] = ("_operands",)
 
     def __init__(self, size, *operands, **kwargs):
```
ehennenfent, line 572: So, everywhere that you create a `BitVecOperation` or `BoolOperation`, you pass in `operands` as a tuple without unpacking it. However, the function signature doesn't change here. So now, by my estimation, `self._operands` becomes a tuple nested two stages deep. Is that intentional? It seems weird to me, but maybe it does something I'm not catching.

manticore/wasm/cli.py
ekilmer: Hmmm. Yeah, I think I just forgot to remove this change before committing. I ran into something where this errored on the missing attribute, but I can't seem to reproduce it now 🤷 

I think adding that line to `wasm/cli.py` would be good in any case though 👍  (Had to delete a corresponding line in the test file)

ekilmer: The process for adding config options seems to be a bit delicate, so I've added some logic to allow the same exact config option to be "added" twice... I'll break this out into a separate PR

manticore/core/smtlib/expression.py
ekilmer: Good question... I think @feliam might have had a use for them? Also asked this question here https://github.com/trailofbits/manticore/pull/1729/files#r585653426

Removing from this PR though 👍

manticore/core/smtlib/expression.py
ekilmer: Good thought. I see what you mean. This was pretty confusing for me when working on it tbh, so I'm glad we can talk about it more 😄 .

I'm going to think out loud a little bit to make sure we're on the same page though:

> everywhere that you create a `BitVecOperation` or `BoolOperation`, you pass in operands as a tuple without unpacking it.

For each subclass of `BitVecOperation` to correctly call its `__init__`, we need to do the following 
```python
super().__init__(size, operand1, operand2, ..., operandn, **kwargs)
```
which would pack `operand1, operand2` into `operands` as a tuple without any nesting. 

On the other hand, we shouldn't be doing this (which is what you're calling to attention, I think):
```python
super().__init__(size, tuple(operand1, operand2, ....), **kwargs)
```
right? This would double-pack the tuple if we left the `*operands` parameter as is in `BitVecOperation`s  `__init__`.

It's a similar deal with `BoolOperation`, just without the first `size` argument.

So now, we should check all subclasses of `BitVecOperation` and `BoolOperation` to see how we call the `super().__init__(...)`. 

... After some searching, I'm unable to manually find where we are packing the operands before passing to `__init__`. In that case, I'm going to add an `assert all(not isinstance(op, tuple) for op in operands)` statement to both `BitVecOperation` and `BoolOperation` `__init__` method and run the tests, which should catch any double tuple packing that happens.

All tests seem to pass on my machine still, so unless I'm still missing something, this PR doesn't do any tuple nesting. Would you like me to commit/add those assertion statements?

ekilmer: The next PR https://github.com/trailofbits/manticore/pull/2395 _does_ take the `operands` as a pre-packed tuple and is keyword-only. That involved a lot more to refactor

manticore/wasm/cli.py
ekilmer: Opened in #2397 and removed those changes in this PR

manticore/core/smtlib/expression.py
ekilmer: ping @ehennenfent for further discussion

ehennenfent: > After some searching, I'm unable to manually find where we are packing the operands before passing to __init__

You're right, I was looking at the various places where we construct `BitVecConcat`, which explicitly asks for a tuple in its type signature, which it then unpacks before passing up the inheritance tree.

ehennenfent (approved):
Looks good to me as a minimal changeset. I do think that the argument unpacking behavior is a little bit too confusing, but my requested fix for that would be to switch to keyword-only arguments. Let's try to merge #2395 very soon after this one.

PR #2395, opened 2021-03-01
Expression use keyword-only arguments for init

ekilmer:

This will help with multiple inheritance in a future commit. Also a better way to enforce correct initialization of the objects.

Follow-up to `slotted` branch/PR #2394


ehennenfent (approved):
Looks good to me! Agree that kwarg-only expressions are the way to go.

PR #2396, opened 2021-03-01
Expressions use multiple inheritance for Constant, Variable, Operation

ekilmer:

Removes some duplicate code

NOTE: This branch will be rebased onto `master` after `expression-keyword-only-init` branch/PR #2395 is merged

### DIFF
--- a/manticore/core/smtlib/constraints.py
+++ b/manticore/core/smtlib/constraints.py
@@ -185,7 +185,7 @@ class ConstraintSet:
                 if (
                     isinstance(expression, BoolEqual)
                     and isinstance(expression.operands[0], Variable)
-                    and isinstance(expression.operands[1], (*Variable, *Constant))
+                    and isinstance(expression.operands[1], (Variable, Constant))
                 ):
                     constant_bindings[expression.operands[0]] = expression.operands[1]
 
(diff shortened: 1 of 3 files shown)


ehennenfent (commented):
I don't see any code problems with this PR, and I think the technique of dynamically creating `__slots__` for multiply-inherited subclasses is really exciting. That said, since this more or less completely reverts #1635, I'd like to get your thoughts on  a couple of matters:

First, changing the inheritance structure reduces the amount of duplicated code, but means we have to do things like the `__xslots__` trick to get good performance. Are we certain the ergonomics of one approach are better than the other?

Second, how much does this change affect performance? #1635 gave us a modest improvement, but not large enough to be noticeable without measuring, so we should measure this branch as well to see how it stacks up.

ekilmer:
After testing with `multiple-styles`, as was done in #1635 I saw a slight slowdown with the use of multiple inheritance. Testing with a microbenchmark that instantiates the `Constant` and `Variable` subclass objects also reveals some performance penalties for the multiple inheritance. I didn't test with `Operation` subclasses, but I suspect it to be a similar story and also more greatly contribute to a performance drop, since `Operation`s are used more in expressions.

I think the `test_Api` test we have to do sanity checking is good enough, and I don't really expect too much churn on these objects to make maintenance unmanageable, so I will close this PR.

PR #2397, opened 2021-03-02
Allow double-adding exact same config option

ekilmer:
Extracted from #2394 

The `config` system is a little brittle when Manticore tries to import all platform modules since we set the config options upon import and declaring the same config option in two different platforms will cause issues.

This is a bit of a band-aid over the real problem, which might look something like moving shared `config` options to a separate file that each platform can import.

Looks to be some related refactoring work here https://github.com/trailofbits/manticore/pull/1636

### DIFF
--- a/manticore/utils/config.py
+++ b/manticore/utils/config.py
@@ -103,6 +103,10 @@ class _Group:
 
         """
         if name in self._vars:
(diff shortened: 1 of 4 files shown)


ehennenfent (approved):
LGTM, thanks!

PR #2398, opened 2021-03-04
Test Python 3.8 and 3.10 in CI

ekilmer:
Things to do before merge:

- [ ] ~Look into notifications for Slack when runs fail (so that we aren't just ignoring failed scheduled runs)~
- [x] Wait for forked https://github.com/trail-of-forks/wasm-tob repo to make a release

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   # needs to run only on pull_request
   lint:
-    runs-on: ubuntu-22.04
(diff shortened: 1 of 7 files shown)


ekilmer:
We'll need to change the `Required` CI tests in the repo settings

ehennenfent:
Glad to see that this works, but I think having 23 jobs per push is going to eat up too much of the organization's GH Actions time. I know we can set filters so that certain actions only run on the scheduled nightly runs; can we do that with matrix options as well?

PR #2409, opened 2021-03-24
Update CI and suggest to use pip3 instead of pip

ggrieco-tob:
A very small change in an error message to use `pip3` instead of `pip`.


ekilmer (approved):
LGTM. Thank you for the CI changes

PR #2414, opened 2021-03-26
Removed use of global solver from Memory

ggrieco-tob:
The Memory class had a global solver hard-coded that resulted in z3 (or the solver the user has installed) always used when a native binary was executed. This PR should fix that.

### DIFF
--- a/manticore/native/memory.py
+++ b/manticore/native/memory.py
@@ -1,6 +1,7 @@
 from abc import ABCMeta, abstractmethod
 from weakref import WeakValueDictionary
 from dataclasses import dataclass, field
+from ..core.state import EventSolver
 from ..core.smtlib import (
     Operators,
     ConstraintSet,
@@ -31,7 +32,6 @@ consts.add(
     description="If True, throws a memory safety error if ANY concretization of a pointer is"
     " out of bounds. Otherwise, forks into valid and invalid memory access states.",
 )
-solver = SelectedSolver.instance()
 
 
 class MemoryException(Exception):
@@ -603,6 +603,7 @@ class Memory(object, metaclass=ABCMeta):
         self.cpu = cpu
         self._page2map: MutableMapping[int, Map] = WeakValueDictionary()  # {page -> ref{MAP}}
         self._recording_stack: List = []
+        self._solver = EventSolver()
         for m in self._maps:
             for i in range(self._page(m.start), self._page(m.end)):
                 assert i not in self._page2map
@@ -1202,7 +1203,7 @@ class SMemory(Memory):
                 assert len(solutions) > 0
             except TooManySolutions as e:
                 self.cpu._publish("will_solve", self.constraints, address, "minmax")
-                m, M = solver.minmax(self.constraints, address)
+                m, M = self._solver.minmax(self.constraints, address)
                 self.cpu._publish("did_solve", self.constraints, address, "minmax", (m, M))
                 logger.debug(
                     f"Got TooManySolutions on a symbolic read. Range [{m:x}, {M:x}]. Not crashing!"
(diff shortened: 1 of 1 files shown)


ekilmer:
> global solver hard-coded that resulted in z3

I just run the following locally with `yices` installed through my system and `SelectedSolver.instance()` returns `YicesSolver`
```python
$ python
Python 3.6.13 (default, Feb 22 2021, 12:20:57)
[GCC 10.2.1 20201125 (Red Hat 10.2.1-9)] on linux
>>> from manticore.core.smtlib import SelectedSolver
>>> solver = SelectedSolver.instance()
>>> solver
<manticore.core.smtlib.solver.YicesSolver object at 0x7f51d999ee48>
```
I also tried inspecting the global solver itself and still got `YicesSolver`
```
$ python
Python 3.6.13 (default, Feb 22 2021, 12:20:57)
[GCC 10.2.1 20201125 (Red Hat 10.2.1-9)] on linux
>>> from manticore.native.memory import solver
>>> solver
<manticore.core.smtlib.solver.YicesSolver object at 0x7fa727dbc780>
```
Am I missing something?

ggrieco-tob:
If you apply this patch:

```diff
diff --git a/manticore/core/smtlib/solver.py b/manticore/core/smtlib/solver.py
index 7e61217..208cfd3 100644
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -761,4 +761,5 @@ class SelectedSolver:
             cls.choice = consts.solver
 
         SelectedSolver = {"cvc4": CVC4Solver, "yices": YicesSolver, "z3": Z3Solver}[cls.choice.name]
+        print("Calling instance() with cls.choice =", cls.choice)
         return SelectedSolver.instance()
```

you can see that running anything using the manticore cli will print `Calling instance() with cls.choice = SolverType.z3` then it will use any other solver you specify (for instance, using `--smt.solver cvc4` will still print that line).

ekilmer:
Ahh, okay, yes I see that now.

I think there's a bigger issue with the CLI options not getting parsed correctly. I'm currently looking into it

manticore/native/memory.py
```diff
@@ -603,6 +603,7 @@ def __init__(self, maps: Optional[Iterable[Map]] = None, cpu=StubCPU()):
         self.cpu = cpu
         self._page2map: MutableMapping[int, Map] = WeakValueDictionary()  # {page -> ref{MAP}}
         self._recording_stack: List = []
+        self._solver = EventSolver()
```
ekilmer, line 606: I think we want to either still use `SelectedSolver` here or remove all of the `self.cpu_publish("will_solve"` and `self.cpu_publish("did_solve"` surrounding the solver calls, because `EventSolver` takes care of publishing those events internally:

https://github.com/trailofbits/manticore/blob/fd83be7a9929d3645a3df995e2f925d3fb3b4b7f/manticore/core/state.py#L121-L125

ekilmer (changes requested):
Okay! I think I finally understand what's going on here...

The global `solver` instance is initialized incorrectly because it calls `SelectedSolver.instance()` before we've parsed the command line flags, due to execution upon import, and so if someone chooses to use a particular solver, it will be ignored for these global instances.

Thank you!

ekilmer (approved):
LGTM!!

manticore/native/memory.py
ggrieco-tob: I removed the duplicated events. I think we should aim to remove code at each PR :smile:

ggrieco-tob:
This looks ready to merge, but again I can't do it since one check (CI / upload) was skipped. Can you merge it @ekilmer ?

PR #2415, opened 2021-03-26
Basic solver stats

ggrieco-tob:
This small PR includes a global dictionary that keeps tracks of when each solver is used succefully, timeouts or returns unknown. At the end, the result will be saved in a plain text file called `global.solver_stats`. If the solver returned either timeout or unknown, the user will be warned about it.

### DIFF
--- a/manticore/core/manticore.py
+++ b/manticore/core/manticore.py
@@ -1106,6 +1106,10 @@ class ManticoreBase(Eventful):
         """
         Runs analysis.
         """
+        # Start measuring the execution time
+        with self.locked_context() as context:
+            context["time_started"] = time.time()
+
         # Delete state cache
         # The cached version of a state may get out of sync if a worker in a
         # different process modifies the state
@@ -1229,6 +1233,17 @@ class ManticoreBase(Eventful):
             )
 
         logger.info("Results in %s", self._output.store.uri)
+
+        time_ended = time.time()
+
+        with self.locked_context() as context:
+            time_elapsed = time_ended - context["time_started"]
+
+            logger.info("Total time: %s", time_elapsed)
+
+            context["time_ended"] = time_ended
+            context["time_elapsed"] = time_elapsed
+
         self.wait_for_log_purge()
 
     def introspect(self) -> typing.Dict[int, StateDescriptor]:
--- a/manticore/core/manticore.py
+++ b/manticore/core/manticore.py
@@ -14,7 +14,7 @@ import functools
 import shlex
 
 from ..core.plugin import Plugin, IntrospectionAPIPlugin, StateDescriptor
-from ..core.smtlib import Expression
+from ..core.smtlib import Expression, SOLVER_STATS
 from ..core.state import StateBase
 from ..core.workspace import ManticoreOutput
 from ..exceptions import ManticoreError
@@ -1218,6 +1218,15 @@ class ManticoreBase(Eventful):
 
         with self._output.save_stream("manticore.yml") as f:
             config.save(f)
+
+        with self._output.save_stream("global.solver_stats") as f:
+            for s, n in sorted(SOLVER_STATS.items()):
+                f.write("%s: %d\n" % (s, n))
+
+        if SOLVER_STATS["timeout"] > 0 or SOLVER_STATS["unknown"] > 0:
+            logger.info(
+                "Warning: the SMT solvers returned timeout or unknown for certain program paths. Results could not cover the entire set of possible paths"
+            )
 
         logger.info("Results in %s", self._output.store.uri)
         self.wait_for_log_purge()
--- a/manticore/core/manticore.py
+++ b/manticore/core/manticore.py
@@ -1228,8 +1228,8 @@ class ManticoreBase(Eventful):
                 f.write("%s: %d\n" % (s, n))
 
         if SOLVER_STATS["timeout"] > 0 or SOLVER_STATS["unknown"] > 0:
-            logger.info(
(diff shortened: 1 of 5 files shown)


manticore/core/smtlib/solver.py
```diff
@@ -78,6 +78,8 @@ class SolverType(config.ConfigEnum):
 )
 RE_MIN_MAX_OBJECTIVE_EXPR_VALUE = re.compile(r"(?P<expr>.*?)\s+\|->\s+(?P<value>.*)", re.DOTALL)
 
+SOLVER_STATS = dict({"unknown": 0, "timeout": 0})
```
montyly, line 81: Manticore relies on multi-threads
https://github.com/trailofbits/manticore/blob/5a258f499098394c0af25e2e3f00b1b603c2334d/manticore/core/manticore.py#L383-L384

So if we share the dict among all the threads, we might suffer from race condition. 

I am not too familiar with the thread architecture, but I suppose we could use a dict with a lock, or create one dict per thread?

ehennenfent: For future reference, you can use the locked context in situations where you need to store data securely across multiple worker threads or processes:
https://github.com/trailofbits/manticore/blob/master/manticore/core/manticore.py#L992-L1021

Unfortunately, buried down in SMTLib as we are, it's a bit difficult to access the locked context from here. As long as the stats are informational only, I would also support storing them at the module level. The built-in types are inherently thread-safe (thanks to the GIL) so as long as we're running in multithreaded mode (the default) this should be fine. 

As a minor nit, `{"unknown": 0, "timeout": 0}` is already a dictionary; no need to explicitly wrap it with the call to `dict()`.

ggrieco-tob: I think using a global dictionary looks like a reasonable solution so far.

> As a minor nit, {"unknown": 0, "timeout": 0} is already a dictionary; no need to explicitly wrap it with the call to dict().

Fixed.

manticore/core/manticore.py
```diff
@@ -1219,7 +1223,28 @@ def save_run_data(self):
         with self._output.save_stream("manticore.yml") as f:
             config.save(f)
 
+        with self._output.save_stream("global.solver_stats") as f:
+            for s, n in sorted(SOLVER_STATS.items()):
+                f.write("%s: %d\n" % (s, n))
+
+        if SOLVER_STATS["timeout"] > 0 or SOLVER_STATS["unknown"] > 0:
+            logger.info(
+                "Warning: the SMT solvers returned timeout or unknown for certain program paths. Results could not cover the entire set of possible paths"
```
ehennenfent, line 1232: * Since the log level for this is `info`, we should remove the "Warning" at the beginning of the message.
* How often do timeouts occur? It looks like the default log level for this file is `info`, so this message will be printed after any run in which a solver times. Personally I think that might be a little too verbose, but we can discuss further.

manticore/core/manticore.py
```diff
@@ -1219,7 +1223,28 @@ def save_run_data(self):
...
+        time_ended = time.time()
+
+        with self.locked_context() as context:
+            if "time_started" in context:
+                time_elapsed = time_ended - context["time_started"]
+                logger.info("Total time: %s", time_elapsed)
+                context["time_ended"] = time_ended
+                context["time_elapsed"] = time_elapsed
+            else:
+                logger.info("Warning: manticore failed to run")
```
ehennenfent, line 1246: * See above re: info/warning disconnect
* "Manticore failed to run" seems a little bit aggressive. Maybe "Run start time not found in context. Was `save_run_data` called without calling `run`?"

manticore/core/smtlib/solver.py
```diff
@@ -583,6 +603,12 @@ def _optimize_fancy(self, constraints: ConstraintSet, x: BitVec, goal: str, max_
             self._smtlib.send("(%s %s)" % (goal, aux.name))
             self._smtlib.send("(check-sat)")
             _status = self._smtlib.recv()
+
+            assert self.sname is not None
+            if self.sname not in SOLVER_STATS:
+                SOLVER_STATS[self.sname] = 0
```
ehennenfent, line 609: For future reference, Python dicts include a builtin method for doing membership checking and initialization in one line. So, this line and the previous would become `SOLVER_STATS.setdefault(self.sname, 0)`

manticore/core/manticore.py
ggrieco-tob: The log level for this was changed to `warning`. Timeouts can happen in any non-trivial smart contract code (specially with the default timeout of two minutes), however this message will be printed once per execution, at the very end, so it should not so be verbose.

manticore/core/manticore.py
ggrieco-tob: The log level for this was changed to `warning`. This error will happen if you provide manticore with incorrect binary code or bytecode. Typically, when a smart contract is not properly compiled. Right now, this is output:

```
2021-04-06 09:15:54,463: [13997] m.c.manticore:INFO: Generated testcase No. 0 - NO STATE RESULT (?)(0 txs)
2021-04-06 09:15:54,464: [13997] m.c.plugin:WARNING: Caught will_solve in state None, but failed to capture its initialization
2021-04-06 09:15:54,475: [13984] m.c.manticore:INFO: Results in /home/g/Data/manticore-benchmarks/mcore_mkg0pep9
```

Here manticore does not provide any good feedback that it failed to run beyond the "NO STATE RESULT (?)", that's why "Manticore failed to run" can be appropriated (but I'm open to suggestions).

ehennenfent (approved):

PR #2417, opened 2021-04-01
Disabled EVM events in testcases by default

### LINKED ISSUE #2416: Feature request: disable event logging by default
When the smart contract code has events, Manticore will try to obtain the concrete values for them during the symbolic execution. This will query the solver a number of time, so if we don't need them, it will be wise to disable them by default.

ggrieco-tob:
This small PR can save a few calls to the SMT solver avoiding getting the EVM events by default. Fixes #2416


ekilmer (approved):
LGTM

PR #2419, opened 2021-04-02
Avoid adding a constrain to take the gas from the balance when oog is "ignore"

ggrieco-tob:
This small PR will eliminate one unnecessary constrain when oog is "ignore"


feliam:
https://github.com/trailofbits/manticore/pull/1823
Out of gas vs. dynamic gas calculation vs. balances

ggrieco-tob:
It is better to merge https://github.com/trailofbits/manticore/pull/1823 instead.

PR #2420, opened 2021-04-03
Portfolio of parallel solvers

ggrieco-tob:
This PR contains:

* A new meta solver called portfolio, that runs several solvers in parallel. Once one solver wins the race, the other ones are stopped, since there is no way to use their results in a meaningful way. 
* As expected, the number of workers is adjusted when using the portfolio to avoid using more than `core.procs` processes. 
* A new `recv` implementation based on non-blocking reads from the `stdout` pipe. I'm still experimenting how each to increase each timeout to avoid busy waiting.
* All the installed solvers but z3 are going to be used in the portfolio. I excluded z3 because I suspect it's the slowest one in almost all the tests but I need some more hard evidence. I'm tracking this in #2449.
* Fixes for #2421


manticore/core/smtlib/solver.py
```diff
@@ -79,6 +63,50 @@ class SolverType(config.ConfigEnum):
...
+
+
+consts.add(
+    "solver",
+    default=SolverType.auto,
+    description="Choose default smtlib2 solver (z3, yices, cvc4, boolector, auto)",
+)
+
+
+class SolverInfo:
```
feliam, line 84: Put the config of each solver in each solver class?

manticore/core/smtlib/solver.py
```diff
@@ -211,10 +243,19 @@ def stop(self):
...
-    def __readline_and_count(self):
+    def __readline_and_count(self, wait):
         assert self._proc
         assert self._proc.stdout
-        buf = self._proc.stdout.readline()  # No timeout enforced here
+        buf = None
+
+        while buf is None:
+            try:
+                buf = self._proc.stdout.readline()
```
feliam, line 253: busy waiting?
I wonder when would the original readline wait forever.

manticore/core/smtlib/solver.py
```diff
@@ -653,7 +700,8 @@ def __init__(self):
         This is implemented using an external z3 solver (via a subprocess).
         See https://github.com/Z3Prover/z3
         """
-        command = f"{consts.z3_bin} -t:{consts.timeout * 1000} -memory:{consts.memory} -smt2 -in"
+        command = info.commands["z3"]
+        self.ncores = 1
```
feliam, line 704: Why z3 needs `ncores` ?

manticore/core/smtlib/solver.py
```diff
@@ -261,6 +305,8 @@ def is_started(self):
 
 
 class SMTLIBSolver(Solver):
+    ncores: Optional[int] = None
```
feliam, line 308: Ahh. Here!  Why the user  of an SMTLIBSolver needs know about the "cores"?

ggrieco-tob: That's necessary to know how much workers create. Originally, every worker used one cpu, but now the portofolio solver can use many of them.

manticore/core/manticore.py
```diff
@@ -380,8 +380,9 @@ def __init__(
             raise TypeError(f"Invalid initial_state type: {type(initial_state).__name__}")
         self._put_state(initial_state)
 
+        nworkers = min(consts.procs / initial_state._solver.ncores, 1)
```
feliam, line 383: And... this is why ncores is part of the interface of an smtlib solver!

manticore/core/smtlib/solver.py
```diff
@@ -735,18 +770,162 @@ def __init__(self):
...
+        if shutil.which(consts.boolector_bin):
+            solvers.append(consts.solver.boolector.name)
+        else:
+            raise SolverException(
+                f"No Solver not found. Install one ({consts.yices_bin}, {consts.z3_bin}, {consts.cvc4_bin}, {consts.boolector_bin})."
+            )
+
+        print("Creating portfolio with solvers", solvers)
+        assert len(solvers) > 0
+        support_reset: bool = False
```
feliam, line 892: As all this is not used, remove.

manticore/core/smtlib/solver.py
```diff
@@ -735,18 +770,162 @@ def __init__(self):
...
-        command = f"{consts.boolector_bin} --time={consts.timeout} -i"
+    def __init__(self, args: List[str] = []):
+        init = info.inits["boolector"]
+        command = info.commands["boolector"]
+        self.ncores = 1
         super().__init__(command=command, init=init)
 
 
+class SmtlibPortfolio:
+    def __init__(self, solvers: List[str], debug: bool = False):
```
feliam, line 788: `SMTLIBSolver` instead of `str`?

manticore/core/smtlib/solver.py
```diff
@@ -735,18 +770,162 @@ def __init__(self):
...
+        """
+        assert len(self._procs) > 0
+        # print(cmd)
+        for proc in self._procs:
+            if not proc.is_started():
+                continue
+
+            proc.send(cmd)
+
+        if "(check-sat)" in cmd:
```
feliam, line 832: What does _check_sat means?
Does this class accept a single check-sat command?

ggrieco-tob: In certain functions like `get_all_values` you check for sat and then ask for some values (with some more constrains):

```python
 while self._is_sat():
    value = self._getvalue(var)
   ...
```

 When you call `check-sat`, you start all the solvers. Eventually one of them will answer.  However, the other ones are going to be busy so when you start adding more constrains, so you need to discard the other solvers until you finish with the call and then restart the solvers. In summary, if `(check-sat)` was used, the portfolio solver should discard the executions of the remaining ones.

feliam: Does that means that after the first `(check-sat)` all commands are in race mode?

Maybe then `_check_sat` should be reset to false after the recv?

ggrieco-tob: Before check-sat all solvers are in race mode. After, I should find a good way to resend all the smtlib2 commands so far.

manticore/core/smtlib/solver.py
ggrieco-tob: Fixed.

manticore/core/smtlib/solver.py
ggrieco-tob: `support_reset` is used

manticore/core/smtlib/solver.py
ggrieco-tob: We keep a list of solver names in several dicts.

manticore/core/smtlib/solver.py
ggrieco-tob: All the other solvers should be stopped after checking for sat, because we cannot use them in parallel with the use of `get-value`. The possibility of interleaving, while small, could result in invalid results (where some values are obtained from one solver and others are obtained from another one).

manticore/core/smtlib/solver.py
```diff
@@ -192,6 +222,11 @@ def start(self):
             close_fds=True,
         )
 
+        # stdout should be non-blocking
+        fl = fcntl.fcntl(self._proc.stdout, fcntl.F_GETFL)
```
ehennenfent, line 226: I won't pretend to have a comprehensive grasp of low-level IO in Python, but I'm more used to seeing the `select` module used than `fcntl`. Did you consider `select.poll`, which is also non-blocking, as a way to handle these file descriptors?

manticore/core/smtlib/solver.py
ehennenfent: As Felipe points out, it'd probably be easier to set some abstract defaults on the `Solver` class. You could use `abc.abstractproperty` on `Solver` and override it on `YicesSolver`, `Z3Solver`, etc.

manticore/core/smtlib/solver.py
ggrieco-tob: I selected `fcntl` instead of `poll` because I did not wanted to add another thread/currency primitive for workers, which already works using threads..

manticore/core/smtlib/solver.py
ggrieco-tob: I believe we cannot use abstract classes here since we need the command line for each solver and they depend on global options. For instance, `"{consts.z3_bin} -t:{consts.timeout * 1000} -memory:{consts.memory} -smt2 -in` depends on `consts.z3_bin` and `consts.timeout`, and `consts.timeout` which are going to be computed when the command line options are parsed.

manticore/core/manticore.py
```diff
@@ -49,7 +49,7 @@
     default=False,
     description="If True enables to run workers over the network UNIMPLEMENTED",
 )
-consts.add("procs", default=10, description="Number of parallel processes to spawn")
+consts.add("procs", default=12, description="Number of parallel processes to spawn")
```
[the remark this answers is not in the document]
ggrieco-tob: Done.

manticore/core/manticore.py
```diff
@@ -380,8 +380,10 @@ def __init__(
             raise TypeError(f"Invalid initial_state type: {type(initial_state).__name__}")
         self._put_state(initial_state)
 
+        nworkers = max(consts.procs / initial_state._solver.ncores, 1)
+        nworkers = int(nworkers)
```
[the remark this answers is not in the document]
ggrieco-tob: Done.

manticore/core/smtlib/solver.py
```diff
@@ -79,6 +64,50 @@ class SolverType(config.ConfigEnum):
...
+    yices = "yices"
+    auto = "auto"
+    portfolio = "portfolio"
+    boolector = "boolector"
+
+
+consts.add(
+    "solver",
+    default=SolverType.auto,
+    description="Choose default smtlib2 solver (z3, yices, cvc4, boolector, auto)",
```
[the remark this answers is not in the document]
ggrieco-tob: Done.

manticore/core/smtlib/solver.py
```diff
@@ -234,17 +257,43 @@ def send(self, cmd: str) -> None:
         self._proc.stdout.flush()  # type: ignore
         self._proc.stdin.write(f"{cmd}\n")  # type: ignore
 
-    def recv(self) -> str:
+    def recv(self, wait=True) -> Optional[str]:
         """Reads the response from the smtlib solver"""
```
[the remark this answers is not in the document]
ggrieco-tob: Done.

manticore/core/smtlib/solver.py
```diff
@@ -735,18 +779,169 @@ def __init__(self):
...
+        # if shutil.which(consts.z3_bin):
+        #    solvers.append(consts.solver.z3.name)
+        if shutil.which(consts.cvc4_bin):
+            solvers.append(consts.solver.cvc4.name)
+        if shutil.which(consts.boolector_bin):
+            solvers.append(consts.solver.boolector.name)
+        if solvers == []:
+            raise SolverException(
+                f"No Solver not found. Install one ({consts.yices_bin}, {consts.z3_bin}, {consts.cvc4_bin}, {consts.boolector_bin})."
+            )
```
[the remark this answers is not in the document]
ggrieco-tob: Done.

manticore/core/smtlib/solver.py
```diff
@@ -234,17 +257,43 @@ def send(self, cmd: str) -> None:
...
+                    return None
+                else:
+                    tries += 1
+
+            if buf == "":
+                continue
+
+            lparen, rparen = map(sum, zip(*((c == "(", c == ")") for c in buf)))
+            if lparen == rparen and buf != "":
+                break
```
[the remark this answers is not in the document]
ekilmer: Ah, I didn't realize it was part of the original code! Thank you for deciphering it for me 😄

ekilmer (approved):

ekilmer:
Looks good on my end, but there are still comments from other people that should be resolved first before merging. Thank you!!

manticore/core/smtlib/solver.py
ggrieco-tob: I followed your advise, but instead of using a property, I had to use a `classmethod`. That was the only way to allow to call `command` and `inits` from the class, without actually create an instance the solvers. Please take a look to my recent changes.

PR #2425, opened 2021-04-06
Delay smt-solving until it is absolutely necessary in EVM

gustavo-grieco:
This PR contains:

* A lazy evaluation mode for `jumpi` constraints enabled by default (so we can see the effects in the tests). This mode cannot be used if the code contains a loop. There is no automatic detection of that, so the user should be careful when using this mode. It is unclear how to handle that in the middle of a manticore run.
* Various fixed for the evm tests to make sure the states are always sound.
* `test_jmpdest_check` test was disabled.
* Some tests using plugins require lazy mode to be disabled.

The last two bullets should be discussed before merging this PR.


manticore/core/state.py
```diff
@@ -360,7 +360,7 @@ def new_symbolic_value(self, nbits, label=None, taint=frozenset()):
         self._input_symbols.append(expr)
         return expr
 
-    def concretize(self, symbolic, policy, maxcount=7):
+    def concretize(self, symbolic, policy, maxcount=7, additional_symbolics=None):
```
ehennenfent, line 363: I'm not thrilled with introducing an additional argument to a user-facing function if it's only used in one other place in Manticore. I'd suggest either:
* Removing the `additional_symbolics` argument from `concretize` and instead calling `concretize` twice in ethereum/manticore.py and zipping the results (since it results in the same number of solver calls regardless)

_Or_

* Adding type hints and documentation items describing each argument, and making `additional_symbolics` keyword-only.

manticore/core/state.py
```diff
@@ -412,9 +412,19 @@ def concretize(self, symbolic, policy, maxcount=7):
                 vals = (True,)
         else:
             assert policy == "ALL"
-            vals = self._solver.get_all_values(
-                self._constraints, symbolic, maxcnt=maxcount, silent=True
-            )
+            if additional_symbolics is not None:
+                print(symbolic, additional_symbolics)
```
ehennenfent, line 416: If you do end up leaving `additional_symbolics`, this should be swapped out for `logger.debug` or `logger.info`

manticore/platforms/evm.py
```diff
@@ -2539,6 +2548,12 @@ def constraints(self, constraints):
         if self.current_vm:
             self.current_vm.constraints = constraints
 
+    def last_ins_was_true_jumpi(self) -> Optional[bool]:
+        return self._last_ins_was_true_jumpi
+
+    def set_last_ins_was_true_jumpi(self, jump_cond: Optional[bool]):
```
ehennenfent, line 2554: Let's convert this getter/setter pairing to a property as well.

manticore/platforms/evm.py
```diff
@@ -863,6 +871,18 @@ def constraints(self, constraints):
     def gas(self):
         return Operators.EXTRACT(self._gas, 0, 256)
 
+    def jumpi_false_branch(self) -> Optional[int]:
```
ehennenfent, line 874: Python has a nifty replacement for getter & setter methods called [Properties](https://docs.python.org/3/library/functions.html#property). Since it looks like these don't explicitly need to be callable, we should convert them to properties.

manticore/ethereum/manticore.py
```diff
@@ -1953,3 +1969,100 @@ def worker_finalize(q):
...
             proc.join()
+
+    def set_lazy_evaluation(self, flag):
+        """
+        Set lazy evaluation flag (only EVM support)
+        :return:
+        """
+        self._lazy_evaluation = flag
+
+    def get_lazy_evaluation(self) -> bool:
```
ehennenfent, line 1980: As before, a property would be perfect here. As for `set_lazy_evaluation`, you could convert that to a property as well, but I think this is intended to be user-facing, and most users will trust `manticore.set_lazy_evaluation(False)` more than `manticore.lazy_evaluation = False`. That said, to make it even more explicit, you could consider splitting this up into `enable_lazy_evaluation` and `disable_lazy_evaluation` methods.

ehennenfent (changes requested):

manticore/core/state.py
ggrieco-tob: Where I should add the type hint exactly for `additional_symbolics`?

manticore/core/state.py
ggrieco-tob: Fixed.

manticore/ethereum/manticore.py
ggrieco-tob: Fixed, now we have `enable/disable_lazy_evaluation`.

manticore/platforms/evm.py
ggrieco-tob: Fixed.

manticore/platforms/evm.py
ggrieco-tob: Fixed

manticore/core/state.py
ehennenfent: Right on this line :). So you'd get something that looks like `def concretize(self, symbolic, policy, maxcount=7, *, additional_symbolics: typing.Optional[typing.List[Expression]]=None):`

ehennenfent:
@ggrieco-tob I can't replicate the failing tests locally. I believe you had the same problem as well. I'd say let's hold this one back from the next release? That will give us more time to figure out what's going on.

PR #2444, opened 2021-04-26
Add support for multiple compilation units

montyly:
See https://github.com/crytic/crytic-compile/pull/167

The changes in `setup.py` are temporary and will be removed before the next release


montyly:
Ready for review

PR #2450, opened 2021-05-09
Start new workers as more states are killed

### LINKED ISSUE #2427: Available workers decrease over time?
It seems that once a state goes from busy to killed, the worker handling that state will not take another one. This cases are not so rare to happen, for instance:

https://github.com/trailofbits/manticore/blob/3da969bf49c69c2a0f00088a96643d8b1b75264f/manticore/platforms/evm.py#L1244-L1257

This impacts the use of manticore on larger targets that require to run for a while, since it is likely that some workers will stop degrading the overall performance of our tool. Also, in the extreme case, where manticore is configured with one worker, it will hang forever (until the user press ctrl+c or the process gets killed)

gustavo-grieco:
Fix for #2427


ehennenfent (commented):
At the meeting, we discussed how checking the number of killed states across each iteration of this loop might be a little bit brittle in a multithreaded context. Fortunately, when a worker dies, Manticore [emits an event](https://github.com/trailofbits/manticore/blob/master/manticore/core/worker.py#L196) called `did_terminate_worker`. Instead of having to check the number of killed states, you could instead create `did_terminate_worker_callback` on `ManticoreBase` and then add `self.subscribe("did_terminate_worker", self.did_terminate_worker_callback)` to `__init__`. That would simplify the code and reduce the risk of race conditions.

The other thing to note is that we may still want some sort of a bail-out mechanism in case things have gone so wrong that all the workers are dying as soon as we re-create them. Perhaps we could have a maximum number of restarts allowed for each worker ID? Or maybe a maximum number of restarts in  five-minute window?

ggrieco-tob:
@ehennenfent I reimplemented this PR using `did_terminate_worker_callback`, but I still need to count the number of killed states to know how many of them I need to restart. This is required since `did_terminate_worker_callback` can be called in a variety of scenarios, where most of them are fine.

On the other hand, some sort of a bail-out mechanism if we restarted too many workers, but keep in mind that manticore should eventually finished because killed states are not revived, so the analysis will finish and manticore will report to the user that some states are unexplored. Perhaps restarting 10 or 20 workers is ok, if you are exploring 2000 states, so it is difficult to determinate. I'm still open to ideas about it.

PR #2454, opened 2021-05-15
Fix incorrect comparison for symbolic file wildcards

ekilmer:
- [x] Add a test that makes sure the `fileio.c` example passes with the correct exploration of the program.
- [x] Add a test that ensures a symbolic file with wildcard characters returns the correct number of detected wildcard characters when opened and read

Addresses https://github.com/trailofbits/manticore/discussions/2453

### DIFF
--- a/examples/script/symbolic_file.py
+++ b/examples/script/symbolic_file.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python
+
+"""
+Example to show usage of introducing a file with symbolic contents
+
+This script should be the equivalent of:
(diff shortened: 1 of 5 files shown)


ekilmer:
I'm not entirely sure how I should test the `examples/linux/fileio.c` file.

Should I write a test in the `tests/native` directory? Should a Manticore script live in the `examples/scripts` directory?

ehennenfent (approved):
Looks good! Thanks for adding a new example.

PR #2464, opened 2021-06-24
Use fixed owner and attacker accounts in multi_tx_analysis

ggrieco-tob:
This small change will hardcode the owner and attacker accounts in `multi_tx_analysis` to make results easier to read and it will force smt constraints to be the same at every run.


ehennenfent (approved):
The specific addresses should have no affect on the actual execution of the smart contract, but make it easier for us to implement caching in Manticore. LGTM!

PR #2467, opened 2021-07-09
Install pinned version of truffle to fix CI

ekilmer:
This pinned version should be reverted when https://github.com/crytic/crytic-compile/issues/199 is fixed.

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -95,7 +95,7 @@ make_wasm_sym_tests(){
 }
 
 install_truffle(){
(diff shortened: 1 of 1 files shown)


ehennenfent (approved):
Let's go ahead and merge. Using a pinned version of truffle isn't a good long-term solution, but until an updated release of crytic-compile is available on PyPI, it's the best way to unblock our development. We should revisit this at future weekly meetings and revert once all the upstream issues are resolved. Thanks for looking into this!

PR #2469, opened 2021-07-19
Chess changes breakout for upstream

ekilmer:
DO NOT MERGE.

This is a draft PR only to comment and point out some separable features from this branch to be potentially submitted as standalone PRs and eventually merged into `master`

PR #2470, opened 2021-07-20
Get Result Register

sschriner:
These features were added for development on the heap tracking module contained in the chess branch.

Related to #2469

### DIFF
--- a/manticore/native/cpu/abstractcpu.py
+++ b/manticore/native/cpu/abstractcpu.py
@@ -294,7 +294,7 @@ class Abi:
         """
         raise NotImplementedError
 
-    def get_result_reg(self):
+    def get_result_reg(self) -> str:
         """
         Extract the location a return value will be written to. Produces
         a string describing a register where the return value is written to.
--- a/manticore/native/cpu/abstractcpu.py
+++ b/manticore/native/cpu/abstractcpu.py
@@ -294,6 +294,15 @@ class Abi:
         """
         raise NotImplementedError
 
+    def get_return_reg(self):
+        """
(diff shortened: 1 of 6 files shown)


manticore/native/cpu/abstractcpu.py
```diff
@@ -294,6 +294,15 @@ def get_arguments(self):
...
         raise NotImplementedError
 
+    def get_result_reg(self):
+        """
+        Extract the location a return value will be written to. Produces
+        a string describing a register where the return value is written to.
+        :return: return register name
+        :rtype: string
+        """
+        raise NotImplementedError
```
ekilmer, line 304: Does mypy give an error if you put the return type like this?
```suggestion
    def get_result_reg(self) -> str:
        """
        Extract the location a return value will be written to. Produces
        a string describing a register where the return value is written to.
        :return: return register name
        :rtype: string
        """
        raise NotImplementedError
```
Could you also add explicit return types to the functions that override this? I'm not sure how smart mypy is or if it even checks untyped definitions.

Looks like there's an options, but I think we will have lots of errors if we turn this on https://mypy.readthedocs.io/en/stable/command_line.html#untyped-definitions-and-calls

ekilmer (changes requested):
Potentially a small improvement to type hints to do, but otherwise LGTM! Thank you!

manticore/native/cpu/abstractcpu.py
sschriner: I didn't get an error :)

ekilmer (approved):

PR #2471, opened 2021-07-20
Track last_pc in StateDescriptors

tcode2k16:
Start tracking `_last_pc` inside StateDescriptors. This is helpful for figuring out which instruction causes a state to fork.


ekilmer (approved):
LGTM!

PR #2472, opened 2021-07-23
Switch to Python 3.7

ehennenfent:
~Three days ago, the symbolic WASM tests started failing on `master`. There were no changes to manticore in between the failing tests and the passing ones. Presumably some kind of upstream update caused the failures, but Yices, Z3, and CVC don't seem to have published any releases that day. Interestingly, the failure always occurs in the `fib_L271`, which is a stress-test designed to test WASM functions with very high call depth. It also seems to take the form of SMTLib errors we've seen before, where an invalid SMT2 string is rendered into text and sent to the solver. For example, in today's failure, the exception is: `2021-07-23 12:12:49,417: [7396] m.c.worker:ERROR: Exception in state 0: SolverException('Error in smtlib: (error "at line 782, column 29: undefined identifier: BitV")',)`. I'm using this PR to try and diagnose the issue because I can't replicate the failure locally.~

This PR started out trying to fix the aforementioned test failures, which have now spontaneously resolved. In the course of working on it though, I noticed that Python 3.6 will be EOL at the end of 2021, which means we're probably okay to bump our minimum tested version of Python up to 3.7.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,10 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python 3.8
+    - name: Set up Python 3.7
       uses: actions/setup-python@v2
       with:
-        python-version: 3.8
+        python-version: 3.7
     - name: Lint
       if: github.event_name == 'pull_request'
       env:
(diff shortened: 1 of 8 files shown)


ekilmer (changes requested):
I think we should also bump the `setup.py` line here as well https://github.com/trailofbits/manticore/blob/1f7b8dc4fcc94db565cba96c9103e91838ad3ea8/setup.py#L64

It's very easy to use new Python3.7+ features, and we shouldn't claim to support a lower version than we actually test.

ehennenfent:
> I think we should also bump the `setup.py` line here as well
> 
> https://github.com/trailofbits/manticore/blob/1f7b8dc4fcc94db565cba96c9103e91838ad3ea8/setup.py#L64
> 
> It's very easy to use new Python3.7+ features, and we shouldn't claim to support a lower version than we actually test.

Fair enough. I bumped all the references to Python3.6 I could find to 3.7.

ekilmer (approved):

PR #2473, opened 2021-07-30
Fix loading custom interpreters for ELF binaries

tcode2k16:
The `interpreter_filename` variable extracted from ELF binaries can contain trailing null bytes. Those null bytes should be removed.


ekilmer (approved):
LGTM! Thank you!

PR #2475, opened 2021-08-02
Add last_executed_pc property to abstract CPU

ekilmer:
Helpful for knowing the exact last executed instruction address.

### DIFF
--- a/manticore/native/cpu/abstractcpu.py
+++ b/manticore/native/cpu/abstractcpu.py
@@ -535,7 +535,12 @@ class Cpu(Eventful):
         self._memory = memory
         self._instruction_cache: Dict[int, Instruction] = {}
         self._icount = 0
+        # _last_pc represents the last PC that was going to be executed, but it
+        # might not have been due to user hooks, exceptions, etc. You probably
+        # want last_executed_pc() or last_executed_insn()
         self._last_pc = None
+        # _last_executed_pc represents the last PC that was executed and
+        # affected the state of the program
         self._last_executed_pc = None
         self._concrete = kwargs.pop("concrete", False)
         self.emu = None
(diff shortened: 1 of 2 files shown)


manticore/native/cpu/abstractcpu.py
```diff
@@ -1009,6 +1021,7 @@ def execute(self):
         # FIXME (theo) why just return here?
         # hook changed PC, so we trust that there is nothing more to do
         if insn.address != self.PC:
+            self._last_executed_pc = self.PC
```
ekilmer, line 1024: Hmmmm. I don't entirely remember when this condition hits... I think I need to take a closer look at this to determine why `self.last_executed_pc` isn't set to `curpc`

ekilmer: See this commit for the change to `insn.address` https://github.com/trailofbits/manticore/pull/2475/commits/db06f2ca149bc1e2dd3b8f977af4932ffc281c7c

manticore/native/cpu/abstractcpu.py
```diff
@@ -985,6 +998,7 @@ def execute(self):
         """
         curpc = self.PC
         if self._delayed_event:
+            self._last_executed_pc = self._last_pc
```
ehennenfent, line 1001: I think lines 1001-1008 could be safely replaced with a call to `self._publish_instruction_as_executed`

manticore/native/cpu/abstractcpu.py
```diff
@@ -527,9 +527,10 @@ def __init__(self, regfile: RegisterFile, memory: Memory, **kwargs):
         super().__init__(**kwargs)
         self._regfile = regfile
         self._memory = memory
-        self._instruction_cache: Dict[int, Any] = {}
+        self._instruction_cache: Dict[int, Instruction] = {}
         self._icount = 0
         self._last_pc = None
+        self._last_executed_pc = None
```
ehennenfent, line 533: Seeing as this comes right after `_last_pc`, it would be worth adding a doc comment explaining the difference between the two, either here or on the property. Without understanding Manticore's internals it'd be easy to get mixed up.

ehennenfent (changes requested):
I think this passes the sanity check assuming I've understood what you're trying to do correctly. To that end, a couple of comments explaining why `last_executed_pc` might not always match `last_pc` would be helpful.

ehennenfent (approved):

PR #2476, opened 2021-08-04
Prioritize LD_LIBRARY_PATH over system default path

tcode2k16:
By prioritizing `LD_LIBRARY_PATH`, it enables the ability to run binaries with custom loaders without any patching. This is useful when trying to run a dynamically linked binary with a specific version of libc which differs from the system default.

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -1366,8 +1366,8 @@ class Linux(Platform):
                         _clean_interp_stream()
                         interpreter = ELFFile(open(interpreter_path_filename, "rb"))
                         break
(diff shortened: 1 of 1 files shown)


ekilmer (approved):
LGTM. Thank you!

ekilmer (commented):
Oh wait, I think this might be a bit more subtle...

If `LD_LIBRARY_PATH` is defined in the environment, then we _always_ try to append the base name of the library file to the end of the paths listed. However, we never check whether the library is an absolute path.

Do you know how this plays with rpath?

I might want to take a closer look at this before merging.

ekilmer (approved):
LGTM! Thank you!

PR #2479, opened 2021-08-16
Track parent in StateDescriptors

tcode2k16:
Track the parent of each state in StateDescriptors. This is helpful when generating a provenance tree for a given state.


ekilmer (approved):
LGTM! Thank you!

Issue #1715, opened 2020-05-24
AttributeError: Can't pickle local object 'ManticoreEVM.finalize.<locals>.worker_finalize'


anhdungle93:
### Summary of the problem
I tried to use manticore (installed in a conda environment) on the following contract

```
pragma solidity >=0.4.24 <0.6.0;
contract Simple {
    function f(uint a) payable public{
        if (a == 65) {
            revert();
        }
    }
}
```
with two manticore versions, one directly from pip and one from git master branch. The one from pip has the error:
```
2020-05-24 13:29:34,985: [30082] m.c.manticore:WARNING: Manticore is only supported on Linux. Proceed at your own risk!
2020-05-24 13:29:35,248: [30082] m.main:INFO: Registered plugins: DetectReentrancySimple, DetectEnvInstruction, DetectUnusedRetVal, DetectSuicidal, DetectExternalCallAndLeak, <class 'manticore.ethereum.plugins.KeepOnlyIfStorageChanges'>, DetectUninitializedMemory, DetectInvalid, DetectReentrancyAdvanced, DetectManipulableBalance, DetectIntegerOverflow, DetectUninitializedStorage, DetectDelegatecall
2020-05-24 13:29:35,249: [30082] m.main:INFO: Beginning analysis
2020-05-24 13:29:35,251: [30082] m.e.manticore:INFO: Starting symbolic create contract
2020-05-24 13:29:36,619: [30082] m.e.manticore:INFO: Starting symbolic transaction: 0
2020-05-24 13:29:36,899: [30082] m.e.manticore:INFO: 1 alive states, 3 terminated states
2020-05-24 13:29:36,964: [30082] m.e.manticore:INFO: Starting symbolic transaction: 1
2020-05-24 13:29:37,326: [30082] m.e.manticore:INFO: 1 alive states, 6 terminated states
Traceback (most recent call last):
  File "/Users/anhdungle/opt/anaconda3/envs/manticore_env/bin/manticore", line 8, in <module>
    sys.exit(main())
  File "/Users/anhdungle/opt/anaconda3/envs/manticore_env/lib/python3.8/site-packages/manticore/__main__.py", line 42, in main
    ethereum_main(args, logger)
  File "/Users/anhdungle/opt/anaconda3/envs/manticore_env/lib/python3.8/site-packages/manticore/ethereum/cli.py", line 123, in ethereum_main
    m.finalize(only_alive_states=args.only_alive_testcases)
[...]

PR #2481, opened 2021-08-24
Use mprocessing config for evm report workers

tcode2k16:
Currently, the evm report workers default to using `multiprocessing.Process` and ignore the `--core.mprocessing` config setting. This is not very desirable and can lead to issues such as #1715 on MacOS.

This pull request fixes this and spawns the report workers correctly based on the `core.mprocessing` setting.


ekilmer (approved):
LGTM! Waiting on CI. Thank you!

ekilmer:
Hmmm. This seems to be failing on an ethereum test (both runs, so it's probably not just a flaky test) https://github.com/trailofbits/manticore/pull/2481/checks?check_run_id=3415474758#step:6:80

PR #2485, opened 2021-08-31
Add more logging messages around solver errors

ekilmer:
Helps with debugging if/when things go bad during solving


ekilmer:

sschriner (approved):

PR #2486, opened 2021-08-31
Add manticore reference property to States

ekilmer:
This was added a while ago when loading states from their serialized
form in https://github.com/trailofbits/manticore/pull/1609 and this PR
makes this feature a bit more robust.


ekilmer:

sschriner (approved):

PR #2488, opened 2021-08-31
Optionally publish events for memory reads

ekilmer:
This is useful if you have multiple plugins and don't want them to
capture memory reads that are executed by other plugins.

Removed some type hints in the comments because they caused cascading
errors, and fixing those cascading errors should be a part of another
PR.


ekilmer:

sschriner (approved):
Also LGTM 😄

PR #2491, opened 2021-09-10
Add methods for getting list of (un)implemented system calls

ekilmer:
This is primarily a helper function that is useful in evaluating the syscall capabilities of Manticore on Linux


manticore/platforms/linux.py
```diff
@@ -3316,6 +3316,44 @@ def _interp_total_size(interp):
...
+        """
+        Get a listing of all concretely implemented system calls for Linux. This
+        does not include whether a symbolic version exists. To get that listing,
+        use the SLinux.implemented_syscalls() method.
+        """
+        import inspect
+
+        return (
+            x[0]
+            for x in inspect.getmembers(cls, predicate=inspect.isfunction)
```
ehennenfent, line 3330: Extremely minor nit: In situations like this, I'd prefer unpacking over indexing since you're going to reuse the piece of data in multiple places, and `for (name, _) in inspect.getmembers(cls, predicate=inspect.isfunction)` makes it a little bit more clear what `getmembers` returns.

ehennenfent (approved):
LGTM! Pointed out one extremely minor nit if you'd like tweak the code, but I'm happy to merge it as-is.

manticore/platforms/linux.py
ekilmer: Yes... Not sure what I was thinking using just `x` lol. Thank you!

PR #2495, opened 2021-10-08
Add walrus operator to shorten code

Velocities:
PLEASE NOTE: This will ONLY function on Python 3.8 or above. If you do not intend running this within those versions, please discard my pull request. Thank you!
This will shorten the code by adding the walrus operator, which can set variables inside of expressions. If there happen to be any other sections that could use this, it could be very helpful in shortening more code in this repository

### DIFF
--- a/tests/auto_generators/trace.py
+++ b/tests/auto_generators/trace.py
@@ -165,8 +165,7 @@ STACK_INSTRUCTIONS = [
 while True:
     try:
         stepped = False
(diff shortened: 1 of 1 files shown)


ehennenfent:
Thanks for the PR! Right now we're still supporting Python 3.7, so I'm afraid we can't incorporate walrus operators just yet. We'll probably make the switch to 3.8 sometime in 2022 when Python 3.7 will be nearing it's End-of-Life date. Nonetheless, thank you for your interest.

PR #2502, opened 2021-11-09
Update Regex in Solver.py

ehennenfent:
Between 2021-10-16 and 2021-10-23, only three dependencies changed their versions according to our CI: `protobuf`, `git`, and `regex`. It's difficult to imagine how this could break the WASM/EVM tests, until you notice the following lines in the 10/16 logs, but _not_ the 10/23 logs:
```
2021-10-16T12:38:09.5387382Z manticore/core/smtlib/solver.py:60
2021-10-16T12:38:09.5388574Z   /home/runner/work/manticore/manticore/manticore/core/smtlib/solver.py:60: DeprecationWarning: invalid escape sequence \(
2021-10-16T12:38:09.5390752Z     "\(([a-zA-Z0-9_]*)[ \\n\\s]*(#b[0-1]*|#x[0-9a-fA-F]*|[\(]?_ bv[0-9]* [0-9]*|true|false)\\)"
2021-10-16T12:38:09.5391316Z 
```

I believe this fixes the character escaping so the regex should work now. Hopefully this fixes the tests as well!

### DIFF
--- a/manticore/core/smtlib/solver.py
+++ b/manticore/core/smtlib/solver.py
@@ -57,7 +57,7 @@ consts.add(
 
 # Regular expressions used by the solver
 RE_GET_EXPR_VALUE_ALL = re.compile(
-    "\(([a-zA-Z0-9_]*)[ \\n\\s]*(#b[0-1]*|#x[0-9a-fA-F]*|[\(]?_ bv[0-9]* [0-9]*|true|false)\\)"
+    r"\(([a-zA-Z0-9_]*)[ \n\s]*(#b[0-1]*|#x[0-9a-fA-F]*|[(]?_ bv[0-9]* [0-9]*|true|false)\)"
 )
 RE_GET_EXPR_VALUE_FMT_BIN = re.compile(r"\(\((?P<expr>(.*))[ \n\s]*#b(?P<value>([0-1]*))\)\)")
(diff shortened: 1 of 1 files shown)


ehennenfent:
Fixing the Regex above removes the warning, but doesn't correct the test failure. In the course of my investigation, I found that the SMT2 definitions were being converted to text properly, but not making it all the way to the solver. This seems to be an issue with Python's unbuffered pipe implementation. I'm not sure what the exact issue is, but switching to buffered I/O and adding explicit calls to `flush` has fixed the tests.

manticore/core/smtlib/solver.py
```diff
@@ -57,7 +53,7 @@
 
 # Regular expressions used by the solver
 RE_GET_EXPR_VALUE_ALL = re.compile(
-    "\(([a-zA-Z0-9_]*)[ \\n\\s]*(#b[0-1]*|#x[0-9a-fA-F]*|[\(]?_ bv[0-9]* [0-9]*|true|false)\\)"
+    r"\(([a-zA-Z0-9_]*)[ \n\s]*(#b[0-1]*|#x[0-9a-fA-F]*|[(]?_ bv[0-9]* [0-9]*|true|false)\)"
```
ekilmer, line 56: Good catch!

ekilmer (approved):
Nice finds! LGTM

PR #2503, opened 2021-11-15
Bump Sphinx version to 4.3.0

ehennenfent:
Our ReadTheDocs builds have been failing due to [an outdated version of Sphinx](https://blog.readthedocs.com/build-errors-docutils-0-18/) that's not compatible with recent versions of Docutils. RtD uses this version by default in order to provide Python2 compatibility, but Manticore has no problem using the latest version instead. 

You can see the fixed build [here](https://readthedocs.org/projects/manticore/builds/15281815/) (It's not part of our CI)


ekilmer (approved):
LGTM

PR #2504, opened 2021-11-17
Add SMT simplifications for bitvec subtraction

Boyan-MILANOV:
This PR adds the following simplifications for expressions: 
`A-A ==> 0`
`A-0 ==> A`

### DIFF
--- a/manticore/core/smtlib/expression.py
+++ b/manticore/core/smtlib/expression.py
@@ -15,6 +15,14 @@ class ExpressionException(SmtlibError):
     pass
 
 
+class ExpressionEvalError(SmtlibError):
+    """Exception raised when an expression can't be concretized, typically
+    when calling __bool__() fails to produce a True/False boolean result
+    """
+
+    pass
+
+
 class XSlotted(type):
     """
     Metaclass that will propagate slots on multi-inheritance classes
@@ -65,7 +73,7 @@ class XSlotted(type):
 
 
 class Expression(object, metaclass=XSlotted, abstract=True):
-    """ Abstract taintable Expression. """
+    """Abstract taintable Expression."""
 
     __xslots__: Tuple[str, ...] = ("_taint",)
 
@@ -214,7 +222,7 @@ class Bool(Expression, abstract=True):
         x = simplify(self)
         if isinstance(x, Constant):
             return x.value
-        raise NotImplementedError("__bool__ for Bool")
+        raise ExpressionEvalError("__bool__ for Bool")
 
 
 class BoolVariable(Bool):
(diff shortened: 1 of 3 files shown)


manticore/core/smtlib/visitors.py
```diff
@@ -739,6 +740,10 @@ def visit_BitVecSub(self, expression, *operands):
                         taint=subright.taint | right.taint,
                     ),
                 )
+        elif isinstance(right, Constant) and right.value == 0:
+            return left
+        elif left is right:
```
ekilmer, line 745: I think you mentioned this during our sync, but this checks to see if they are the same instance in memory? Is there a reason why `==` isn't used? A `==` would be more correct. Maybe add a `TODO` or some sort of comment to keep track of this reason

Boyan-MILANOV: I want to check that `left` and `right`  are the same. For `BitVec` objects,  `==` is overloaded to create a new equality constraint instead of actually checking for equality, so it won't work AFAIU:   
https://github.com/trailofbits/manticore/blob/abbfa01cdc790f7ccaf3c24d6726362e53f88e79/manticore/core/smtlib/expression.py#L444-L445

ekilmer: Right, but then `BoolEqual` is used in the `if`, which I believe will call `BoolEqual.__bool__`, which is here https://github.com/trailofbits/manticore/blob/52007a8fa59e2234b4e69ff9167e8c35916aff6b/manticore/core/smtlib/expression.py#L210-L217

Of course, this would involve a potentially large runtime cost for the simplification.

The `is` check is better than nothing, so I'm fine leaving it like this 👍

ekilmer (approved):
LGTM!

manticore/core/smtlib/visitors.py
Boyan-MILANOV: Oh you're right, I did miss that. `BoolEqual.__bool__` will ultimately call [visit_BoolEqual](https://github.com/trailofbits/manticore/blob/52007a8fa59e2234b4e69ff9167e8c35916aff6b/manticore/core/smtlib/visitors.py#L509) which tests for equality using `is` but handles other cases as well, so I agree that using `==` is a better alternative

manticore/core/smtlib/expression.py
```diff
@@ -15,6 +15,14 @@ class ExpressionException(SmtlibError):
     pass
 
 
+class ExpressionEvalError(SmtlibError):
```
Boyan-MILANOV, line 18: I didn't really know how to name this exception. Something like `ExpressionConcretizeError` would be ambiguous IMO since `Concretize` is the exception that Manticore uses to fork and is totally unrelated to the error we want to address here.
Feel free to suggest a better name if you find one!

ekilmer: This is probably fine for now 👍

ekilmer (approved):
LGTM!

PR #2501, opened 2021-11-07
Improves `namedtuple` definition

sobolevn:
I am working on [new `mypy` feature](https://github.com/python/mypy/pull/11206) and it identified a problem with your `namedtuple` definitions.
https://github.com/python/mypy/pull/11206#issuecomment-962573633

By standard first string arg should match variable name.

Refs https://github.com/python/mypy/pull/11206

PR #2506, opened 2021-12-04
Improves namedtuple definition

sobolevn:
While working on https://github.com/python/mypy/pull/11206 I found that `Regspec` definition is not ideal.
It is recommended to use the same string name, as variable name. 
For example, it affects how `pickle` works.

Related https://github.com/trailofbits/manticore/pull/2501
Related https://github.com/trailofbits/manticore/commit/6e036f3034b76bbecf36b5469e49294e038ff705


ekilmer (approved):
Thank you!

PR #2511, opened 2021-12-16
x86 FXSAVE & FXRSTOR support

Boyan-MILANOV:
This PR adds support for the FXSAVE and FXRSTOR instructions

### DIFF
--- a/manticore/core/smtlib/expression.py
+++ b/manticore/core/smtlib/expression.py
@@ -15,6 +15,14 @@ class ExpressionException(SmtlibError):
     pass
 
 
+class ExpressionEvalError(SmtlibError):
+    """Exception raised when an expression can't be concretized, typically
+    when calling __bool__() fails to produce a True/False boolean result
+    """
+
+    pass
+
+
 class XSlotted(type):
     """
     Metaclass that will propagate slots on multi-inheritance classes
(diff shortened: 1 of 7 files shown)


Boyan-MILANOV:
The emulation tests are currently failing because the new `x87` registers that this PR adds are not supported by unicorn. I opened an [dedicated issue](https://github.com/unicorn-engine/unicorn/issues/1524) to ask about this.

Boyan-MILANOV:
Note to myself: when merging this PR we should open a new manticore issue that reminds us to track unicorn releases so that when they merge the support for the additional floating point registers (`FIP`, `FDP`, ...) we can stop ignoring them in the fallback emulator and include them in the `manticore <-> unicorn` context switch.

Boyan-MILANOV:
This is finally ready for review! 
I apologise for the terrible commit history on the PR - it's mostly due to me testing using Github's CI before figuring how to run it locally.

Summary of this PR:
- Adds support for `FXSAVE` and `FXRSTOR` x86 instructions
- Adds support for reading floating point registers as bitfields. For Unicorn-Engine compatibility reasons, Manticore represents floating point registers as tuples `(mantissa,exponent)`. However in order to write FP registers to memory (as done in `FXSAVE`) we need to translate the floating point tuple value into actual bits. I have added the `X86CPU.read_register_as_bitfield()` method for this purpose, which is a simple wrapper around `read_register()` that packs FP tuple values to a 80-bit python `int` if needed
- Adds a simple test that checks that doing `FXSAVE` -> clobber CPU -> `FXRSTOR` restores the CPU registers properly

ehennenfent (approved):
This looks good! I appreciate the way you encoded the layouts as tables so we'll be able to easily tweak them later. I was almost certain that `FOP` being 11 bits was a typo, but upon googling it turns out that x87 is just weirder than I expected.

PR #2513, opened 2021-12-16
Add pip-audit action workflow

elopez:
This adds a new GitHub Actions workflow. This workflow runs pip-audit
on the repository on each new commit and pull request to the `master`
branch, as well as every Tuesday morning. If any known vulnerable
dependency is found, this workflow will fail.


ehennenfent (approved):
Thanks for the PR!

PR #2514, opened 2021-12-17
Add EXPLICIT fork policy

Boyan-MILANOV:
Add a new fork policy called "EXPLICIT" that allows to specify arbitrary concrete values for the expression on whom we are forking.

### DIFF
--- a/manticore/core/state.py
+++ b/manticore/core/state.py
@@ -1,7 +1,7 @@
 import copy
 import logging
 
-from typing import List, Tuple, Sequence
+from typing import List, Tuple, Sequence, Optional, Any
 
 from .smtlib import solver, Bool, issymbolic, BitVecConstant
 from .smtlib.expression import Expression
@@ -63,7 +63,15 @@ class Concretize(StateException):
         "EXPLICIT",
     ]
 
-    def __init__(self, message, expression, setstate=None, policy=None, values=None, **kwargs):
+    def __init__(
+        self,
+        message,
+        expression,
+        setstate=None,
+        policy=None,
+        values: Optional[List[Any]] = None,
+        **kwargs,
+    ):
         if policy is None:
             policy = "ALL"
         if policy not in self._ValidPolicies:
@@ -376,7 +384,7 @@ class StateBase(Eventful):
         self._input_symbols.append(expr)
         return expr
 
-    def concretize(self, symbolic, policy, maxcount=7, explicit_values=None):
+    def concretize(self, symbolic, policy, maxcount=7, explicit_values: Optional[List[Any]] = None):
         """This finds a set of solutions for symbolic using policy.
 
         This limits the number of solutions returned to `maxcount` to avoid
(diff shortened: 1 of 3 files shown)


manticore/core/state.py
```diff
@@ -415,6 +426,8 @@ def concretize(self, symbolic, policy, maxcount=7):
             else:
                 # We assume the path constraint was feasible to begin with
                 vals = (True,)
+        elif policy == "EXPLICIT":
+            vals = explicit_values[:maxcount]
```
feliam, line 430: Note. This _could_ be the source of not feasible states. We are trusting API user to provide valid solutions. This is different to the other policies.

Boyan-MILANOV: That's true. 
I can add solver checks to ensure feasibility for each explicit value. It should't cause too much overhead compared to all the computation required by forking.

ehennenfent: I'm not opposed to adding optional feasibility checking functionality, but I don't think it's necessary for this PR. The policy API is buried fairly deep, and users are unlikely to stumble over this new policy. 

What I _would_ like to see is `_ValidPolicies` replaced with an Enum so that we can document each of the variants individually. That also doesn't have to be in this PR (since making it backwards compatible will require adding a bunch of type checking) but we should do it soon because I don't think any of the policies are documented right now.

manticore/core/state.py
```diff
@@ -415,6 +426,12 @@ def concretize(self, symbolic, policy, maxcount=7):
             else:
                 # We assume the path constraint was feasible to begin with
                 vals = (True,)
+        elif policy == "EXPLICIT":
+            for val in explicit_values:
+                if self._solver.can_be_true(self._constraints, val == symbolic):
+                    vals.append(val)
+                if len(vals) >= maxcount:
+                    break
```
Boyan-MILANOV, line 434: Added the feasibility checks here

manticore/core/state.py
```diff
@@ -415,6 +434,13 @@ def concretize(self, symbolic, policy, maxcount=7):
             else:
                 # We assume the path constraint was feasible to begin with
                 vals = (True,)
+        elif policy == "EXPLICIT":
+            if explicit_values:
+                for val in explicit_values:
```
ekilmer, line 439: Do we want to warn or `assert` here if `explicit_values` is `None` or an empty list?

ekilmer (changes requested):
Minor changes for more type hints. (Might break formatting lint when applying the suggestions from GitHub).

manticore/core/state.py
ekilmer: Never mind, we'll just return an empty list if the above is true.

manticore/core/state.py
```diff
@@ -376,7 +395,7 @@ def concretize(self, symbolic, policy, maxcount=7):
         assert self.constraints == self.platform.constraints
         symbolic = self.migrate_expression(symbolic)
 
-        vals = []
+        vals: Any = []
```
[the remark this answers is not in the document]
Boyan-MILANOV: `vals` gets assigned `Tuple` for some of the policies and `List` for others, which is why I used `Any`. Using `List[Any]` will require changing a bit more code but that shouldn't do any harm :)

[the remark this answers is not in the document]
ekilmer: 😬  Ugh. Okay. Maybe `Union[List, Tuple]`? I think if you don't specify the types, it defaults to `Any`. Would like to not have an `Any` next to an assignment of an empty list.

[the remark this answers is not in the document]
Boyan-MILANOV: When I didn't specify the type mypy seemed to infer the type based on the first assignment encountered. I will try to change the tuples into lists and see how it goes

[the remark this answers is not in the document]
Boyan-MILANOV: That worked. `concretize` now works with lists internally and converts the list into a tuple to return it, just as it did before.

ekilmer (approved):
Thank you! LGTM.

PR #2524, opened 2022-01-20
Fix `BitVecExtract` simplification for constant folding

Boyan-MILANOV:
This PR fixes an SMT simplification pattern for `BitVecExtract` on constants. 
The current implementation was ignoring the most significant bit of the extracted value.


ekilmer (approved):
LGTM!

Issue #2518, opened 2022-01-08
Manticore fails basic command on multiple python versions


Jovonni:
### Summary of the problem
The following basic command has not been running on Mac. I changed the python version several times, from 3.9, to 3.7, to 3.6.

### Manticore version

Version: 0.3.6

### Python version

Tried on multiple versions with ```pyenv versions```
```bash
3.6.0
3.6.15
* 3.6.5 
3.7.6
```

### OS / Environment

MacOS 10.15.7

### Dependencies

```bash
crytic-compile==0.2.2
cytoolz==0.11.2
dataclasses==0.8
eth-hash==0.3.2
eth-typing==3.0.0
eth-utils==2.0.0
future==0.18.2
importlib-metadata==4.8.3
intervaltree==3.1.0
manticore==0.3.6
ply==3.11
prettytable==2.5.0
protobuf==3.19.1
pyevmasm==0.2.3
pysha3==1.0.2
PyYAML==6.0
rlp==3.0.0
sortedcontainers==2.4.0
toolz==0.11.2
typing-extensions==4.0.1
wasm==1.2
wcwidth==0.2.5
z3-solver==4.8.14.0
zipp==3.6.0
```

### Step to reproduce the behavior
```bash
python -m manticore <contract>.sol --contract <contract_name> --verbose-trace
```

### Expected behavior
Maticore should successfully run the same way its running in the docker container.

### Actual behavior
```bash
2022-01-08 14:25:25,514: [6436] m.c.manticore:WARNING: Manticore is only supported on Linux. Proceed at your own risk!
2022-01-08 14:25:25,813: [6436] m.main:INFO: Registered plugins: IntrospectionAPIPlugin, <class 'manticore.ethereum.plugins.SkipRevertBasicBlocks'>, <class 'manticore.ethereum.plugins.VerboseTrace'>, <class 'manticore.ethereum.plugins.FilterFunctions'>
2022-01-08 14:25:25,814: [6436] m.main:INFO: Beginning analysis
2022-01-08 14:25:25,833: [6436] m.e.manticore:INFO: Starting symbolic create contract
Process Process-1:
Traceback (most recent call last):
  File "/Users/<user>/.pyenv/versions/3.6.5/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/Users/<user>/.pyenv/versions/3.6.5/lib/python3.6/multiprocessing/process.py", line 93, in run
[...]

PR #2527, opened 2022-01-26
Optimise forking when there is only 1 solution

Boyan-MILANOV:
This PR improves the forking process when there is only one possible solution for the symbolic expression upon which we fork. In that case, we re-use the same `State`  instead of copying the entire state to create a new one.

I don't think that Manticore has many tests for forks except this one: https://github.com/trailofbits/manticore/blob/93211002a9cea6074790c0de974fb9e603881577/tests/native/test_models.py#L227-L263 

I did test this PR in CHESS, but we should add a Manticore testcase for forking on  a single solution since it hits a rarer and slightly different code path than regular forking.

### DIFF
--- a/manticore/core/manticore.py
+++ b/manticore/core/manticore.py
@@ -523,30 +523,48 @@ class ManticoreBase(Eventful):
 
         self._publish("will_fork_state", state, expression, solutions, policy)
 
-        # Build and enqueue a state for each solution
+        # Create new states
         children = []
-        for new_value in solutions:
-            with state as new_state:
-                new_state.constrain(expression == new_value)
+        if len(solutions) == 1:
+            # If only one solution don't copy the state but update the current
+            # state instead
+            # Add state constraint
+            new_value = solutions[0]
+            state.constrain(expression == new_value)
+            setstate(state, new_value)
 
-                # and set the PC of the new state to the concrete pc-dest
-                # (or other register or memory address to concrete)
-                setstate(new_state, new_value)
+            # Put the state back in ready list
+            self._put_state(state)  # Doesn't change state id
+            self._publish("did_fork_state", state, expression, solutions, policy, [])
 
-                # enqueue new_state, assign new state id
-                new_state_id = self._put_state(new_state)
+            # Remove the state from busy list
+            with self._lock:
+                self._busy_states.remove(state.id)
+                self._lock.notify_all()
+        else:
+            # Build and enqueue a state for each solution
+            for new_value in solutions:
+                with state as new_state:
+                    new_state.constrain(expression == new_value)
 
-                # maintain a list of children for logging purpose
-                children.append(new_state_id)
+                    # and set the PC of the new state to the concrete pc-dest
+                    # (or other register or memory address to concrete)
+                    setstate(new_state, new_value)
 
-        self._publish("did_fork_state", state, expression, solutions, policy, children)
-        logger.debug("Forking current state %r into states %r", state.id, children)
+                    # enqueue new_state, assign new state id
+                    new_state_id = self._put_state(new_state)
 
-        with self._lock:
-            self._busy_states.remove(state.id)
-            self._remove(state.id)
-            state._id = None
-            self._lock.notify_all()
+                    # maintain a list of children for logging purpose
+                    children.append(new_state_id)
+
+            self._publish("did_fork_state", state, expression, solutions, policy, children)
+            logger.debug("Forking current state %r into states %r", state.id, children)
+
+            with self._lock:
+                self._busy_states.remove(state.id)
+                self._remove(state.id)
+                state._id = None
+                self._lock.notify_all()
 
     @staticmethod
     @deprecated("Use utils.log.set_verbosity instead.")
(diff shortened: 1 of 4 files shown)


Boyan-MILANOV:
I added a test for forking on a unique solution.
For the sake of simplicity I re-used the `strcpy` symbolic fork test and added a similar one with additional constraints on the input so that we trigger the case where there is only a unique solution when forking on a symbolic input byte.

tests/native/test_models.py
```diff
@@ -262,6 +263,39 @@ def strlen_model(state):
...
+        )
+        tmp_dir = tempfile.TemporaryDirectory(prefix="mcore_test_sym_")
+        m = Manticore(BIN_PATH, stdin_size=10, workspace_url=str(tmp_dir.name))
+
+        addr_of_strlen = 0x04404D0
+
+        @m.hook(addr_of_strlen)
+        def strlen_model(state):
+            stdin_1 = ArrayProxy(array=state.constraints.get_variable("STDIN")).select(1)
+            state.constrain(stdin_1 == 0)
```
Boyan-MILANOV, line 280: This constrains the second symbolic byte of the input to be zero, causing Manticore to fork on an expression that has only one possible solution.

Kokokokoka:
this PR helps with latest python3.9.9 on gentoo, one can successfully use manticore on a contract

Boyan-MILANOV:
@Kokokokoka: would you mind elaborating on your comment? Does this PR fix an opened issue, or just a bug that you was running into but didn't report?

Kokokokoka:
I've seen similar bug reports #2518 seems to be the same, maybe there were others (which had mac os in them). I had this bug both on mac m1 and on my gentoo pc. Now it works

tests/native/test_models.py
ehennenfent: I'm pretty sure I didn't know you could use `ArrayProxy` to access a specific byte of `stdin` like this. Great find!

manticore/core/manticore.py
```diff
@@ -523,30 +523,48 @@ def setstate(x, y):
...
+            # If only one solution don't copy the state but update the current
+            # state instead
+            # Add state constraint
+            new_value = solutions[0]
+            state.constrain(expression == new_value)
+            setstate(state, new_value)
+
+            # Put the state back in ready list
+            self._put_state(state)  # Doesn't change state id
+            self._publish("did_fork_state", state, expression, solutions, policy, [])
```
ehennenfent, line 538: This could be controversial (we're not forking, we're just concretizing data within the current state) but I think it's the most appropriate thing to do here. We don't have a special event for concretizing without forking. 

One thing devs of downstream tools might need to be aware of is that code that consumes the introspection API (like MUI) might need to be tweaked. The state introspection API consumes this event and tracks which states fork from which other states, so with this tweak it'll appear that we're getting multiple fork events that produce the same state.

tests/native/test_models.py
```diff
@@ -262,6 +263,39 @@ def strlen_model(state):
...
+            state.invoke_model(strlen_exact)
+
+        m.run()
+        m.finalize()
+
+        # Expected stdout outputs
+        expected = {"Length of string is: 0", "Length of string is: 1"}
```
ehennenfent, line 287: Would we expect to see different results for this test before the changes to our forking strategy? If we're forking on the length of the string, and constrain the _second_ byte to be zero, then it makes sense that valid strings would be of length zero or one. That's two possibilities, which doesn't seem like it would test the new forking policy. 

Would it make more sense to inspect the fork events using a plugin, or inspect the state ID numbers to confirm that we aren't creating new states when we concretize a symbol with a single possible value?

ehennenfent:
The implementation looks good to me! Once we clarify my final question on the model tests I'm happy to approve & merge.

ehennenfent (commented):
Hmm, submitting via github.dev dropped all my comments, so let's try that again

manticore/core/manticore.py
Boyan-MILANOV: When publishing the fork event the `children` parameter is set to `[]` to indicate that this fork doesn't result in new states being created. 

I agree that this is not really a fork since there is only one solution but in most cases the users would probably still want their callbacks to be executed.

tests/native/test_models.py
Boyan-MILANOV: Yeah that's spot on. I hadn't noticed that the `strlen_exact` callback doesn't concretise the individual bytes but uses the solver API to check whether a byte is `null`, avoiding any forks. Definitely an oversight on my part.

I will replace that test with one that creates a Manticore instance and then raises a `ConcretizeException` on a constrained symbolic variable, then just checks that no additional states were created using `will_fork_state_callback`.

Boyan-MILANOV: I have replaced the test with another one here: https://github.com/trailofbits/manticore/blob/003ce0ad620046e9602e95c1eb2c763ba66cedf9/tests/other/test_fork.py 

The new test executes the simple `hello_world` program and artificially concretises a symbolic variable (constrained to one unique value) from a hook, and finally asserts that no additional states were created when forking. Let me know what you think!

ehennenfent (approved):
The new test looks great!

Issue #2528, opened 2022-02-04
Fix Manticore <-> Unicorn context switch


Boyan-MILANOV:
### Summary of the problem
Opening this issue for traceability after merging #2511.
#2511 added new FPU registers (`FIP`, `FDP`, ...) to manticore. Those registers were also missing in Unicorn and have been added by their team since, although in their `dev` branch which is not yet merged into master. Thus, manticore currently ignores those registers when syncing the CPU context with unicorn: 
https://github.com/trailofbits/manticore/blob/bf2fba399045d33dc574dab937a4d0d1fddb8fdf/manticore/utils/emulate.py#L92

When Unicorn has a new release and adds the new registers to their python API we can stop ignoring them during the context switch.

Boyan-MILANOV:
#2564 updates to unicorn `2.0` which now supports all these registers except `MXCSR_MASK`.

PR #2531, opened 2022-02-10
Ignore missing unicorn registers in the fallback emulator

Boyan-MILANOV:
As described in https://github.com/trailofbits/manticore/issues/2528, we need to exclude some FPU registers from the unicorn <-> manticore context switch when using the emulator, because those registers aren't supported in unicorn yet.

It turns out that the FPU registers also need to be ignored in the fallback emulator (different than the regular emulator), which is what this PR does.


ekilmer (approved):
LGTM!

PR #2533, opened 2022-02-16
Split off ENDBR32/64 from CHESS branch

ehennenfent:
Seeing as they're fancy NOP's, I don't think there's any reason not to merge them into `master`, and thus avoid aggressively concretizing the state in order to emulate them under Unicorn.

Thanks for the implementation, @lordidiot


Boyan-MILANOV (approved):
LGTM!

PR #2534, opened 2022-02-18
bugfix in solidity_create_contract

lordidiot:
Current implementation will cause `KeyError` for contracts with non-payable constructors.


lordidiot:
Oops, I think this fix has issues, will need to revisit this.

PR #2535, opened 2022-02-24
fix typo in Manticore.linux constructor docstring

kokrui:
noticed on the sphinx-generated docs that `entry_symbol` had no type specified and passing a string to `envp` didn't work, turns out there's a typo


ekilmer (approved):
LGTM. Thank you for the update!

PR #2542, opened 2022-04-25
Fix bug in register_log_callback

kokrui:
Currently a test in `test_tui_api.py` is failing and the TUI itself doesn't send logs over as a protobuf message because the LogCaptureWorker's `log_callback` isn't being registered correctly. This PR fixes that!


ekilmer (approved):
This is great!! Thank you for figuring this out! So nice to have passing tests now 🚀

Issue #489, opened 2017-09-08
Linker can't find C library, ignores .rpath?


251:
### OS / Environment
Docker: Ubuntu 16.04

Installed C libraries:
```
/lib/x86_64-linux-gnu/libc.so.6
/usr/lib/x86_64-linux-gnu/libc.so
/opt/i686-unknown-linux-gnu/i686-unknown-linux-gnu/sysroot/lib/libc.so.6
/opt/i686-unknown-linux-gnu/i686-unknown-linux-gnu/sysroot/usr/lib/libc.so
/opt/i686-unknown-linux-musl/i686-unknown-linux-musl/sysroot/usr/lib/libc.so
/opt/x86_64-unknown-linux-gnu/x86_64-unknown-linux-gnu/sysroot/lib64/libc.so.6
/opt/x86_64-unknown-linux-gnu/x86_64-unknown-linux-gnu/sysroot/usr/lib64/libc.so
/opt/x86_64-unknown-linux-musl/x86_64-unknown-linux-musl/sysroot/usr/lib/libc.so
```

### Manticore version
Version: 0.1.4

### Python version
Python 2.7.12

### Dependencies
capstone==3.0.5rc2
manticore==0.1.4
ply==3.10
pyelftools==0.24
unicorn==1.0.1

### Summary of the problem
Running natively with `LD_DEBUG=libs` results in:
```
      2309:	find library=libcheck-g3-O0-s0.so [0]; searching
[...]

PR #2544, opened 2022-05-03
Fix newstat and newfstat syscalls

lordidiot:
Two changes in this PR

## newstat
- Currently `sys_newstat`'s implementation is based on `sys_stat64`
- This is inaccurate because `sys_stat64` is expected to return a `struct stat64` back to userland [ref1](https://elixir.bootlin.com/linux/v5.17.5/source/fs/stat.c#L521) [ref2](https://elixir.bootlin.com/linux/v5.17.5/source/arch/x86/include/uapi/asm/stat.h#L42)
- Instead, `sys_newstat` is supposed to return a `struct stat` [ref1](https://elixir.bootlin.com/linux/v5.17.5/source/fs/stat.c#L380) [ref2](https://elixir.bootlin.com/linux/v5.17.5/source/arch/x86/include/uapi/asm/stat.h#L83)
- This causes issues because the two structs have different definitions, and in practice causes errors in the loader. For example "LD_LIBRARY_PATH" environment variable does not work. I suspect https://github.com/trailofbits/manticore/issues/489 might be related as well but I have yet to test.
- To fix this we can reuse the `sys_newfstat` implementation, which returns the correct `struct stat` structure to userland

## newfstat 
- Currently `sys_newfstat` returns a structure based on the x86_64 version of `struct stat` [ref1](https://elixir.bootlin.com/linux/v5.17.5/source/arch/x86/include/uapi/asm/stat.h#L83)
- This does not account for the different `struct stat` definition on 32-bit x86 [ref1](https://elixir.bootlin.com/linux/v5.17.5/source/arch/x86/include/uapi/asm/stat.h#L10) (Notice the `#ifdef`)

### DIFF
--- a/manticore/platforms/linux.py
+++ b/manticore/platforms/linux.py
@@ -3282,11 +3282,19 @@ class Linux(Platform):
 
         bufstat = add(nw, stat.st_dev)  # long st_dev
         bufstat += add(nw, stat.st_ino)  # long st_ino
-        bufstat += add(nw, stat.st_nlink)  # long st_nlink
-        bufstat += add(4, stat.st_mode)  # 32 mode
-        bufstat += add(4, stat.st_uid)  # 32 uid
-        bufstat += add(4, stat.st_gid)  # 32 gid
-        bufstat += add(4, 0)  # 32 _pad
+
+        if self.current.address_bit_size == 64:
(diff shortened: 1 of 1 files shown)


Boyan-MILANOV (approved):
LGTM!

PR #2553, opened 2022-05-24
Add `psubq` instruction (x86)

lordidiot:
Adds support for `psubq` x86 instruction. Implementation very similar to `psubb`.

### DIFF
--- a/manticore/native/cpu/x86.py
+++ b/manticore/native/cpu/x86.py
@@ -5903,7 +5903,7 @@ class X86Cpu(Cpu):
         for i in reversed(range(0, dest.size, 64)):
             a = Operators.EXTRACT(value_a, i, 64)
             b = Operators.EXTRACT(value_b, i, 64)
-            result.append((a - b) & 0xFFFFFFFFFFFFFFFF)
+            result.append(a - b)
         dest.write(Operators.CONCAT(dest.size, *result))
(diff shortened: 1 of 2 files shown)


manticore/native/cpu/x86.py
```diff
@@ -5881,6 +5881,31 @@ def PSUBB(cpu, dest, src):
...
+        :param src: source operand.
+        """
+        result = []
+        value_a = dest.read()
+        value_b = src.read()
+
+        for i in reversed(range(0, dest.size, 64)):
+            a = Operators.EXTRACT(value_a, i, 64)
+            b = Operators.EXTRACT(value_b, i, 64)
+            result.append((a - b) & 0xFFFFFFFFFFFFFFFF)
```
Boyan-MILANOV, line 5906: We are subtracting on 64 bits already so I don't think masking with `0xffffffffffffffff` is necessary?

Boyan-MILANOV (approved):
Left one small comment but it looks good to me otherwise!

manticore/native/cpu/x86.py
lordidiot: Ok! I thought it should be used to explicitly mask any negative results, but I think this is also done in `CONCAT` so I will remove it.

PR #2555, opened 2022-05-31
Regenerate protobuf code with protoc>=3.19.0

kokrui:
As of protobuf 4.21.0, generated code must be generated with protoc>=3.19.0

This PR generates the TUI state protobuf code with protoc==3.21.1


kokrui:
seems like upgrading to protobuf 4 requires some syntax changes to the protobuf spec and the API is slightly different, so I'll just pin the protobuf version to 3.20.1 for now

PR #2556, opened 2022-05-31
Pin protobuf version to 3.20.1

kokrui:
Protobuf introduced major version 4.21.0, which contains breaking changes 

This is a quick fix which pins protobuf to 3.20.1 before we eventually migrate the codebase to proto4


ekilmer:
closing in favor of using better dependency version constraints in #2558

PR #2562, opened 2022-07-01
Use python3.7 instead of python3

gsm1011:
`python3` will give `ModuleNotFoundError: No module named 'manticore'` error.


Boyan-MILANOV:
Thanks you @gsm1011 for this PR. Would you mind signing the Contributor Licence Agreement so that we can merge it ?

PR #2564, opened 2022-07-20
Update to Unicorn v2.0.0

ekilmer:
Fix CPUID errors with latest glibc

Add stub for new `sys_rseq` syscall in glibc


Boyan-MILANOV (approved):
LGTM! Not sure adding the `rseq` syscall stub belongs to this PR but it's not a big deal.

PR #2566, opened 2022-07-22
Patch out wallet requirement for truffle test

ekilmer:
Patch is from this PR https://github.com/truffle-box/metacoin-box/pull/32


ekilmer:
This is fixed by https://github.com/truffle-box/metacoin-box/pull/33

PR #2568, opened 2022-07-22
Loosen test dependency version constraints

ekilmer:
Only use a lower bound (not sure if the lower bound actually means
anything with regards to what Manticore _requires_, but better keep it
around just in case)


Boyan-MILANOV (approved):
LGTM!

PR #2663, opened 2025-05-19
Fix docker (try to use z3 prebuilt binary)

Yustynn:
Problem: Couldn't use docker on my M1 mac due to failure to build `z3-solver` wheel.

Solution: try to use a prebuilt binary if possible.

It works fine with this.

Original erorr message attached for context
[z3-error.txt](https://github.com/user-attachments/files/20298072/z3-error.txt)


disconnect3d:
This makes no sense. Any of this should go to pyproject.toml

PR #2693, opened 2026-05-25
fix: encode static arrays and dynamic tuples in ABI

### LINKED ISSUE #1219: ethereum.ABI incorrectly (de)serializes certain types

TUPM96:
Fixes #1219

## Summary
- classify ABI types as static or dynamic according to the Solidity ABI spec
- encode fixed-size arrays of static elements inline instead of as dynamic arrays
- encode and decode nested tuples with dynamic members through head offsets
- add regression coverage for `uint[1]` and nested dynamic tuples

## Validation
- `python -m black --check manticore/ethereum/abi.py tests/ethereum/test_general.py`
- `python -m compileall manticore/ethereum/abi.py tests/ethereum/test_general.py`
- custom ABI regression script covering fixed arrays, dynamic tuples, dynamic arrays, fixed arrays with dynamic elements, and function-call encoding/decoding
- `git diff --check`

Full pytest could not be run on this Windows checkout because Manticore imports POSIX-only `fcntl`; WSL is present but lacks pip/ensurepip in this environment.


disconnect3d:
See https://github.com/trailofbits/manticore/pull/2694#issuecomment-4741794812


Happy to review and maybe merge if u sign in CLA. Since lots of time has passed w/o it, I am closing it.

Issue #1251, opened 2018-11-03
Implement EVM refunds


feliam:
There is not support for refunds.
For example writing a zero to a storage slot that previously had a value some gas is refunded.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 100.0 DAI (100.0 USD @ $1.0/DAI)  attached to it as part of the ECF fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/trailofbits/manticore/1251/1701).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/trailofbits/manticore/1251/1701).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $56,333.57 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 10 months, 1 week from now.
Please review their action plans below:


**1) [jalextowle](https://gitcoin.co/profile/jalextowle) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/trailofbits/manticore/1251/1701?mutate_worker_action=approve&worker=jalextowle) | [reject worker](https://gitcoin.co/issue/trailofbits/manticore/1251/1701?mutate_worker_action=reject&worker=jalextowle))_.**

The opcodes that refund gas are sstore and selfdestruct. The Ethereum Yellow Paper specifies this behavior, so my first step would include solidifying my understanding of EVM level refunds using the Yellow Paper. Next I would use the specific numbers for gas refunds within the implementation of the sstore and selfdestruct opcodes in Manticore.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/trailofbits/manticore/1251/1701).

gitcoinbot:
@jalextowle Hello from Gitcoin Core - are you still working on this issue? Please submit a WIP PR or comment back within the next 3 days or you will be removed from this ticket and it will be returned to an ‘Open’ status. Please let us know if you have questions!
* [x] reminder (3 days)
[...]

PR #2694, opened 2026-05-25
fix: discard EVM refunds on reverted transactions

TUPM96:
Part of #1251

## Summary
- discard accumulated EVM gas refunds when a transaction rolls back via `REVERT`/rollback paths
- keep unused gas behavior intact for `REVERT`, while `THROW`/`TXERROR` still consume all gas
- add a regression test proving an `SSTORE` clear followed by `REVERT` restores storage and charges the caller without applying the storage refund

## Why
Manticore already tracks storage/selfdestruct refunds, but a reverted transaction should not keep refunds from state changes that were rolled back. Before this patch, clearing a non-zero storage slot and then reverting charged only half the expected gas because the SSTORE refund survived the rollback.

## Validation
- `python -m black --check manticore/platforms/evm.py tests/ethereum/test_general.py`
- `python -m compileall manticore/platforms/evm.py tests/ethereum/test_general.py`
- `git diff --check`
- `pytest tests/ethereum/test_general.py::EthSpecificTxIntructionTests -q` with Windows compatibility stubs for `fcntl`/`sha3` and a solver-binary discovery shim; result: `7 passed`


disconnect3d:
Please sign in CLA if you want us to review and possibly merge your PRs.

> Payout details can be provided privately on acceptance.

What do you mean by that? I don't think we are going to provide any bounties for Manticore contributions at this point.

Issue #1365, opened 2019-01-23
Need a way to save EVM state


drortirosh:
### OS / Environment
	Ubuntu 16.04.5 LTS

### Manticore version
0.2.4
### Python version
3.7

### Summary of the problem

Our contract takes a lot of time to start the first time.
just calling solidity_create_contract() (and some initialization transaciton) takes very long, before running each state.

It would be very nice to SAVE the EVM state at this point,  and be able to restore it on each run, when we modify the state processing code.

e.g.
```
m = ManticoreEVM(..)
if !reload:
   myContract = m.solidity_create_contract(...)
   myContract.init (...)
else:
  m.reloadState(...)

...
```

PR #2695, opened 2026-05-25
fix: support saving and resuming EVM states

TUPM96:
Addresses #1365

## Summary
- add a public `ManticoreEVM.save_state()` helper for serializing a ready EVM state
- override `ManticoreEVM.from_saved_state()` so EVM saved states resume through an EVM manager instead of being passed as the `plugins` positional argument
- add regression coverage that saves an initialized EVM state, restores it, and successfully continues with a transaction

## Validation
- `python -m black --check manticore/ethereum/manticore.py tests/ethereum/test_general.py`
- `python -m compileall manticore/ethereum/manticore.py tests/ethereum/test_general.py`
- `git diff --check`
- `pytest tests/ethereum/test_general.py::EthPluginTests::test_save_and_resume_evm_state -q` with Windows compatibility stubs for `fcntl`/`sha3` and a solver-binary discovery shim; result: `1 passed`

Note: running the adjacent Solidity-backed `test_checkpoint` on this Windows checkout is blocked by the existing local `solc`/crytic-compile temp-file issue before this patch's assertions run.


disconnect3d:
See https://github.com/trailofbits/manticore/pull/2694#issuecomment-4741794812


Happy to review and maybe merge if u sign in CLA. Since lots of time has passed w/o it, I am closing it.

